PR #2, opened 2020-02-20
Update .gitignore for GoLand 

aravindhp:
Add GoLand config directory to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,3 +84,6 @@ tags
 .vscode/*
 .history
 # End of https://www.gitignore.io/api/go,vim,emacs,visualstudiocode
+
+# GoLand
+.idea/


.gitignore
```diff
@@ -11,5 +11,9 @@
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
 
+# Directory of IDE settings
+*.idea/*
+*.vscode/*
```
sebsoto: From git ignore docs:

> The pattern foo/ will match a directory foo and paths underneath it, but will not match a regular file or a symbolic link foo (this is consistent with the way how pathspec works in general in Git)

Do you need the trailing `*`?

Especially since:


> The pattern "foo/*", matches "foo/test.json" (a regular file), "foo/bar" (a directory), but it does not match "foo/bar/hello.c" (a regular file), as the asterisk in the pattern does not match "bar/hello.c" which has a slash in it.

aravindhp: I copied it over from the WMCB repo. We should fix it there too.

PR #31, opened 2020-04-02
[docs] Fix link for SshKeyPair in README

mansikulkarni96:
Updated Readme to point the correct link for updating shheyPair,
now mentioned in create_test.go file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ export AWS_SHARED_CREDENTIALS_FILE=<path to aws credentials file>
 export CLUSTER_ADDR=<cluster_name, eg: ravig211.devcluster.openshift.com>
 export KUBE_SSH_KEY_PATH=<path to ssh key>
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -21,7 +21,7 @@ export AWS_SHARED_CREDENTIALS_FILE=<path to aws credentials file>
 export CLUSTER_ADDR=<cluster_name, eg: ravig211.devcluster.openshift.com>
 export KUBE_SSH_KEY_PATH=<path to ssh key>
 ```
-- Update the keypair mentioned at [sshKeyPair](https://github.com/openshift/windows-machine-config-operator/blob/42593c5d2eb798b572c58b5debafc4c392d1f967/test/e2e/wmco_test.go#L59) to match with what 
+- Update the sshKeyPair mentioned in [create_test.go](https://github.com/openshift/windows-machine-config-operator/blob/master/test/e2e/create_test.go) to match with what
```
PratikMahajan, line 24: you can point to the line number where this needs to be updated

aravindhp: I suggest we don't do that. Every time the code is moved around we have to remember to update the README.

mansikulkarni96: @PratikMahajan refrained from doing that for the same reason.

PR #33, opened 2020-04-03
[ci] Add local dev flags to e2e script

ravisantoshgudimetla:
As of now, after the test suite has been run, the
Windows nodes would be deleted as well. This commit
ensures that
- We have a flag which retains the Windows node
 created as part of test suite. Please note that
 we're skipping the part where the delete
 suite is being called in case this flag has been
 set.
- We have flag to specify the number of number of
  Windows nodes to be created as part of test run
- We won't download operator-sdk if it's already
  available in the PATH

```
hack/run-ci-e2e-test.sh -s -k "openshift-dev" -n 2"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,9 +21,6 @@ export AWS_SHARED_CREDENTIALS_FILE=<path to aws credentials file>
 export CLUSTER_ADDR=<cluster_name, eg: ravig211.devcluster.openshift.com>
 export KUBE_SSH_KEY_PATH=<path to ssh key>
 ```
-- Update the sshKeyPair mentioned in [create_test.go](https://github.com/openshift/windows-machine-config-operator/blob/master/test/e2e/create_test.go) to match with what 
-is being used in KUBE_SSH_KEY_PATH. Please note that we're using libra as keypair
-for our CI purposes.
 - Ensure that /payload directory exists and is accessible by the user account. The directory needs to be populated with the following files. Please see the [Dockerfile](https://github.com/openshift/windows-machine-config-operator/blob/master/build/Dockerfile) for figuring where to download and build these binaries. It is up to the user to keep these files up to date.
 ```
 /payload/
@@ -42,9 +39,21 @@ for our CI purposes.
 ```
 Once the above variables are set and the /payload directory has been populated, run the following script:
 ```shell script
-hack/run-ci-e2e-test.sh
+hack/run-ci-e2e-test.sh -k "openshift-dev"
 ```
+We assume that the developer uses `openshift-dev` as the key pair in the aws cloud
 
+Additional flags that can be passed to `hack/run-ci-e2e-test.sh` are
+- `-s` to skip the deletion of Windows nodes that are created as part of test suite run
+- `-n` to represent the number of Windows nodes to be created for test run
+- `-k` to represent the AWS specific key pair that will be used during e2e run and it should map to the private key
+       that we have in `KUBE_SSH_KEY_PATH`. The default value points to `libra` which we use in our CI
+       
+Example command to spin up 2 Windows nodes and retain them after test run:
+```
+hack/run-ci-e2e-test.sh -s -k "openshift-dev" -n 2      
+```
+        
 ## Bundling the Windows Machine Config Operator
 This directory contains resources related to installing the WMCO onto a cluster using OLM.
 
--- a/hack/run-ci-e2e-test.sh
+++ b/hack/run-ci-e2e-test.sh
@@ -7,21 +7,61 @@ set -o pipefail
 # TODO: This needs to be removed as part of https://issues.redhat.com/browse/WINC-351
 oc patch network.operator cluster --type=merge -p '{"spec":{"defaultNetwork":{"ovnKubernetesConfig":{"hybridOverlayConfig":{"hybridClusterNetwork":[{"cidr":"10.132.0.0/14","hostPrefix":23}]}}}}}'
 
+NODE_COUNT=""
+SKIP_NODE_DELETION=""
+KEY_PAIR_NAME=""
+
 WMCO_ROOT=$(dirname "${BASH_SOURCE}")/..
 
 export CGO_ENABLED=0
 export CLUSTER_ADDR=$(oc cluster-info | head -n1 | sed 's/.*\/\/api.//g'| sed 's/:.*//g')
 
+while getopts ":n:k:s" opt; do
+  case ${opt} in
+    n ) # process option for the node count
+      NODE_COUNT=$OPTARG
+      ;;
+    k ) # process option for the keypair to be used by AWS cloud provider
+      KEY_PAIR_NAME=$OPTARG
+      ;;
+    s ) # process option for skipping deleting Windows VMs created by test suite
+      SKIP_NODE_DELETION="-skip-node-deletion"
+      ;;
+    \? )
+      echo "Usage: $0 [-n] [-k] [-s]"
+      exit 0
+      ;;
+  esac
+done
+
 # Copy the cloud credentials and KUBESSH key path so that they can be used by operator
 cp $AWS_SHARED_CREDENTIALS_FILE /etc/cloud/credentials
 cp $KUBE_SSH_KEY_PATH /etc/private-key/private-key.pem
 
-# Get operator-sdk binary.
-# TODO: Make this to download the same version we have in go dependencies in gomod
-wget -O /tmp/operator-sdk https://github.com/operator-framework/operator-sdk/releases/download/v0.15.2/operator-sdk-v0.15.2-x86_64-linux-gnu && chmod +x /tmp/operator-sdk
+# Download the operator-sdk binary only if it is not already available
+# We do not validate the version of operator-sdk if it is available already
+if ! type operator-sdk > /dev/null; then
+  # TODO: Make this download the same version we have in go dependencies in gomod
+  wget -O /tmp/operator-sdk https://github.com/operator-framework/operator-sdk/releases/download/v0.15.2/operator-sdk-v0.15.2-x86_64-linux-gnu && chmod +x /tmp/operator-sdk
+  # To expand alias. We'll add an alias for operator-sdk to be in `/tmp/operator-sdk`
+  shopt -s expand_aliases
+  alias operator-sdk=/tmp/operator-sdk
+fi
+
+# Set default values for the flags. Without this operator-sdk flags are getting
+# polluted. For example, if KEY_PAIR_NAME is not passed or passed as empty value
+# the value is literally taken as "" instead of empty-string so default values we
+# specified in main_test.go has literally no effect. Not sure, if this is because of
+# the way operator-sdk testing is done using `go test []string{}
+NODE_COUNT=${NODE_COUNT:-1}
+SKIP_NODE_DELETION=${SKIP_NODE_DELETION:-"-skip-node-deletion=false"}
+KEY_PAIR_NAME=${KEY_PAIR_NAME:-"libra"}
+
 cd $WMCO_ROOT
-oc create -f deploy/namespace.yaml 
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -42,9 +39,14 @@ for our CI purposes.
...
 ```shell script
-hack/run-ci-e2e-test.sh
+hack/run-ci-e2e-test.sh -k "openshift-dev"
 ```
-
+Additional flags that can be passed to `hack/run-ci-e2e-test.sh` are
+- `-s` to skip the deletion of Windows nodes that are created as part of test suite run
+- `-n` to represent the number of Windows nodes to be created for test run
+- `-k` to represent the key pair that will be used during e2e run. This private key path in this pair should match with 
+       what we have in `KUBE_SSH_KEY_PATH` The default value points to `libra` which we use in our CI
```
aravindhp, line 48: This is confusing. From the example above it seems you are asking for the key pair name in the cloud account but you are not being clear about the connection with `KUBE_SSH_KEY_PATH`.

Please add the example you have included in the commit message here too.

hack/run-ci-e2e-test.sh
```diff
@@ -3,21 +3,54 @@ set -o errexit
...
 
 export CGO_ENABLED=0
 export CLUSTER_ADDR=$(oc cluster-info | head -n1 | sed 's/.*\/\/api.//g'| sed 's/:.*//g')
 
+while getopts ":n:k:s" opt; do
+  case ${opt} in
+    n ) # process option for providing existing VM credentials
+      NODE_COUNT=$OPTARG
+      ;;
+    k )
```
aravindhp, line 21: Missing doc

README.md
ravisantoshgudimetla: `This private key path in this pair should match with what we have in `KUBE_SSH_KEY_PATH``. You mean I should elaborate on that?

hack/run-ci-e2e-test.sh
```diff
@@ -3,21 +3,54 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+# TODO: Add input validation
```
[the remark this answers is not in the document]
ravisantoshgudimetla: I think it's ok even if we don't the validation as long as the documentation is clear. I'll remove the TODO

hack/run-ci-e2e-test.sh
ravisantoshgudimetla: Will add it

README.md
aravindhp: What does `This private key path in this pair` even mean? You should make it less confusing. Reading the explanation it seemed like `-k` can be used to specify the key itself. You need to call out that this specific to key-pair names in the AWS account and that it should map to the private key specified in `KUBE_SSH_KEY_PATH`.

hack/run-ci-e2e-test.sh
```diff
@@ -3,21 +3,54 @@ set -o errexit
...
+done
+
 # Copy the cloud credentials and KUBESSH key path so that they can be used by operator
 cp $AWS_SHARED_CREDENTIALS_FILE /etc/cloud/credentials
 cp $KUBE_SSH_KEY_PATH /etc/private-key/private-key.pem
 
 # Get operator-sdk binary.
-# TODO: Make this to download the same version we have in go dependencies in gomod
-wget -O /tmp/operator-sdk https://github.com/operator-framework/operator-sdk/releases/download/v0.15.2/operator-sdk-v0.15.2-x86_64-linux-gnu && chmod +x /tmp/operator-sdk
+# We don't care about the operator-sdk version we're using if running locally.
```
mansikulkarni96, line 39: Could be reframed, a little confusing as of now something like, we can use any operator-sdk version if we are using it for running locally.

hack/run-ci-e2e-test.sh
```diff
@@ -3,21 +3,54 @@ set -o errexit
...
+
+# Set default value for node count
+NODE_COUNT=${NODE_COUNT:-1}
+
 cd $WMCO_ROOT
-oc create -f deploy/namespace.yaml 
-/tmp/operator-sdk test local ./test/e2e --up-local --namespace=windows-machine-config-operator --go-test-flags "-v -timeout 60m"
+oc create -f deploy/namespace.yaml
+# The bool flags in golang does not respect key value pattern. They follow key=value pattern.
+# -flag x  // non-boolean flags only(https://golang.org/pkg/flag/)
```
mansikulkarni96, line 52: this doc is a little confusing.

test/e2e/main_test.go
```diff
@@ -27,6 +32,10 @@ var (
 type globalContext struct {
 	// numberOfNodes to be used for the test suite.
 	numberOfNodes int
+	// skipNodeDeletion flag allows skipping deletion of nodes within the test suite
```
mansikulkarni96, line 35: should this doc be consistent with above doc for skipNodeDeletion?

ravisantoshgudimetla: no preference but I can change it.

hack/run-ci-e2e-test.sh
```diff
@@ -3,21 +3,54 @@ set -o errexit
...
 cp $KUBE_SSH_KEY_PATH /etc/private-key/private-key.pem
 
 # Get operator-sdk binary.
-# TODO: Make this to download the same version we have in go dependencies in gomod
-wget -O /tmp/operator-sdk https://github.com/operator-framework/operator-sdk/releases/download/v0.15.2/operator-sdk-v0.15.2-x86_64-linux-gnu && chmod +x /tmp/operator-sdk
+# We don't care about the operator-sdk version we're using if running locally.
+if ! type operator-sdk > /dev/null; then
+  # TODO: Make this to download the same version we have in go dependencies in gomod
+  wget -O /tmp/operator-sdk https://github.com/operator-framework/operator-sdk/releases/download/v0.15.2/operator-sdk-v0.15.2-x86_64-linux-gnu && chmod +x /tmp/operator-sdk
+  export PATH=$PATH:/tmp
```
sebsoto, line 43: I'm not sure if this should be done, it could cause other side effects. Might be better to create an alias or something?

ravisantoshgudimetla: I thought about this but I looked at the pod in CI which is controlled environment and it shouldn't cause an issue and in our machine when we are running it might be ok considering the `PATH` env variable is limited to scope of this bash script. However if you want this changed I can do it.

hack/run-ci-e2e-test.sh
ravisantoshgudimetla: I mean we're not validating if the `operator-sdk` which developer has in his/her machine matches with 0.15.2 which is what we have tested for.

test/e2e/main_test.go
mansikulkarni96: This would be just a nit.

hack/run-ci-e2e-test.sh
ravisantoshgudimetla: Which part?

README.md
```diff
@@ -42,9 +39,19 @@ for our CI purposes.
 ```
 Once the above variables are set and the /payload directory has been populated, run the following script:
 ```shell script
-hack/run-ci-e2e-test.sh
+hack/run-ci-e2e-test.sh -k "openshift-dev"
```
aravindhp, line 42: Add a line mentioning that we are assuming developers are using the `openshift-dev` key in their AWS account.

hack/run-ci-e2e-test.sh
```diff
@@ -7,21 +7,61 @@ set -o pipefail
...
 cp $AWS_SHARED_CREDENTIALS_FILE /etc/cloud/credentials
 cp $KUBE_SSH_KEY_PATH /etc/private-key/private-key.pem
 
-# Get operator-sdk binary.
-# TODO: Make this to download the same version we have in go dependencies in gomod
-wget -O /tmp/operator-sdk https://github.com/operator-framework/operator-sdk/releases/download/v0.15.2/operator-sdk-v0.15.2-x86_64-linux-gnu && chmod +x /tmp/operator-sdk
+# Download the operator-sdk binary only if not is not already available
+# We do not validate the version of operator-sdk if it is available already
+if ! type operator-sdk > /dev/null; then
+  # TODO: Make this to download the same version we have in go dependencies in gomod
```
aravindhp, line 44: `Make this download...` No need for the `to`.

hack/run-ci-e2e-test.sh
```diff
@@ -7,21 +7,61 @@ set -o pipefail
...
+  wget -O /tmp/operator-sdk https://github.com/operator-framework/operator-sdk/releases/download/v0.15.2/operator-sdk-v0.15.2-x86_64-linux-gnu && chmod +x /tmp/operator-sdk
+  # To expand alias
+  shopt -s expand_aliases
```
aravindhp, line 47: Why is this needed?

hack/run-ci-e2e-test.sh
```diff
@@ -7,21 +7,61 @@ set -o pipefail
...
+KEY_PAIR_NAME=${KEY_PAIR_NAME:-"libra"}
+
 cd $WMCO_ROOT
-oc create -f deploy/namespace.yaml 
-/tmp/operator-sdk test local ./test/e2e --up-local --namespace=windows-machine-config-operator --go-test-flags "-v -timeout 60m"
+oc create -f deploy/namespace.yaml
+# The bool flags in golang does not respect key value pattern. They follow key=value pattern.
+# -flag x  // non-boolean flags only(https://golang.org/pkg/flag/)
+operator-sdk test local ./test/e2e --up-local --namespace=windows-machine-config-operator --go-test-flags "-v -timeout=60m -node-count=$NODE_COUNT $SKIP_NODE_DELETION -ssh-key-pair=$KEY_PAIR_NAME"
 oc delete -f deploy/namespace.yaml
```
PratikMahajan, line 65: if the test fails, it crashed the script and fails to delete the namespace. We can catch the error and then delete the namespace.

aravindhp: That needs to fixed in a separate PR and not in this one. While fixing that we also need to only create the namespace if it doesn't already exist to help with local development.

README.md
ravisantoshgudimetla: Added a line a below.

hack/run-ci-e2e-test.sh
```diff
@@ -7,21 +7,61 @@ set -o pipefail
...
+done
+
 # Copy the cloud credentials and KUBESSH key path so that they can be used by operator
 cp $AWS_SHARED_CREDENTIALS_FILE /etc/cloud/credentials
 cp $KUBE_SSH_KEY_PATH /etc/private-key/private-key.pem
 
-# Get operator-sdk binary.
-# TODO: Make this to download the same version we have in go dependencies in gomod
-wget -O /tmp/operator-sdk https://github.com/operator-framework/operator-sdk/releases/download/v0.15.2/operator-sdk-v0.15.2-x86_64-linux-gnu && chmod +x /tmp/operator-sdk
+# Download the operator-sdk binary only if not is not already available
```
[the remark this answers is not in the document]
aravindhp: You have not addressed this.

hack/run-ci-e2e-test.sh
```diff
@@ -7,21 +7,61 @@ set -o pipefail
...
+NODE_COUNT=${NODE_COUNT:-1}
+SKIP_NODE_DELETION=${SKIP_NODE_DELETION:-"-skip-node-deletion=false"}
+KEY_PAIR_NAME=${KEY_PAIR_NAME:-"libra"}
+
 cd $WMCO_ROOT
-oc create -f deploy/namespace.yaml 
-/tmp/operator-sdk test local ./test/e2e --up-local --namespace=windows-machine-config-operator --go-test-flags "-v -timeout 60m"
+oc create -f deploy/namespace.yaml
+# The bool flags in golang does not respect key value pattern. They follow -flag=x pattern.
+# -flag x is only allowed for non-boolean flags only(https://golang.org/pkg/flag/)
```
mansikulkarni96, line 63: extra only in this line.

PR #36, opened 2020-04-07
Added worker label and windows taint for windows node

PratikMahajan:
Refactor:
This PR refactors nodeConfig and adds node object to nodeConfig
struct. It also includes validation for windowsVM and node object.
Also, moved WindowsOSLabel to nodeconfig.go

WMCO:
This PR adds a worker label for windows node
and adds e2e test to check worker label on windows node

Also, adds e2e test to check presence of windows taint on windows node

### DIFF
--- a/pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
+++ b/pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
@@ -27,6 +27,8 @@ const (
 	HybridOverlayMac = "k8s.ovn.org/hybrid-overlay-distributed-router-gateway-mac"
 	// WindowsOSLabel is the label that is applied by WMCB to identify the Windows nodes bootstrapped via WMCB
 	WindowsOSLabel = "node.openshift.io/os_id=Windows"
+	// WorkerLabel is the label that needs to be applied to the Windows node to make it worker node
+	WorkerLabel = "node-role.kubernetes.io/worker"
 )
 
 // nodeConfig holds the information to make the given VM a kubernetes node. As of now, it holds the information
@@ -68,7 +70,10 @@ func (nc *nodeConfig) Configure() error {
 		return errors.Wrap(err, fmt.Sprintf("error getting node object for VM %s",
 			nc.GetCredentials().GetInstanceId()))
 	}
-
+	// Apply worker labels
+	if err := nc.applyWorkerLabel(); err != nil {
+		return errors.Wrap(err, "failed applying worker label")
+	}
 	// Now that basic kubelet configuration is complete, configure networking in the node
 	if err := nc.configureNetwork(); err != nil {
 		return errors.Wrap(err, "configuring node network failed")
@@ -101,6 +106,21 @@ func (nc *nodeConfig) configureNetwork() error {
 		return errors.Wrap(err, fmt.Sprintf("error waiting for %s node annotation for %s", HybridOverlayMac,
 			nc.node.GetName()))
 	}
+	return nil
+}
+
+// applyWorkerLabel applies the worker label to the Windows Node we created.
+func (nc *nodeConfig) applyWorkerLabel() error {
+	if _, found := nc.node.Labels[WorkerLabel]; found {
+		log.V(1).Info("worker label %s already present on node %s", WorkerLabel, nc.node.GetName())
+		return nil
+	}
+	nc.node.Labels[WorkerLabel] = ""
+	node, err := nc.k8sclientset.CoreV1().Nodes().Update(nc.node)
+	if err != nil {
+		return errors.Wrap(err, "error applying worker label to node object")
+	}
+	nc.node = node
 	return nil
 }
 
--- a/pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
+++ b/pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
@@ -25,6 +25,8 @@ const (
 	HybridOverlaySubnet = "k8s.ovn.org/hybrid-overlay-node-subnet"
 	// HybridOverlayMac is an annotation applied by the hybrid-overlay
 	HybridOverlayMac = "k8s.ovn.org/hybrid-overlay-distributed-router-gateway-mac"
+	// WindowsOSLabel is the label that is applied by WMCB to identify the Windows nodes bootstrapped via WMCB
+	WindowsOSLabel = "node.openshift.io/os_id=Windows"
 )
 
 // nodeConfig holds the information to make the given VM a kubernetes node. As of now, it holds the information
@@ -34,6 +36,8 @@ type nodeConfig struct {
 	k8sclientset *kubernetes.Clientset
 	// Windows holds the information related to the windows VM
 	*windows.Windows
+	// Node holds the information related to node object
+	node *v1.Node
 }
 
 // NewNodeConfig creates a new instance of nodeConfig to be used by the caller.
@@ -45,11 +49,24 @@ var log = logf.Log.WithName("nodeconfig")
 
 // Configure configures the Windows VM to make it a Windows worker node
 func (nc *nodeConfig) Configure() error {
+	var err error
+	// Validate WindowsVM
+	if err := nc.Windows.Validate(); err != nil {
+		return errors.Wrap(err, "error validating Windows VM")
+	}
+
 	if err := nc.Windows.Configure(); err != nil {
 		return errors.Wrap(err, "configuring the Windows VM failed")
 	}
 	if err := nc.handleCSRs(); err != nil {
 		return errors.Wrap(err, "handling CSR for the given node failed")
+	}
+
+	// populate node object in nodeConfig
+	err = nc.getNode()
+	if err != nil {
+		return errors.Wrap(err, fmt.Sprintf("error getting node object for VM %s",
+			nc.GetCredentials().GetInstanceId()))
 	}
 
 	// Now that basic kubelet configuration is complete, configure networking in the node
@@ -61,43 +78,29 @@ func (nc *nodeConfig) Configure() error {
 }
 
 // configureNetwork configures k8s networking in the node
+// we are assuming that the WindowsVM and node objects are valid
 func (nc *nodeConfig) configureNetwork() error {
-	if nc.GetCredentials() == nil {
-		return fmt.Errorf("nil credentials for VM")
-	}
-
-	if nc.GetCredentials().GetIPAddress() == "" {
-		return fmt.Errorf("empty IP for VM: %v", nc.GetCredentials())
-	}
-
-	node, err := nc.getNode(nc.GetCredentials().GetIPAddress())
-	if err != nil {
-		return errors.Wrap(err, fmt.Sprintf("error getting node object for VM %s",
-			nc.GetCredentials().GetInstanceId()))
-	}
-
 	// Wait until the node object has the hybrid overlay subnet annotation. Otherwise the hybrid-overlay will fail to
 	// start
-	if err = nc.waitForNodeAnnotation(node.GetName(), HybridOverlaySubnet); err != nil {
+	if err := nc.waitForNodeAnnotation(HybridOverlaySubnet); err != nil {
 		return errors.Wrap(err, fmt.Sprintf("error waiting for %s node annotation for %s", HybridOverlaySubnet,
-			node.GetName()))
+			nc.node.GetName()))
 	}
 
 	// NOTE: Investigate if we need to introduce a interface wrt to the VM's networking configuration. This will
 	// become more clear with the outcome of https://issues.redhat.com/browse/WINC-343
 
 	// Configure the hybrid overlay in the Windows VM
-	if err = nc.Windows.ConfigureHybridOverlay(node.GetName()); err != nil {
-		return errors.Wrap(err, fmt.Sprintf("error configuring hybrid overlay for %s", node.GetName()))
+	if err := nc.Windows.ConfigureHybridOverlay(nc.node.GetName()); err != nil {
+		return errors.Wrap(err, fmt.Sprintf("error configuring hybrid overlay for %s", nc.node.GetName()))
 	}
 
 	// Wait until the node object has the hybrid overlay MAC annotation. This is required for the CNI configuration to
 	// start.
-	if err = nc.waitForNodeAnnotation(node.GetName(), HybridOverlayMac); err != nil {
+	if err := nc.waitForNodeAnnotation(HybridOverlayMac); err != nil {
 		return errors.Wrap(err, fmt.Sprintf("error waiting for %s node annotation for %s", HybridOverlayMac,
-			node.GetName()))
+			nc.node.GetName()))
 	}
-
 	return nil
 }
 
@@ -209,39 +212,30 @@ func (nc *nodeConfig) handleCSR(requestorFilter string) error {
 	return nil
 }
 
-// GetNode returns a pointer to the node object associated with the external IP provided
-func (nc *nodeConfig) getNode(externalIP string) (*v1.Node, error) {
-	var matchedNode *v1.Node
-
-	nodes, err := nc.k8sclientset.CoreV1().Nodes().List(metav1.ListOptions{})
+// getNode returns a pointer to the node object associated with the instance id provided
+func (nc *nodeConfig) getNode() error {
+	nodes, err := nc.k8sclientset.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: WindowsOSLabel})
 	if err != nil {
-		return nil, errors.Wrap(err, "could not get list of nodes")
+		return errors.Wrap(err, "could not get list of nodes")
 	}
 	if len(nodes.Items) == 0 {
-		return nil, fmt.Errorf("no nodes found")
+		return fmt.Errorf("no nodes found")
 	}
-
-	// Find the node that has the given IP
+	// get the node with given instance id
+	instanceID := nc.GetCredentials().GetInstanceId()
 	for _, node := range nodes.Items {
-		for _, address := range node.Status.Addresses {
-			if address.Type == "ExternalIP" && address.Address == externalIP {
-				matchedNode = &node
-				break
-			}
-		}
-		if matchedNode != nil {
-			break
+		if instanceID == getInstanceIDfromProviderID(node.Spec.ProviderID) {
+			nc.node = &node
+			return nil
 		}
 	}
-	if matchedNode == nil {
-		return nil, fmt.Errorf("could not find node with IP: %s", externalIP)
-	}
-	return matchedNode, nil
+	return errors.Errorf("unable to find node for instanceID %s", instanceID)
 }
 
 // waitForNodeAnnotation checks if the node object has the given annotation and waits for retry.Interval seconds and
 // returns an error if the annotation does not appear in that time frame.
-func (nc *nodeConfig) waitForNodeAnnotation(nodeName, annotation string) error {
+func (nc *nodeConfig) waitForNodeAnnotation(annotation string) error {
+	nodeName := nc.node.GetName()
 	var found bool
 	err := wait.Poll(retry.Interval, retry.Timeout, func() (bool, error) {
 		node, err := nc.k8sclientset.CoreV1().Nodes().Get(nodeName, metav1.GetOptions{})
@@ -250,6 +244,8 @@ func (nc *nodeConfig) waitForNodeAnnotation(nodeName, annotation string) error {
 		}
 		_, found := node.Annotations[annotation]
 		if found {
+			//update node to avoid staleness
+			nc.node = node
 			return true, nil
 		}
 		return false, nil
@@ -259,4 +255,11 @@ func (nc *nodeConfig) waitForNodeAnnotation(nodeName, annotation string) error {
 		return errors.Wrap(err, fmt.Sprintf("timeout waiting for %s node annotation", annotation))
 	}
 	return nil
+}
+
+// getInstanceIDfromProviderID gets the instanceID of VM for a given cloud provider ID
+// Ex: aws:///us-east-1e/i-078285fdadccb2eaa. We always want the last entry which is the instanceID
+func getInstanceIDfromProviderID(providerID string) string {
+	providerTokens := strings.Split(providerID, "/")
+	return providerTokens[len(providerTokens)-1]
 }
--- a/pkg/controller/windowsmachineconfig/windows/windows.go
+++ b/pkg/controller/windowsmachineconfig/windows/windows.go
@@ -51,6 +51,20 @@ func (vm *Windows) Configure() error {
 	return vm.runBootstrapper()
 }
 
+// validate the WindowsVM node object
+func (vm *Windows) Validate() error {
+	if vm.GetCredentials() == nil {
+		return fmt.Errorf("nil credentials for VM")
+	}
+	if vm.GetCredentials().GetIPAddress() == "" {
+		return fmt.Errorf("empty IP for VM: %v", vm.GetCredentials())
+	}
+	if vm.GetCredentials().GetInstanceId() == "" {
+		return fmt.Errorf("empty instance id for VM: %v", vm.GetCredentials())
+	}
+	return nil
+}
+
 // runBootstrapper copies the bootstrapper and runs the code on the remote Windows VM
 func (vm *Windows) runBootstrapper() error {
 	if err := vm.CopyFile(wkl.WmcbPath, remoteDir); err != nil {
--- a/test/e2e/create_test.go
+++ b/test/e2e/create_test.go
@@ -35,6 +35,7 @@ func creationTestSuite(t *testing.T) {
 	t.Run("Secrets validation", func(t *testing.T) { testValidateSecrets(t) })
 	t.Run("Network validation", testNetwork)
 	t.Run("Label validation", func(t *testing.T) { testWorkerLabel(t) })
+	t.Run("NodeTaint validation", func(t *testing.T) { testNodeTaint(t) })
 }
 
 // testWindowsNodeCreation tests the Windows node creation in the cluster
--- a/test/e2e/create_test.go
+++ b/test/e2e/create_test.go
@@ -34,6 +34,7 @@ func creationTestSuite(t *testing.T) {
 	t.Run("ConfigMap validation", func(t *testing.T) { testConfigMapValidation(t) })
 	t.Run("Secrets validation", func(t *testing.T) { testValidateSecrets(t) })
 	t.Run("Network validation", testNetwork)
+	t.Run("Label validation", func(t *testing.T) { testWorkerLabel(t) })
 }
 
 // testWindowsNodeCreation tests the Windows node creation in the cluster
--- a/test/e2e/create_test.go
+++ b/test/e2e/create_test.go
@@ -7,7 +7,6 @@ import (
 	"time"
 
 	operator "github.com/openshift/windows-machine-config-operator/pkg/apis/wmc/v1alpha1"
-	wmc "github.com/openshift/windows-machine-config-operator/pkg/controller/windowsmachineconfig"
 	"github.com/openshift/windows-machine-config-operator/pkg/controller/windowsmachineconfig/nodeconfig"
 	framework "github.com/operator-framework/operator-sdk/pkg/test"
 	"github.com/stretchr/testify/require"
@@ -77,7 +76,7 @@ func (tc *testContext) waitForWindowsNode() error {
 	// As per testing, each windows VM is taking roughly 12 minutes to be shown up in the cluster, so to be on safe
 	// side, let's make it as 20 minutes per node. The value comes from nodeCreationTime variable
 	err := wait.Poll(nodeRetryInterval, time.Duration(gc.numberOfNodes)*nodeCreationTime, func() (done bool, err error) {
-		nodes, err = tc.kubeclient.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: wmc.WindowsOSLabel})
+		nodes, err = tc.kubeclient.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: nodeconfig.WindowsOSLabel})
 		if err != nil {
 			if apierrors.IsNotFound(err) {
 				log.Printf("waiting for %d Windows nodes", gc.numberOfNodes)
--- a/test/e2e/validation_test.go
+++ b/test/e2e/validation_test.go
@@ -284,3 +284,25 @@ func testWorkerLabel(t *testing.T) {
 		assert.Contains(t, node.Labels, nc.WorkerLabel, "expected node label %s was not present on %s", nc.WorkerLabel, node.GetName())
 	}
 }
+
+// testNodeTaint tests if the Windows node has the Windows taint
+func testNodeTaint(t *testing.T) {
+	// windowsTaint is the taint that needs to be applied to the Windows node
+	windowsTaint := corev1.Taint{
+		Key:    "os",
+		Value:  "Windows",
+		Effect: corev1.TaintEffectNoSchedule,
+	}
+
+	for _, node := range gc.nodes {
+		hasTaint := func() bool {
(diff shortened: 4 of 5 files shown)


pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
```diff
@@ -48,6 +51,53 @@ func (nc *nodeConfig) Configure() error {
...
+		if instanceID == GetInstanceID(node.Spec.ProviderID) {
+			return &node, nil
+		}
+	}
+	return nil, errors.Errorf("unable to find node for instanceID %s", instanceID)
+}
+
+// GetInstanceID gets the instanceID of VM for a given cloud provider ID
+// Ex: aws:///us-east-1e/i-078285fdadccb2eaa. We always want the last entry which is the instanceID
+func GetInstanceID(providerID string) string {
```
ravisantoshgudimetla, line 90: You already have this method in test suite. I made it public to be shared with the test suite, is that why you made it public while copying too?

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
```diff
@@ -48,6 +51,53 @@ func (nc *nodeConfig) Configure() error {
...
+
+	// Apply worker labels
+	if err := nc.applyWorkerLabel(node); err != nil {
+		return errors.Wrap(err, "failed applying worker label")
+	}
+	return nil
+}
+
+func (nc *nodeConfig) getNodeObjectAssociated(instanceID string) (*v1.Node, error) {
+	nodes, err := nc.k8sclientset.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: "node.openshift.io/os_id=Windows"})
```
ravisantoshgudimetla, line 76: Why are you have `node.openshift.io/os_id` as string? Are you having a cyclic dependency causing this problem?

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
PratikMahajan: should i remove the one in test suite and use this one instead ?

ravisantoshgudimetla: You should, if you want to re-use it there unless @aravindhp has any concerns.

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
```diff
@@ -48,6 +51,53 @@ func (nc *nodeConfig) Configure() error {
...
+	// As the CSR has been approved get the Kubernetes node object associated
+	// the Windows VM created
+	if nc.Windows.GetCredentials() != nil && len(nc.Windows.GetCredentials().GetInstanceId()) > 0 {
+		instanceID = nc.Windows.GetCredentials().GetInstanceId()
+	}
+
+	// Get the node object associated with node
+	node, err := nc.getNodeObjectAssociated(instanceID)
+	if err != nil {
+		return errors.Wrap(err, "node object associated with the node")
```
aravindhp, line 65: Please fix the error message.

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
```diff
@@ -48,6 +51,53 @@ func (nc *nodeConfig) Configure() error {
...
 		return errors.Wrap(err, "handling CSR for the given node failed")
 	}
+
+	var instanceID string
+	// As the CSR has been approved get the Kubernetes node object associated
+	// the Windows VM created
+	if nc.Windows.GetCredentials() != nil && len(nc.Windows.GetCredentials().GetInstanceId()) > 0 {
+		instanceID = nc.Windows.GetCredentials().GetInstanceId()
+	}
+
```
aravindhp, line 61: So what happens if you are unable to get an `instanceID`? You need a separate error check for that.

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
```diff
@@ -48,6 +51,53 @@ func (nc *nodeConfig) Configure() error {
 	if err := nc.handleCSRs(); err != nil {
 		return errors.Wrap(err, "handling CSR for the given node failed")
 	}
+
+	var instanceID string
+	// As the CSR has been approved get the Kubernetes node object associated
+	// the Windows VM created
+	if nc.Windows.GetCredentials() != nil && len(nc.Windows.GetCredentials().GetInstanceId()) > 0 {
```
aravindhp, line 58: Can't you check for `nc.Windows.GetCredentials().GetInstanceId() != ""` instead of the `len`?

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
aravindhp: I prefer not sharing test and main code.

test/e2e/validation_test.go
```diff
@@ -273,3 +274,15 @@ func createWindowsMachineConfig(namespace string, isReplicasFieldRequired bool,
...
 	return wmc
 }
+
+// testWorkerLabel tests if the worker label has been applied properly
+func testWorkerLabel(t *testing.T, nodeCount int) {
+	testCtx := framework.NewTestCtx(t)
+	defer testCtx.Cleanup()
+	nodes, err := framework.Global.KubeClient.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: wmc.WindowsOSLabel})
+	require.NoError(t, err, "error querying the nodes")
+	require.Equal(t, len(nodes.Items), nodeCount, "mismatched node count")
```
aravindhp, line 284: @ravisantoshgudimetla I see why you want to add `nodes` to the global context. I will add it and hopefully we don't run into any staleness issues.

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
PratikMahajan: yes

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
aravindhp: You need to handle all the error cases. If the credentials are nil this will fall through. Please rewrite your check above to return an error if credentials are nil or instanceID is blank.

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
aravindhp: Then we need to find a better place for the constant. Adding it as a string is not acceptable.

@ravisantoshgudimetla we might have to move away from the sub-packages to make our lives easier.

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
```diff
@@ -48,6 +51,56 @@ func (nc *nodeConfig) Configure() error {
...
+func getInstanceID(providerID string) string {
+	providerTokens := strings.Split(providerID, "/")
+	return providerTokens[len(providerTokens)-1]
+}
+
+// applyWorkerLabel applies the worker label to the Windows Node we created.
+func (nc *nodeConfig) applyWorkerLabel(node *v1.Node) error {
+	node.Labels[WorkerLabel] = ""
+	if _, err := nc.k8sclientset.CoreV1().Nodes().Update(node); err != nil {
+		return errors.Wrap(err, "error updating node object")
```
aravindhp, line 102: Please make this error message more specific.

test/e2e/create_test.go
```diff
@@ -27,6 +27,7 @@ func creationTestSuite(t *testing.T) {
 	t.Run("ConfigMap validation", func(t *testing.T) { testConfigMapValidation(t) })
 	t.Run("Secrets validation", func(t *testing.T) { testValidateSecrets(t) })
 	t.Run("Label validation", func(t *testing.T) { testWorkerLabel(t, numberOfNodes) })
+	t.Run("WindowsTaint validation", func(t *testing.T) { testWindowsTaint(t) })
```
aravindhp, line 30: Please make this `Node taint validation`

test/e2e/validation_test.go
```diff
@@ -286,3 +286,26 @@ func testWorkerLabel(t *testing.T, nodeCount int) {
 		assert.Contains(t, node.Labels, nc.WorkerLabel, "expected node label to be present")
 	}
 }
+
+// testWindowsTaint tests if the Windows node has the Windows taint
+func testWindowsTaint(t *testing.T) {
```
aravindhp, line 291: `testNodeTaint`

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
```diff
@@ -48,6 +51,56 @@ func (nc *nodeConfig) Configure() error {
...
+		if instanceID == getInstanceID(node.Spec.ProviderID) {
+			return &node, nil
+		}
+	}
+	return nil, errors.Errorf("unable to find node for instanceID %s", instanceID)
+}
+
+// getInstanceID gets the instanceID of VM for a given cloud provider ID
+// Ex: aws:///us-east-1e/i-078285fdadccb2eaa. We always want the last entry which is the instanceID
+func getInstanceID(providerID string) string {
```
ravisantoshgudimetla, line 93: nit: I want to explain a pattern that I've noticed wrt testing, if you are writing functionality for A and B. TestA and TestB represent testing functions associated with A and B.

A -> TestA should not share any code with A, meaning don't call the very method you are testing to validate.
B -> TestB should not share any code with B, same as above

TestA -> can have methods that are part of B. This will be case where you already have functionality for B and you have to duplicate a lot of code in TestA unless you use it from B, so this ok if the method is already public. In this case, it would have been ok say if `GetInstanceID` is already public because it is needed by another package.

test/e2e/validation_test.go
```diff
@@ -273,3 +274,38 @@ func createWindowsMachineConfig(namespace string, isReplicasFieldRequired bool,
...
+
+// testNodeTaint tests if the Windows node has the Windows taint
+func testNodeTaint(t *testing.T) {
+	// windowsTaint is the taint that needs to be applied to the Windows node
+	windowsTaint := corev1.Taint{
+		Key:    "os",
+		Value:  "Windows",
+		Effect: corev1.TaintEffectNoSchedule,
+	}
+	nodes, err := framework.Global.KubeClient.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: wmc.WindowsOSLabel})
```
ravisantoshgudimetla, line 298: You could have just setup the context which might already have the information related to nodes.

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
PratikMahajan: Kept it as private as we are currently not using it outside package. We can always change it to public when needed.

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
```diff
@@ -209,10 +230,9 @@ func (nc *nodeConfig) handleCSR(requestorFilter string) error {
 	return nil
 }
 
-// GetNode returns a pointer to the node object associated with the external IP provided
-func (nc *nodeConfig) getNode(externalIP string) (*v1.Node, error) {
-	var matchedNode *v1.Node
-
+// getNode returns a pointer to the node object associated with the external IP provided
+func (nc *nodeConfig) getNode() (*v1.Node, error) {
```
ravisantoshgudimetla, line 248: This doesn't need to return anything, you can directly populate nc.node here in this method.

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
```diff
@@ -103,7 +109,22 @@ func (nc *nodeConfig) configureNetwork() error {
...
+	providerTokens := strings.Split(providerID, "/")
+	return providerTokens[len(providerTokens)-1]
+}
 
+// applyWorkerLabel applies the worker label to the Windows Node we created.
+func (nc *nodeConfig) applyWorkerLabel() error {
+	nc.node.Labels[WorkerLabel] = ""
+	if _, err := nc.k8sclientset.CoreV1().Nodes().Update(nc.node); err != nil {
+		return errors.Wrap(err, "error applying worker label to node object")
+	}
```
aravindhp, line 127: Don't you need to update `nc.node` after the `Update` call with a `Get`? Otherwise it can be stale.

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
```diff
@@ -115,13 +136,6 @@ func (nc *nodeConfig) validateNode() error {
 	return nil
 }
 
-// getInstanceID gets the instanceID of VM for a given cloud provider ID
-// Ex: aws:///us-east-1e/i-078285fdadccb2eaa. We always want the last entry which is the instanceID
-func getInstanceID(providerID string) string {
-	providerTokens := strings.Split(providerID, "/")
-	return providerTokens[len(providerTokens)-1]
-}
-
```
aravindhp: Please don't move code around across commits. Put in the location where you want and leave it there. You are causing unnecessary code churn and forcing the reviewer to look at the same code multiple times.

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
```diff
@@ -221,34 +241,25 @@ func (nc *nodeConfig) getNode(externalIP string) (*v1.Node, error) {
...
 }
 
 // waitForNodeAnnotation checks if the node object has the given annotation and waits for retry.Interval seconds and
 // returns an error if the annotation does not appear in that time frame.
 func (nc *nodeConfig) waitForNodeAnnotation(nodeName, annotation string) error {
 	var found bool
 	err := wait.Poll(retry.Interval, retry.Timeout, func() (bool, error) {
-		node, err := nc.k8sclientset.CoreV1().Nodes().Get(nodeName, metav1.GetOptions{})
+		var err error
+		nc.node, err = nc.k8sclientset.CoreV1().Nodes().Get(nodeName, metav1.GetOptions{})
```
ravisantoshgudimetla, line 272: You're updating the nc.node directly here, you should wait till you get the node object, imagine an intermediate state where `Get` returns nil and you've now nc.node is nil, if there is another method which tries to access the nodeObject in meanwhile it will be problem, so update the nc.node only when you're sure about it.

test/e2e/validation_test.go
```diff
@@ -275,3 +275,15 @@ func createWindowsMachineConfig(namespace string, isReplicasFieldRequired bool,
 	}
 	return wmc
 }
+
+// testWorkerLabel tests if the worker label has been applied properly
+func testWorkerLabel(t *testing.T, nodeCount int) {
+	testCtx := framework.NewTestCtx(t)
+	defer testCtx.Cleanup()
+	nodes, err := framework.Global.KubeClient.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: nc.WindowsOSLabel})
```
aravindhp, line 283: Why aren't you using the the nodes in the global context?

test/e2e/validation_test.go
```diff
@@ -275,3 +275,15 @@ func createWindowsMachineConfig(namespace string, isReplicasFieldRequired bool,
...
+	require.NoError(t, err, "error querying the nodes")
+	require.Equal(t, len(nodes.Items), nodeCount, "mismatched node count")
```
aravindhp, line 285: This is already being tested elsewhere. Why are you repeating the test?

test/e2e/validation_test.go
```diff
@@ -287,3 +287,26 @@ func testWorkerLabel(t *testing.T, nodeCount int) {
...
+
+// testNodeTaint tests if the Windows node has the Windows taint
+func testNodeTaint(t *testing.T) {
+	// windowsTaint is the taint that needs to be applied to the Windows node
+	windowsTaint := corev1.Taint{
+		Key:    "os",
+		Value:  "Windows",
+		Effect: corev1.TaintEffectNoSchedule,
+	}
+	nodes, err := framework.Global.KubeClient.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: nc.WindowsOSLabel})
```
aravindhp, line 299: Again, why aren't you using the nodes in the gloabal context?

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
```diff
@@ -45,59 +51,88 @@ var log = logf.Log.WithName("nodeconfig")
...
+// getInstanceID gets the instanceID of VM for a given cloud provider ID
+// Ex: aws:///us-east-1e/i-078285fdadccb2eaa. We always want the last entry which is the instanceID
+func getInstanceID(providerID string) string {
+	providerTokens := strings.Split(providerID, "/")
+	return providerTokens[len(providerTokens)-1]
+}
+
+// applyWorkerLabel applies the worker label to the Windows Node we created.
+func (nc *nodeConfig) applyWorkerLabel() error {
+	nc.node.Labels[WorkerLabel] = ""
```
ravisantoshgudimetla, line 124: Just to be on the safe side, have a guard if the label is already present and continue..

test/e2e/validation_test.go
```diff
@@ -287,3 +287,26 @@ func testWorkerLabel(t *testing.T, nodeCount int) {
...
+	require.NoError(t, err, "error querying the nodes")
+	hasTaint := func() bool {
```
aravindhp, line 301: I would like to see cases where one node having the taint and another not having it being clearly called. This way of coding hides that.

test/e2e/validation_test.go
```diff
@@ -287,3 +287,26 @@ func testWorkerLabel(t *testing.T, nodeCount int) {
...
+		for _, node := range nodes.Items {
+			for _, taint := range node.Spec.Taints {
+				if taint.Key == windowsTaint.Key && taint.Value == windowsTaint.Value && taint.Effect == windowsTaint.Effect {
+					return true
+				}
+			}
+		}
+		return false
+	}()
+	assert.Equal(t, hasTaint, true, "expected Windows Taint to be present on the Windows Node")
```
aravindhp, line 311: You are assuming a single Windows node in your message.

test/e2e/validation_test.go
```diff
@@ -275,3 +275,38 @@ func createWindowsMachineConfig(namespace string, isReplicasFieldRequired bool,
 	}
 	return wmc
 }
+
+// testWorkerLabel tests if the worker label has been applied properly
+func testWorkerLabel(t *testing.T, nodeCount int) {
+	testCtx := framework.NewTestCtx(t)
+	defer testCtx.Cleanup()
+	nodes, err := framework.Global.KubeClient.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: nc.WindowsOSLabel})
```
ravisantoshgudimetla, line 283: You can use kubeclient from testCtx. And you can get nodes from the global context `gc`

test/e2e/validation_test.go
```diff
@@ -275,3 +275,38 @@ func createWindowsMachineConfig(namespace string, isReplicasFieldRequired bool,
...
+
+// testNodeTaint tests if the Windows node has the Windows taint
+func testNodeTaint(t *testing.T) {
+	// windowsTaint is the taint that needs to be applied to the Windows node
+	windowsTaint := corev1.Taint{
+		Key:    "os",
+		Value:  "Windows",
+		Effect: corev1.TaintEffectNoSchedule,
+	}
+	nodes, err := framework.Global.KubeClient.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: nc.WindowsOSLabel})
```
ravisantoshgudimetla, line 299: Same as above.

test/e2e/validation_test.go
```diff
@@ -275,3 +275,38 @@ func createWindowsMachineConfig(namespace string, isReplicasFieldRequired bool,
...
+	require.NoError(t, err, "error querying the nodes")
+	hasTaint := func() bool {
+		for _, node := range nodes.Items {
```
ravisantoshgudimetla, line 302: You're are returning if any of the nodes have this taint, you should ensure all the windows nodes have taints.

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
```diff
@@ -45,59 +51,88 @@ var log = logf.Log.WithName("nodeconfig")
...
+			nc.node.GetName()))
 	}
 
 	// NOTE: Investigate if we need to introduce a interface wrt to the VM's networking configuration. This will
 	// become more clear with the outcome of https://issues.redhat.com/browse/WINC-343
 
 	// Configure the hybrid overlay in the Windows VM
-	if err = nc.Windows.ConfigureHybridOverlay(node.GetName()); err != nil {
-		return errors.Wrap(err, fmt.Sprintf("error configuring hybrid overlay for %s", node.GetName()))
+	if err := nc.Windows.ConfigureHybridOverlay(nc.node.GetName()); err != nil {
```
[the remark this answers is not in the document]
PratikMahajan: I think we had discussion on this. As we have circular imports, we decided on passing the argument for now.

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
```diff
@@ -76,6 +78,10 @@ func (nc *nodeConfig) Configure() error {
 	if err := nc.configureNetwork(); err != nil {
 		return errors.Wrap(err, "configuring node network failed")
 	}
+	// Apply worker labels
+	if err := nc.applyWorkerLabel(); err != nil {
+		return errors.Wrap(err, "failed applying worker label")
+	}
```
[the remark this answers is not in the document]
PratikMahajan: i had a discussion with @ravisantoshgudimetla  regarding this. We decided on applying the worker label after network config is complete. We can change it to apply before network config to keep it in sync with what we are doing in WSU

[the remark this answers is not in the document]
aravindhp: >We decided on applying the worker label after network config is complete. We can change it to apply before network config to keep it in sync with what we are doing in WSU

You statements are contradictory. What approach are you going with?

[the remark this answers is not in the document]
PratikMahajan: i have changed it to before network config

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
```diff
@@ -45,13 +49,26 @@ var log = logf.Log.WithName("nodeconfig")
 
 // Configure configures the Windows VM to make it a Windows worker node
 func (nc *nodeConfig) Configure() error {
+	var err error
+	// Validate WindowsVM
+	if err := nc.Windows.Validate(); err != nil {
+		return errors.Wrap(err, "Unable to validate Windows VM")
```
aravindhp, line 57: `error validating Windows VM`

The pattern is to not start messages with capitals.

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
```diff
@@ -61,46 +78,52 @@ func (nc *nodeConfig) Configure() error {
...
+
+	for _, node := range nodes.Items {
+		if instanceID == getInstanceIDfromProviderID(node.Spec.ProviderID) {
+			nc.node = &node
+			return nil
+		}
+	}
+	return errors.Errorf("unable to find node for instanceID %s", instanceID)
+}
+
```
aravindhp, line 146: Why are you moving code around unnecessarily?

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
```diff
@@ -209,39 +232,10 @@ func (nc *nodeConfig) handleCSR(requestorFilter string) error {
...
-		if matchedNode != nil {
-			break
-		}
-	}
-	if matchedNode == nil {
-		return nil, fmt.Errorf("could not find node with IP: %s", externalIP)
-	}
-	return matchedNode, nil
-}
-
```
aravindhp, line 241: Please make the `getNode` changes here.

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
```diff
@@ -250,6 +244,7 @@ func (nc *nodeConfig) waitForNodeAnnotation(nodeName, annotation string) error {
 		}
 		_, found := node.Annotations[annotation]
 		if found {
+			nc.node = node
```
aravindhp, line 267: Please add a comment saying why you are updating the `nc.node`.

pkg/controller/windowsmachineconfig/windows/windows.go
```diff
@@ -51,6 +51,17 @@ func (vm *Windows) Configure() error {
 	return vm.runBootstrapper()
 }
 
+// validate the WindowsVM node object
+func (vm *Windows) Validate() error {
+	if vm.GetCredentials() == nil {
```
aravindhp, line 56: You are using `vm.GetCredentials().GetInstanceId()` in multiple places without validating it here. Please add that.

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
```diff
@@ -104,6 +109,21 @@ func (nc *nodeConfig) configureNetwork() error {
 	return nil
 }
 
+// applyWorkerLabel applies the worker label to the Windows Node we created.
+func (nc *nodeConfig) applyWorkerLabel() error {
+	if _, found := nc.node.Labels[WorkerLabel]; found {
+		log.Info("Worker label already present")
```
aravindhp, line 115: Please reference the worker label as defined in the message. I am also not sure if it needs to be `Info`.

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
aravindhp: I am assuming the `Update()` call returns the latest object

test/e2e/validation_test.go
```diff
@@ -275,3 +275,12 @@ func createWindowsMachineConfig(namespace string, isReplicasFieldRequired bool,
 	}
 	return wmc
 }
+
+// testWorkerLabel tests if the worker label has been applied properly
+func testWorkerLabel(t *testing.T) {
+	testCtx := framework.NewTestCtx(t)
+	defer testCtx.Cleanup()
+	for _, node := range gc.nodes {
+		assert.Contains(t, node.Labels, nc.WorkerLabel, "expected node label to be present")
```
aravindhp, line 284: `expected node label was not present`

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
PratikMahajan: wrt https://github.com/openshift/windows-machine-config-operator/pull/36#discussion_r408389472

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
PratikMahajan: wrt https://github.com/openshift/windows-machine-config-operator/pull/36#discussion_r408389472

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
PratikMahajan: was mislead by https://github.com/openshift/windows-machine-config-operator/pull/36#discussion_r408389472

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
PratikMahajan: was misled by https://github.com/openshift/windows-machine-config-operator/pull/36#discussion_r408389472

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
aravindhp: In https://github.com/openshift/windows-machine-config-operator/pull/36#discussion_r408389472 I was asking for the variable `instanceID` to be defined closer to the scope of its usage. It had nothing to do with moving the function. You still need to move the variable closer to the usage scope.

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
PratikMahajan: yes

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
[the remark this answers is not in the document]
ravisantoshgudimetla: Ahh got it. Let's leave it as is for now.

mansikulkarni96:
LGTM

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
```diff
@@ -209,39 +232,30 @@ func (nc *nodeConfig) handleCSR(requestorFilter string) error {
 	return nil
 }
 
-// GetNode returns a pointer to the node object associated with the external IP provided
-func (nc *nodeConfig) getNode(externalIP string) (*v1.Node, error) {
-	var matchedNode *v1.Node
-
+// getNode returns a pointer to the node object associated with the instance id provided
+func (nc *nodeConfig) getNode() error {
 	nodes, err := nc.k8sclientset.CoreV1().Nodes().List(metav1.ListOptions{})
```
ravisantoshgudimetla, line 237: nit: Sorry, it's a miss on my side in earlier reviews but shouldn't you `List` only the Windows nodes? This way will limit the items in our `nodes` object.

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
```diff
@@ -61,43 +83,44 @@ func (nc *nodeConfig) Configure() error {
...
-			node.GetName()))
+			nc.node.GetName()))
 	}
+	return nil
+}
 
+// applyWorkerLabel applies the worker label to the Windows Node we created.
+func (nc *nodeConfig) applyWorkerLabel() error {
+	if _, found := nc.node.Labels[WorkerLabel]; found {
+		log.V(5).Info("worker label %s already present on node %s", WorkerLabel, nc.node.GetName())
```
ravisantoshgudimetla, line 115: nit: Why is this V(5)? @mansikulkarni96 I think you were against this?

test/e2e/validation_test.go
```diff
@@ -275,3 +275,34 @@ func createWindowsMachineConfig(namespace string, isReplicasFieldRequired bool,
 	}
 	return wmc
 }
+
+// testWorkerLabel tests if the worker label has been applied properly
+func testWorkerLabel(t *testing.T) {
+	testCtx := framework.NewTestCtx(t)
+	defer testCtx.Cleanup()
+	for _, node := range gc.nodes {
+		assert.Contains(t, node.Labels, nc.WorkerLabel, "expected node label was not present")
```
ravisantoshgudimetla, line 284: on node `%s`, nodeName?

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
aravindhp: This is not a nit. Please fix.

pkg/controller/windowsmachineconfig/nodeconfig/nodeconfig.go
aravindhp: Please fix and follow whatever standard we have

mansikulkarni96: yes missed this, it is recommended by logr package we use V(1) if we want them to be debug logs instead of higher levels.

PR #43, opened 2020-04-21
[ci] Fix waits in VM creation failure scenarios

aravindhp:
**Problem**
In cases where the VM creation failed and then the scale down tests are executed, there is potential for waitForTrackerConfigMap() and waitForWindowsNode() to wait until the go test time out is reached instead of the expected 10 minutes.

**Solution**
The reason for the issues is the way we created the duration based on the number of nodes. Correct this to account for the scale down to 0 nodes scenarios.

### DIFF
--- a/test/e2e/create_test.go
+++ b/test/e2e/create_test.go
@@ -3,6 +3,7 @@ package e2e
 import (
 	"context"
 	"log"
+	"math"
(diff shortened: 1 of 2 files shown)


mansikulkarni96:
LGTM

PR #53, opened 2020-05-15
Bug 1835166: [wmco] fix validation for replicas

PratikMahajan:
[wmco] remove validation for replicas
fixes validation for replicas field as it was being interpreted as
float while validating.
this was because replicas was defined as `int` in place of `int32`
which is causing the `operator-sdk generate crds` to not create a
`format` field in crd. This is a known issue in
operator-sdk 0.17.

Created [ticket](https://issues.redhat.com/browse/WINC-407) to revert this change 
once we bump operator sdk to 0.18 or above 
operator-sdk [issue](https://github.com/operator-framework/operator-sdk/issues/2952) 

[Bug](https://bugzilla.redhat.com/show_bug.cgi?id=1835166)


mansikulkarni96:
LGTM

VaishnaviHire:
LGTM

PR #62, opened 2020-06-01
[wmco] Bump kube-node binaries to v1.18.3

aravindhp:
Update the download location of the kube-node binaries to v1.18.3.


mansikulkarni96:
LGTM

PR #67, opened 2020-06-08
Bug 1840837: Get reconcile error on wmco reboot

mansikulkarni96:
This PR fixes the wmco reboot error that occurs due to
a stale resource read of the object from client cache for a
subsequent update. This PR replaces the default client
with a non-default-client that does not read from cache and
only updates the status to reconciling when we scale up or down.
Ref: [opertaor-sdk-non-default-client]( https://github.com/operator-framework/operator-sdk/blob/3efa98cb74a8388877724c574473c291677f2a18/website/content/en/docs/golang/references/client.md#non-default-client)

### DIFF
--- a/pkg/controller/windowsmachineconfig/status.go
+++ b/pkg/controller/windowsmachineconfig/status.go
@@ -48,8 +48,11 @@ func (s *StatusManager) updateStatus() error {
 		return errors.Wrapf(err, "could not get %v", s.wmcName)
 	}
 
-	object.Status.JoinedVMCount = s.joinedVMCount
 	for _, condition := range s.conditionsToSet {
+		// update the joined VM count if we are done reconciling
+		if condition.Type == wmcapi.Reconciling && condition.Status == corev1.ConditionFalse {
+			object.Status.JoinedVMCount = s.joinedVMCount
+		}
 		object.Status.SetWindowsMachineConfigCondition(condition)
 	}
 	if (s.degradedCondition != wmcapi.WindowsMachineConfigCondition{}) {
--- a/pkg/controller/windowsmachineconfig/windowsmachineconfig_controller.go
+++ b/pkg/controller/windowsmachineconfig/windowsmachineconfig_controller.go
@@ -17,6 +17,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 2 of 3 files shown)


pkg/controller/windowsmachineconfig/status.go
```diff
@@ -62,6 +62,24 @@ func (s *StatusManager) updateStatus() error {
 	return nil
 }
 
+// updateStatusReconciling updates the status of the WMC to indicate that
+// it is reconciling
+func (s *StatusManager) updateStatusReconciling() error {
+	object := &wmcapi.WindowsMachineConfig{}
+	err := s.client.Get(context.TODO(), s.wmcName, object)
```
VaishnaviHire, line 69: Please use `if err := s.client.Get(context.TODO(), s.wmcName, object); err !=nil` format

pkg/controller/windowsmachineconfig/status.go
```diff
@@ -62,6 +62,24 @@ func (s *StatusManager) updateStatus() error {
 	return nil
 }
 
+// updateStatusReconciling updates the status of the WMC to indicate that
+// it is reconciling
+func (s *StatusManager) updateStatusReconciling() error {
```
ravisantoshgudimetla, line 67: Why to have a separate method for reconciling, why can't it be part of `updateStatus` method?

pkg/controller/windowsmachineconfig/windowsmachineconfig_controller.go
```diff
@@ -47,6 +48,20 @@ func Add(mgr manager.Manager, clusterServiceCIDR string) error {
 func newReconciler(mgr manager.Manager, clusterServiceCIDR string) (reconcile.Reconciler, error) {
 	// TODO: This should be moved out to validation for reconciler struct.
 	// 		Jira story: https://issues.redhat.com/browse/WINC-277
+	// The default client serves read requests from the cache which
+	// could be stale and result in a get call to return an older version
+	// of the object. Hence we are using a non-default-client referenced
+	// by operator-sdk.
+	cfg, err := config.GetConfig()
```
ravisantoshgudimetla, line 55: Why can't this be mgt.GetConfig() and then build all necessary clients?

aravindhp: This is what the [docs](https://sdk.operatorframework.io/docs/golang/references/client/#non-default-client) suggests to do

pkg/controller/windowsmachineconfig/windowsmachineconfig_controller.go
```diff
@@ -192,13 +207,14 @@ func (r *ReconcileWindowsMachineConfig) Reconcile(request reconcile.Request) (re
...
-		// Don't continue reconciling and requeue on status error
-		return reconcile.Result{}, err
+	if currentCountOfWindowsVMs != int(instance.Spec.Replicas) {
+		// Update status to reflect that operator is reconciling
+		r.statusMgr.setStatusConditions([]wmcapi.WindowsMachineConfigCondition{
+			*wmcapi.NewWindowsMachineConfigCondition(wmcapi.Reconciling, corev1.ConditionTrue, "", "")})
+		if err = r.statusMgr.updateStatusReconciling(); err != nil {
+			// Don't continue reconciling and requeue on status error
+			return reconcile.Result{}, err
+		}
```
aravindhp, line 217: Should this be done inside `reconcileWindowsNodes`?

pkg/controller/windowsmachineconfig/status.go
mansikulkarni96: this can be a part of update status method, just wanted to avoid unnecessary setting of status fields for object when they are not set in the struct, but I can add validations in the updateStatus method to set the values only if they are set in the struct

pkg/controller/windowsmachineconfig/windowsmachineconfig_controller.go
mansikulkarni96: could be moved there, kept it here since all status sets are being done in reconcile

pkg/controller/windowsmachineconfig/status.go
```diff
@@ -47,10 +47,13 @@ func (s *StatusManager) updateStatus() error {
 	if err != nil {
 		return errors.Wrapf(err, "could not get %v", s.wmcName)
 	}
-
-	object.Status.JoinedVMCount = s.joinedVMCount
-	for _, condition := range s.conditionsToSet {
-		object.Status.SetWindowsMachineConfigCondition(condition)
+	if s.joinedVMCount != 0 {
+		object.Status.JoinedVMCount = s.joinedVMCount
+	}
```
aravindhp, line 52: Can't the `joinedVMCount` be 0 on a scale down operation?

mansikulkarni96: missed this scenario, updating the joinedVmCount with current value while reconciling the status, in order to avoid setting incorrect values.

PR #84, opened 2020-07-23
[e2e] Fail on creation test failure

aravindhp:
- Don't proceed if the creation test fails
- Add additional guard that checks that the operator is running
- Update get_WMCO_logs with --tail=-1 to ensure we get all logs

### DIFF
--- a/hack/run-ci-e2e-test.sh
+++ b/hack/run-ci-e2e-test.sh
@@ -13,7 +13,7 @@ KEY_PAIR_NAME=""
 export CGO_ENABLED=0
 
 get_WMCO_logs() {
-  oc logs -l name=windows-machine-config-operator -n windows-machine-config-operator
+  oc logs -l name=windows-machine-config-operator -n windows-machine-config-operator --tail=-1
 }
 
 # This function runs operator-sdk test with certain go test arguments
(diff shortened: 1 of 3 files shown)


mansikulkarni96:
LGTM

VaishnaviHire:
LGTM

hack/run-ci-e2e-test.sh
```diff
@@ -74,13 +74,19 @@ KEY_PAIR_NAME=${KEY_PAIR_NAME:-"openshift-dev"}
...
+  exit 1
+fi
 
 # The bool flags in golang does not respect key value pattern. They follow -flag=x pattern.
 # -flag x is allowed for non-boolean flags only(https://golang.org/pkg/flag/)
 # Run the creation tests and skip deletion of the Windows VMs
 OSDK_WMCO_test $OSDK "-run=TestWMCO/create -v -timeout=90m -node-count=$NODE_COUNT -skip-node-deletion -ssh-key-pair=$KEY_PAIR_NAME"
 
+# Get logs on success before running the destroy tests
+get_WMCO_logs
```
sebsoto, line 88: This will result in the logs being gathered twice, once here, and then once on either success or failure of the deletion suite. Since the operator pod is the same between the creation and deletion suites the logs will have duplicate information.

aravindhp: Good point. I added this because I could not figure out why all logs were not being shown but I think the main issue was with `get_WMCO_logs` itself which I fixed by adding the `-tail` option.

PR #88, opened 2020-08-03
[wmco] Add WMCO version to Windows node

sebsoto:
This commit adds the version of the WMCO instance that configured each
node to that node's annotations. This will be used in the WMCO upgrade
process, as this allows WMCO to see which nodes were not configured by
the current version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,13 +63,17 @@ Once the above variables are set, run the following script:
 ```shell script
 hack/run-ci-e2e-test.sh -k "openshift-dev"
 ```
-We assume that the developer uses `openshift-dev` as the key pair in the aws cloud
+We assume that the developer uses `openshift-dev` as the key pair in the aws cloud.
 
 Additional flags that can be passed to `hack/run-ci-e2e-test.sh` are
 - `-s` to skip the deletion of Windows nodes that are created as part of test suite run
 - `-n` to represent the number of Windows nodes to be created for test run
 - `-k` to represent the AWS specific key pair that will be used during e2e run and it should map to the private key
        that we have in `KUBE_SSH_KEY_PATH`. The default value points to `openshift-dev` which we use in our CI
+- `-b` gives an alternative path to the WMCO binary. This option overridden in OpenShift CI.
+       When building the operator locally, the WMCO binary is created as part of the operator image build process and
+       can be found at `build/_output/bin/windows-machine-config-operator`, this is the default value of this option.
+
        
 Example command to spin up 2 Windows nodes and retain them after test run:
 ```
--- a/pkg/controller/windowsmachine/nodeconfig/nodeconfig.go
+++ b/pkg/controller/windowsmachine/nodeconfig/nodeconfig.go
@@ -10,6 +10,7 @@ import (
 	"github.com/openshift/windows-machine-config-operator/pkg/controller/retry"
 	wkl "github.com/openshift/windows-machine-config-operator/pkg/controller/wellknownlocations"
 	"github.com/openshift/windows-machine-config-operator/pkg/controller/windowsmachine/windows"
+	"github.com/openshift/windows-machine-config-operator/version"
 	"github.com/pkg/errors"
 	"golang.org/x/crypto/ssh"
 	"k8s.io/api/core/v1"
@@ -28,6 +29,8 @@ const (
 	WindowsOSLabel = "node.openshift.io/os_id=Windows"
 	// WorkerLabel is the label that needs to be applied to the Windows node to make it worker node
 	WorkerLabel = "node-role.kubernetes.io/worker"
+	// VersionAnnotation indicates the version of WMCO that configured the node
+	VersionAnnotation = "windowsmachineconfig.openshift.io/version"
 )
 
 // nodeConfig holds the information to make the given VM a kubernetes node. As of now, it holds the information
@@ -129,10 +132,16 @@ func (nc *nodeConfig) Configure() error {
 	if err := nc.setNode(); err != nil {
 		return errors.Wrapf(err, "error getting node object for VM %s", nc.ID())
 	}
-	// Apply worker labels
-	if err := nc.applyWorkerLabel(); err != nil {
-		return errors.Wrap(err, "failed applying worker label")
+	// Apply labels and annotations to nc.node
+	nc.addWorkerLabel()
+	nc.addVersionAnnotation()
+	// update node object
+	node, err := nc.k8sclientset.CoreV1().Nodes().Update(context.TODO(), nc.node, metav1.UpdateOptions{})
+	if err != nil {
+		return errors.Wrap(err, "error updating node labels and annotations")
 	}
+	nc.node = node
+
 	// Now that basic kubelet configuration is complete, configure networking in the node
 	if err := nc.configureNetwork(); err != nil {
 		return errors.Wrap(err, "configuring node network failed")
@@ -176,19 +185,18 @@ func (nc *nodeConfig) configureNetwork() error {
 	return nil
 }
 
-// applyWorkerLabel applies the worker label to the Windows Node we created.
-func (nc *nodeConfig) applyWorkerLabel() error {
+// addVersionAnnotation adds the version annotation to nc.node
+func (nc *nodeConfig) addVersionAnnotation() {
+	nc.node.Annotations[VersionAnnotation] = version.Get()
+}
+
+// addWorkerLabel adds the worker label to nc.node
+func (nc *nodeConfig) addWorkerLabel() {
 	if _, found := nc.node.Labels[WorkerLabel]; found {
 		log.V(1).Info("worker label %s already present on node %s", WorkerLabel, nc.node.GetName())
-		return nil
+		return
 	}
 	nc.node.Labels[WorkerLabel] = ""
-	node, err := nc.k8sclientset.CoreV1().Nodes().Update(context.TODO(), nc.node, metav1.UpdateOptions{})
-	if err != nil {
-		return errors.Wrap(err, "error applying worker label to node object")
-	}
-	nc.node = node
-	return nil
 }
 
 // setNode identifies the node from the instanceID provided and sets the node object in the nodeconfig.
--- a/test/e2e/validation_test.go
+++ b/test/e2e/validation_test.go
@@ -2,6 +2,8 @@ package e2e
 
 import (
 	"context"
+	"fmt"
+	"os/exec"
 	"strings"
 	"testing"
 
@@ -40,6 +42,35 @@ func testWorkerLabel(t *testing.T) {
 	defer testCtx.Cleanup()
 	for _, node := range gc.nodes {
 		assert.Contains(t, node.Labels, nc.WorkerLabel, "expected node label %s was not present on %s", nc.WorkerLabel, node.GetName())
+	}
+}
+
+// getWMCOVersion returns the version of the operator. This is sourced from the WMCO binary used to create the operator image.
+// This function will return an error if the binary is missing.
+func getWMCOVersion() (string, error) {
+	cmd := exec.Command(wmcoPath, "version")
+	out, err := cmd.Output()
+	if err != nil {
+		return "", errors.Wrapf(err, "error running %s", cmd.String())
+	}
+	// out is formatted like:
+	// ./build/_output/bin/windows-machine-config-operator version: "0.0.1+4165dda-dirty", go version: "go1.13.7 linux/amd64"
+	versionSplit := strings.Split(string(out), "\"")
+	if len(versionSplit) < 3 {
+		return "", fmt.Errorf("unexpected version output")
+	}
+	return versionSplit[1], nil
+}
+
+// testVersionAnnotation tests all nodes are annotated with the version of the currently deployed WMCO
+func testVersionAnnotation(t *testing.T) {
+	operatorVersion, err := getWMCOVersion()
+	require.NoError(t, err, "could not get WMCO version")
+	for _, node := range gc.nodes {
+		t.Run(node.GetName(), func(t *testing.T) {
+			require.Containsf(t, node.Annotations, nc.VersionAnnotation, "node %s missing version annotation", node.GetName())
+			assert.Equal(t, operatorVersion, node.Annotations[nc.VersionAnnotation], "WMCO version annotation mismatch")
+		})
 	}
 }
 
--- a/test/e2e/wmco_test.go
+++ b/test/e2e/wmco_test.go
@@ -1,6 +1,7 @@
 package e2e
 
 import (
+	"os"
 	"testing"
 	"time"
 
@@ -32,6 +33,11 @@ func TestWMCO(t *testing.T) {
 	gc.numberOfNodes = int32(numberOfNodes)
 	gc.skipNodeDeletion = skipNodeDeletion
 	gc.sshKeyPair = sshKeyPair
+	// When the OPENSHIFT_CI env var is set to true, the test is running within CI
+	if inCI := os.Getenv("OPENSHIFT_CI"); inCI == "true" {
+		// In the CI container the WMCO binary will be found here
+		wmcoPath = "/usr/local/bin/windows-machine-config-operator"
+	}
 
 	t.Run("create", creationTestSuite)
 	if !gc.skipNodeDeletion {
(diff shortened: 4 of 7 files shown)


test/e2e/validation_test.go
```diff
@@ -43,6 +44,14 @@ func testWorkerLabel(t *testing.T) {
 	}
 }
 
+// testVersionAnnotation tests all nodes are annotated with the version of the currently deployed WMCO
+func testVersionAnnotation(t *testing.T) {
+	for _, node := range gc.nodes {
+		require.Containsf(t, node.Annotations, nc.VersionAnnotation, "node %s missing version annotation", node.Name)
```
VaishnaviHire, line 50: nit: `node.GetName()`

pkg/controller/windowsmachine/nodeconfig/nodeconfig.go
```diff
@@ -178,13 +181,14 @@ func (nc *nodeConfig) configureNetwork() error {
...
+		nc.node.Labels[WorkerLabel] = ""
+	} else {
 		log.V(1).Info("worker label %s already present on node %s", WorkerLabel, nc.node.GetName())
-		return nil
 	}
-	nc.node.Labels[WorkerLabel] = ""
+	nc.node.Annotations[VersionAnnotation] = version.Get()
 	node, err := nc.k8sclientset.CoreV1().Nodes().Update(context.TODO(), nc.node, metav1.UpdateOptions{})
 	if err != nil {
 		return errors.Wrap(err, "error applying worker label to node object")
```
mansikulkarni96, line 194: error message needs to be updated for version annotation

hack/common.sh
```diff
@@ -91,7 +91,7 @@ cleanup_WMCO() {
 # returns the operator version in `Version+GitHash` format
 get_version() {
   OPERATOR_VERSION=0.0.1
-  GIT_COMMIT=$(git rev-parse --short HEAD)
+  GIT_COMMIT=$(git rev-parse HEAD)
```
ravisantoshgudimetla, line 94: Why this change?

test/e2e/wmco_test.go
```diff
@@ -48,3 +53,18 @@ func setupWMCOResources() error {
 	}
 	return nil
 }
+
+// wmcoVersion returns the version that should be annotated on each Windows node
+func wmcoVersion() (string, error) {
```
ravisantoshgudimetla, line 58: nit: I'd have injected this value into the custom struct by calling `wmco version` command or something like that.

cmd/manager/main.go
```diff
@@ -152,8 +152,8 @@ func main() {
 		os.Exit(1)
 	}
 
-	// watch `openshift-machine-api` namespace along with wmco namespace
-	namespaces := []string{"openshift-machine-api", namespace}
+	// watch `""` (nodes), `openshift-machine-api` (machines), namespaces, along with the wmco namespace
```
aravindhp, line 155: It is unclear what `""` implies. Please make it `WMCO` given you are editing this line.

hack/common.sh
aravindhp: Is this because of `The version format was changed to full hash instead of shortened hash to
remove possibility of collisions and allow for better validation.`? I am -1 on having the full hash added to the version. We get reasonable collision prevnetion and validation with the short hash and for version that is good enough.

pkg/controller/add_node.go
```diff
@@ -0,0 +1,10 @@
+package controller
```
aravindhp, line 1: Did you add this manually or run some SDK command?

pkg/controller/node/node_controller.go
```diff
@@ -0,0 +1,127 @@
+package node
+
+import (
+	"context"
+
+	"github.com/openshift/windows-machine-config-operator/pkg/controller/windowsmachine/nodeconfig"
+	"github.com/openshift/windows-machine-config-operator/version"
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
```
aravindhp, line 9: `core` is sufficient.

pkg/controller/node/node_controller.go
```diff
@@ -0,0 +1,127 @@
...
+			return false
+		},
+		GenericFunc: func(e event.GenericEvent) bool {
+			return false
+		},
+	}
+
+	// Watch for changes to primary resource Node
+	err = c.Watch(&source.Kind{Type: &corev1.Node{}}, &handler.EnqueueRequestForObject{}, predicateFilter)
+	if err != nil {
```
aravindhp, line 78: One line please

pkg/controller/node/node_controller.go
```diff
@@ -0,0 +1,127 @@
...
+
+// Add creates a new Node Controller and adds it to the Manager. The Manager will set fields on the Controller
+// and Start it when the Manager is Started.
+func Add(mgr manager.Manager, _ string) error {
+	return add(mgr, newReconciler(mgr))
+}
+
+// newReconciler returns a new reconcile.Reconciler
+func newReconciler(mgr manager.Manager) reconcile.Reconciler {
+	return &ReconcileNode{client: mgr.GetClient(), scheme: mgr.GetScheme()}
```
aravindhp, line 33: Are you sure we don't need the non-cached client?

pkg/controller/node/node_controller.go
```diff
@@ -0,0 +1,127 @@
...
+
+// ReconcileNode reconciles a Node object
+type ReconcileNode struct {
+	// This client, initialized using mgr.Client() above, is a split client
+	// that reads objects from the cache and writes to the apiserver
+	client client.Client
+	scheme *runtime.Scheme
+}
+
+// Reconcile reads that state of the cluster for a Node object and makes changes based on the state read
```
aravindhp, line 96: I would just say `state of a Node object`

pkg/controller/node/node_controller.go
```diff
@@ -0,0 +1,127 @@
...
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, err
+	}
+
+	// Set version annotation to the proper value
+	instance.Annotations[nodeconfig.VersionAnnotation] = version.Get()
+	err = r.client.Update(context.TODO(), instance, &client.UpdateOptions{})
+	if err != nil {
```
aravindhp, line 121: Shouldn't this update be retried on failure?

pkg/controller/windowsmachine/nodeconfig/nodeconfig.go
```diff
@@ -30,6 +31,8 @@ const (
 	WindowsOSLabel = "node.openshift.io/os_id=Windows"
 	// WorkerLabel is the label that needs to be applied to the Windows node to make it worker node
 	WorkerLabel = "node-role.kubernetes.io/worker"
+	// VersionAnnotation indicates the version of WMCO that configured the node
+	VersionAnnotation = "windowsmachineconfiguration.openshift.io/config-version"
```
aravindhp, line 35: I am not so sure if `windowmachineconfiguration` is the right choice. Is there a reason for `config-version` vs `version`? I prefer the shorter `version`. Plus there is some stuttering going on. How about `windowsmachineconfig.openshift.io/version`?

pkg/controller/windowsmachine/nodeconfig/nodeconfig.go
```diff
@@ -131,9 +134,8 @@ func (nc *nodeConfig) Configure() error {
 	if err := nc.setNode(); err != nil {
 		return errors.Wrapf(err, "error getting node object for VM %s", nc.ID())
 	}
-	// Apply worker labels
-	if err := nc.applyWorkerLabel(); err != nil {
-		return errors.Wrap(err, "failed applying worker label")
+	if err := nc.applyLabelsAndAnnotations(); err != nil {
```
aravindhp, line 137: Please have separate functions for labels and annotations. The `Update` can be moved out. This will give us more flexibility in the future.

test/e2e/validation_test.go
```diff
@@ -43,6 +44,40 @@ func testWorkerLabel(t *testing.T) {
...
+		require.NoError(t, err, "error updating node objects")
+	}
+
+	// repopulate gc.nodes with new values
+	nodes, err := tc.kubeclient.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{LabelSelector: nc.WindowsOSLabel})
+	require.NoError(t, err, "error listing Windows nodes")
+	require.Lenf(t, nodes.Items, int(gc.numberOfNodes), "Expected num of Windows nodes does not match: %v", nodes.Items)
+	gc.nodes = nodes.Items
+	for _, node := range gc.nodes {
+		require.Containsf(t, node.Annotations, nc.VersionAnnotation, "node %s missing version annotation", node.GetName())
```
aravindhp, line 75: It will be a good data point for us if one node gets updated and not the other. So the loop should not stop on the first failure.

test/e2e/wmco_test.go
aravindhp: I would rather we use `wmco version`. You can further validate by comparing with `git` if you want but `wmco version` should be the source of truth.

hack/common.sh
sebsoto: from the commit message:

"The version format was changed to full hash instead of shortened hash to
remove possibility of collisions and allow for better validation."

cmd/manager/main.go
```diff
@@ -152,8 +152,8 @@ func main() {
...
+	namespaces := []string{"", "openshift-machine-api", namespace}
```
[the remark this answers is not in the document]
sebsoto: Updated comment to make that more clear

pkg/controller/add_node.go
sebsoto: I used the sdk command to add a controller

sebsoto:
Noticed a weirdness happening in the test, Windows node failed to be found but bootstrapping reported completion and must-gather had the node's kubelet logs. Retested to check if this is a flake, and if so we may need to increase the node list timeout

README.md
```diff
@@ -61,15 +61,19 @@ export OPERATOR_IMAGE=<registry url for remote WMCO image>
...
 ```
-We assume that the developer uses `openshift-dev` as the key pair in the aws cloud
+We assume that the developer uses `openshift-dev` as the key pair in the aws cloud.
 
 Additional flags that can be passed to `hack/run-ci-e2e-test.sh` are
 - `-s` to skip the deletion of Windows nodes that are created as part of test suite run
 - `-n` to represent the number of Windows nodes to be created for test run
 - `-k` to represent the AWS specific key pair that will be used during e2e run and it should map to the private key
        that we have in `KUBE_SSH_KEY_PATH`. The default value points to `openshift-dev` which we use in our CI
+- `-b` gives an alternative path to the WMCO binary. If the -b option is not given outside of CI, certain validation
```
aravindhp, line 73: - Is this is a required parameter when running locally? If yes, then it should be called out.
- Ideally it would be nice for this to be an option parameter with a sensible default.
- What validations tests fail? It would help to call that out here.

test/e2e/validation_test.go
```diff
@@ -43,6 +45,35 @@ func testWorkerLabel(t *testing.T) {
 	}
 }
 
+// wmcoVersion returns the version of the operator. This is sourced from the WMCO binary used to create the operator image.
+// This function will return an error if the binary is missing.
+func wmcoVersion() (string, error) {
```
aravindhp, line 50: Given this is not associated with a struct, getWMCOVersion() would be more appropriate.

pkg/controller/windowsmachine/nodeconfig/nodeconfig.go
```diff
@@ -129,10 +132,16 @@ func (nc *nodeConfig) Configure() error {
...
 	}
-	// Apply worker labels
-	if err := nc.applyWorkerLabel(); err != nil {
-		return errors.Wrap(err, "failed applying worker label")
+	// Apply labels and annotations to nc.node
+	nc.addWorkerLabel()
+	nc.addVersionAnnotation()
+	// update node object
+	node, err := nc.k8sclientset.CoreV1().Nodes().Update(context.TODO(), nc.node, metav1.UpdateOptions{})
+	if err != nil {
```
aravindhp, line 140: Should we fail immediately or retry a few times? It is very possible that a change could be in flight when this update is made.

README.md
sebsoto: Changed this to use the normal dev workflow as default.

pkg/controller/windowsmachine/nodeconfig/nodeconfig.go
sebsoto: Its most likely safer to just requeue on error

README.md
```diff
@@ -63,13 +63,17 @@ Once the above variables are set, run the following script:
...
 ```
-We assume that the developer uses `openshift-dev` as the key pair in the aws cloud
+We assume that the developer uses `openshift-dev` as the key pair in the aws cloud.
 
 Additional flags that can be passed to `hack/run-ci-e2e-test.sh` are
 - `-s` to skip the deletion of Windows nodes that are created as part of test suite run
 - `-n` to represent the number of Windows nodes to be created for test run
 - `-k` to represent the AWS specific key pair that will be used during e2e run and it should map to the private key
        that we have in `KUBE_SSH_KEY_PATH`. The default value points to `openshift-dev` which we use in our CI
+- `-b` gives an alternative path to the WMCO binary. This option is not acknowledged when running in CI.
```
aravindhp, line 73: Nit: I don't follow what you mean by `This option is not acknowledged when running in CI.`.

README.md
```diff
@@ -63,13 +63,17 @@ Once the above variables are set, run the following script:
...
+       When building the operator locally, the WMCO binary is created as part of the operator image build process and
+       can be found at `build/_output/bin/windows-machine-config-operator`, this is the default value of this option.
```
aravindhp, line 75: Where is this default being set?

test/e2e/validation_test.go
```diff
@@ -43,6 +45,35 @@ func testWorkerLabel(t *testing.T) {
...
+	return versionSplit[1], nil
+}
+
+// testVersionAnnotation tests all nodes are annotated with the version of the currently deployed WMCO
+func testVersionAnnotation(t *testing.T) {
+	operatorVersion, err := getWMCOVersion()
+	require.NoError(t, err, "could not get WMCO version")
+	for _, node := range gc.nodes {
+		t.Run(node.GetName(), func(t *testing.T) {
+			require.Containsf(t, node.Annotations, nc.VersionAnnotation, "node %s missing version annotation", node.GetName())
```
aravindhp, line 71: I thought I mentioned this before. We should not stop the test on the first failure. It will be a good data point for us if there is a scenario where one node has the correct version annotation while the other does not.

pkg/controller/add_node.go
aravindhp: Please add the SDK command to your commit message.

test/e2e/validation_test.go
sebsoto: We are not stopping at the first failure here, each node is checked within a separate subtest https://github.com/openshift/windows-machine-config-operator/pull/88/files#diff-1a876af15bc4507294c2bc7e5f81dc37R70 

When one node fails, its subtest fails and the rest of the nodes are tested with their subtests.

README.md
sebsoto: Being set here: https://github.com/openshift/windows-machine-config-operator/pull/88/files#diff-e8664074431f2653c7da0fc9019a1935R95

aravindhp: Sorry I missed that.

test/e2e/validation_test.go
aravindhp: You are right. I missed this too.

README.md
sebsoto: Reworded

README.md
```diff
@@ -63,13 +63,17 @@ Once the above variables are set, run the following script:
...
 ```
-We assume that the developer uses `openshift-dev` as the key pair in the aws cloud
+We assume that the developer uses `openshift-dev` as the key pair in the aws cloud.
 
 Additional flags that can be passed to `hack/run-ci-e2e-test.sh` are
 - `-s` to skip the deletion of Windows nodes that are created as part of test suite run
 - `-n` to represent the number of Windows nodes to be created for test run
 - `-k` to represent the AWS specific key pair that will be used during e2e run and it should map to the private key
        that we have in `KUBE_SSH_KEY_PATH`. The default value points to `openshift-dev` which we use in our CI
+- `-b` gives an alternative path to the WMCO binary. This option overrided in OpenShift CI.
```
aravindhp, line 73: `is overridden in...`

VaishnaviHire:
LGTM

PR #92, opened 2020-08-06
[WIP][wmco] Watch userData secret

VaishnaviHire:
This PR introduces a new controller which is responsible for watching
`userData` secret and reacting to changes to the secret.


pkg/controller/userdatasecret/userdatasecret_controller.go
```diff
@@ -0,0 +1,217 @@
...
+
+	signer, err := wmc.CreateSigner()
+	if err != nil {
+		return nil, errors.Wrapf(err, "error creating signer using private key: %v", wkl.PrivateKeyPath)
+	}
+	reconciler := &ReconcileUserDataSecret{client: client, scheme: mgr.GetScheme(), signer: signer}
+
+	// Create userData secret
+	userDataSecret, _ := reconciler.createUserDataSecret()
+	err = reconciler.client.Create(context.TODO(), userDataSecret)
```
mansikulkarni96, line 69: aren't we creating this as a part of reconcile?

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -225,71 +225,30 @@ func (r *ReconcileWindowsMachine) addWorkerNode(ipAddress, instanceID string) er
...
-			$pubKeyLocationConf = (Get-Content -path C:\ProgramData\ssh\sshd_config) -replace 'Match Group administrators','#Match Group administrators'
-			$pubKeyLocationConf | Set-Content -Path C:\ProgramData\ssh\sshd_config
-			Restart-Service sshd
-			New-item -Path $env:USERPROFILE -Name .ssh -ItemType Directory -force
-			echo "` + string(pubKeyBytes[:]) + `"| Out-File $env:USERPROFILE\.ssh\authorized_keys -Encoding ascii
-			</powershell>
-			<persist>true</persist>`),
-		},
+	if err != nil {
+		return errors.Errorf("could not find windows user data secret in required namespace: %v", err)
```
mansikulkarni96, line 239: windows-> Windows for all Windows occurences

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -225,71 +225,30 @@ func (r *ReconcileWindowsMachine) addWorkerNode(ipAddress, instanceID string) er
...
-		return errors.Wrap(err, "error creating windows user data secret")
+	secretData := string(instance.Data["userData"][:])
+	if !strings.Contains(secretData, string(pubKeyBytes[:])) {
+		return errors.Errorf("expected user data not present in required namespace")
 	}
 	return nil
+
 }
 
 // createSigner creates a signer using the private key from the privateKeyPath
```
mansikulkarni96, line 250: createSigner -> CreateSigner

test/e2e/delete_test.go
```diff
@@ -63,4 +64,8 @@ func testWindowsNodeDeletion(t *testing.T) {
 			Namespace: "openshift-machine-api"}, machineSet)
 		assert.NoError(t, framework.Global.Client.Delete(context.TODO(), machineSet))
 	}
+
+	// Cleanup secrets created by us.
+	err = framework.Global.KubeClient.CoreV1().Secrets("openshift-machine-api").Delete(context.TODO(), "windows-user-data", metav1.DeleteOptions{})
+	require.NoError(t, err, "could not delete userData secret")
```
mansikulkarni96, line 70: the messages can be standardized to using userData or user data in all places.

test/e2e/validation_test.go
```diff
@@ -96,3 +97,39 @@ func testUserData(t *testing.T) {
 	require.NoError(t, err, "could not find windows user data secret in required namespace")
 	assert.Contains(t, string(found.Data["userData"][:]), string(pubKeyBytes[:]), "expected user data not present in required namespace")
 }
+
+// testUserDataUpdate tests if userData reverts to previous value if updated
+func testUserDataUpdate(t *testing.T) {
+	secretInstance := &corev1.Secret{}
+	err := framework.Global.Client.Get(context.TODO(), kubeTypes.NamespacedName{Name: "windows-user-data", Namespace: "openshift-machine-api"}, secretInstance)
+	require.NoError(t, err, "could not find windows user data secret in required namespace")
```
mansikulkarni96, line 105: windows-> Windows

test/e2e/validation_test.go
```diff
@@ -96,3 +97,39 @@ func testUserData(t *testing.T) {
...
+				require.NoError(t, err, "could not update userData secret")
+				err = framework.Global.Client.Get(context.TODO(), kubeTypes.NamespacedName{Name: "windows-user-data", Namespace: "openshift-machine-api"}, secretInstance)
+				require.NoError(t, err, "could not find windows user data secret in required namespace")
+				require.Equal(t, validUserData, secretInstance.Data["userData"], "unable to revert userData Secret")
+			}
+			if tt.operation == "Delete" {
+				err := framework.Global.KubeClient.CoreV1().Secrets("openshift-machine-api").Delete(context.TODO(), "windows-user-data", metav1.DeleteOptions{})
+				require.NoError(t, err, "could not delete userData secret")
+				err = framework.Global.Client.Get(context.TODO(), kubeTypes.NamespacedName{Name: "windows-user-data", Namespace: "openshift-machine-api"}, secretInstance)
+				require.NoError(t, err, "could not find windows user data secret in required namespace")
```
mansikulkarni96, line 131: windows-> Windows

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -225,71 +225,30 @@ func (r *ReconcileWindowsMachine) addWorkerNode(ipAddress, instanceID string) er
...
-				return errors.Wrap(err, "error creating windows user data secret")
-			}
-		}
-		return errors.Wrap(err, "error creating windows user data secret")
+	secretData := string(instance.Data["userData"][:])
+	if !strings.Contains(secretData, string(pubKeyBytes[:])) {
+		return errors.Errorf("expected user data not present in required namespace")
 	}
 	return nil
+
```
mansikulkarni96, line 247: extra line

pkg/controller/userdatasecret/userdatasecret_controller.go
```diff
@@ -0,0 +1,217 @@
...
+			return false
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			secretName := e.MetaNew.GetName()
+			if secretName == userDataSecret {
+				return true
+			}
+			return false
+		},
+		// ignore delete event, as we will get an update event while deleting
```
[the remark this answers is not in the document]
VaishnaviHire: Its not, I will change the comment

pkg/controller/userdatasecret/userdatasecret_controller.go
```diff
@@ -0,0 +1,217 @@
...
+	userData, err := r.createUserDataSecret()
+	if err != nil {
+		return reconcile.Result{}, errors.Wrapf(err, "error creating user data secret")
+
+	}
+	// Fetch UserData instance
+	err = r.client.Get(context.TODO(), kubeTypes.NamespacedName{Name: userData.Name, Namespace: userData.Namespace}, &corev1.Secret{})
+	if err != nil && k8sapierrors.IsNotFound(err) {
+		// Secret is deleted
+		reqLogger.Info("Creating a new Secret", "Secret.Namespace", userData.Namespace, "Secret.Name", userData.Name)
```
sebsoto, line 147: Creating a new Secret -> Creating userdata secret

pkg/controller/userdatasecret/userdatasecret_controller.go
```diff
@@ -0,0 +1,217 @@
...
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		// Secret created successfully - don't requeue
+		return reconcile.Result{}, nil
+	} else if err != nil {
+		return reconcile.Result{}, err
+	} else {
+		// secret is updated
+		reqLogger.Info("Updating the secret", "Secret.Namespace", userData.Namespace, "Secret.Name", userData.Name)
```
sebsoto, line 158: Updating the secret -> Updating userdata secret

pkg/controller/userdatasecret/userdatasecret_controller.go
```diff
@@ -0,0 +1,217 @@
...
+		}
+		// Secret updated successfully
+		return reconcile.Result{}, nil
+	}
+}
+
+// createUserDataSecret creates a secret 'windows-user-data' in 'openshift-machine-api'
+// namespace. This secret will be used to inject cloud provider user data for creating
+// windows machines
+func (r *ReconcileUserDataSecret) createUserDataSecret() (*corev1.Secret, error) {
```
sebsoto, line 171: This isnt actually creating a userdata secret, it is returning a secret that can be created.

sebsoto: You can change the name to reflect the functionality, or change the function to reflect the name.

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -225,71 +225,30 @@ func (r *ReconcileWindowsMachine) addWorkerNode(ipAddress, instanceID string) er
...
-	}
-
+// validateUserData validates userData secret. It returns error if the secret doesn`t
+// contain expected public key bytes.
+func (r *ReconcileWindowsMachine) validateUserData() error {
 	pubKeyBytes := ssh.MarshalAuthorizedKey(r.signer.PublicKey())
 	if pubKeyBytes == nil {
 		return errors.Errorf("failed to retrieve public key using signer for private key: %v", wkl.PrivateKeyPath)
 	}
+	instance := &core.Secret{}
```
sebsoto, line 235: please use a more descriptive name, like `userdata`

pkg/controller/userdatasecret/userdatasecret_controller.go
VaishnaviHire: The primary resource for this controller is `Secret`. The reconcile will be triggered only for `Update` and `Delete` events. The other alternative for this could be creating `UserData` in `windowsmachine-controller`

test/e2e/validation_test.go
sebsoto, line 131: I would restructure this test block. [Table driven tests](https://dave.cheney.net/2013/06/09/writing-table-driven-tests-in-go) are useful if you performing the same actions with different parameters and expect specific results. The way you are using it does not fall under this use case.

In order to make this test follow the table driven test pattern, you could change your `tests` struct definition to have: `{name, apiCall, expectedSecret}` where apiCall is a function pointer, and expectedSecret is the expected userData object.

If the above suggestion works thats cool, but if it doesn't theres no reason to try and force the table driven test pattern in a place it doesnt work. You can separate these tests into different test functions, so you'd have 
```
testUserdataDelete(t *testing.T) { ...}
testUserdataModify(t *testing.T) {...}
```
And then you could have a larger test that calls these two test functions as subtests to organize the test results:
```
testUserdataTamper(t *testing.T) {
t.Run("test that a user cannot delete the userdata secret", testUserdataDelete)
t.Run("test that a user cannot modify the userdata secret", testUserdataModify)
}
```

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -225,71 +225,30 @@ func (r *ReconcileWindowsMachine) addWorkerNode(ipAddress, instanceID string) er
...
-		if k8sapierrors.IsNotFound(err) {
-			log.Info("Creating a new Secret", "Secret.Namespace", userDataSecret.Namespace, "Secret.Name", userDataSecret.Name)
-			err = r.client.Create(context.TODO(), userDataSecret)
-			if err != nil {
-				return errors.Wrap(err, "error creating windows user data secret")
-			}
-		}
-		return errors.Wrap(err, "error creating windows user data secret")
+	secretData := string(instance.Data["userData"][:])
+	if !strings.Contains(secretData, string(pubKeyBytes[:])) {
```
ravisantoshgudimetla, line 243: Are you just checking for pubKeyBytes? I think we also need to look at the entire existing secret or is the scope of this controller limited to checking the pulickey?

test/e2e/validation_test.go
ravisantoshgudimetla: I agree with most of what @sebsoto told except for if this should be a tabledriven test or not. To me, this is ok considering we are not writing units but e2es but we should have expectedResult for example if the test supposed to recreate a secret or not?

pkg/controller/userdatasecret/userdatasecret_controller.go
```diff
@@ -0,0 +1,217 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	client, err := client.New(cfg, client.Options{Scheme: mgr.GetScheme()})
+	if err != nil {
+		return nil, err
+	}
+
+	signer, err := wmc.CreateSigner()
```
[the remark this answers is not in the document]
VaishnaviHire: @ravisantoshgudimetla Should I be moving it to `pkg/internal/utils` ?

pkg/controller/windowsmachine/windowsmachine_controller.go
VaishnaviHire: Its not. I will update the check.

pkg/controller/userdatasecret/userdatasecret_controller.go
aravindhp: Please fix the doc too. `construct` or `generate` might be more appropriate here.

pkg/controller/userdatasecret/userdatasecret_controller.go
```diff
@@ -0,0 +1,217 @@
...
+// The Controller will requeue the Request to be processed again if the returned error is non-nil or
+// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
+func (r *ReconcileUserDataSecret) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+	reqLogger.Info("Reconciling UserDataSecret")
+
+	userData, err := r.createUserDataSecret()
+	if err != nil {
+		return reconcile.Result{}, errors.Wrapf(err, "error creating user data secret")
+
```
aravindhp, line 141: Stray line

pkg/controller/userdatasecret/userdatasecret_controller.go
aravindhp: `reqLogger` should anynow show the name and namespace. It needed be repeated here. Also `Restoring...` would be more informative.

pkg/controller/userdatasecret/userdatasecret_controller.go
aravindhp: The log should reflect that we are recreating it.

test/e2e/create_test.go
```diff
@@ -31,6 +31,7 @@ func creationTestSuite(t *testing.T) {
 	t.Run("Label validation", func(t *testing.T) { testWorkerLabel(t) })
 	t.Run("NodeTaint validation", func(t *testing.T) { testNodeTaint(t) })
 	t.Run("User Data validation", func(t *testing.T) { testUserData(t) })
+	t.Run("Update User Data", func(t *testing.T) { testUserDataUpdate(t) })
```
aravindhp, line 34: A more appropriate name like `User data idempotent check` would be preferable.

test/e2e/delete_test.go
```diff
@@ -9,6 +9,7 @@ import (
 	framework "github.com/operator-framework/operator-sdk/pkg/test"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
aravindhp, line 12: `meta`

test/e2e/validation_test.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"golang.org/x/crypto/ssh"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
aravindhp, line 15: `meta`

pkg/controller/userdatasecret/userdatasecret_controller.go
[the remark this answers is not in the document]
VaishnaviHire: Moved the `createSigner` to a top level location. I am not sure if modifying `createSigner` to take `cloud-private-key` secret should be done as part of this PR or `WINC-316`, hence added a TODO

lib/secrets/secrets.go
```diff
@@ -0,0 +1,81 @@
+package secrets
```
ravisantoshgudimetla, line 1: This can go in `/pkg`

test/e2e/validation_test.go
```diff
@@ -91,8 +93,44 @@ func testUserData(t *testing.T) {
...
+				updatedSecret.Data["userData"] = []byte("invalid data")
+				err := framework.Global.Client.Update(context.TODO(), updatedSecret)
+				require.NoError(t, err, "could not update userData secret")
+			}
+			if tt.operation == "Delete" {
+				err := framework.Global.KubeClient.CoreV1().Secrets("openshift-machine-api").Delete(context.TODO(), "windows-user-data", meta.DeleteOptions{})
+				require.NoError(t, err, "could not delete userData secret")
+			}
+
+			err = framework.Global.Client.Get(context.TODO(), kubeTypes.NamespacedName{Name: "windows-user-data",
```
ravisantoshgudimetla, line 130: I think you need to have wait with a timeout logic here as the apiserver update can take some time.

pkg/controller/secret/secret_controller.go
```diff
@@ -0,0 +1,145 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	ControllerName = "controller_secret"
```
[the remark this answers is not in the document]
VaishnaviHire: Yes it`s generated, but I can change it.

test/e2e/validation_test.go
```diff
@@ -91,8 +92,45 @@ func testUserData(t *testing.T) {
...
+				updatedSecret := validUserDataSecret.DeepCopy()
+				updatedSecret.Data["userData"] = []byte("invalid data")
+				err := framework.Global.Client.Update(context.TODO(), updatedSecret)
+				require.NoError(t, err, "could not update userData secret")
+			}
+			if tt.operation == "Delete" {
+				err := framework.Global.KubeClient.CoreV1().Secrets("openshift-machine-api").Delete(context.TODO(), "windows-user-data", meta.DeleteOptions{})
+				require.NoError(t, err, "could not delete userData secret")
+			}
+			// the sleep allows for userData secret creation / update to take effect.
```
ravisantoshgudimetla, line 128: No this is not what I was asking for, something similar to [this](https://github.com/openshift/windows-machine-config-operator/blob/master/test/e2e/create_test.go#L107)

pkg/controller/secret/secret_controller.go
```diff
@@ -0,0 +1,145 @@
...
+
+	client, err := client.New(cfg, client.Options{Scheme: mgr.GetScheme()})
+	if err != nil {
+		return nil, err
+	}
+
+	reconciler := &ReconcileSecret{client: client, scheme: mgr.GetScheme()}
+
+	// Create userData secret
+	userData, _ := secrets.GenerateUserData()
```
ravisantoshgudimetla, line 57: You don't want to catch the error or does this mean, we don't care if there is userData secret creation in here as the subsequent reconciler would take care of it?

test/e2e/validation_test.go
```diff
@@ -2,17 +2,18 @@ package e2e
...
 	framework "github.com/operator-framework/operator-sdk/pkg/test"
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"golang.org/x/crypto/ssh"
-	corev1 "k8s.io/api/core/v1"
+	core "k8s.io/api/core/v1"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kubeTypes "k8s.io/apimachinery/pkg/types"
+	"strings"
```
ravisantoshgudimetla, line 14: nit: Ordering?

pkg/controller/secret/secret_controller.go
```diff
@@ -0,0 +1,145 @@
...
+	err = client.Get(context.TODO(), kubeTypes.NamespacedName{Name: userDataSecret,
+		Namespace: userData.Namespace}, &core.Secret{})
+	if err != nil {
```
ravisantoshgudimetla, line 60: So to be clear, when the operator restarts, if the secret exists and it doesn't match with the one we want, we would ignore it and handle it in the subsequent reconciliation. I think we might not catch it as we ignore events other than Update.

test/e2e/validation_test.go
VaishnaviHire: I did look into `wait.Poll`. However, would it make sense to have a `conditionFunc` here, as in case of update operation, we would still be able to do a `Get()` . Maybe, we also need to have a check for the contents

pkg/controller/secret/secret_controller.go
VaishnaviHire: I think you are right. We could recreate the secret, if the operator restarts irrespective of the secret content.

pkg/lib/secrets/secrets.go
```diff
@@ -0,0 +1,81 @@
+package secrets
```
aravindhp, line 1: Do we need a pkg called `lib`? I was under the impression it is best to avoid generic higher level packages.

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -225,79 +225,23 @@ func (r *ReconcileWindowsMachine) addWorkerNode(ipAddress, instanceID string) er
...
-			echo "` + string(pubKeyBytes[:]) + `"| Out-File $env:USERPROFILE\.ssh\authorized_keys -Encoding ascii
-			</powershell>
-			<persist>true</persist>`),
-		},
-	}
+// validateUserData validates userData secret. It returns error if the secret doesn`t
+// contain expected public key bytes.
+func (r *ReconcileWindowsMachine) validateUserData() error {
+	secret := &core.Secret{}
+	err := r.client.Get(context.TODO(), kubeTypes.NamespacedName{Name: "windows-user-data", Namespace: "openshift-machine-api"}, secret)
```
aravindhp, line 232: Please use `userDataSecret`. We should also create a constant for `openshift-machine-api`.

pkg/lib/secrets/secrets.go
```diff
@@ -0,0 +1,81 @@
...
+	if pubKeyBytes == nil {
+		return nil, errors.Errorf("failed to retrieve public key using signer for private key: %v", wkl.PrivateKeyPath)
+	}
+
+	// sshd service is started to create the default sshd_config file. This file is modified
+	// for enabling publicKey auth and the service is restarted for the changes to take effect.
+	userDataSecret := &core.Secret{
+		ObjectMeta: meta.ObjectMeta{
+			Name:      "windows-user-data",
+			Namespace: "openshift-machine-api",
```
aravindhp, line 49: Please use constants. Applies to all places where you are using these.

pkg/lib/secrets/secrets.go
VaishnaviHire: The functions / code common across controllers will be stored in `lib`

aravindhp: Then why not in the `controllers` pkg?

pkg/controller/secret/secret_controller.go
```diff
@@ -0,0 +1,145 @@
...
+		err = r.client.Create(context.TODO(), userData)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		// Secret created successfully - don't requeue
+		return reconcile.Result{}, nil
+	} else if err != nil {
+		return reconcile.Result{}, err
+	} else {
+		// secret is updated
```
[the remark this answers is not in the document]
VaishnaviHire: Modified predicate to get update events only when userData content is changed

pkg/controller/secret/secret_controller.go
```diff
@@ -0,0 +1,170 @@
...
+		err = r.client.Update(context.TODO(), userData)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		// Secret updated successfully
+		return reconcile.Result{}, nil
+	}
+}
+
+// createValidUserData creates userData secret if it is not present in the required namespace.
```
aravindhp, line 144: I would rename this to `createUserData()`. It being valid is implicit as there is no method to create an invalid one.

pkg/controller/secret/secret_controller.go
```diff
@@ -0,0 +1,170 @@
...
+	if err != nil && k8sapierrors.IsNotFound(err) {
+		// Create valid userData secret
+		if err = r.client.Create(context.TODO(), validUserDataSecret); err != nil {
+			return errors.Wrapf(err, "error creating userData secret")
+		}
+	} else if err != nil {
+		return errors.Wrapf(err, "error retrieving userData secret")
+	}
+	// Update userData secret if the content is invalid
+	if string(userData.Data["userData"]) != string(validUserDataSecret.Data["userData"]) {
```
aravindhp, line 164: Do you need this check now that you checking in the predicate? Or is there some corner case that is not obvious? If there is a corner case that you are accounting for, please mention it here.

pkg/controller/secrets/secrets.go
```diff
@@ -0,0 +1,87 @@
...
+
+const (
+	userDataSecret    = "windows-user-data"
+	userDataNamespace = "openshift-machine-api"
+)
+
+// CreateSigner creates a signer using the private key from the privateKeyPath
+// TODO: As a part of https://issues.redhat.com/browse/WINC-316 , modify CreateSigner to take private key secret
+//  and return the signer
+func CreateSigner() (ssh.Signer, error) {
```
aravindhp, line 21: Nit: from a purity perspective, this would be in a separate package called `signer`.

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -225,79 +225,23 @@ func (r *ReconcileWindowsMachine) addWorkerNode(ipAddress, instanceID string) er
...
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to find private key from path: %v", wkl.PrivateKeyPath)
+		return err
 	}
-
-	signer, err := ssh.ParsePrivateKey(privateKeyBytes)
-	if err != nil {
-		return nil, errors.Wrapf(err, "unable to parse private key: %v", wkl.PrivateKeyPath)
+	if string(desiredUserDataSecret.Data["userData"][:]) != secretData {
+		return errors.Errorf("expected userData secret not present in required namespace")
```
aravindhp, line 244: This seems to imply the secret is not present. It is just the contents that is invalid.

test/e2e/validation_test.go
```diff
@@ -91,8 +95,56 @@ func testUserData(t *testing.T) {
...
+				err := framework.Global.Client.Update(context.TODO(), updatedSecret)
+				require.NoError(t, err, "could not update userData secret")
+			}
+			if tt.operation == "Delete" {
+				err := framework.Global.KubeClient.CoreV1().Secrets("openshift-machine-api").Delete(context.TODO(), "windows-user-data", meta.DeleteOptions{})
+				require.NoError(t, err, "could not delete userData secret")
+			}
+
+			// wait for userData secret creation / update to take effect.
+			err := wait.Poll(5*time.Second, 20*time.Second, func() (done bool, err error) {
```
aravindhp, line 133: Do we need reusable constants for these times?

pkg/controller/secret/secret_controller.go
VaishnaviHire: This func is called in `newReconciler`.  There might be a case where user stops the operator and creates a secret with invalid contents. In that case, we need to check this condition.

aravindhp: Won't that cause an update event which is caught by the operator when it comes up?

test/e2e/validation_test.go
VaishnaviHire: I am not sure, as these are only used once.

aravindhp: I am fine leaving it as is then but we need to ensure the constants gets added on the next usage.

pkg/controller/secret/secret_controller.go
VaishnaviHire: I did not see any Update event when I tried this. Still see [this](https://github.com/openshift/windows-machine-config-operator/pull/92#discussion_r470152208), if the check is not in place

aravindhp: I am a bit puzzled by this. Please leave it as it is for now and we can discuss this during parking lot.

VaishnaviHire: Discussed about this in the parking lot. This usecase will generate a `Create` event.  I will modify the controller to handle create events.

PR #101, opened 2020-08-27
[wmco] Delete Machines with mismatched version annotations

VaishnaviHire:
This PR makes sure that machine controller handles changes in version
annotation on Windows nodes. The version annotation can be changed in the following scenarios -

- During upgrade process, where `operatorVersion` would not match with the version used in the annotation
- When node annotation is tampered by an user

The controller makes sure to delete the machines and subsequently the nodes, when annotation is mismatched and the number of nodes after deletion are greater than the predetermined `minHealthyCount`

This PR also increases the value of `nodeCreation` in the e2e tests, to take into consideration time required to delete the node and recreate it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -299,3 +299,17 @@ spec:
           userDataSecret:
             name: windows-user-data
 ```
+### Windows nodes Kubernetes component upgrade
+
+When a new version of WMCO is released that is compatible with the current cluster version, an operator upgrade will 
+take place which will result in the Kubernetes components in the Windows Machine to be upgraded. For a non-disruptive 
+upgrade, WMCO terminates the Windows Machines configured by previous version of WMCO and recreates them using the
+current version. This is done by deleting the Machine object that results in the drain and deletion of the Windows node.
+To facilitate an upgrade, WMCO adds a version annotation to all the configured nodes. During an upgrade, a mismatch in
+version annotation will result in deletion and recreation of Windows Machine. In order to have minimal service 
+disruption during an upgrade, WMCO makes sure that the cluster will have atleast 1 Windows Machine per MachineSet in the
+running state.
+
+WMCO is not responsible for Windows operating system updates. The cluster administrator provides the Window image while
+creating the VMs and hence, the cluster administrator is responsible for providing an updated image. The cluster 
+administrator can provide an updated image by changing the image in the MachineSet spec.
--- a/deploy/olm-catalog/windows-machine-config-operator/manifests/windows-machine-config-operator.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/windows-machine-config-operator/manifests/windows-machine-config-operator.clusterserviceversion.yaml
@@ -81,6 +81,15 @@ spec:
           - get
           - list
           - watch
+          - delete
+        - apiGroups:
+          - machine.openshift.io
+          resources:
+          - machinesets
+          verbs:
+          - list
+          - get
+          - watch
         serviceAccountName: windows-machine-config-operator
       deployments:
       - name: windows-machine-config-operator
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -137,3 +137,12 @@ rules:
      - get
      - list
      - watch
+     - delete
+ - apiGroups:
+     - machine.openshift.io
+   resources:
+     - machinesets
+   verbs:
+     - list
+     - get
+     - watch
--- a/hack/run-ci-e2e-test.sh
+++ b/hack/run-ci-e2e-test.sh
@@ -102,17 +102,25 @@ OSDK_WMCO_test $OSDK "-run=TestWMCO/operator_deployed_without_private_key_secret
 
 # Run the creation tests of the Windows VMs
 OSDK_WMCO_test $OSDK "-run=TestWMCO/create -v -timeout=90m -node-count=$NODE_COUNT --private-key-path=$KUBE_SSH_KEY_PATH $WMCO_PATH_OPTION"
+# Get logs for the creation tests
+printf "\n####### WMCO logs for creation tests #######\n"
+get_WMCO_logs
+
+# Run the upgrade tests and skip deletion of the Windows VMs
+OSDK_WMCO_test $OSDK "-run=TestWMCO/upgrade -v -timeout=90m -node-count=$NODE_COUNT --private-key-path=$KUBE_SSH_KEY_PATH $WMCO_PATH_OPTION"
 
 # Run the deletion tests while testing operator restart functionality. This will clean up VMs created
 # in the previous step
 if ! $SKIP_NODE_DELETION; then
   OSDK_WMCO_test $OSDK "-run=TestWMCO/destroy -v -timeout=60m --private-key-path=$KUBE_SSH_KEY_PATH"
   # Get logs on success before cleanup
+  printf "\n####### WMCO logs for upgrade and deletion tests #######\n"
   get_WMCO_logs
   # Cleanup the operator resources
   cleanup_WMCO $OSDK
 else
   # Get logs on success
+  printf "\n####### WMCO logs for upgrade tests #######\n"
   get_WMCO_logs
 fi
 
--- a/pkg/controller/windowsmachine/windowsmachine_controller.go
+++ b/pkg/controller/windowsmachine/windowsmachine_controller.go
@@ -4,15 +4,16 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"time"
 
 	mapi "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
-	"github.com/openshift/windows-machine-config-operator/pkg/clusternetwork"
 	"github.com/pkg/errors"
 	"golang.org/x/crypto/ssh"
 	core "k8s.io/api/core/v1"
 	k8sapierrors "k8s.io/apimachinery/pkg/api/errors"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	kubeTypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/record"
@@ -27,14 +28,22 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	"github.com/openshift/windows-machine-config-operator/pkg/clusternetwork"
 	"github.com/openshift/windows-machine-config-operator/pkg/controller/secrets"
 	"github.com/openshift/windows-machine-config-operator/pkg/controller/signer"
 	"github.com/openshift/windows-machine-config-operator/pkg/controller/windowsmachine/nodeconfig"
+	"github.com/openshift/windows-machine-config-operator/version"
 )
 
 const (
 	// ControllerName is the name of the WindowsMachine controller
 	ControllerName = "windowsmachine-controller"
+	// minHealthyCount is the minimum number of nodes that are required to be in running phase at a given time.
+	minHealthyCount = 1
+	// windowsOSLabel is the label used to identify the Windows Machines.
+	windowsOSLabel = "machine.openshift.io/os-id"
+	// requeueDuration is the time after which the request is requeued.
+	requeueDuration = time.Minute * 5
 )
 
 var log = logf.Log.WithName(ControllerName)
@@ -94,8 +103,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return errors.Wrapf(err, "could not create %s", ControllerName)
 	}
 	// Watch for the Machine objects with label defined by windowsOSLabel
-	windowsOSLabel := "machine.openshift.io/os-id"
-	predicateFilter := predicate.Funcs{
+	machinePredicate := predicate.Funcs{
 		// ignore create event for all Machines as WMCO should for Machine getting provisioned
 		CreateFunc: func(e event.CreateEvent) bool {
 			return false
@@ -117,11 +125,80 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 
 	err = c.Watch(&source.Kind{Type: &mapi.Machine{
 		ObjectMeta: meta.ObjectMeta{Namespace: "openshift-machine-api"},
-	}}, &handler.EnqueueRequestForObject{}, predicateFilter)
+	}}, &handler.EnqueueRequestForObject{}, machinePredicate)
 	if err != nil {
 		return errors.Wrap(err, "could not create watch on Machine objects")
 	}
 
+	err = c.Watch(&source.Kind{Type: &core.Node{
+		ObjectMeta: meta.ObjectMeta{Namespace: ""},
+	}}, &handler.EnqueueRequestsFromMapFunc{ToRequests: newNodeToMachineMapper(mgr.GetClient())}, predicate.Funcs{
+		CreateFunc: func(createEvent event.CreateEvent) bool {
+			if createEvent.Meta.GetAnnotations()[nodeconfig.VersionAnnotation] != version.Get() {
+				return true
+			}
+			return false
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			if e.MetaNew.GetAnnotations()[nodeconfig.VersionAnnotation] != version.Get() {
+				return true
+			}
+			return false
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return false
+		},
+	})
+	if err != nil {
+		return errors.Wrap(err, "could not create watch on node objects")
+	}
+
+	return nil
+}
+
+// nodeToMachineMapper fulfills the mapper interface and allows for the mapping from a node to the associated Machine
+type nodeToMachineMapper struct {
+	client client.Client
+}
+
+// newNodeToMachineMapper returns a pointer to a new nodeToMachineMapper
+func newNodeToMachineMapper(client client.Client) *nodeToMachineMapper {
+	return &nodeToMachineMapper{client: client}
+}
+
+// Map maps Windows nodes to machines
+func (m *nodeToMachineMapper) Map(object handler.MapObject) []reconcile.Request {
+	node := core.Node{}
+
+	// If for some reason this mapper is called on an object which is not a Node, return
+	if kind := object.Object.GetObjectKind().GroupVersionKind(); kind.Kind != node.Kind {
+		return nil
+	}
+	if object.Meta.GetLabels()[core.LabelOSStable] != "windows" {
+		return nil
+	}
+
+	// Map the Node to the associated Machine through the Node's UID
+	machines := &mapi.MachineList{}
+	err := m.client.List(context.TODO(), machines,
+		client.MatchingLabels(map[string]string{windowsOSLabel: "Windows"}))
+	if err != nil {
+		log.Error(err, "could not get a list of machines")
+	}
+	for _, machine := range machines.Items {
+		if machine.Status.NodeRef != nil && machine.Status.NodeRef.UID == object.Meta.GetUID() {
+			return []reconcile.Request{
+				{
+					NamespacedName: types.NamespacedName{
+						Namespace: machine.GetNamespace(),
+						Name:      machine.GetName(),
+					},
+				},
+			}
+		}
+	}
+
+	// Node doesn't match a machine, return
 	return nil
 }
 
@@ -203,7 +280,26 @@ func (r *ReconcileWindowsMachine) Reconcile(request reconcile.Request) (reconcil
 		}
 
 		if _, present := node.Annotations[nodeconfig.VersionAnnotation]; present {
-			// version annotation exists, node is fully configured, do nothing.
+			// version annotation doesn`t match the current operator version
+			if node.Annotations[nodeconfig.VersionAnnotation] != version.Get() {
+
+				if !r.isAllowedDeletion(machine) {
+					r.recorder.Eventf(machine, core.EventTypeWarning, "MachineDeletionRestricted", "Machine %v deletion restricted due to exceeded number of unhealthy machines. ", machine.Name)
+					return reconcile.Result{RequeueAfter: requeueDuration}, nil
+				}
+				if !machine.GetDeletionTimestamp().IsZero() {
+					// Delete already initiated
+					return reconcile.Result{}, nil
+				}
+
+				if err := r.client.Delete(context.TODO(), machine); err != nil {
+					r.recorder.Eventf(machine, core.EventTypeWarning, "MachineDeletionFailed", "Machine %v deletion failed: unable to delete Machine object: %v", machine.Name, err)
+					return reconcile.Result{}, err
+				}
+				r.recorder.Eventf(machine, core.EventTypeNormal, "MachineDeleted", "Machine %v has been remediated by requesting to delete Machine object", machine.Name)
+				return reconcile.Result{}, nil
+			}
+			// version annotation exists with a valid value, node is fully configured, do nothing.
 			return reconcile.Result{}, nil
 		}
 	} else if *machine.Status.Phase != provisionedPhase {
@@ -295,4 +391,44 @@ func (r *ReconcileWindowsMachine) validateUserData(privateKey []byte) error {
 		return errors.Errorf("invalid content for userData secret")
 	}
 	return nil
+}
+
+// isAllowedDeletion determines if the number of machines after deletion of the given machine doesn`t fall below the
+// minHealthyCount
+func (r *ReconcileWindowsMachine) isAllowedDeletion(machine *mapi.Machine) bool {
+	if len(machine.OwnerReferences) == 0 {
+		return false
+	}
+	machinesetName := machine.OwnerReferences[0].Name
+
+	machines := &mapi.MachineList{}
+	err := r.client.List(context.TODO(), machines,
+		client.MatchingLabels(map[string]string{windowsOSLabel: "Windows"}))
+	if err != nil {
+		return false
+	}
+
+	// get Windows MachineSet
+	windowsMachineSet := &mapi.MachineSet{}
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: machinesetName,
+		Namespace: "openshift-machine-api"}, windowsMachineSet)
+	if err != nil {
+		return false
+	}
+
+	if minHealthyCount == *(windowsMachineSet.Spec.Replicas) {
+		return true
+	}
+
+	totalHealthy := 0
+	for _, ma := range machines.Items {
+		// Machines are determined as Healthy when they are part of Windows MachineSet and are
+		// in the Running Status
+		if len(ma.OwnerReferences) != 0 && ma.OwnerReferences[0].Name == machinesetName {
+			if ma.Status.Phase != nil && *ma.Status.Phase == "Running" && ma.Status.NodeRef != nil {
+				totalHealthy += 1
+			}
+		}
+	}
+	return totalHealthy-1 >= minHealthyCount
 }
--- a/test/e2e/create_test.go
+++ b/test/e2e/create_test.go
@@ -26,7 +26,6 @@ func creationTestSuite(t *testing.T) {
 	testCtx, err := NewTestContext(t)
 	require.NoError(t, err)
 	require.NoError(t, testCtx.createPrivateKeySecret(), "could not create private key secret")
-	defer testCtx.deletePrivateKeySecret()
 
 	// The order of tests here are important. testValidateSecrets is what populates the windowsVMs slice in the gc.
 	// testNetwork needs that to check if the HNS networks have been installed. Ideally we would like to run testNetwork
@@ -76,7 +75,7 @@ func testWindowsNodeCreation(t *testing.T) {
 		if err := testCtx.createWindowsMachineSet(test.replicas, test.isWindows); err != nil {
 			t.Fatalf("failed to create Windows MachineSet %v ", err)
 		}
-		err := testCtx.waitForWindowsNodes(test.replicas, true, !test.isWindows)
+		err := testCtx.waitForWindowsNodes(test.replicas, true, !test.isWindows, false)
 
 		if err != nil && test.isWindows {
 			t.Fatalf("windows node creation failed  with %v", err)
@@ -104,10 +103,11 @@ func (tc *testContext) createWindowsMachineSet(replicas int32, windowsLabel bool
 // if expectError = true, the function will wait for duration of 5 minutes for the nodes as the error would be thrown
 // immediately, else we will wait for the duration given by nodeCreationTime variable which is 20 minutes increasing
 // the overall wait time in test suite
-func (tc *testContext) waitForWindowsNodes(nodeCount int32, waitForAnnotations, expectError bool) error {
+func (tc *testContext) waitForWindowsNodes(nodeCount int32, waitForAnnotations, expectError, checkVersion bool) error {
 	var nodes *v1.NodeList
 	annotations := []string{nodeconfig.HybridOverlaySubnet, nodeconfig.HybridOverlayMac, nodeconfig.VersionAnnotation}
 	var creationTime time.Duration
+	startTime := time.Now()
 	if expectError {
 		// The time we expect to wait, if the windowsLabel is
 		// not used while creating nodes.
@@ -148,6 +148,16 @@ func (tc *testContext) waitForWindowsNodes(nodeCount int32, waitForAnnotations, 
 					return false, nil
 				}
 			}
+			if checkVersion {
+				operatorVersion, err := getWMCOVersion()
+				if err != nil {
+					log.Printf("error getting operator version : %v", err)
+					return false, nil
+				}
+				if node.Annotations[nodeconfig.VersionAnnotation] != operatorVersion {
+					return false, nil
+				}
+			}
 		}
 
 		return true, nil
@@ -155,6 +165,9 @@ func (tc *testContext) waitForWindowsNodes(nodeCount int32, waitForAnnotations, 
 
 	// Initialize/update nodes to avoid staleness
 	gc.nodes = nodes.Items
+	// Log the time elapsed while waiting for creation of the nodes
+	endTime := time.Now()
+	log.Printf("%v time is required to configure %v nodes", endTime.Sub(startTime), gc.numberOfNodes)
 
 	return err
 }
@@ -187,9 +200,4 @@ func (tc *testContext) createPrivateKeySecret() error {
 	}
 	_, err = tc.kubeclient.CoreV1().Secrets(tc.namespace).Create(context.TODO(), &privateKeySecret, metav1.CreateOptions{})
 	return err
-}
-
-// deletePrivateKeySecret deletes the private key secret
-func (tc *testContext) deletePrivateKeySecret() error {
-	return tc.kubeclient.CoreV1().Secrets(tc.namespace).Delete(context.TODO(), secrets.PrivateKeySecret, metav1.DeleteOptions{})
 }
--- a/test/e2e/delete_test.go
+++ b/test/e2e/delete_test.go
@@ -12,6 +12,8 @@ import (
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/openshift/windows-machine-config-operator/pkg/controller/secrets"
 )
 
 func deletionTestSuite(t *testing.T) {
@@ -52,7 +54,7 @@ func testWindowsNodeDeletion(t *testing.T) {
 		t.Fatalf("error updating windowsMachineSet custom resource  %v", err)
 	}
 	// As per testing, each windows VM is taking roughly 12 minutes to be shown up in the cluster
-	err = testCtx.waitForWindowsNodes(gc.numberOfNodes, true, true)
+	err = testCtx.waitForWindowsNodes(gc.numberOfNodes, true, true, false)
 	if err != nil {
 		t.Fatalf("windows node deletion failed  with %v", err)
 	}
@@ -68,4 +70,8 @@ func testWindowsNodeDeletion(t *testing.T) {
 	// Cleanup secrets created by us.
 	err = framework.Global.KubeClient.CoreV1().Secrets("openshift-machine-api").Delete(context.TODO(), "windows-user-data", meta.DeleteOptions{})
 	require.NoError(t, err, "could not delete userData secret")
+
+	err = framework.Global.KubeClient.CoreV1().Secrets("windows-machine-config-operator").Delete(context.TODO(), secrets.PrivateKeySecret, meta.DeleteOptions{})
+	require.NoError(t, err, "could not delete privateKey secret")
+
 }
--- a/test/e2e/upgrade_test.go
+++ b/test/e2e/upgrade_test.go
@@ -1,0 +1,150 @@
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"testing"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	nc "github.com/openshift/windows-machine-config-operator/pkg/controller/windowsmachine/nodeconfig"
+)
+
+const (
+	// deploymentRetryInterval is the retry time for WMCO deployment to scale up/down
+	deploymentRetryInterval = time.Second * 10
+	// deploymentTimeout is the maximum duration to update WMCO deployment
+	deploymentTimeout = time.Minute * 1
+	// resourceName is the name of a resource in the watched namespace (e.g pod name, deployment name)
+	resourceName = "windows-machine-config-operator"
+)
+
+// upgradeTestSuite tests behaviour of the operator when an upgrade takes place.
+func upgradeTestSuite(t *testing.T) {
+	testCtx, err := NewTestContext(t)
+	require.NoError(t, err)
+
+	// apply configuration steps before running the upgrade tests
+	err = testCtx.configureUpgradeTest()
+	require.NoError(t, err, "error configuring upgrade")
+
+	t.Run("Operator version upgrade", testUpgradeVersion)
+	t.Run("Version annotation tampering", testTamperAnnotation)
+}
+
+// testUpgradeVersion tests the upgrade scenario of the operator. The node version annotation is changed when
+// the operator is shut-down. The function tests if the operator on restart deletes the machines and recreates
+// them on version annotation mismatch.
+func testUpgradeVersion(t *testing.T) {
+	testCtx, err := NewTestContext(t)
+	require.NoError(t, err)
+
+	err = testCtx.waitForWindowsNodes(gc.numberOfNodes, true, false, true)
+	require.NoError(t, err, "windows node upgrade failed")
+	// Test if the version annotation corresponds to the current operator version
+	testVersionAnnotation(t)
+}
+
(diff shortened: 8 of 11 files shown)


test/e2e/validation_test.go
```diff
@@ -49,9 +49,9 @@ func testWorkerLabel(t *testing.T) {
 	}
 }
 
-// getWMCOVersion returns the version of the operator. This is sourced from the WMCO binary used to create the operator image.
+//// getWMCOVersion returns the version of the operator. This is sourced from the WMCO binary used to create the operator image.
```
PratikMahajan, line 52: unnecessary `//`

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -313,3 +363,45 @@ func (r *ReconcileWindowsMachine) validateUserData() error {
...
 }
+
+// isAllowedDeletion determines if the number of machines after deletion of the given machine doesn`t fall below the
+// minHealthyCount
+func (r *ReconcileWindowsMachine) isAllowedDeletion(machine *mapi.Machine) bool {
+	machinesetName := machine.OwnerReferences[0].Name
+
+	machines := &mapi.MachineList{}
+	err := r.client.List(context.TODO(), machines,
+		client.MatchingLabels(map[string]string{"machine.openshift.io/os-id": "Windows"}))
```
PratikMahajan, line 374: can we make the label as a constant?

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -130,6 +137,12 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	err = c.Watch(&source.Kind{Type: &core.Node{
 		ObjectMeta: meta.ObjectMeta{Namespace: ""},
 	}}, &handler.EnqueueRequestsFromMapFunc{ToRequests: newNodeToMachineMapper(mgr.GetClient())}, predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
```
sebsoto, line 140: Create events need to handled as well, when the operator comes up the existing machines will be given as create events

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -30,11 +26,22 @@ import (
...
+	"github.com/openshift/windows-machine-config-operator/version"
 )
 
 const (
 	// ControllerName is the name of the WindowsMachine controller
 	ControllerName = "windowsmachine-controller"
+	// minHealthyCount is the minimum number of nodes that are required to be in running phase at a given time.
+	minHealthyCount = 1
+	// VersionAnnotation indicates the version of WMCO that configured the node
+	VersionAnnotation = "windowsmachineconfig.openshift.io/version"
```
sebsoto, line 44: Why not use the exported VersionAnnotation from the nodeconfig package? We should try and keep constants like this defined in one place so they are easy to change if needed

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -313,3 +363,45 @@ func (r *ReconcileWindowsMachine) validateUserData() error {
...
+
+// isAllowedDeletion determines if the number of machines after deletion of the given machine doesn`t fall below the
+// minHealthyCount
+func (r *ReconcileWindowsMachine) isAllowedDeletion(machine *mapi.Machine) bool {
+	machinesetName := machine.OwnerReferences[0].Name
+
+	machines := &mapi.MachineList{}
+	err := r.client.List(context.TODO(), machines,
+		client.MatchingLabels(map[string]string{"machine.openshift.io/os-id": "Windows"}))
+
```
sebsoto, line 375: extra newline

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -313,3 +363,45 @@ func (r *ReconcileWindowsMachine) validateUserData() error {
...
+	if err != nil {
+		return false
+	}
+
+	// get Windows MachineSet
+	windowsMachineSet := &mapi.MachineSet{}
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: machinesetName,
+		Namespace: "openshift-machine-api"}, windowsMachineSet)
+
```
sebsoto, line 384: extra newline

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,131 @@
...
+	"math"
+	"testing"
+	"time"
+
+	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	nc "github.com/openshift/windows-machine-config-operator/pkg/controller/windowsmachine/nodeconfig"
```
sebsoto, line 15: Please try to preserve the 3 import groups @LorbusChris talked about in standup

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,131 @@
...
+	framework "github.com/operator-framework/operator-sdk/pkg/test"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"log"
+)
+
+func upgradeTestSuite(t *testing.T) {
+
```
sebsoto, line 24: extra newline

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,131 @@
...
+	nc "github.com/openshift/windows-machine-config-operator/pkg/controller/windowsmachine/nodeconfig"
+	framework "github.com/operator-framework/operator-sdk/pkg/test"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"log"
+)
+
+func upgradeTestSuite(t *testing.T) {
```
sebsoto, line 23: Comment needed

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+	t.Run("Node deletion during upgrade", func(t *testing.T) { testUpgradeVersion(t) })
```
sebsoto, line 25: This name leads me to believe that this is testing if a user deletes a node during the upgrade process

sebsoto (changes requested):

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,131 @@
...
+	for _, node := range nodes.Items {
+		t.Run(node.GetName(), func(t *testing.T) {
+			require.Containsf(t, node.Annotations, nc.VersionAnnotation, "node %s missing version annotation", node.GetName())
+			assert.Equal(t, operatorVersion, node.Annotations[nc.VersionAnnotation], "Node recreation failed. WMCO version annotation mismatch")
+		})
+	}
+}
+
+// waitForWindowsNodesAnnotation waits until the given number of nodes are created with the expected version annotation
+func (tc *testContext) waitForWindowsNodesAnnotation(nodeCount int32, operatorVersion string) error {
```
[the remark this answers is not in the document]
VaishnaviHire: Yes, I did not use `waitForWindowsNodes` because it takes time for nodes to be deleted and due to that `waitForWindowsNodes` will immediately retrurn true. To avoid that, I had to add [this](https://github.com/openshift/windows-machine-config-operator/pull/101/files#diff-225c6dc1116ed17311dbd31aef8e09e9R119) condition and hence needed to create this function.

pkg/controller/windowsmachine/windowsmachine_controller.go
VaishnaviHire: Create events are being handled for node objects. Is your suggestion to handle `create` events for Machine objects as well?

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -159,7 +172,7 @@ func (m *nodeToMachineMapper) Map(object handler.MapObject) []reconcile.Request
 	if kind := object.Object.GetObjectKind().GroupVersionKind(); kind.Kind != node.Kind {
 		return nil
 	}
-	if object.Meta.GetLabels()[core.LabelOSStable] != "windows" {
+	if object.Meta.GetLabels()["node.openshift.io/os_id"] != "Windows" {
```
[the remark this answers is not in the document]
VaishnaviHire: core.LabelOSStable equals to `"kubernetes.io/os"`, did not find that label in one of nodes while testing.

pkg/controller/windowsmachine/windowsmachine_controller.go
VaishnaviHire: Same as [here](https://github.com/openshift/windows-machine-config-operator/pull/101#discussion_r486633614)

test/e2e/wmco_test.go
```diff
@@ -33,13 +40,22 @@ func TestWMCO(t *testing.T) {
...
+	gc.skipNodeUpgrade = skipNodeUpgrade
 	// When the OPENSHIFT_CI env var is set to true, the test is running within CI
 	if inCI := os.Getenv("OPENSHIFT_CI"); inCI == "true" {
 		// In the CI container the WMCO binary will be found here
 		wmcoPath = "/usr/local/bin/windows-machine-config-operator"
 	}
 
 	t.Run("create", creationTestSuite)
+
+	if !gc.skipNodeUpgrade {
```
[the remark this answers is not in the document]
VaishnaviHire: If this flag is not added, we would go ahead and do the `preconfiguration` required by the `upgrade` test suite

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -230,6 +242,44 @@ func (r *ReconcileWindowsMachine) Reconcile(request reconcile.Request) (reconcil
...
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		if node.Annotations[VersionAnnotation] != version.Get() {
+
+			if !r.isAllowedDeletion(machine) {
+				r.recorder.Eventf(machine, core.EventTypeWarning,
+					"MachineDeletionRestricted", "Machine %v deletion restricted due to exceeded number of unhealthy machines. ",
+					machine.Name)
+				return reconcile.Result{RequeueAfter: time.Minute * 5}, nil
```
aravindhp, line 258: Please make `time.Minute * 5` a constant.

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -230,6 +242,44 @@ func (r *ReconcileWindowsMachine) Reconcile(request reconcile.Request) (reconcil
...
+
+			if err := r.client.Delete(context.TODO(), machine); err != nil {
+				r.recorder.Eventf(
+					machine,
+					core.EventTypeWarning,
+					"MachineDeletionFailed",
+					"Machine %v deletion failed: unable to delete Machine object: %v",
+					machine.Name,
+					err,
+				)
```
aravindhp, line 273: Is there a reason for different formatting of the event here vs other places?

pkg/controller/windowsmachine/windowsmachine_controller.go
aravindhp: Don't you need to check the version annotation in the create events also?

deploy/olm-catalog/windows-machine-config-operator/manifests/windows-machine-config-operator.clusterserviceversion.yaml
```diff
@@ -81,6 +81,15 @@ spec:
...
           - watch
+          - delete
+        - apiGroups:
+          - machine.openshift.io
+          resources:
+          - machinesets
+          verbs:
+          - list
+          - get
+          - watch
```
aravindhp, line 92: Why do we need `watch`?

deploy/role.yaml
```diff
@@ -137,3 +137,12 @@ rules:
      - get
      - list
      - watch
+     - delete
+ - apiGroups:
+     - "machine.openshift.io"
```
aravindhp, line 142: No need for quotes.

hack/run-ci-e2e-test.sh
```diff
@@ -95,8 +100,11 @@ fi
 
 # The bool flags in golang does not respect key value pattern. They follow -flag=x pattern.
 # -flag x is allowed for non-boolean flags only(https://golang.org/pkg/flag/)
-# Run the creation tests and skip deletion of the Windows VMs
-OSDK_WMCO_test $OSDK "-run=TestWMCO/create -v -timeout=90m -node-count=$NODE_COUNT -skip-node-deletion -ssh-key-pair=$KEY_PAIR_NAME $WMCO_PATH_OPTION"
+# Run the creation tests and skip deletion and upgrade of the Windows VMs
+OSDK_WMCO_test $OSDK "-run=TestWMCO/create -v -timeout=90m -node-count=$NODE_COUNT -skip-node-upgrade -skip-node-deletion -ssh-key-pair=$KEY_PAIR_NAME $WMCO_PATH_OPTION"
+
```
aravindhp, line 105: Don't you need to change the version annotation before running the upgrade tests?

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,131 @@
...
+func testUpgradeVersion(t *testing.T) {
+	testCtx, err := NewTestContext(t)
+	require.NoError(t, err)
+
+	operatorVersion, err := testCtx.getWMCOVersion()
+	require.NoError(t, err, "could not get WMCO version")
+
+	err = testCtx.waitForWindowsNodesAnnotation(gc.numberOfNodes, operatorVersion)
+	if err != nil {
+		t.Fatalf("windows node upgrade failed  with %v", err)
```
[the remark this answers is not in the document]
aravindhp: Correct unless you want to hard fail here.

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,131 @@
...
+	operatorVersion, err := testCtx.getWMCOVersion()
+	require.NoError(t, err, "could not get WMCO version")
+
+	// tamper node annotation
+	nodes, err := framework.Global.KubeClient.CoreV1().Nodes().List(context.TODO(),
+		metav1.ListOptions{LabelSelector: nc.WindowsOSLabel})
+	require.NoError(t, err)
+
+	for _, node := range nodes.Items {
+		patchData := fmt.Sprintf(`{"metadata":{"annotations":{"%s":"%s"}}}`, nc.VersionAnnotation, "0.0.2+baa8c1f-dirty")
```
aravindhp, line 71: There is potential for `0.0.2+baa8c1f-dirty` to be an actual version :-) I suggest something completely bogus like "bad-version"

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,131 @@
...
+	"log"
+)
+
+func upgradeTestSuite(t *testing.T) {
+
+	t.Run("Node deletion during upgrade", func(t *testing.T) { testUpgradeVersion(t) })
+	t.Run("Node deletion during annotation tampering", func(t *testing.T) { testTamperAnnotation(t) })
+}
+
+// testUpgradeVersion tests if the operator deletes machines and recreats them, if the node annotation doesn`t match
```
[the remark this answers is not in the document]
aravindhp: Please call out that this is testing an actual upgrade scenario i.e. the operator is shutdown, node versions are changed and operator is restarted.

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,131 @@
...
+	for _, node := range nodes.Items {
+		t.Run(node.GetName(), func(t *testing.T) {
+			require.Containsf(t, node.Annotations, nc.VersionAnnotation, "node %s missing version annotation", node.GetName())
+			assert.Equal(t, operatorVersion, node.Annotations[nc.VersionAnnotation], "Node recreation failed. WMCO version annotation mismatch")
+		})
+	}
+}
+
+// testTamperAnnotation tests if the operator deletes machines and recreats them, if the node annotation is changed to an invalid value
+// with the expected annotation
```
aravindhp, line 57: Please call out that, this is when the operator is running.

test/e2e/upgrade_test.go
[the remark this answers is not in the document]
aravindhp: Is there a way to share the common code by adding a parameter input? Only if that parameter is true will it check `if currAnn != operatorVersion`

test/e2e/wmco_test.go
```diff
@@ -54,3 +70,27 @@ func setupWMCOResources() error {
 	}
 	return nil
 }
+
+// configureForUpgrade changes the node annotations of all the configured Windows nodes.
+func configureForUpgrade() error {
```
aravindhp, line 75: How do you guarantee that the operator is not running at this point? I feel this should be done outside of the test suite to simulate the upgrade scenario.

test/e2e/wmco_test.go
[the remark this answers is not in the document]
aravindhp: I was thinking along the same lines as @sebsoto. We should move away from this pattern of adding more flags if possible.

@ravisantoshgudimetla I forget why you added the  flag given you are explicitly saying `-run=TestWMCO/destroy`.

aravindhp (changes requested):
Nice work, @VaishnaviHire.

You don't need separate commits in this case. The reason for having separate commits is to separate clearly distinct pieces of work needed in the same PR. Here the second commit won't take effect without the first.

pkg/controller/windowsmachine/windowsmachine_controller.go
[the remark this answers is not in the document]
sebsoto: If this is true this is an issue, the must-gather changes I made required that label https://github.com/openshift/oc/pull/509/files

I was under the impression that `stable` parts of the API should not just be changed like that?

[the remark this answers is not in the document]
VaishnaviHire: @sebsoto Yes you are right. I see the "kubernetes.io/os" label for every node recreated in the tests. I will revert this change.

deploy/olm-catalog/windows-machine-config-operator/manifests/windows-machine-config-operator.clusterserviceversion.yaml
VaishnaviHire: [Here](https://github.com/openshift/windows-machine-config-operator/pull/101/files#diff-9e63f7b0791edcb4a53c944552b88970R155) we are allowing for watching  of resources within the `openshift-machine-api`. So if watch is not added we get ` Failed to watch *unstructured.Unstructured: unknown` error

aravindhp: This is strange given that we only want to watch Machines in the `openshift-machine-api` namespace.

VaishnaviHire: Looked more into the error. Looks like cache informers use `list` and `watch` calls.  https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/cache/internal/informers_map.go#L207

test/e2e/upgrade_test.go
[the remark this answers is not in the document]
VaishnaviHire: I think we should hard fail here. The `err` implies that on version change, the nodes are not recreated and configured as expected. So the further tests would fail anyway

test/e2e/upgrade_test.go
VaishnaviHire: Comments added in individual test runs

test/e2e/create_test.go
```diff
@@ -136,6 +137,15 @@ func (tc *testContext) waitForWindowsNodes(nodeCount int32, waitForAnnotations,
 					return false, nil
 				}
 			}
+			if checkVersion {
+				operatorVersion, err := getWMCOVersion()
+				if err != nil {
+					return false, nil
```
aravindhp, line 143: s/nil/err

test/e2e/upgrade_test.go
aravindhp: I suggest adding a comment saying what feature the test suite is exercising.

test/e2e/wmco_test.go
```diff
@@ -40,6 +40,9 @@ func TestWMCO(t *testing.T) {
 	}
 
 	t.Run("create", creationTestSuite)
+
```
aravindhp, line 43: I suggest removing this line.

aravindhp (changes requested):

hack/run-ci-e2e-test.sh
```diff
@@ -98,14 +107,18 @@ fi
 # Run the creation tests and skip deletion of the Windows VMs
 OSDK_WMCO_test $OSDK "-run=TestWMCO/create -v -timeout=90m -node-count=$NODE_COUNT -skip-node-deletion -ssh-key-pair=$KEY_PAIR_NAME $WMCO_PATH_OPTION"
 
-# Run the deletion tests while testing operator restart functionality. This will clean up VMs created
-# in the previous step
+## Run the upgrade tests and skip deletion of the Windows VMs
+update_version_annotation
```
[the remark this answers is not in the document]
VaishnaviHire: Yes. I can explicitly stop the operator and restart it. I am not sure about the [restart](https://github.com/openshift/windows-machine-config-operator/pull/101/files/423f1b25a021f776288a4b0b5016a06ce9fa876e#diff-98085f3fd979233fe093e8f731f11240R114) here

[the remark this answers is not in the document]
aravindhp: I don't think we were ever restarting :-) but I could be mistaken.

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,170 @@
...
+const (
+	// deploymentRetryInterval is the retry time for WMCO deployment to scale up/down
+	deploymentRetryInterval = time.Second * 10
+	// deploymentTimeout is the maximum duration to update WMCO deployment
+	deploymentTimeout = time.Minute * 1
+	// resourceName is the name of a resource in the watched namespace (e.g pod name, deployment name)
+	resourceName = "windows-machine-config-operator"
+)
+
+// upgradeTestSuite tests behaviour of the operator when an upgrade takes place. It
```
aravindhp, line 31: Please remove the trailing `It`

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,170 @@
...
+// The steps include -
+// 1. Scale down the operator to 0.
+// 2. Change Windows node version annotation to an invalid value
+// 3. Scale up the operator to 1
+func configureUpgradeTest() error {
+	// Scale down the WMCO deployment to 0
+	if err := scaleDeployment(0); err != nil {
+		return err
+	}
+
```
aravindhp, line 118: To really simulate the upgrade scenario, you should wait for the deployment to scale down to 0.

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,170 @@
...
+	nodes, err = framework.Global.KubeClient.CoreV1().Nodes().List(context.TODO(),
+		metav1.ListOptions{LabelSelector: nc.WindowsOSLabel})
+	require.NoError(t, err)
+
+	for _, node := range nodes.Items {
+		t.Run(node.GetName(), func(t *testing.T) {
+			require.Containsf(t, node.Annotations, nc.VersionAnnotation, "node missing version annotation")
+			assert.Equal(t, operatorVersion, node.Annotations[nc.VersionAnnotation], "Node recreation failed. WMCO version annotation mismatch")
+		})
+	}
```
aravindhp, line 105: This seems to be a repeated pattern and should be modularized.

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,170 @@
...
+			metav1.GetOptions{})
+
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				log.Printf("error getting deployment %v", err)
+				return false, nil
+			}
+			return false, err
+		}
+		if deployment.Spec.Replicas != &desiredReplicas {
```
aravindhp, line 158: Not sure if this check is needed. The patch won't take effect if `deployment.Spec.Replicas == &desiredReplicas`.

aravindhp (changes requested):

test/e2e/upgrade_test.go
VaishnaviHire: I am waiting for the deployment to scale up/down [here](https://github.com/openshift/windows-machine-config-operator/pull/101/files#diff-225c6dc1116ed17311dbd31aef8e09e9R147)

aravindhp: But you are only checking `Spec.Replicas` there. That is going to take effect as soon as you patch deployment. Or am I missing something?

test/e2e/upgrade_test.go
VaishnaviHire: I added that check because, we won`t need to patch the WMCO deployment if the number of replicas is equal to the given input.

aravindhp: True, I am just calling out it is a bit redundant. I am fine if you want to retain it.

test/e2e/upgrade_test.go
VaishnaviHire: Right. Thanks for catching that.

aravindhp (approved):
Nice work, @VaishnaviHire 

/approve

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -26,15 +24,22 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
+	"strings"
```
LorbusChris, line 27: this line needs to stay further up

test/e2e/delete_test.go
```diff
@@ -2,6 +2,7 @@ package e2e
 
 import (
 	"context"
+	"github.com/openshift/windows-machine-config-operator/pkg/controller/secrets"
```
LorbusChris, line 5: please move this into a third imports group

LorbusChris (changes requested):
Thanks @VaishnaviHire! Just two small nits.

hack/run-ci-e2e-test.sh
```diff
@@ -108,6 +108,9 @@ OSDK_WMCO_test $OSDK "-run=TestWMCO/operator_deployed_without_private_key_secret
 # Run the creation tests and skip deletion of the Windows VMs
 OSDK_WMCO_test $OSDK "-run=TestWMCO/create -v -timeout=90m -node-count=$NODE_COUNT -skip-node-deletion -ssh-key-pair=$KEY_PAIR_NAME --private-key-path=$KUBE_SSH_KEY_PATH $WMCO_PATH_OPTION"
 
+# Run the upgrade tests and skip deletion of the Windows VMs
+OSDK_WMCO_test $OSDK "-run=TestWMCO/upgrade -v -timeout=90m -node-count=$NODE_COUNT -skip-node-deletion -ssh-key-pair=$KEY_PAIR_NAME --private-key-path=$KUBE_SSH_KEY_PATH $WMCO_PATH_OPTION"
```
sebsoto, line 112: -skip-node-deletion is not needed. since you specified `-run=...upgrade` the deletion tests will not run

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -296,3 +388,40 @@ func (r *ReconcileWindowsMachine) validateUserData(privateKey []byte) error {
 	}
 	return nil
 }
+
+// isAllowedDeletion determines if the number of machines after deletion of the given machine doesn`t fall below the
+// minHealthyCount
+func (r *ReconcileWindowsMachine) isAllowedDeletion(machine *mapi.Machine) bool {
+	machinesetName := machine.OwnerReferences[0].Name
```
sebsoto, line 395: I am not sure if we should use this.
We may want to instead use the `machine.openshift.io/cluster-api-machineset` label on the machine.
The label seems like it is more likely to supported for longer

test/e2e/delete_test.go
```diff
@@ -66,6 +67,10 @@ func testWindowsNodeDeletion(t *testing.T) {
 	}
 
 	// Cleanup secrets created by us.
-	err = framework.Global.KubeClient.CoreV1().Secrets("openshift-machine-api").Delete(context.TODO(), "windows-user-data", meta.DeleteOptions{})
+	err = framework.Global.KubeClient.CoreV1().Secrets(testCtx.namespace).Delete(context.TODO(), "windows-user-data", meta.DeleteOptions{})
```
sebsoto, line 70: testCtx.namespace would be the tests created to deploy pods, not the machine-api namespace. "windows-user-data" wouldnt be in there.

test/e2e/delete_test.go
```diff
@@ -66,6 +67,10 @@ func testWindowsNodeDeletion(t *testing.T) {
...
 	require.NoError(t, err, "could not delete userData secret")
+
+	err = framework.Global.KubeClient.CoreV1().Secrets(testCtx.namespace).Delete(context.TODO(), secrets.PrivateKeySecret, meta.DeleteOptions{})
```
sebsoto, line 73: testCtx.namespace would be the tests created to deploy pods, not the machine-api namespace. the private key wouldnt be in there.

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,150 @@
...
+	if err := scaleDeployment(1); err != nil {
+		return err
+	}
+	return nil
+}
+
+// scaleDeployment scales the WMCO operator to the given replicas. If the deployment is managed by OLM, updating the
+// replicas only scales the deployment to 0 or 1. If we want to scale the deployment to more than 1 replicas, we need to
+// make changes in replicas defined in the corresponding CSV.
+func scaleDeployment(desiredReplicas int32) error {
```
sebsoto, line 122: please rename this to `scaleWMCODeployment` so its clear what is being scaled

test/e2e/upgrade_test.go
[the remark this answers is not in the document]
sebsoto: [Fatal](https://golang.org/pkg/testing/#T.Fatal) is just a logging wrapper around `FailNow()`. This is the same as [require](https://github.com/stretchr/testify/blob/master/require/require.go#L1199) 

The two are functionally identical but with require we get cleaner code :)

If you want to prevent future tests from running if this test fails, you should restructure the tests making the dependent tests subtests.

cc @aravindhp

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,150 @@
...
+		_, err := framework.Global.KubeClient.CoreV1().Nodes().Patch(context.TODO(), node.Name, types.MergePatchType, []byte(patchData), metav1.PatchOptions{})
+		require.NoError(t, err)
+		if err == nil {
+			break
+		}
+	}
+
+	err = testCtx.waitForWindowsNodes(1, true, false, true)
+	if err != nil {
+		t.Fatal(t, "windows node upgrade failed  with %v", err)
```
sebsoto, line 76: Please use require see other comment about fatal for explanation

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,150 @@
...
+	deploymentTimeout = time.Minute * 1
+	// resourceName is the name of a resource in the watched namespace (e.g pod name, deployment name)
+	resourceName = "windows-machine-config-operator"
+)
+
+// upgradeTestSuite tests behaviour of the operator when an upgrade takes place.
+func upgradeTestSuite(t *testing.T) {
+
+	if err := configureUpgradeTest(); err != nil {
+		t.Fatalf("error configuring upgrade %v", err)
```
sebsoto, line 33: Please use require see other comment about fatal for explanation

sebsoto (changes requested):
Thanks @VaishnaviHire, PR is looking very good! A few comments mainly about the test suite

pkg/controller/windowsmachine/windowsmachine_controller.go
VaishnaviHire: Yes, I tried using the label earlier. However, it doesn`t hold true for our test MachineSets.
e2e Tests -

` MachineSet Name : e2e-windows-machineset-with-windows-label-6ptb`

`machine.openshift.io/cluster-api-machineset=e2e-windows-machineset-with-windows-label-us-west-1a`

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -29,11 +26,24 @@ import (
...
+	"github.com/openshift/windows-machine-config-operator/version"
 )
 
 const (
 	// ControllerName is the name of the WindowsMachine controller
 	ControllerName = "windowsmachine-controller"
+	// minHealthyCount is the minimum number of nodes that are required to be in running phase at a given time.
+	minHealthyCount = 1
+	// windowsOSLabel is the label used to identify the Windows Machines.
+	windowsOSLabel = "machine.openshift.io/os-id"
```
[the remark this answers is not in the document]
VaishnaviHire: Did not find any in `github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1`

hack/run-ci-e2e-test.sh
VaishnaviHire: Removing this from upgrade here. Creating a separate PR to remove the flag itself.

aravindhp: @VaishnaviHire yes I too think the `-skip-node-deletion` flag is not needed in general. So removing it in a separate PR is fine. However you don't need to add it to this line if it is not needed.

README.md
```diff
@@ -299,3 +299,15 @@ spec:
           userDataSecret:
             name: windows-user-data
 ```
+### Windows nodes k8s component upgrade
```
aravindhp, line 302: Please use `Kubernetes`

README.md
```diff
@@ -299,3 +299,15 @@ spec:
...
+
+When a new version of WMCO is released that is compatible with the current cluster version, an update can take place.
```
aravindhp, line 304: `, an operator upgrade will take place which will result in the Kubernetes components in the Windows machine to be upgraded.`

README.md
```diff
@@ -299,3 +299,15 @@ spec:
...
+For a non-disruptive upgrade, WMCO terminates the Windows VMs configured by previous version of WMCO and recreates them
```
aravindhp, line 305: Lets go with using `Machine` instead of VM.

README.md
```diff
@@ -299,3 +299,15 @@ spec:
...
 ```
+### Windows nodes k8s component upgrade
+
+When a new version of WMCO is released that is compatible with the current cluster version, an update can take place.
+For a non-disruptive upgrade, WMCO terminates the Windows VMs configured by previous version of WMCO and recreates them 
+using the current version. This is done by deleting the Machine object that results in the drain and deletion of the 
+Windows node. To facilitate an upgrade, WMCO adds a version annotation to all the configured nodes. During an upgrade, 
+a mismatch in version annotation will result in deletion and recreation of Windows VM. WMCO takes into consideration the
+`minHealthy`(default=1) count while deleting the nodes. So, at any time during upgrade, the cluster will have
+`minHealthy` Windows machines per MachineSet.
```
aravindhp, line 310: Please don't call out `minHealthy` here as it is not configurable. Just mention how we will maintain 1 healthy node during upgrades.

README.md
```diff
@@ -299,3 +299,15 @@ spec:
...
+
+WMCO is not responsible for Windows Operating System updates. The cluster administrator provides the Window image while
```
aravindhp, line 312: `operating system`

README.md
```diff
@@ -299,3 +299,15 @@ spec:
...
+creating the VMs and hence, the cluster administrator is responsible for providing an updated image
```
aravindhp, line 313: `The cluster administrator can provide an updated image by changing the `image` in the MachineSet...`

aravindhp (changes requested):

aravindhp (approved):
/approve

pkg/controller/windowsmachine/windowsmachine_controller.go
aravindhp: I just saw this. I think you need to add a guard against `machine.OwnerReferences[0]` not existing at least.

test/e2e/create_test.go
```diff
@@ -26,7 +26,6 @@ func creationTestSuite(t *testing.T) {
 	testCtx, err := NewTestContext(t)
 	require.NoError(t, err)
 	require.NoError(t, testCtx.createPrivateKeySecret(), "could not create private key secret")
-	defer testCtx.deletePrivateKeySecret()
```
aravindhp, line 29: Ah now I see the issue we were seeing before :-) Good catch.

aravindhp (changes requested):

hack/run-ci-e2e-test.sh
```diff
@@ -103,6 +103,9 @@ OSDK_WMCO_test $OSDK "-run=TestWMCO/operator_deployed_without_private_key_secret
 # Run the creation tests of the Windows VMs
 OSDK_WMCO_test $OSDK "-run=TestWMCO/create -v -timeout=90m -node-count=$NODE_COUNT --private-key-path=$KUBE_SSH_KEY_PATH $WMCO_PATH_OPTION"
 
+# Run the upgrade tests and skip deletion of the Windows VMs
```
aravindhp, line 106: Please call `get_WMCO_logs` after running the creation test suite or we will lose the logs from the first pod.

pkg/controller/windowsmachine/windowsmachine_controller.go
aravindhp: >We may want to instead use the machine.openshift.io/cluster-api-machineset label on the machine.
The label seems like it is more likely to supported for longer

I prefer owner references as they are more authoritative. The only problem would be if `MachineSets` are in one namespace and the `Machines` are in another at which point owner refs are not supposed to be added due to the way [GC works](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#owners-and-dependents).

aravindhp (approved):
/approve

README.md
```diff
@@ -299,3 +299,17 @@ spec:
           userDataSecret:
             name: windows-user-data
 ```
+### Windows nodes Kubernetes component upgrade
```
sebsoto, line 302: Upgrading Kubernetes node components on Windows nodes

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -117,11 +125,80 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+	}}, &handler.EnqueueRequestForObject{}, machinePredicate)
 	if err != nil {
 		return errors.Wrap(err, "could not create watch on Machine objects")
 	}
 
+	err = c.Watch(&source.Kind{Type: &core.Node{
+		ObjectMeta: meta.ObjectMeta{Namespace: ""},
+	}}, &handler.EnqueueRequestsFromMapFunc{ToRequests: newNodeToMachineMapper(mgr.GetClient())}, predicate.Funcs{
+		CreateFunc: func(createEvent event.CreateEvent) bool {
+			if createEvent.Meta.GetAnnotations()[nodeconfig.VersionAnnotation] != version.Get() {
```
sebsoto, line 137: nit: This will filter out all linux nodes by default, but it might be good to be more verbose, and return if the node is not a Windows node

applies to all predicate funcs

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,146 @@
...
+)
+
+// upgradeTestSuite tests behaviour of the operator when an upgrade takes place.
+func upgradeTestSuite(t *testing.T) {
+	testCtx, err := NewTestContext(t)
+	require.NoError(t, err)
+
+	if err := testCtx.configureUpgradeTest(); err != nil {
+		require.NoError(t, err, "error configuring upgrade")
+	}
```
sebsoto, line 35: `if` is redundant, try:
```
err = testCtx.configureUpgradeTest()
require.NoError(t, err, "error configuring upgrade")
```

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,146 @@
...
+			types.MergePatchType, []byte(patchData), metav1.PatchOptions{})
+		if err != nil {
+			log.Printf("error patching operator deployment : %v", err)
+			return false, nil
+		}
+		return true, nil
+	})
+
+	// wait for the windows-machine-config-operator to scale up/down
+	if err == nil {
```
sebsoto, line 134: Instead of this, you should do
```
if err != nil {
  return ...
}
err = wait.Poll(...
```
This is a better way of doing this as
1) it reduces indentation. The rest of the function is indented which is something that is nice to avoid
2) the control flow is more obvious. If we get an error we want to return immediately. If this function was to be added to in the future things would get pretty messy.

sebsoto (changes requested):
Looks great, just a few comments

README.md
sebsoto: Eh actually I dont know how i feel about repeating `nodes` twice. Use your best judgement here

pkg/controller/windowsmachine/windowsmachine_controller.go
VaishnaviHire: We are checking for label [here](https://github.com/openshift/windows-machine-config-operator/pull/101/files#diff-d5d428eaf9d5786d351222f5184d99d9R177)

README.md
VaishnaviHire: I am keeping it as is for now

aravindhp:
/hold

@VaishnaviHire `aws-e2e-operator-with-custom-vxlan-port` has been consistently failing. Please look into it.

hack/run-ci-e2e-test.sh
```diff
@@ -102,6 +102,11 @@ OSDK_WMCO_test $OSDK "-run=TestWMCO/operator_deployed_without_private_key_secret
 
 # Run the creation tests of the Windows VMs
 OSDK_WMCO_test $OSDK "-run=TestWMCO/create -v -timeout=90m -node-count=$NODE_COUNT --private-key-path=$KUBE_SSH_KEY_PATH $WMCO_PATH_OPTION"
+# Get logs for the creation tests
```
aravindhp, line 105: Please add a `echo \n####### WMCO logs for creation tests #######\n` or something along those lines here. Finding it hard to figure out which pod logs belongs to which set of tests. Something similar would be needed before the other `get_WMCO_logs` also: `echo \n####### WMCO logs for upgrade and deletion tests #######\n`.

test/e2e/upgrade_test.go
```diff
@@ -45,7 +45,10 @@ func testUpgradeVersion(t *testing.T) {
 	testCtx, err := NewTestContext(t)
 	require.NoError(t, err)
 
+	startTime := time.Now()
 	err = testCtx.waitForWindowsNodes(gc.numberOfNodes, true, false, true)
+	endTime := time.Now()
+	log.Printf("%v time is required to configure %v nodes", endTime.Sub(startTime), gc.numberOfNodes)
```
aravindhp: Might be useful to add this in `waitForWindowsNodes` itself, so that we can see the time in all cases.

VaishnaviHire: Should I squash the second commit ?

aravindhp: Yes, please. I think none of it is throw away work.

test/e2e/wmco_test.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 var (
-	nodeCreationTime     = time.Minute * 20
+	nodeCreationTime     = time.Minute * 30
```
aravindhp, line 17: Please add a comment about this in the PR and commit message.

aravindhp (approved):
Thanks for working on this, @VaishnaviHire. Please run the tests successfully 5 times before merging.

/approve

aravindhp:
@VaishnaviHire we are getting a solid green on the e2es. Please remove the hold wherever you are ready.

aravindhp:
openshift/openshift-team-windows-containers we are no longer seeing `read TCP errors` and the VM configuration time has reduced by around 50%. It's possible that the latest OVN merge addressed that. 
/shrug

PR #108, opened 2020-09-10
[wmco] Watch cloud-private-key secret

sebsoto:
This commit is stopping the mounting of the cloud-private-key secret,
and instead fetching it within the Windows Machine and secret
controllers.

This has two main benefits. Firstly, this allows the
operator to be deployed without the cloud-private-key existing. This
will allow the operator to be properly deployed via the OpenShift
console, allowing for a much better UX. Machines will not be configured
until the private key secret is created, as they will not enter the
provisoned state until the userData secret is created by WMCO.
Secondly, this commit allows for our operator to update the userData
secret immediately after a change is made to the private key.

This commit made the private key secret the primary resource of the
secret controller. This is more fitting than the userData being the
primary resource, as we create the userData based off the contents of
the private key. The private key is the source of truth and therefore
should be the primary resource.

Changes were also made to the CSV in order to remove the mounting of
the private key.

This commit only causes future Windows machines to be configured with
a new private key when it is specified, and does not deal with Windows
machines which were configured with a previous private key. That needs
to be handled in a future commit.

### DIFF
--- a/hack/run-ci-e2e-test.sh
+++ b/hack/run-ci-e2e-test.sh
@@ -57,6 +57,12 @@ while getopts ":n:k:b:s" opt; do
   esac
 done
 
+# KUBE_SSH_KEY_PATH needs to be set in order to create the cloud-private-key secret
+if [ -z "$KUBE_SSH_KEY_PATH" ]; then
+    echo "env KUBE_SSH_KEY_PATH not found"
+    return 1
+fi
+
 OSDK=$(get_operator_sdk)
 
 # Set default values for the flags. Without this operator-sdk flags are getting
@@ -95,12 +101,16 @@ fi
 
 # The bool flags in golang does not respect key value pattern. They follow -flag=x pattern.
 # -flag x is allowed for non-boolean flags only(https://golang.org/pkg/flag/)
+
+# Test that the operator is running when the private key secret is not present
+OSDK_WMCO_test $OSDK "-run=TestWMCO/operator_deployed_without_private_key_secret -v -node-count=$NODE_COUNT -skip-node-deletion -ssh-key-pair=$KEY_PAIR_NAME --private-key-path=$KUBE_SSH_KEY_PATH $WMCO_PATH_OPTION"
+
 # Run the creation tests and skip deletion of the Windows VMs
-OSDK_WMCO_test $OSDK "-run=TestWMCO/create -v -timeout=90m -node-count=$NODE_COUNT -skip-node-deletion -ssh-key-pair=$KEY_PAIR_NAME $WMCO_PATH_OPTION"
+OSDK_WMCO_test $OSDK "-run=TestWMCO/create -v -timeout=90m -node-count=$NODE_COUNT -skip-node-deletion -ssh-key-pair=$KEY_PAIR_NAME --private-key-path=$KUBE_SSH_KEY_PATH $WMCO_PATH_OPTION"
 
 # Run the deletion tests while testing operator restart functionality. This will clean up VMs created
 # in the previous step
-OSDK_WMCO_test $OSDK "-run=TestWMCO/destroy -v -timeout=60m -ssh-key-pair=$KEY_PAIR_NAME"
+OSDK_WMCO_test $OSDK "-run=TestWMCO/destroy -v -timeout=60m -ssh-key-pair=$KEY_PAIR_NAME --private-key-path=$KUBE_SSH_KEY_PATH"
 
 # Get logs on success before cleanup
 get_WMCO_logs
--- a/pkg/controller/secret/secret_controller.go
+++ b/pkg/controller/secret/secret_controller.go
@@ -3,8 +3,6 @@ package secret
 import (
 	"context"
 
-	"github.com/openshift/windows-machine-config-operator/pkg/clusternetwork"
-	"github.com/openshift/windows-machine-config-operator/pkg/controller/secrets"
 	"github.com/pkg/errors"
 	core "k8s.io/api/core/v1"
 	k8sapierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -21,6 +19,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	"github.com/openshift/windows-machine-config-operator/pkg/clusternetwork"
+	"github.com/openshift/windows-machine-config-operator/pkg/controller/secrets"
 )
 
 const (
@@ -33,12 +34,12 @@ var log = logf.Log.WithName(ControllerName)
 
 // Add creates a new Secret Controller and adds it to the Manager. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
-func Add(mgr manager.Manager, _ clusternetwork.ClusterNetworkConfig) error {
+func Add(mgr manager.Manager, _ clusternetwork.ClusterNetworkConfig, watchNamespace string) error {
 	reconciler, err := newReconciler(mgr)
 	if err != nil {
 		return errors.Wrapf(err, "could not create %s reconciler", ControllerName)
 	}
-	return add(mgr, reconciler)
+	return add(mgr, reconciler, watchNamespace)
 }
 
 // newReconciler returns a new reconcile.Reconciler
@@ -54,38 +55,65 @@ func newReconciler(mgr manager.Manager) (reconcile.Reconciler, error) {
 	}
 
 	reconciler := &ReconcileSecret{client: client, scheme: mgr.GetScheme()}
-
-	// Create userData secret
-	if err := reconciler.createUserData(); err != nil {
-		return nil, errors.Wrapf(err, "error creating primary resource : %v", userDataSecret)
-	}
 	return reconciler, nil
 }
 
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
-func add(mgr manager.Manager, r reconcile.Reconciler) error {
+func add(mgr manager.Manager, r reconcile.Reconciler, watchNamespace string) error {
 	// Create a new controller
 	c, err := controller.New(ControllerName, mgr, controller.Options{Reconciler: r})
 	if err != nil {
 		return errors.Wrapf(err, "could not create the controller: %v", ControllerName)
 	}
-	predicateFilter := predicate.Funcs{
-		// get update event only when secret data is changed
-		UpdateFunc: func(e event.UpdateEvent) bool {
-			if string(e.ObjectOld.(*core.Secret).Data["userData"][:]) != string(e.ObjectNew.(*core.Secret).Data["userData"][:]) {
-				return true
-			}
-			return false
-		},
+
+	// Check that the private key exists, if it doesn't, log a warning
+	_, err = secrets.GetPrivateKey(kubeTypes.NamespacedName{Namespace: watchNamespace, Name: secrets.PrivateKeySecret}, mgr.GetClient())
+	if err != nil {
+		log.Error(err, "Unable to retrieve private key, please ensure it is created")
 	}
 
-	// Watch for changes to primary resource UserDataSecret
-	err = c.Watch(&source.Kind{Type: &core.Secret{ObjectMeta: meta.ObjectMeta{Namespace: userDataNamespace,
-		Name: userDataSecret}}}, &handler.EnqueueRequestForObject{}, predicateFilter)
+	// Watch for changes to the userData secret and enqueue the cloud-private-key if changed
+	// Name and namespace cannot be used to watch for specific secrets, so we filter out all the other secrets we
+	// dont care about.
+	// https://github.com/kubernetes-sigs/controller-runtime/issues/244
+	err = c.Watch(&source.Kind{Type: &core.Secret{}},
+		&handler.EnqueueRequestsFromMapFunc{ToRequests: newUserDataMapper(watchNamespace)},
+		predicate.Funcs{
+			CreateFunc: func(e event.CreateEvent) bool {
+				return isPrivateKeySecret(e.Meta, watchNamespace) || isUserDataSecret(e.Meta)
+			},
+			DeleteFunc: func(e event.DeleteEvent) bool {
+				return isPrivateKeySecret(e.Meta, watchNamespace) || isUserDataSecret(e.Meta)
+			},
+			UpdateFunc: func(e event.UpdateEvent) bool {
+				// get update event only when secret data is changed
+				if isPrivateKeySecret(e.MetaNew, watchNamespace) {
+					if string(e.ObjectOld.(*core.Secret).Data[secrets.PrivateKeySecretKey]) !=
+						string(e.ObjectNew.(*core.Secret).Data[secrets.PrivateKeySecretKey]) {
+						return true
+					}
+				} else if isUserDataSecret(e.MetaNew) {
+					if string(e.ObjectOld.(*core.Secret).Data["userData"][:]) != string(e.ObjectNew.(*core.Secret).Data["userData"][:]) {
+						return true
+					}
+				}
+				return false
+			},
+		})
 	if err != nil {
 		return err
 	}
 	return nil
+}
+
+// isUserDataSecret returns true if the object meta indicates that the object is the userData Secret
+func isUserDataSecret(meta meta.Object) bool {
+	return meta.GetName() == userDataSecret && meta.GetNamespace() == userDataNamespace
+}
+
+// isPrivateKeySecret returns true if the object meta indicates that the object is the private key secret
+func isPrivateKeySecret(meta meta.Object, keyNamespace string) bool {
+	return meta.GetName() == secrets.PrivateKeySecret && meta.GetNamespace() == keyNamespace
 }
 
 // blank assignment to verify that ReconcileSecret implements reconcile.Reconciler
@@ -105,9 +133,19 @@ type ReconcileSecret struct {
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
 func (r *ReconcileSecret) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
-	reqLogger.Info("Reconciling " + userDataSecret)
 
-	validUserData, err := secrets.GenerateUserData()
+	privateKey, err := secrets.GetPrivateKey(request.NamespacedName, r.client)
+	if err != nil {
+		if k8sapierrors.IsNotFound(err) {
+			// Request object not found, could have been deleted after reconcile request.
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			// Return and don't requeue
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, errors.Wrapf(err, "unable to get secret %s", request.NamespacedName)
+	}
+	// Generate expected userData based on the existing private key
+	validUserData, err := secrets.GenerateUserData(privateKey)
 	if err != nil {
 		return reconcile.Result{}, errors.Wrapf(err, "error generating userData secret")
 	}
@@ -117,7 +155,7 @@ func (r *ReconcileSecret) Reconcile(request reconcile.Request) (reconcile.Result
 	err = r.client.Get(context.TODO(), kubeTypes.NamespacedName{Name: userDataSecret, Namespace: userDataNamespace}, userData)
 	if err != nil && k8sapierrors.IsNotFound(err) {
 		// Secret is deleted
-		reqLogger.Info("UserData secret not found, recreating the secret.")
+		reqLogger.Info("UserData secret not found, creating the secret.")
 		err = r.client.Create(context.TODO(), validUserData)
 		if err != nil {
 			return reconcile.Result{}, err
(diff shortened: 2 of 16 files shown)


pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -181,6 +180,27 @@ func (r *ReconcileWindowsMachine) Reconcile(request reconcile.Request) (reconcil
 		return reconcile.Result{}, nil
 	}
 
+	// Get the private key that will be used to configure the instance
```
VaishnaviHire, line 183: Should this code block be moved before fetching machine instance? We should be checking for private key before checking machine status. It would allow us to see why it did not go in `Provisioned` state.

VaishnaviHire (commented):
Thanks for working on this @sebsoto . Mostly looks good to me except one question.

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -69,23 +70,23 @@ func newReconciler(mgr manager.Manager, networkConfig clusternetwork.ClusterNetw
...
 
-	signer, err := signer.Create()
+	serviceCIDR, err := networkConfig.GetServiceCIDR()
 	if err != nil {
-		return nil, errors.Wrapf(err, "error creating signer using private key: %v", wkl.PrivateKeyPath)
+		return nil, errors.Wrap(err, "error getting service CIDR")
 	}
 
-	serviceCIDR, err := networkConfig.GetServiceCIDR()
+	namespace, err := k8sutil.GetWatchNamespace()
```
aravindhp, line 78: Nit: Ideally we should do this in one place at a higher level and pass it down.

pkg/controller/secret/secret_controller.go
```diff
@@ -69,25 +66,65 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
-	predicateFilter := predicate.Funcs{
-		// get update event only when secret data is changed
-		UpdateFunc: func(e event.UpdateEvent) bool {
-			if string(e.ObjectOld.(*core.Secret).Data["userData"][:]) != string(e.ObjectNew.(*core.Secret).Data["userData"][:]) {
-				return true
-			}
-			return false
-		},
+
+	// Get namespace that the operator is deployed in
```
aravindhp, line 70: Comment does not match the API you are using.

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -69,23 +70,23 @@ func newReconciler(mgr manager.Manager, networkConfig clusternetwork.ClusterNetw
...
 	}
 
 	return &ReconcileWindowsMachine{client: client,
 			scheme:             mgr.GetScheme(),
 			k8sclientset:       clientset,
 			clusterServiceCIDR: serviceCIDR,
-			signer:             signer,
 			vxlanPort:          networkConfig.VXLANPort(),
 			recorder:           mgr.GetEventRecorderFor(ControllerName),
+			opNamespace:        namespace,
```
aravindhp, line 89: `opNamespace` is not appropriately named given it is the namespace the operator is watching. In our case it just so happens to be the same as the namespace the operator is running in.

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -149,6 +150,8 @@ type ReconcileWindowsMachine struct {
 	vxlanPort string
 	// recorder to generate events
 	recorder record.EventRecorder
+	// opNamespace is the namespace the operator is deployed in
```
aravindhp, line 153: Incorrect based on the API you are using to get the namespace.

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -181,6 +180,27 @@ func (r *ReconcileWindowsMachine) Reconcile(request reconcile.Request) (reconcil
...
+		Name: secrets.PrivateKeySecret}, r.client)
+	if err != nil {
+		if k8sapierrors.IsNotFound(err) {
+			// Private key was removed, requeue
+			return reconcile.Result{}, errors.Wrapf(err, "%s does not exist, please create it", secrets.PrivateKeySecret)
+		}
+		return reconcile.Result{}, errors.Wrapf(err, "unable to get secret %s", request.NamespacedName)
+	}
+	// Update the signer with the existing privateKey
+	newSigner, err := signer.Create(privateKey)
```
aravindhp, line 194: `newSigner` is a misleading as it could be a function name. How about `signer` or `privateKeySigner`? Or just use `r.signer` directly.

test/e2e/wmco_test.go
```diff
@@ -33,12 +36,15 @@ func TestWMCO(t *testing.T) {
...
 	gc.sshKeyPair = sshKeyPair
+	require.NotEmpty(t, privateKeyPath, "private-key-path is not set")
+	gc.privateKeyPath = privateKeyPath
 	// When the OPENSHIFT_CI env var is set to true, the test is running within CI
 	if inCI := os.Getenv("OPENSHIFT_CI"); inCI == "true" {
 		// In the CI container the WMCO binary will be found here
 		wmcoPath = "/usr/local/bin/windows-machine-config-operator"
 	}
 
+	t.Run("operator deployed", testOperatorDeployed)
```
aravindhp, line 47: What is the reason for adding this test? Ideally this should be a separate commit or PR.

aravindhp (changes requested):

test/e2e/wmco_test.go
sebsoto: The reasons is to make sure the operator can deploy without the secret already created, and then use a secret created after the operator is running

aravindhp: Please add that as a comment as it is not obvious.

pkg/controller/secret/secret_controller.go
LorbusChris: `watchNamespace` maybe?

pkg/controller/secret/secret_controller.go
```diff
@@ -69,25 +66,65 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
-	err = c.Watch(&source.Kind{Type: &core.Secret{ObjectMeta: meta.ObjectMeta{Namespace: userDataNamespace,
-		Name: userDataSecret}}}, &handler.EnqueueRequestForObject{}, predicateFilter)
+	// Check that the private key exists, if it doesn't, log a warning
+	_, err = secrets.GetPrivateKey(kubeTypes.NamespacedName{Namespace: namespace, Name: secrets.PrivateKeySecret}, mgr.GetClient())
+	if err != nil {
+		log.Info("Unable to retrieve private key, please ensure it is created")
+	}
+
+	// Watch for changes to the userData secret and enqueue the cloud-private-key if changed
+	err = c.Watch(&source.Kind{Type: &core.Secret{}},
```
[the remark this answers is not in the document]
LorbusChris: let's add the metadata here and add a comment pointing to the OSDK issue.

pkg/controller/secret/secret_controller.go
```diff
@@ -142,23 +189,20 @@ func (r *ReconcileSecret) Reconcile(request reconcile.Request) (reconcile.Result
...
-	// generate valid userData secret
-	validUserDataSecret, err := secrets.GenerateUserData()
-	if err != nil {
-		return errors.Wrapf(err, "error generating valid userData secret")
-	}
+// userDataMapper is a simple implementation of the Mapper interface allowing for the mapping from the userData secret
+// to the private key secret
+type userDataMapper struct {
+	// opNamespace is the namespace the operator is deployed within
+	opNamespace string
```
LorbusChris, line 196: I think this should be `watchNamespace` here as well

pkg/controller/secrets/secrets.go
```diff
@@ -1,29 +1,49 @@
...
 )
 
 const (
 	userDataSecret    = "windows-user-data"
 	userDataNamespace = "openshift-machine-api"
+	PrivateKeySecret  = "cloud-private-key"
 )
 
+// GetPrivateKey fetches the specified secret and extracts the private key data
+func GetPrivateKey(secret kubeTypes.NamespacedName, c client.Client) ([]byte, error) {
```
LorbusChris, line 23:
```suggestion
func GetPrivateKey(namespace kubeTypes.NamespacedName, c client.Client) ([]byte, error) {
```
or maybe `secretNamespace`

pkg/controller/windowsmachine/windowsmachine_controller.go
LorbusChris: agree, I think moving it up makes sense

pkg/controller/windowsmachine/windowsmachine_controller.go
LorbusChris: +1 for `privateKeySigner` or `desiredKeySigner`

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -181,6 +180,27 @@ func (r *ReconcileWindowsMachine) Reconcile(request reconcile.Request) (reconcil
...
+	if err != nil {
+		return reconcile.Result{}, errors.Wrap(err, "error creating signer")
+	}
+	r.signer = newSigner
```
LorbusChris, line 198: we only really need to update this if `newSigner != r.signer` - maybe it's helpful to log when this update is happening?

test/e2e/wmco_test.go
LorbusChris: `"operator deployed without private key secret"`

test/e2e/wmco_test.go
```diff
@@ -54,3 +60,13 @@ func setupWMCOResources() error {
 	}
 	return nil
 }
+
+// testOperatorDeployed tests that the operator pod is running
+func testOperatorDeployed(t *testing.T) {
```
LorbusChris, line 65:
```suggestion
func testOperatorDeployedWithoutPrivateKeySecret(t *testing.T) {
```

hack/run-ci-e2e-test.sh
```diff
@@ -96,11 +102,11 @@ fi
 # The bool flags in golang does not respect key value pattern. They follow -flag=x pattern.
 # -flag x is allowed for non-boolean flags only(https://golang.org/pkg/flag/)
 # Run the creation tests and skip deletion of the Windows VMs
-OSDK_WMCO_test $OSDK "-run=TestWMCO/create -v -timeout=90m -node-count=$NODE_COUNT -skip-node-deletion -ssh-key-pair=$KEY_PAIR_NAME $WMCO_PATH_OPTION"
+OSDK_WMCO_test $OSDK "-run=TestWMCO -v -timeout=90m -node-count=$NODE_COUNT -skip-node-deletion -ssh-key-pair=$KEY_PAIR_NAME --private-key-path=$KUBE_SSH_KEY_PATH $WMCO_PATH_OPTION"
```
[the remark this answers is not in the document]
sebsoto: So that it runs all tests except the deletion suite. This will cause it to run the operatorDeployed test

pkg/controller/secret/secret_controller.go
```diff
@@ -69,25 +66,66 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+				return false
+			},
+		})
 	if err != nil {
 		return err
 	}
 	return nil
 }
 
+// isUserData returns true if the object meta indicates that the object is the userData Secret
```
mansikulkarni96, line 119: *isUserDataSecret

pkg/controller/secret/secret_controller.go
```diff
@@ -69,25 +66,66 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+func isUserDataSecret(meta meta.Object) bool {
+	return meta.GetName() == userDataSecret && meta.GetNamespace() == userDataNamespace
+}
+
+// isUserData returns true if the object meta indicates that the object is the private key secret
```
mansikulkarni96, line 124: *isPrivateKeySecret

test/e2e/wmco_test.go
```diff
@@ -1,13 +1,16 @@
...
+	"context"
 	"os"
 	"testing"
 	"time"
 
 	mapi "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
 	framework "github.com/operator-framework/operator-sdk/pkg/test"
 	"github.com/pkg/errors"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
mansikulkarni96, line 13: metav1->*meta

pkg/controller/secret/secret_controller.go
```diff
@@ -69,25 +66,66 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+	if err != nil {
+		log.Error(err, "Unable to retrieve private key, please ensure it is created")
+	}
+
+	// Watch for changes to the userData secret and enqueue the cloud-private-key if changed
+	err = c.Watch(&source.Kind{Type: &core.Secret{}},
+		&handler.EnqueueRequestsFromMapFunc{ToRequests: newUserDataMapper(namespace)},
+		predicate.Funcs{
+			CreateFunc: func(e event.CreateEvent) bool {
+				if isPrivateKeySecret(e.Meta, namespace) || isUserDataSecret(e.Meta) {
```
ravisantoshgudimetla, line 88: nit: This can be further reduced to single return statement.

pkg/controller/secret/secret_controller.go
```diff
@@ -69,25 +66,66 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
-		Name: userDataSecret}}}, &handler.EnqueueRequestForObject{}, predicateFilter)
+	// Check that the private key exists, if it doesn't, log a warning
+	_, err = secrets.GetPrivateKey(kubeTypes.NamespacedName{Namespace: namespace, Name: secrets.PrivateKeySecret}, mgr.GetClient())
+	if err != nil {
+		log.Error(err, "Unable to retrieve private key, please ensure it is created")
+	}
+
+	// Watch for changes to the userData secret and enqueue the cloud-private-key if changed
+	err = c.Watch(&source.Kind{Type: &core.Secret{}},
+		&handler.EnqueueRequestsFromMapFunc{ToRequests: newUserDataMapper(namespace)},
```
ravisantoshgudimetla, line 85: nit: You don't need this enqueuing as technically you would like to react upon the private secret change anyways.

pkg/controller/secret/secret_controller.go
```diff
@@ -105,9 +143,19 @@ type ReconcileSecret struct {
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
 func (r *ReconcileSecret) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
-	reqLogger.Info("Reconciling " + userDataSecret)
 
-	validUserData, err := secrets.GenerateUserData()
+	privateKey, err := secrets.GetPrivateKey(request.NamespacedName, r.client)
```
ravisantoshgudimetla, line 147: Don't pass client around, if possible create a struct around it make them methods, so that you can share

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -181,6 +191,16 @@ func (r *ReconcileWindowsMachine) Reconcile(request reconcile.Request) (reconcil
 		return reconcile.Result{}, nil
 	}
 
+	// Update the signer with the existing privateKey
+	r.signer, err = signer.Create(privateKey)
```
ravisantoshgudimetla, line 195: I think the signer should be moved elsewhere, not sure if it'd make sense to have it in reconciler struct considering, this is not static anymore. Perhaps we can get the value at the call site.

ravisantoshgudimetla (commented):
Thank you for working on this @sebsoto. Left a partial review. Will look more into the test cases tomorrow.

pkg/controller/secret/secret_controller.go
sebsoto: Are you suggesting moving GetPrivateKey() out of the secrets library? That would be the only way to stop passing the client around

test/e2e/wmco_test.go
sebsoto: I made the change for the test name, i am hesitant to change the function name as it does not actually test that the private key secret does not exist.

pkg/controller/windowsmachine/windowsmachine_controller.go
sebsoto: I dont know if it makes sense passing the privatekey all the way down to the nodeconfig

pkg/controller/secret/secret_controller.go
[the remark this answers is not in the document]
sebsoto: we cannot add the actual metadata as we need to watch for two distinct secrets. It depends on how the "fix" is implemented. we may need two sperate watches.

pkg/controller/secret/secret_controller.go
ravisantoshgudimetla: Nope, I meant don't pass the client to every method. Create a struct called 
```suggestion
type secretConfig struct {
   privateKey []byte
   secret *corev1.Secret
   clientSet
}
func NewSecretConfig(clientSet) *SecretConfig {
// Construct Secret Config
}
     // Call NewSecretConfig()
      privateKey, err := secretConfig.GetPrivateKey(request.NamespacedName, r.client)
```
With this association becomes easier and would be easier to map secret to a privateKey. You don't have to implement this.

pkg/controller/windowsmachine/windowsmachine_controller.go
ravisantoshgudimetla: Ideally I'd like to get the secret there in the nodeConfig instead of checking in the reconcile method. This way we don't have to pass the objects around. The downside to this approach is we cannot fail early in the reconcile. Previously we were assuming that signer will be invariant in all reconciles which is not the case anymore. So, to me it is better to fail at the location where we are acting up on it. However, this is a nit again. I don't see that much of a problem of with this approach. Just that it looks a bit different to what I would have implemented

sebsoto: I think that approach is valid, however I think failing early is more important right now, im inclined to leave the code as it is.

pkg/controller/secret/secret_controller.go
sebsoto: I think this adds unnecessary abstraction, we still would have to create a new secretClient every time. I'd rather leave it as it is for now

pkg/controller/secrets/secrets.go
```diff
@@ -1,29 +1,49 @@
...
+	// Fetch the private key secret
+	privateKeySecret := &core.Secret{}
+	if err := c.Get(context.TODO(), secret, privateKeySecret); err != nil {
+		// Error reading the object - requeue the request.
+		return []byte{}, err
+	}
+	privateKey, ok := privateKeySecret.Data["private-key.pem"]
```
aravindhp, line 30: Sorry I should have caught this before. Please make `private-key.pem` a constant.

pkg/controller/windowsmachine/windowsmachine_controller.go
aravindhp: This has not been addressed.

test/e2e/wmco_test.go
```diff
@@ -33,12 +36,17 @@ func TestWMCO(t *testing.T) {
...
+	gc.privateKeyPath = privateKeyPath
 	// When the OPENSHIFT_CI env var is set to true, the test is running within CI
 	if inCI := os.Getenv("OPENSHIFT_CI"); inCI == "true" {
 		// In the CI container the WMCO binary will be found here
 		wmcoPath = "/usr/local/bin/windows-machine-config-operator"
 	}
 
+	// test that the operator can deploy without the secret already created, we can later use a secret created by the
+	// individual test suites after the operator is running
+	t.Run("operator deployed without private key secret", testOperatorDeployed)
```
aravindhp, line 49: Nit: I am not a fan of these long test names. But I can't think of a better name :-)

hack/run-ci-e2e-test.sh
```diff
@@ -95,12 +101,16 @@ fi
 
 # The bool flags in golang does not respect key value pattern. They follow -flag=x pattern.
 # -flag x is allowed for non-boolean flags only(https://golang.org/pkg/flag/)
+
+# Test that the operator is running
```
aravindhp, line 105: `...when the private key is not present

aravindhp (changes requested):

pkg/controller/windowsmachine/windowsmachine_controller.go
sebsoto: You're right, i had moved it up but not enough, doing that now

ravisantoshgudimetla (approved):
/lgtm

hack/run-ci-e2e-test.sh
```diff
@@ -95,12 +101,16 @@ fi
 
 # The bool flags in golang does not respect key value pattern. They follow -flag=x pattern.
 # -flag x is allowed for non-boolean flags only(https://golang.org/pkg/flag/)
+
+# Test that the operator is running when the private key secret is not present
+OSDK_WMCO_test $OSDK "-run=TestWMCO/operator_deployed_without_private_key_secret -v -timeout=90m -node-count=$NODE_COUNT -skip-node-deletion -ssh-key-pair=$KEY_PAIR_NAME --private-key-path=$KUBE_SSH_KEY_PATH $WMCO_PATH_OPTION"
```
aravindhp, line 106: Why do we need a 90 minute timeout?

ravisantoshgudimetla:
/hold

Please unhold after someone else reviews it

aravindhp (approved):
/approve

Thanks for working on this, @sebsoto. Approved with the caveat you address any changes requested by other folks.

ravisantoshgudimetla:
/hold

for now.

PR #112, opened 2020-09-15
WIP: [wmco] Download worker ignition to operator

LorbusChris:
[wmco] Download worker ignition to operator, copy to VM from there

This changes the windowsmachine controller to download the worker
ignition config itself, and then copy it over to the Windows VM.

Before, the config was downloaded by a powershell script running
directly on the VM, using the VM's host network.
With this change, the Ignition config is downloaded onto the WMCO
container first, and from there the SSH connection is used for
transferring it to the VM, without the VM needing access to the network
other than through the SSH conection.

The `wget-ignore-cert.ps1` PowerShell script is not needed anymore,
and is removed.


LorbusChris:
This approach will not work as-is. In order to get rid of the powershell script, I still favour the approach from https://github.com/openshift/windows-machine-config-bootstrapper/pull/232, at least as a temporary solution
/close

PR #116, opened 2020-09-18
[wmco] Handle ignition file download errors

sebsoto:
This commit makes it so that the powershell script we use to download
the ignition file properly returns an error when it fails. Previously
the actual error was also being set to nil, and we only had
stdout/stderr to tell if something went wrong


sebsoto:
This should only be merged if the operator starts erroring on ignition file downloads instead of continuing.

PR #121, opened 2020-09-28
[hack] add MachineSet creation/deletion script

PratikMahajan:
this script makes the creation of MachineSet on AWS and Azure platform
less of a task.
Requires an existing logged in cluster to create MachineSet
Currently works only for AWS and AZURE platform

USAGE
`machineset.sh apply/delete`

OPTIONS
`$1           Action (optional)             apply/delete the MachineSet`

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,17 @@ spec:
 Example MachineSet for other cloud providers:
 - [AWS](docs/machineset-aws.md)
 
+Alternatively, the [hack/machineset.sh](hack/machineset.sh) script can be used to generate MachineSets for AWS and Azure platforms.
+The hack script will generate a `MachineSet.yaml` file which can be edited before using or can be used as it is.
+The script takes optional arguments `apply` and `delete` to directly create/delete MachineSet on the cluster without 
+generating a `yaml` file.
+
+Usage:
+```
+./hack/machineset.sh                 # to generate yaml file
+./hack/machineset.sh apply/delete    # to create/delete MachineSet directly on cluster
+```
+
 ## Windows nodes Kubernetes component upgrade
 
 When a new version of WMCO is released that is compatible with the current cluster version, an operator upgrade will 
--- a/hack/machineset.sh
+++ b/hack/machineset.sh
@@ -1,0 +1,207 @@
+#!/usr/bin/env bash
+
+# machineset.sh - create a Windows MachineSet for Windows Machine Config Operator
+# Creates a MachineSet.yaml file and apply/delete the MachineSet if optional action is provided
+#
+# USAGE
+#    machineset.sh
+# OPTIONS
+#    $1      Action       (Optional) apply/delete the MachineSet
+# PREREQUISITES
+#    oc                   to fetch cluster info and apply/delete MachineSets on the cluster(cluster should be logged in)
+#    aws                  to fetch Windows AMI id for AWS platform (only required for clusters running on AWS)
+set -euo pipefail
+
+ACTION=${1:-}
+
+error-exit() {
+    echo "Error: $*" >&2
+    exit 1
+}
+
+# get_spec returns the template yaml common for all cloud providers
+get_spec() {
+
+  if [ "$#" -lt 3 ]; then
+    error-exit incorrect parameter count for get_spec $#
+  fi
+
+  local infraID=$1
+  local az=$2
+  local provider=$3
+
+  machineSetName="$infraID"-windows-worker-"$az"
+  if [ "$provider" = "azure" ]; then
+    # Shorter name for azure as VMs with more than 15 characters in name does not come up
+    machineSetName="winworker"
+  fi
+
+  cat <<EOF
+apiVersion: machine.openshift.io/v1beta1
+kind: MachineSet
+metadata:
+  labels:
+    machine.openshift.io/cluster-api-cluster: ${infraID}
+  name: ${machineSetName}
+  namespace: openshift-machine-api
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      machine.openshift.io/cluster-api-cluster: ${infraID}
+      machine.openshift.io/cluster-api-machineset: ${machineSetName}
+  template:
+    metadata:
+      labels:
+        machine.openshift.io/cluster-api-cluster: ${infraID}
+        machine.openshift.io/cluster-api-machine-role: worker
+        machine.openshift.io/cluster-api-machine-type: worker
+        machine.openshift.io/cluster-api-machineset: ${machineSetName}
+        machine.openshift.io/os-id: Windows
+    spec:
+      metadata:
+        labels:
+          node-role.kubernetes.io/worker: ""
+EOF
+}
+
+# get_aws_ms creates a MachineSet for AWS Cloud Provider
+get_aws_ms() {
+
+  if [ "$#" -lt 4 ]; then
+    error-exit incorrect parameter count for get_spec $#
+  fi
+
+  local infraID=$1
+  local region=$2
+  local az=$3
+  local provider=$4
+
+  # get the AMI id for the Windows VM
+  ami_date="2020.09.09"
+  ami_id=$(aws ec2 describe-images --filters Name=name,Values=Windows_Server-2019-English-Full-ContainersLatest-${ami_date} --region ${region} --query 'Images[*].[ImageId]' --output text)
+
+  cat <<EOF
+$(get_spec $infraID $az $provider)
+      providerSpec:
+        value:
+          ami:
+            id: ${ami_id}
+          apiVersion: awsproviderconfig.openshift.io/v1beta1
+          blockDevices:
+            - ebs:
+                iops: 0
+                volumeSize: 120
+                volumeType: gp2
+          credentialsSecret:
+            name: aws-cloud-credentials
+          deviceIndex: 0
+          iamInstanceProfile:
+            id: ${infraID}-worker-profile
+          instanceType: m5a.large
+          kind: AWSMachineProviderConfig
+          placement:
+            availabilityZone: ${az}
+            region: ${region}
+          securityGroups:
+            - filters:
+                - name: tag:Name
+                  values:
+                    - ${infraID}-worker-sg
+          subnet:
+            filters:
+              - name: tag:Name
+                values:
+                  - ${infraID}-private-${az}
+          tags:
+            - name: kubernetes.io/cluster/${infraID}
+              value: owned
+          userDataSecret:
+            name: windows-user-data
+EOF
+}
+
+# get_azure_ms creates a MachineSet for Azure Cloud Provider
+get_azure_ms() {
+
+  if [ "$#" -lt 4 ]; then
+    error-exit incorrect parameter count for get_spec $#
+  fi
+
+  local infraID=$1
+  local region=$2
+  local az=$3
+  local provider=$4
+
+  cat <<EOF
+$(get_spec $infraID $az $provider)
+      providerSpec:
+        value:
+          apiVersion: azureproviderconfig.openshift.io/v1beta1
+          credentialsSecret:
+            name: azure-cloud-credentials
+            namespace: openshift-machine-api
+          image:
+            offer: WindowsServer
+            publisher: MicrosoftWindowsServer
+            resourceID: ""
+            sku: 2019-Datacenter-with-Containers
+            version: latest
+          kind: AzureMachineProviderSpec
+          location: ${region}
+          managedIdentity: ${infraID}-identity
+          metadata:
+            creationTimestamp: null
+          networkResourceGroup: ${infraID}-rg
+          osDisk:
+            diskSizeGB: 128
+            managedDisk:
+              storageAccountType: Premium_LRS
+            osType: Windows
+          publicIP: false
+          resourceGroup: ${infraID}-rg
+          subnet: ${infraID}-worker-subnet
+          userDataSecret:
+            name: windows-user-data
+            namespace: openshift-machine-api
+          vmSize: Standard_D2s_v3
+          vnet: ${infraID}-vnet
+          zone: "${az}"
+EOF
+}
+
+# Retrieves the Cloud Provider for the OpenShift Cluster
+provider="$(oc -n openshift-kube-apiserver get configmap config -o json | jq -r '.data."config.yaml"' | jq '.apiServerArguments."cloud-provider"' | jq -r '.[]')"
+
+# Gets the Infrastructure Id for the cluster like `pmahajan-azure-68p9l-gv45m`
+infraID="$(oc get -o jsonpath='{.status.infrastructureName}{"\n"}' infrastructure cluster)"
+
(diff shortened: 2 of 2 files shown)


hack/machineset.sh
```diff
@@ -0,0 +1,181 @@
...
+#
+# USAGE
+#    machineset.sh apply/delete
+# OPTIONS
+#    $1      Action       apply/delete the MachineSet
+# PREREQUISITES
+#    oc                   to apply MachineSets on the cluster(cluster should be logged in)
+#    aws                  to fetch Windows AMI id for AWS platform (only required for clusters running on AWS)
+
+ACTION=$1
```
aravindhp, line 13: I suggest the script should just generate the yaml. Applying / deleting should be left to the user through oc. This will also allow us to make changes to it before applying.

hack/machineset.sh
```diff
@@ -0,0 +1,181 @@
...
+if [[ ! "$ACTION" =~ ^apply|delete$ ]]; then
+    error-exit "Action (1st parameter) must be \"apply\" or \"delete\""
+fi
+
+# Creates a MachineSet for AWS Cloud Provider
+get_aws_ms(){
```
aravindhp, line 19: Having this per provider is going to hard to maintain given the majority of the spec is the same across providers. How about a single function that takes the provider as input and generates the yaml differently based on the provider?

aravindhp (changes requested):
Thanks for working on this, @PratikMahajan. This will really help improve our workflow. You can also link to this script from the README section about MachineSets with an explanation.

hack/machineset.sh
PratikMahajan: The only thing we are actually changing for the cloud-provider is the `provider spec` we can keep the rest of `yaml` as it is for all cloud providers... 
But except `user-data` part in provider spec, rest everything is different for different cloud providers. I mean, the code duplication can be reduced in this case, but i think we might still have to have these functions for different cloud providers to make the code look good and understandable, but i am open for suggestions ,,,

aravindhp: Sure the `providerSpec` will be different, however the rest should be a common template where you replace things like `infraID` etc. I am not averse to having different functions for each provider but there needs to be more in common.

This can also be tackled in a follow up PR.

hack/machineset.sh
```diff
@@ -0,0 +1,181 @@
...
+
+local infraID=$1
+local region=$2
+local az=$3
+
+# get the AMI id for the Windows VM
+ami_date="2020.06.10"
```
aravindhp, line 26: Is this Oct 10th or June 10th? :-) We do need to pick the AMI which has the VXLAN patch present.

PratikMahajan: My Bad, i thought this one is the latest. 
We have June 10, Sept 9, Aug 12, July 15 as our options. 
I'll replace this with the latest one i.e. Sept 9

hack/machineset.sh
PratikMahajan: I'll reduce the code duplication in this PR. Wont take much effort.. 
Rest of the things can be done in follow up PRs

hack/machineset.sh
PratikMahajan: This is based on the fact that i am considering the middle part is month and not date. I'm confused as well....

aravindhp: OK, once the confusion about the date format is resolved, please ensure you pick one on or after Aug 12, 2020. Anything before that does not have VXLAN patch. Latest is good too.

hack/machineset.sh
```diff
@@ -0,0 +1,181 @@
...
+        machine.openshift.io/os-id: Windows
+    spec:
+      metadata:
+        labels:
+          node-role.kubernetes.io/worker: ""
+EOF
+}
+
+# get_aws_ms creates a MachineSet for AWS Cloud Provider
+get_aws_ms(){
```
aravindhp, line 57: Please indent the function body

hack/machineset.sh
```diff
@@ -0,0 +1,181 @@
...
+          tags:
+            - name: kubernetes.io/cluster/$(echo $infraID)
+              value: owned
+          userDataSecret:
+            name: windows-user-data
+EOF
+}
+
+# get_azure_ms creates a MachineSet for Azure Cloud Provider
+get_azure_ms(){
```
aravindhp, line 108: Same here

hack/machineset.sh
```diff
@@ -0,0 +1,181 @@
+#!/usr/bin/env bash
+
+# machineset.sh - create a Windows MachineSet for Windows Machine Config Operator
```
aravindhp, line 3: Please update the MachineSet section of the README with help regarding this script.

aravindhp (changes requested):

LorbusChris:
@PratikMahajan I'll give this a try on Azure. I expect we'll be hitting an error there because the generated Machine names will be too long. Will report back.

hack/machineset.sh
PratikMahajan: Sorry, forgot to update. 
I've made the change which creates a YAML file if no action is provided and will create a machineset directly if an action is provided. 
This due to the fact that, we wont be changing the machineset always and will not result in an additional file

hack/machineset.sh
PratikMahajan: Kept the indentation at 0 because of 
```
cat <<EOF
...
EOF
```
In case, i indent, I'll have to keep the data in cat part at 0 so that it'll not result in a incorrectly indented yaml file

hack/machineset.sh
PratikMahajan: Forgot about this part... adding while addressing next wave of changes

PratikMahajan:
Sure @LorbusChris I created a machineset on azure using the script and it worked without any issue . But let me know in case the name is too big so that i can change it

hack/machineset.sh
aravindhp: Multiple solutions for this:
- You can use `echo -e` and use escape characters
- use `-EOF` with `cat
- use `sed` to fix things up after redirecting
You can experiment and figure out what works best for you.

PratikMahajan: i'll try these out and update....

LorbusChris:
On Azure, the MachineSet is created successfully, but the Machines aren't: 

Reason: MachineCreationFailed

> InvalidConfiguration: failed to reconcile machine "cglombek-az-whzk8-windows-worker-3-tfnzw": failed to create vm cglombek-az-whzk8-windows-worker-3-tfnzw: failure sending request for machine cglombek-az-whzk8-windows-worker-3-tfnzw: cannot create vm: compute.VirtualMachinesClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="InvalidParameter" Message="Windows computer name cannot be more than 15 characters long, be entirely numeric, or contain the following characters: ` ~ ! @ # $ % ^ & * ( ) = + _ [ ] { } \\ | ; : . ' \" , < > / ?." Target="computerName"

hack/machineset.sh
```diff
@@ -0,0 +1,194 @@
...
+# get_spec returns the template yaml common for all cloud providers
+get_spec(){
+
+local infraID=$1
+local az=$2
+local provider=$3
+
+machineSetName="$infraID"-windows-worker-"$az"
+if [ "$provider" = "azure" ]; then
+  machineSetName="winworker"
```
LorbusChris, line 30: a comment on why this is necessary would be great

hack/machineset.sh
```diff
@@ -0,0 +1,194 @@
+#!/usr/bin/env bash
```
LorbusChris, line 1: we probably want to `set -euo pipefail` here.
See also http://redsymbol.net/articles/unofficial-bash-strict-mode/

aravindhp:
I am getting error on an AWS cluster when trying to create the resulting `MachineSet`

```
oc create -f MachineSet.yaml
Error from server (providerSpec.ami: Required value: expected either providerSpec.ami.arn or providerSpec.ami.filters or providerSpec.ami.id to be populated): error when creating "MachineSet.yaml": admission webhook "validation.machineset.machine.openshift.io" denied the request: providerSpec.ami: Required value: expected either providerSpec.ami.arn or providerSpec.ami.filters or providerSpec.ami.id to be populated
```
Please [MachineSet.yaml](https://gist.github.com/aravindhp/9a4ed4cb09045b02e3ab6bbc7ff764c5)

hack/machineset.sh
```diff
@@ -0,0 +1,194 @@
...
+get_aws_ms(){
+
+local infraID=$1
+local region=$2
+local az=$3
+local provider=$4
+
+# get the AMI id for the Windows VM
+ami_date="2020.09.09"
+ami_id=$(aws ec2 describe-images --filters Name=name,Values=Windows_Server-1909-English-Core-ContainersLatest-${ami_date} --region ${region} | awk '{print $17}' | head -n1)
```
aravindhp, line 72: The bug that I [found](https://github.com/openshift/windows-machine-config-operator/pull/121#issuecomment-704480941) is with `print $17`. I suggest using `jq` instead to parse the output.

hack/machineset.sh
PratikMahajan: done

hack/machineset.sh
PratikMahajan: Left the heredoc part as it is and indented the rest of body.. 
Was not able to make the indentation work as required in the heredoc

hack/machineset.sh
```diff
@@ -0,0 +1,196 @@
...
+  local infraID=$1
+  local region=$2
+  local az=$3
+  local provider=$4
+
+  # get the AMI id for the Windows VM
+  ami_date="2020.09.09"
+  ami_id=$(aws ec2 describe-images --filters Name=name,Values=Windows_Server-1909-English-Core-ContainersLatest-${ami_date} --region ${region} --query 'Images[*].[ImageId]' | head -n1)
+
+cat <<EOF
```
LorbusChris, line 76: this line should also be indented. Leaving the heredoc itself without identation is fine I think, at least I've seen it done this way before elsewhere.

hack/machineset.sh
```diff
@@ -0,0 +1,196 @@
...
+
+# get_azure_ms creates a MachineSet for Azure Cloud Provider
+get_azure_ms(){
+
+  local infraID=$1
+  local region=$2
+  local az=$3
+  local provider=$4
+
+cat <<EOF
```
LorbusChris, line 124: same as above, needs indentation on this line

README.md
```diff
@@ -298,6 +298,17 @@ spec:
 Example MachineSet for other cloud providers:
 - [AWS](docs/machineset-aws.md).
 
+Alternatively, you can use `hack/machineset.sh` script to generate MachineSets for AWS and Azure platform.
```
LorbusChris, line 301: nit
```suggestion
Alternatively, the `hack/machineset.sh` script can be used to generate MachineSets for AWS and Azure platforms.
```

hack/machineset.sh
PratikMahajan: Fixed

hack/machineset.sh
```diff
@@ -0,0 +1,196 @@
...
+get_aws_ms(){
+
+  local infraID=$1
+  local region=$2
+  local az=$3
+  local provider=$4
+
+  # get the AMI id for the Windows VM
+  ami_date="2020.09.09"
+  ami_id=$(aws ec2 describe-images --filters Name=name,Values=Windows_Server-2019-English-Full-ContainersLatest-${ami_date} --region ${region} --query 'Images[*].[ImageId]' --output text)
```
aravindhp, line 74: Will this work if multiple images are listed?

aravindhp (changes requested):

hack/machineset.sh
```diff
@@ -0,0 +1,196 @@
...
+
+  cat <<EOF
+$(get_spec $infraID $az $provider)
+      providerSpec:
+        value:
+          ami:
+            id: $(echo $ami_id)
```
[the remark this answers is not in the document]
PratikMahajan: If we dont do the `echo` it considers the variable as a command and throws an error 
```
./machineset.sh: line 76: ami-002ad2301d1a7322d: command not found

```
Similarly for all the other places with `echo`

README.md
```diff
@@ -298,6 +298,17 @@ spec:
 Example MachineSet for other cloud providers:
 - [AWS](docs/machineset-aws.md).
 
+Alternatively, the `hack/machineset.sh` script can be used to generate MachineSets for AWS and Azure platforms.
+The hack script will generate a `MachineSet.yaml` file which can be edited before using or can be used as it is.
+The script takes optional arguments like `apply` and `delete` to directly create/delete MachineSet on the cluster without 
+generating a `yaml` file.
```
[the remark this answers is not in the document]
PratikMahajan: if the optional argument is provided, it'll directly interact with cluster and will not create a `yaml` file... 
if the optional argument is not provided, it'll create a `MachineSet.yaml` which the user can edit before applying.. 
this is considering that not everyone will be editing the `MachineSet` before applying(because they might just be testing it with one VM and can easily scale up machineset from console or terminal) and providing argument will be faster and wont require cleaning up created files afterwords...

hack/machineset.sh
PratikMahajan: We are just searching for one specific(unique) image here and getting its ami-id

aravindhp: It might be fine in this case but there is potential for multiple images depending on how the AMIs are published. But it is fine to leave it as is for now.

hack/machineset.sh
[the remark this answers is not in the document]
aravindhp: Please fix as discussed over Slack:
```
#!/bin/sh
set -eo pipefail

foo="test"
cat <<EOF
${foo}
EOF
```

aravindhp (changes requested):

hack/machineset.sh
```diff
@@ -0,0 +1,195 @@
...
+
+ACTION=${1:-}
+
+function error-exit() {
+    echo "Error: $*" >&2
+    exit 1
+}
+
+# get_spec returns the template yaml common for all cloud providers
+get_spec(){
```
aravindhp, line 23: Space before `{`. Comment applies to `get_aws_ms` and `get_azure_ms` also.

Nit: You ideally want a guard that the helper is called with expected number of inputs.

hack/machineset.sh
```diff
@@ -0,0 +1,195 @@
...
+# OPTIONS
+#    $1      Action       (Optional) apply/delete the MachineSet
+# PREREQUISITES
+#    oc                   to fetch cluster info and apply/delete MachineSets on the cluster(cluster should be logged in)
+#    aws                  to fetch Windows AMI id for AWS platform (only required for clusters running on AWS)
+set -euo pipefail
+
+ACTION=${1:-}
+
+function error-exit() {
```
aravindhp, line 17: Please remove the `function` directive given we are not using it everywhere.

aravindhp (changes requested):

aravindhp (approved):
Thanks for working on this, @PratikMahajan 

/approve

PR #142, opened 2020-10-20
WINC-474: Configure Prometheus to scrape exposed metric Endpoint

VaishnaviHire:
This PR consists of changes to configure Prometheus.  It includes:

- Adding required permissions to the operator
- Adding metrics port to create Service and ServiceMonitor
- Adding `metrics` package, so that prometheus can be configured outside nodeconfig
- Adding Endpoints object with entries for all the Windows Nodes created by WMCO


pkg/controller/windowsmachine/nodeconfig/nodeconfig.go
```diff
@@ -189,6 +195,51 @@ func (nc *nodeConfig) configureNetwork() error {
...
+						Name: node.Name,
+					},
+				})
+				break
+			}
+		}
+	}
+	// return list and error
+	return nodeIPAddress, nil
+
```
LorbusChris, line 240: superfluous new line

pkg/controller/windowsmachine/nodeconfig/nodeconfig.go
```diff
@@ -274,3 +325,64 @@ func getInstanceIDfromProviderID(providerID string) string {
...
 }
+
+// ConfigurePrometheus configures Prometheus monitoring on Windows node
+func ConfigurePrometheus(k8sclientset *kubernetes.Clientset) error {
+	// check if we can get IP address
+	windowsIPList, err := getIPAddress(k8sclientset)
+	if err != nil {
+		return fmt.Errorf("could not get IP address")
+	}
+
```
LorbusChris, line 336: superfluous new line

pkg/controller/windowsmachine/nodeconfig/nodeconfig.go
```diff
@@ -274,3 +325,64 @@ func getInstanceIDfromProviderID(providerID string) string {
...
+						Protocol:    v1.ProtocolTCP,
+						AppProtocol: nil,
+					},
+				},
+			},
+		},
+	}
+	// Retrieve endpoints object
+	_, err := k8sclientset.CoreV1().Endpoints("openshift-windows-machine-config-operator").Get(context.TODO(), "windows-machine-config-operator-metrics", metav1.GetOptions{})
+
```
LorbusChris, line 371: superfluous new line

pkg/controller/windowsmachine/nodeconfig/nodeconfig.go
```diff
@@ -274,3 +325,64 @@ func getInstanceIDfromProviderID(providerID string) string {
...
+			}
+		}
+		return errors.Wrap(err, "error getting Endpoints object")
+	}
+
+	_, err = k8sclientset.CoreV1().Endpoints("openshift-windows-machine-config-operator").Update(context.TODO(), endpointsSpec, metav1.UpdateOptions{})
+	if err != nil {
+		return errors.Wrap(err, "error updating Endpoints object")
+	}
+	return nil
```
LorbusChris, line 387: please add a new line here before `return nil`

pkg/controller/windowsmachine/nodeconfig/nodeconfig.go
```diff
@@ -274,3 +325,64 @@ func getInstanceIDfromProviderID(providerID string) string {
...
+	// check if we can get IP address
+	windowsIPList, err := getIPAddress(k8sclientset)
+	if err != nil {
+		return fmt.Errorf("could not get IP address")
+	}
+
+	if err := updateMetricsEndpoint(windowsIPList, k8sclientset); err != nil {
+		return err
+	}
+	return nil
```
LorbusChris, line 340: please add a new line here before `return nil`

test/e2e/prometheus_test.go
```diff
@@ -0,0 +1,62 @@
...
+			if node.Name == endpointAddress.TargetRef.Name {
+				foundNode = true
+				break
+			}
+		}
+		if !foundNode {
+			return errors.New("target node not found in Endpoints object ")
+		}
+	}
+	return nil
```
LorbusChris, line 61: please add a new line here before `return nil`

LorbusChris (changes requested):
Looks great! I only found a few very minor nits that I'd like to be addressed.

Also please add a space in the message of the first commit, after [wmco]

deploy/olm-catalog/windows-machine-config-operator/manifests/windows-exporter-role_binding.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: prometheus-k8s
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: prometheus-k8s
+subjects:
+  - kind: ServiceAccount
+    name: prometheus-k8s
+    namespace: openshift-monitoring
```
LorbusChris, line 12: now there's the final newline missing :D (sorry to bug y'all with this)

deploy/namespace.yaml
```diff
@@ -2,3 +2,5 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: openshift-windows-machine-config-operator
+  labels:
+    openshift.io/cluster-monitoring: "true"
```
LorbusChris, line 6: same here, missing newline (GH won't show single trailing new lines in this view, but will add the :no_entry: symbol if it's missing)

deploy/olm-catalog/windows-machine-config-operator/manifests/windows-exporter-role_binding.yaml
alinaryan: whoops, adding those back in now.. Thanks for catching this!

cmd/manager/main.go
```diff
@@ -215,6 +210,19 @@ func addMetrics(ctx context.Context, cfg *rest.Config, namespace string) {
...
 
+	oclient, err := k8sclient.NewForConfig(cfg)
+	if err != nil {
+		log.Info("could not create config clientset", "error", err.Error())
+	}
+
+	// Remove selector from Metrics Service to avoid reverting changes in the Endpoints object.
+	patchData := fmt.Sprintf(`{"spec":{"selector": null }}`)
+	service, err = oclient.CoreV1().Services(namespace).Patch(ctx, service.Name, types.MergePatchType, []byte(patchData), metav1.PatchOptions{})
+
```
aravindhp, line 221: Please remove extra line

cmd/manager/main.go
```diff
@@ -215,6 +210,19 @@ func addMetrics(ctx context.Context, cfg *rest.Config, namespace string) {
...
+	if err != nil {
+		log.Info("could not remove selector from metrics service", "error", err.Error())
```
aravindhp, line 223: Please use `log.Error()`

cmd/manager/main.go
```diff
@@ -215,6 +210,19 @@ func addMetrics(ctx context.Context, cfg *rest.Config, namespace string) {
 		log.Info("could not create metrics Service", "error", err.Error())
 	}
 
+	oclient, err := k8sclient.NewForConfig(cfg)
+	if err != nil {
+		log.Info("could not create config clientset", "error", err.Error())
```
aravindhp, line 215: Please use `log.Error()`. I am not sure why the SDK generated code that used `log.Info()` for error cases.

cmd/manager/main.go
```diff
@@ -215,6 +210,19 @@ func addMetrics(ctx context.Context, cfg *rest.Config, namespace string) {
 		log.Info("could not create metrics Service", "error", err.Error())
```
aravindhp, line 210: Please feel free to update this to `log.Error()`

deploy/olm-catalog/windows-machine-config-operator/manifests/windows-exporter-role.yaml
```diff
@@ -0,0 +1,15 @@
+# Roles needed by prometheus to scrape windows metrics endpoint
```
aravindhp, line 1: Is there no way to define this inside the CSV? Moreover this is now missing from the `deploy` folder which implies, we have to use OLM to deploy the operator and can't do a direct deployment.

deploy/olm-catalog/windows-machine-config-operator/manifests/windows-exporter-role_binding.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
aravindhp, line 1: Is there no way to define this inside the CSV? Moreover this is now missing from the `deploy` folder which implies, we have to use OLM to deploy the operator and can't do a direct deployment.

aravindhp (changes requested):
Thanks for working on this, @VaishnaviHire @alinaryan. Please address my comments.

deploy/olm-catalog/windows-machine-config-operator/manifests/windows-exporter-role.yaml
VaishnaviHire: Tested by appending `role` and `role-binding` to existing files and it did not work.

However, when moving to operator-sdk v1.0, we could add them in the [`config/rbac` ](https://sdk.operatorframework.io/docs/building-operators/golang/migration/)folder.

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -320,6 +320,14 @@ func (r *ReconcileWindowsMachine) Reconcile(request reconcile.Request) (reconcil
 			log.Info("machine has current version", "name", machine.GetName(),
 				"version", node.Annotations[nodeconfig.VersionAnnotation])
 			// version annotation exists with a valid value, node is fully configured, do nothing.
+
+			// When the operator pod is restarted, the metrics Service is recreated reverting the changes in the Endpoints object, even when
+			// the Windows Nodes are fully configured. The reason is selector in the Metrics Service should be removed
+			// before creation of Endpoints, to avoid losing the updates. Configuring prometheus again, will allow
+			// the Endpoints object to be updated.
```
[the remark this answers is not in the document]
VaishnaviHire: Discussed this with @aravindhp . Decided to not update the Endpoints object with  the IP addresses of all the nodes every time. Instead, for every node event update the Endpoints object with IP address corresponding to that node. 

This ensures that we only update Endpoints object, only when required.

test/e2e/prometheus_test.go
```diff
@@ -0,0 +1,63 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"testing"
+
+	monitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
+	framework "github.com/operator-framework/operator-sdk/pkg/test"
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/require"
+)
+
+func testPrometheus(t *testing.T) {
```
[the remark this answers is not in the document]
VaishnaviHire: As a part of [WINC-528](https://issues.redhat.com/browse/WINC-528), we will be adding Prometheus rules. I think that should be testing the metrics as well

pkg/controller/windowsmachine/nodeconfig/nodeconfig.go
```diff
@@ -281,3 +328,66 @@ func getInstanceIDfromProviderID(providerID string) string {
...
+	if err := updateMetricsEndpoint(windowsIPList, k8sclientset); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// updateMetricsEndpoint upates the endpoint object with the new list of ip addresses from the Windows nodes.
+// It creates an endpoint object if the object is not found.
+func updateMetricsEndpoint(ipAddresses []v1.EndpointAddress, k8sclientset *kubernetes.Clientset) error {
```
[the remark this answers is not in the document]
VaishnaviHire: This will be addressed as a part of [this](https://github.com/openshift/windows-machine-config-operator/pull/142#discussion_r520943875) implementation

cmd/manager/main.go
```diff
@@ -195,24 +198,30 @@ func main() {
 // addMetrics will create the Services and Service Monitors to allow the operator export the metrics by using
 // the Prometheus operator
 func addMetrics(ctx context.Context, cfg *rest.Config, namespace string) {
```
sebsoto, line 200: Why not define this function within the metrics package?

sebsoto: Along with the metricsHost + metricsPort variables

cmd/manager/main.go
```diff
@@ -35,16 +37,17 @@ import (
 
 // Change below variables to serve metrics on different host or port.
 var (
-	metricsHost               = "0.0.0.0"
-	metricsPort         int32 = 8383
-	operatorMetricsPort int32 = 8686
+	metricsHost       = "0.0.0.0"
+	metricsPort int32 = 9182
```
sebsoto, line 41: This is being defined in two different places

cmd/manager/main.go
VaishnaviHire: This function is generated by `operator-sdk` here and is just creating Services and Service Monitor.  I can move it to `metrics` package

cmd/manager/main.go
```diff
@@ -195,24 +198,30 @@ func main() {
...
-		{Port: metricsPort, Name: metrics.OperatorPortName, Protocol: v1.ProtocolTCP, TargetPort: intstr.IntOrString{Type: intstr.Int, IntVal: metricsPort}},
-		{Port: operatorMetricsPort, Name: metrics.CRPortName, Protocol: v1.ProtocolTCP, TargetPort: intstr.IntOrString{Type: intstr.Int, IntVal: operatorMetricsPort}},
+		{Port: metricsPort, Name: metricsPortName, Protocol: v1.ProtocolTCP, TargetPort: intstr.IntOrString{Type: intstr.Int, IntVal: metricsPort}},
 	}
 
 	// Create Service object to expose the metrics port(s).
 	service, err := metrics.CreateMetricsService(ctx, cfg, servicePorts)
 	if err != nil {
-		log.Info("could not create metrics Service", "error", err.Error())
+		log.Error(err, "could not create metrics Service", "error")
```
mansikulkarni96, line 210: key error here has no value assigned, "error" can be removed.

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -0,0 +1,134 @@
...
+	nodePosition := -1
+	for index, endpointAddress := range endpoints.Subsets[0].Addresses {
+		if pc.hostname == endpointAddress.TargetRef.Name {
+			foundNode = true
+			nodePosition = index
+			break
+		}
+	}
+	if !foundNode {
+		return "", errors.New(fmt.Sprintf("target node %s not found in Endpoints object %s",
```
mansikulkarni96, line 93: if this contains a format string, errors.wrap can be used.

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -264,11 +265,13 @@ func (r *ReconcileWindowsMachine) Reconcile(request reconcile.Request) (reconcil
 	provisionedPhase := "Provisioned"
 	// runningPhase is the status of the machine when it is in the `Running` state, indicating that it is configured into a node
 	runningPhase := "Running"
+	// runningPhase is the status of the machine when it is in the `Deleting` state
```
mansikulkarni96, line 268: comment needs to be updated

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -264,11 +265,13 @@ func (r *ReconcileWindowsMachine) Reconcile(request reconcile.Request) (reconcil
...
+	deletingPhase := "Deleting"
 	if machine.Status.Phase == nil {
 		// Phase is nil and should be ignored by WMCO until phase is set
 		// TODO: Instead of requeuing ignore certain events: https://issues.redhat.com/browse/WINC-500
 		return reconcile.Result{}, fmt.Errorf("could not get the phase associated with machine %s", machine.Name)
-	} else if *machine.Status.Phase == runningPhase {
+	} else if *machine.Status.Phase == runningPhase || *machine.Status.Phase == deletingPhase {
```
mansikulkarni96, line 274: why is the deleting phase check and what are we changing with the upgrade here?

mansikulkarni96 (commented):
Thanks for working on this @VaishnaviHire , PTAL at some of my comments

pkg/controller/windowsmachine/windowsmachine_controller.go
VaishnaviHire: Changed the predicate to get Delete events. We need delete events, so that we could remove the deleted node entry in the Endpoints object

pkg/controller/windowsmachine/windowsmachine_controller.go
aravindhp, line 268: Please fix the comment

pkg/controller/windowsmachine/windowsmachine_controller.go
aravindhp, line 274: Please call out why we care about the `deletingPhase` now

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -283,41 +286,63 @@ func (r *ReconcileWindowsMachine) Reconcile(request reconcile.Request) (reconcil
...
-						"Machine %v deletion restricted as the maximum unhealthy machines can`t exceed %v count",
-						machine.Name, maxUnhealthyCount)
-					return reconcile.Result{Requeue: true}, nil
-				}
-				if !machine.GetDeletionTimestamp().IsZero() {
-					// Delete already initiated
+		// Get prometheus configuration
+		addresses, err := r.getNodeAddress(machine)
+		if err != nil {
+			log.Error(err, "could not get ip address and hostname associated with the ", "node", node.Name, "adresses", node.Status.Addresses)
```
aravindhp, line 292: Please don't try to form sentences with the logr approach. State the error like `"could not IP address and hostname associated with the node", "name", node.Name...`,

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -341,15 +366,9 @@ func (r *ReconcileWindowsMachine) Reconcile(request reconcile.Request) (reconcil
...
-	ipAddress := ""
-	for _, address := range machine.Status.Addresses {
-		if address.Type == core.NodeInternalIP {
-			ipAddress = address.Address
-		}
-	}
-	if len(ipAddress) == 0 {
-		return reconcile.Result{}, errors.Errorf("no internal ip address associated with machine %s",
-			machine.Name)
+	addresses, _ := r.getNodeAddress(machine)
```
aravindhp, line 369: Ideally we need a struct like `MachineInfo` that encapsulates all the metadata that we care about the Machine and pass that around.

test/e2e/create_test.go
```diff
@@ -47,7 +47,7 @@ func creationTestSuite(t *testing.T) {
 	t.Run("UserData validation", func(t *testing.T) { testUserData(t) })
 	t.Run("UserData idempotent check", func(t *testing.T) { testUserDataTamper(t) })
 	t.Run("Node Logs", func(t *testing.T) { testNodeLogs(t) })
-	t.Run("Windows Exporter metrics test", func(t *testing.T) { testWindowsExporter(t) })
+	t.Run("metrics validation", testMetrics)
```
aravindhp, line 50: `Metrics validation`

test/e2e/metrics_test.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	core "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func testMetrics(t *testing.T) {
+
```
aravindhp, line 19: Please remove

test/e2e/upgrade_test.go
```diff
@@ -95,6 +97,8 @@ func testTamperAnnotation(t *testing.T) {
 	require.NoError(t, err, "windows node upgrade failed")
 	// Test if the version annotation corresponds to the current operator version
 	testVersionAnnotation(t)
+	// Test if prometheus is reconfigured with ip address of newly configured node
+	testPrometheus(t)
```
aravindhp, line 101: Don't you need a test after all nodes have been deleted to check if the enpoints are blank?

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -0,0 +1,134 @@
...
+	return &prometheusConfig{
+		k8sclientset:    clientset,
+		hostname:        hostname,
+		endpointAddress: endpointAddress}
+
+}
+
+// Configure patches the endpoint object to reflect the current list Windows nodes. When addIPAddress is
+// true it adds the endpoint address for the current node, else it removes that endpoint address
+func (pc *prometheusConfig) Configure(addIPAddress bool) error {
```
aravindhp, line 46: Why does the caller need to decided if it is a situation of adding or removing an IP address? Shouldn't you be able to decide that based on the `endpointAddress`?

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -0,0 +1,134 @@
...
+
+const (
+	windowsMetricsEndpoints          = "windows-machine-config-operator-metrics"
+	windowsMetricsEndpointsNamespace = "openshift-windows-machine-config-operator"
+	metricsPort                      = 9182
+)
+
+// prometheusConfig holds the information required to configure prometheus, so that it can scrape metrics from the
+// given endpoint address
+type prometheusConfig struct {
```
aravindhp, line 22: Is this `prometheusConfig` or `prometheusNodeConfig`?

aravindhp (changes requested):
Thanks for working on this, @VaishnaviHire @alinaryan. This looks really good. Please address my comments.

pkg/controller/windowsmachine/windowsmachine_controller.go
aravindhp: This is a big change. Please call this out in the commit message mentioning why you had to make the change.

pkg/controller/windowsmachine/metrics/metrics.go
VaishnaviHire: `endpointAddress` just has the ipAddress and node info. The addition or removal of ip address depends on the events.

Given that we have a case where we need to reconfigure during pod restart, we would not be able to decide based on `endpointAddress`

test/e2e/metrics_test.go
```diff
@@ -0,0 +1,111 @@
...
+			winNodeInternalIP := ""
+			for _, address := range winNode.Status.Addresses {
+				if address.Type == core.NodeInternalIP {
+					winNodeInternalIP = address.Address
+				}
+			}
+			require.Greaterf(t, len(winNodeInternalIP), 0, "test requires Windows node %s to have internal IP", winNode.Name)
+
+			// This will curl the windows server. curl must be present in the container image.
+			linuxCurlerCommand := []string{"bash", "-c", "curl http://" + winNodeInternalIP + ":9182/metrics"}
```
saifshaikh48, line 44: The `"9182/metrics"` can be replaced with `metricsPort` and `metricsPortName` variables

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -0,0 +1,196 @@
...
+		return errors.Wrapf(err, "could not remove node %v from endpoints object", pc.hostname)
+	}
+	log.Info("address removed from endpoints object", "IPAddress", pc.endpointAddress.IP)
+	return nil
+}
+
+// addMetrics will create the Services and Service Monitors to allow the operator export the metrics by using
+// the Prometheus operator
+func AddMetrics(ctx context.Context, cfg *rest.Config, namespace string) {
+
```
aravindhp, line 156: Please remove

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -0,0 +1,196 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "could not remove node %v from endpoints object", pc.hostname)
+	}
+	log.Info("address removed from endpoints object", "IPAddress", pc.endpointAddress.IP)
+	return nil
+}
+
+// addMetrics will create the Services and Service Monitors to allow the operator export the metrics by using
+// the Prometheus operator
+func AddMetrics(ctx context.Context, cfg *rest.Config, namespace string) {
```
aravindhp, line 155: This will now become just `Add()` so that there is not stuttering in the call sites: `metrics.Add()`.

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	// CreateServiceMonitors will automatically create the prometheus-operator ServiceMonitor resources
+	// necessary to configure Prometheus to scrape metrics from this operator.
+	services := []*v1.Service{service}
+	_, err = metrics.CreateServiceMonitors(cfg, namespace, services)
+	if err != nil {
+		log.Info("could not create ServiceMonitor object", "error", err.Error())
+		// If this operator is deployed to a cluster without the prometheus-operator running, it will return
+		// ErrServiceMonitorNotPresent, which can be used to safely skip ServiceMonitor creation.
```
aravindhp, line 189: In this case won't the endpoint update also fail causing node configuration to always be retried but never succeed? I think you need to figure out if metrics is enabled before proceeding with `Configure()`.

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -110,9 +111,9 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		UpdateFunc: func(e event.UpdateEvent) bool {
 			return isWindowsMachine(e.MetaNew.GetLabels())
 		},
-		// ignore delete event for all Machines as WMCO does not react to node getting deleted
+		// We need delete events, so that we could remove the deleted node entry in the metrics Endpoints object
```
aravindhp, line 114: `we can remove...`

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -264,60 +265,87 @@ func (r *ReconcileWindowsMachine) Reconcile(request reconcile.Request) (reconcil
...
 
-				if err := r.client.Delete(context.TODO(), machine); err != nil {
-					r.recorder.Eventf(machine, core.EventTypeWarning, "MachineDeletionFailed",
-						"Machine %v deletion failed: %v", machine.Name, err)
-					return reconcile.Result{}, err
 				}
-				log.Info("machine has been remediated by deletion", "name", machine.GetName())
-				r.recorder.Eventf(machine, core.EventTypeNormal, "MachineDeleted",
-					"Machine %v has been remediated by deleting the Machine object", machine.Name)
+				log.Info("prometheus is reconfigured for the node", "name", node.Name)
```
aravindhp, line 337: Please move this log message to spot where the actual reconfiguration takes place. Otherwise this message will always appear even if we are not updating the endpoint object.

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -264,60 +265,87 @@ func (r *ReconcileWindowsMachine) Reconcile(request reconcile.Request) (reconcil
...
-			// version annotation exists with a valid value, node is fully configured, do nothing.
+		} else {
+			// Machine needs to be checked for deletion. If the machine is being deleted, remove its IP Address from the
+			// list of addresses in metrics endpoint object.
+			// Reconfigure prometheus to remove endpoint address of the deleted node in Endpoints object
+			if err := pc.Configure(false); err != nil {
+				return reconcile.Result{}, errors.Wrapf(err, "failed to configure Prometheus for node %s",
+					addresses[core.NodeHostName])
+			}
+			log.Info("prometheus is reconfigured for the node", "name", addresses[core.NodeHostName])
```
aravindhp, line 348: Same comment here.

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -485,3 +512,21 @@ func (r *ReconcileWindowsMachine) isWindowsMachineHealthy(machine *mapi.Machine)
 
 	return true
 }
+
+// getNodeAddress returns map of ip address and hostname associated with the node.
```
aravindhp, line 516: `IP addresses`

aravindhp (changes requested):
Sorry @VaishnaviHire I should have caught some of these issues sooner.

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -0,0 +1,196 @@
...
+func (pc *prometheusNodeConfig) Configure(addIPAddress bool) error {
+	// Retrieve endpoints object
+	endpoints, err := pc.k8sclientset.CoreV1().Endpoints(windowsMetricsEndpointsNamespace).Get(context.TODO(),
+		windowsMetricsEndpoints, metav1.GetOptions{})
+	if err != nil {
+		return errors.Wrap(err, "error getting Endpoints object")
+	}
+
+	// Add Metrics Port if not present.
+	if len(endpoints.Subsets[0].Ports) == 0 || endpoints.Subsets[0].Ports[0].Port != 9182 {
```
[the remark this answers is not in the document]
VaishnaviHire: Yes, Subsets is a required field and it will always be present

pkg/controller/windowsmachine/metrics/metrics.go
VaishnaviHire: I can check if `ServiceMonitor` exists in Configure()

pkg/controller/windowsmachine/metrics/metrics.go
VaishnaviHire: Kept it as is since I wanted error object

aravindhp: `fmt.Errorf` is what we have been doing similar situations.

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -0,0 +1,210 @@
...
+			pc.hostname, endpoints.Name))
+	}
+
+	return strconv.Itoa(nodePosition), nil
+}
+
+// addEndpointAddress adds node ip address to the list of subsets in Endpoints object
+func (pc *prometheusNodeConfig) addEndpointAddress() error {
+	_, err := pc.isNodeInEndpoints()
+	if err != nil {
```
sebsoto, line 132: See above comments

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -0,0 +1,210 @@
...
+	if !foundNode {
+		return "", errors.New(fmt.Sprintf("target node %s not found in Endpoints object %s",
+			pc.hostname, endpoints.Name))
+	}
+
+	return strconv.Itoa(nodePosition), nil
+}
+
+// addEndpointAddress adds node ip address to the list of subsets in Endpoints object
+func (pc *prometheusNodeConfig) addEndpointAddress() error {
```
sebsoto, line 130: See above comment on function signature change

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -0,0 +1,210 @@
...
+
+	// When a selector is present in a headless service i.e. spec.ClusterIP=None, Kubernetes manages the
+	// list of endpoints reverting all the changes made by the operator. Remove selector from Metrics Service to avoid
+	// reverting changes in the Endpoints object.
+	patchData := fmt.Sprintf(`{"spec":{"selector": null }}`)
+	service, err = oclient.CoreV1().Services(namespace).Patch(ctx, service.Name, types.MergePatchType,
+		[]byte(patchData), metav1.PatchOptions{})
+	if err != nil {
+		log.Error(err, "could not remove selector from metrics service")
+	}
```
aravindhp, line 193: Do we need to this if metrics is not enabled? Maybe you should create the `ServiceMonitor` first and only if that succeeds proceed to do this?

cmd/manager/main.go
```diff
@@ -181,7 +173,7 @@ func main() {
 	}
 
 	// Add the Metrics Service
-	addMetrics(ctx, cfg, namespace)
+	winmetrics.Add(ctx, cfg, namespace)
```
sebsoto, line 176: Are we considering metrics an optional feature? Is that why we are not going to retry if we cant start metrics services?

aravindhp (changes requested):

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -0,0 +1,210 @@
...
+func Add(ctx context.Context, cfg *rest.Config, namespace string) {
+	// Add to the below struct any other metrics ports you want to expose.
+	servicePorts := []v1.ServicePort{
+		{Port: Port, Name: metricsPortName, Protocol: v1.ProtocolTCP, TargetPort: intstr.IntOrString{Type: intstr.Int, IntVal: Port}},
+	}
+
+	// Create Service object to expose the metrics port(s).
+	service, err := metrics.CreateMetricsService(ctx, cfg, servicePorts)
+	if err != nil {
+		log.Error(err, "could not create metrics Service")
```
sebsoto, line 178: Why not error out here? It seems wrong to me to continue configuring metrics if such big errors occur. I think returning an error and deciding how handle it at the call site is a better way to approach this. If we dont want metrics failures to stop the operator from coming up thats fine, but we can return these errors and the log them at the call site.

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -264,60 +265,89 @@ func (r *ReconcileWindowsMachine) Reconcile(request reconcile.Request) (reconcil
...
 		// TODO: Instead of requeuing ignore certain events: https://issues.redhat.com/browse/WINC-500
 		return reconcile.Result{}, fmt.Errorf("could not get the phase associated with machine %s", machine.Name)
-	} else if *machine.Status.Phase == runningPhase {
-		// Machine has been configured into a node, we need to ensure that the version annotation exists. If it doesn't
-		// the machine was not fully configured and needs to be configured properly.
-		if machine.Status.NodeRef == nil {
-			// NodeRef missing. Requeue and hope it is created. It never being created indicates an issue with the
-			// machine api operator
-			return reconcile.Result{}, fmt.Errorf("ready Windows machine %s missing NodeRef", machine.GetName())
+	} else if *machine.Status.Phase == runningPhase || *machine.Status.Phase == deletingPhase {
```
sebsoto, line 274: What happens if the machine is fully deleted before we can reconcile it?

sebsoto: https://github.com/openshift/windows-machine-config-operator/pull/142/files?file-filters%5B%5D=.go&file-filters%5B%5D=.txt&file-filters%5B%5D=No+extension#diff-f6cf75653deff7ed72be94f059acc292d15113ed7c9f9ce0fdc5e842fa71aa72R256-R259

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -0,0 +1,210 @@
...
+		namespace:       windowsMetricsEndpointsNamespace}, err
+
+}
+
+// Configure patches the endpoint object to reflect the current list Windows nodes. When addIPAddress is
+// true it adds the endpoint address for the current node, else it removes that endpoint address
+func (pc *prometheusNodeConfig) Configure(addIPAddress bool) error {
+	// Check if metrics are enabled in current cluster
+	if !metricsEnabled {
+		return errors.New("install-prometheus operator to enable prometheus configuration")
```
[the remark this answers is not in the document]
VaishnaviHire: Is that because we would keep `Reconciling` ? 

Makes sense, I will log the error

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -380,7 +404,7 @@ func (r *ReconcileWindowsMachine) Reconcile(request reconcile.Request) (reconcil
 }
 
 // addWorkerNode configures the given Windows VM, adding it as a node object to the cluster
-func (r *ReconcileWindowsMachine) addWorkerNode(ipAddress, providerName, instanceID string) error {
+func (r *ReconcileWindowsMachine) addWorkerNode(ipAddress, providerName, instanceID, hostname string) error {
```
sebsoto, line 407: This is going to conflict with the vSphere changes I think https://github.com/openshift/windows-machine-config-operator/pull/149/commits/410ad4c900fe4bdc383723f567bb21a052fb7227

pkg/controller/windowsmachine/metrics/metrics.go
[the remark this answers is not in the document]
aravindhp: Yes, we will keep reconciling otherwise.

sebsoto (changes requested):

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+	// Retrieve endpoints object
+	endpoints, err := pc.k8sclientset.CoreV1().Endpoints(pc.namespace).Get(context.TODO(),
+		windowsMetricsEndpoints, metav1.GetOptions{})
+	if err != nil {
+		return errors.Wrap(err, "error getting Endpoints object")
+	}
+
+	// Add Metrics Port if not present.
+	if len(endpoints.Subsets[0].Ports) == 0 || endpoints.Subsets[0].Ports[0].Port != 9182 {
```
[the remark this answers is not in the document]
VaishnaviHire: Yes we are overwriting the value in the sense, we are providing the port number 9182. This check is if we add different ports and for some reason, do not have 9182, we should create it.

I can remove it, if this seems unnecessary.

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -0,0 +1,210 @@
...
+		patchData := fmt.Sprintf(
+			"[{\"op\": \"add\", \"path\": \"/subsets/0/ports/-\",\"value\":{\"name\": \"metrics\", "+
+				"\"port\":\"%s\",\"protocol\": \"%s\"}}]", strconv.Itoa(int(Port)), v1.ProtocolTCP)
+
+		_, err := pc.k8sclientset.CoreV1().Endpoints(pc.namespace).Patch(context.TODO(),
+			windowsMetricsEndpoints, types.JSONPatchType, []byte(patchData), metav1.PatchOptions{})
+		if err != nil {
+			return err
+		}
+		return nil
```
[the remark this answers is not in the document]
VaishnaviHire: yes. Because `windows-exporter` is only exposed on port 9182

pkg/controller/windowsmachine/metrics/metrics.go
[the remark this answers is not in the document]
aravindhp: @VaishnaviHire and I discussed this. What eventually needs to happen is to have an endpoints controller that always stomps this back to the way we want it to look. In addition we would also need a node controller to keep the endpoints in sync. @VaishnaviHire please create stories for this.

pkg/controller/windowsmachine/windowsmachine_controller.go
aravindhp: I suggest we normalize around requiring just a single input if possible.

pkg/controller/windowsmachine/metrics/metrics.go
VaishnaviHire: This is called in `main.go`. Do you want me to return error and `log` the message there ?

pkg/controller/windowsmachine/windowsmachine_controller.go
VaishnaviHire: Yes, would need some changes. Does the changed hostname in vsphere matches the Node name ?

pkg/controller/windowsmachine/metrics/metrics.go
sebsoto: I think that would be ideal

pkg/controller/windowsmachine/windowsmachine_controller.go
sebsoto: It ensures hostname of the Windows VM matches the machine name, only for vSphere

pkg/controller/windowsmachine/metrics/metrics.go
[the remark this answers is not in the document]
VaishnaviHire: I am removing the patching for port. [metrics.CreateService](https://github.com/openshift/windows-machine-config-operator/blob/master/cmd/manager/main.go#L213) ensures for creating endpoints with desired ports

cmd/manager/main.go
aravindhp: Metrics is dependent on the cluster having prometheus installed. So it is optional in that sense.

pkg/controller/windowsmachine/windowsmachine_controller.go
VaishnaviHire: As long as the passed value matches the targeted node name, we should be good

pkg/controller/windowsmachine/windowsmachine_controller.go
VaishnaviHire: @sebsoto Did not see this scenario in testing. I am not sure about it

cmd/manager/main.go
```diff
@@ -31,14 +27,10 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
-)
 
-// Change below variables to serve metrics on different host or port.
-var (
-	metricsHost               = "0.0.0.0"
-	metricsPort         int32 = 8383
-	operatorMetricsPort int32 = 8686
+	winmetrics "github.com/openshift/windows-machine-config-operator/pkg/controller/windowsmachine/metrics"
```
aravindhp, line 31: Please group this correctly.

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -0,0 +1,219 @@
...
+	"k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/client-go/kubernetes"
+	k8sclient "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+	"github.com/openshift/windows-machine-config-operator/pkg/controller/windowsmachine/nodeconfig"
```
aravindhp, line 20: Please group this correctly

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -0,0 +1,219 @@
...
+	// Host is the host address used by Windows metrics
+	Host = "0.0.0.0"
+	// Port is the port number on which windows-exporter is exposed.
+	Port int32 = 9182
+)
+
+// prometheusNodeConfig holds the information required to configure prometheus, so that it can scrape metrics from the
+// given endpoint address
+type prometheusNodeConfig struct {
+	// k8sclientset holds the information related to kubernetes clientset
```
aravindhp, line 43: Nit: its a handle that allows you to interact with the Kubernetes API.

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	return &prometheusNodeConfig{
+		k8sclientset: clientset,
+		namespace:    windowsMetricsEndpointsNamespace,
+	}, err
+}
+
+// Add will create the Services and Service Monitors to allow the operator export the metrics by using
+// the Prometheus operator
```
aravindhp, line 73: `Add will create the Services and Service Monitors that allows the operator to export the metrics...`

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -0,0 +1,219 @@
...
+	windowsMetricsEndpoints = "windows-machine-config-operator-metrics"
+	// metricsPortName specifies the portname used for prometheus monitoring
+	PortName = "metrics"
+	// Host is the host address used by Windows metrics
+	Host = "0.0.0.0"
+	// Port is the port number on which windows-exporter is exposed.
+	Port int32 = 9182
+)
+
+// prometheusNodeConfig holds the information required to configure prometheus, so that it can scrape metrics from the
```
aravindhp, line 40: `Prometheus`

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -0,0 +1,219 @@
...
+	service, err = oclient.CoreV1().Services(namespace).Patch(ctx, service.Name, types.MergePatchType,
+		[]byte(patchData), metav1.PatchOptions{})
+	if err != nil {
+		return errors.Wrap(err, "could not remove selector from metrics service")
+	}
+
+	return nil
+}
+
+// updateMetricsEndpoint upates the endpoint object with the new list of ip addresses from the Windows nodes.
```
saifshaikh48, line 118: nit: `syncMetricsEndpoint updates ...`

aravindhp, line 118: `IP`

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -0,0 +1,219 @@
...
+	if err != nil {
+		return errors.Wrap(err, "unable to get patch data in bytes")
+	}
+
+	_, err = pc.k8sclientset.CoreV1().Endpoints(pc.namespace).
+		Patch(context.TODO(), windowsMetricsEndpoints, types.JSONPatchType, patchDataBytes, metav1.PatchOptions{})
+	return errors.Wrap(err, "unable to sync metrics endpoints")
+}
+
+// Configure configures patches the endpoint object to reflect the current list Windows nodes.
```
aravindhp, line 138: `Configure patches...`

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -0,0 +1,219 @@
...
+	endpoints, err := pc.k8sclientset.CoreV1().Endpoints(pc.namespace).Get(context.TODO(),
+		windowsMetricsEndpoints, metav1.GetOptions{})
+	if err != nil {
+		return errors.Wrapf(err, "could not get metrics endpoints %v", windowsMetricsEndpoints)
+	}
+
+	if !pc.isEndpointsValid(nodes, endpoints) {
+		// check if we can get list of endpoint addresses
+		windowsIPList := pc.getNodeEndpointAddresses(nodes)
+		log.Info("get ip list", "list", windowsIPList)
```
aravindhp, line 162: Do we really want to log this for the release version?

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -0,0 +1,219 @@
...
+		if err := pc.syncMetricsEndpoint(windowsIPList); err != nil {
+			return errors.Wrap(err, "error updating endpoints object with list of endpoint addresses")
+		}
+	}
+	log.Info("prometheus configured", "endpoints", windowsMetricsEndpoints, "port", Port, "name", PortName)
+	return nil
+}
+
+// getNodeEndpointAddresses returns a list of endpoint addresses according to the given list of Windows nodes
+func (pc *prometheusNodeConfig) getNodeEndpointAddresses(nodes *v1.NodeList) []v1.EndpointAddress {
```
aravindhp, line 173: Why is this method part of `prometheusNodeConfig`?

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -0,0 +1,219 @@
...
+			}
+		}
+	}
+	// return list and error
+	return nodeIPAddress
+}
+
+// isEndpointsValid returns true if Endpoints object has entries for all the Windows nodes in the cluster
+// else returns false.
+func (pc *prometheusNodeConfig) isEndpointsValid(nodes *v1.NodeList, endpoints *v1.Endpoints) bool {
```
aravindhp, line 200: Why is this method part of `prometheusNodeConfig`?

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -0,0 +1,219 @@
...
+	if len(nodes.Items) != len(endpoints.Subsets[0].Addresses) {
+		return false
+	}
+
+	for _, node := range nodes.Items {
+		nodeFound := false
+		for _, address := range endpoints.Subsets[0].Addresses {
+			if address.TargetRef.Name == node.Name {
+				nodeFound = true
+				break
```
aravindhp, line 211: Can't you just `return true` here? With that you won't need `nodeFound`.

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -0,0 +1,219 @@
...
+			}
+		}
+		if !nodeFound {
+			return false
+		}
+	}
+	return true
```
aravindhp, line 218: And this would be `return false`

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -317,11 +324,19 @@ func (r *ReconcileWindowsMachine) Reconcile(request reconcile.Request) (reconcil
...
+			if err := pc.Configure(); err != nil {
+				return reconcile.Result{}, errors.Wrap(err, "unable to configure Prometheus")
+			}
 			return reconcile.Result{}, nil
 		}
 	} else if *machine.Status.Phase != provisionedPhase {
 		log.V(1).Info("machine not provisioned", "phase", *machine.Status.Phase)
+		// configure prometheus if it is not configured
+		if err := pc.Configure(); err != nil {
+			return reconcile.Result{}, errors.Wrap(err, "unable to configure Prometheus")
```
aravindhp, line 338: Why are we doing this for Machines that are not in the `provisionedPhase`? This implies the comment should explain why it is needed.

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -248,6 +249,12 @@ func (r *ReconcileWindowsMachine) Reconcile(request reconcile.Request) (reconcil
 		return reconcile.Result{}, errors.Wrapf(err, "unable to get secret %s", request.NamespacedName)
 	}
 
+	// Initialize prometheus configuration
+	pc, err := metrics.NewPrometheusNodeConfig(r.k8sclientset)
```
aravindhp, line 253: Is it possible to move this closer to the scope of its usage?

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -375,7 +390,10 @@ func (r *ReconcileWindowsMachine) Reconcile(request reconcile.Request) (reconcil
 	}
 	r.recorder.Eventf(machine, core.EventTypeNormal, "MachineSetup",
 		"Machine %s configured successfully", machine.Name)
-
+	// configure prometheus if it is not configured
```
aravindhp, line 393: Again, the comment should explain why. The main issue is that it is not clear why we call it in multiple places. I feel we are being overzealous in calling `pc.Configure()`.

test/e2e/metrics_test.go
```diff
@@ -0,0 +1,113 @@
...
+			winNodeInternalIP := ""
+			for _, address := range winNode.Status.Addresses {
+				if address.Type == core.NodeInternalIP {
+					winNodeInternalIP = address.Address
+				}
+			}
+			require.Greaterf(t, len(winNodeInternalIP), 0, "test requires Windows node %s to have internal IP", winNode.Name)
+
+			// This will curl the windows server. curl must be present in the container image.
+			linuxCurlerCommand := []string{"bash", "-c", "curl http://" + winNodeInternalIP + ":" + metrics.Port + "/" + metrics.PortName}
```
aravindhp, line 45: 120 col

aravindhp (changes requested):

pkg/controller/windowsmachine/metrics/metrics.go
VaishnaviHire: Not sure of the regrouping here . I have regrouped as -

First group : go packages
Second group : external repo
third group : wmco

pkg/controller/windowsmachine/metrics/metrics.go
VaishnaviHire: I will remove this

pkg/controller/windowsmachine/metrics/metrics.go
VaishnaviHire: I want to return `true` only after all nodes are found and return if `single` node is not found

pkg/controller/windowsmachine/metrics/metrics.go
aravindhp: OK, I thought the second group was `wmco`. We need to add this grouping to the contribution guide.

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -317,11 +324,19 @@ func (r *ReconcileWindowsMachine) Reconcile(request reconcile.Request) (reconcil
 			}
 			log.Info("machine has current version", "name", machine.GetName(),
 				"version", node.Annotations[nodeconfig.VersionAnnotation])
-			// version annotation exists with a valid value, node is fully configured, do nothing.
+			// version annotation exists with a valid value, node is fully configured.
+			// configure prometheus if it is not configured
+			if err := pc.Configure(); err != nil {
```
saifshaikh48, line 329: I am wondering why we make a call to `pc.Configure()` regardless of the machine's status. For my learning, could you please explain why we'd want to configure Prometheus again if the machine is already `Running` for example? I am assuming here that a machine would already have Prometheus configured if it has reached the `Running` state.

pkg/controller/windowsmachine/metrics/metrics.go
aravindhp: Please call that out explicitly in the doc then.

pkg/controller/windowsmachine/windowsmachine_controller.go
VaishnaviHire: This call is required when a WMCO pod restarts. It creates a [headless service](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) which causes Endpoints object to revert all the changes.

We are creating that service [here](https://github.com/openshift/windows-machine-config-operator/pull/142/files#diff-ccf7cfc3caf950d956bbe879bfbd3fd978235b1e00781499c93510d8e95ff994R81). Hence we need to reconfigure

pkg/controller/windowsmachine/windowsmachine_controller.go
VaishnaviHire: Since the configuration is being called at different places in the controller, I am tending to keep this here

sebsoto: Instead of doing this here, what about making this part of the reconciler object and doing it when creating the reconciler? The object isnt changing on each reconcile is it?

pkg/controller/windowsmachine/windowsmachine_controller.go
sebsoto: I almost feel like we need a separate node controller for this, as this doesn't deal with the machine object at all, and isnt fitting in well within this controller. I think a node controller makes more sense, and will help clean up the code, as its getting somewhat convoluted in my opinion.

sebsoto (changes requested):

sebsoto:
/lgtm
Good work, sorry for the design changes

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -317,11 +327,21 @@ func (r *ReconcileWindowsMachine) Reconcile(request reconcile.Request) (reconcil
 			}
 			log.Info("machine has current version", "name", machine.GetName(),
 				"version", node.Annotations[nodeconfig.VersionAnnotation])
-			// version annotation exists with a valid value, node is fully configured, do nothing.
+			// version annotation exists with a valid value, node is fully configured.
+			// configure Prometheus when we have Ready Windows Nodes. This is required to update Endpoints object if
+			// it gets reverted during pod restart.
```
aravindhp, line 332: `operator pod restarts`

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -317,11 +327,21 @@ func (r *ReconcileWindowsMachine) Reconcile(request reconcile.Request) (reconcil
 			}
 			log.Info("machine has current version", "name", machine.GetName(),
 				"version", node.Annotations[nodeconfig.VersionAnnotation])
-			// version annotation exists with a valid value, node is fully configured, do nothing.
+			// version annotation exists with a valid value, node is fully configured.
+			// configure Prometheus when we have Ready Windows Nodes. This is required to update Endpoints object if
```
aravindhp, line 331: Why is it `Ready`? I suggest saying `already configured Windows Nodes`.

aravindhp (changes requested):

aravindhp (approved):
/approve

PR #143, opened 2020-10-21
[DOCS] Add a quick start example to prep base windows server

usrbinkat:
Proposal:
Provide a quick start guide providing minimum viable prototype steps to prepare a windows server core node.

This pull request contains rough sketched notes from my own experience stumbling through rough lessons learned while collecting cross industry docs and peer tribal knowledge and may serve as a starting point for accumulating content from other invested individuals as well.

### DIFF
--- a/docs/ms-server-core-docker-quickstart.md
+++ b/docs/ms-server-core-docker-quickstart.md
@@ -1,0 +1,28 @@
+# Quick Start MS Windows Server Datacenter Core with Containers 
+This guide assumes the user has a current version of MS Windows Server 2019
+Datacenter (core) with containers ready to prepare as a worker node for
+OpenShift.
+
+## Enable ssh service
+  - Run in admin powershell
+```sh
+Add-WindowsCapability -Online -Name OpenSSH.Server~~~~0.0.1.0
+Add-WindowsCapability -Online -Name OpenSSH.Client~~~~0.0.1.0
+
+Start-Service ssh-agent
+Start-Service sshd
+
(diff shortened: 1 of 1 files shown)


ravisantoshgudimetla:
@usrbinkat - Thank you for opening a PR.

Looks like you're following ansible approach in 4.4? Can you try the operator which is in this repo?

usrbinkat:
@ravisantoshgudimetla yes, identified in channel that I had some preconceived ideas from the previous bootstrap operator. I am currently targeting airgap/restricted usage of WMCO on bare metal & vmware deployments of OCP so having all image prep steps outlined in a central location is high value task for me.

To clarify actual requirements, a windows core server image must meet these requirements:
  - public key passed into image
  - ssh private key kubernetes secret declared in machineset
  - docker service pre-installed in image

Those are the windows core server prereqs i understand to be required. What I still dont understand is how the public key is passed into the image? I understand it can be done via user-data in some scenarios but do not see where you are doing that yet. 

Currently I'm baking the pub key into the golden image but that seems like an anti-pattern and inconsistent with the cloudy paradigm.

@aravindhp looking for colab points [here](https://github.com/openshift/windows-machine-config-operator/blob/d3cbfd069e79b5d0a0bef0177ceda1f7ae16dd1a/README.md) today. Thank you for pointing that out.

aravindhp:
/close
The information here is reflected in https://github.com/openshift/windows-machine-config-operator/pull/149. @usrbinkat if we have missed anything please let us know.

PR #151, opened 2020-10-29
WINC-525: [WMCO] Handle OpenSSHUtils deprecation

PratikMahajan:
OpenSSHUtils is getting deprecated
MicrosoftDocs/windowsserverdocs#3806
We need to move away from using OpenSSHUtils to configure ssh
access to Windows VMs

Remove installing OpenSSHUtils and its usage for setting the
authorized keys and uses the built-in ssh mechanism supported
by Windows Server 1709 and above

### DIFF
--- a/pkg/controller/secrets/secrets.go
+++ b/pkg/controller/secrets/secrets.go
@@ -64,23 +64,23 @@ func GenerateUserData(privateKey []byte) (*core.Secret, error) {
 			$firewallRuleName = "ContainerLogsPort"
 			$containerLogsPort = "10250"
 			New-NetFirewallRule -DisplayName $firewallRuleName -Direction Inbound -Action Allow -Protocol TCP -LocalPort $containerLogsPort -EdgeTraversalPolicy Allow
-			Install-PackageProvider -Name NuGet -MinimumVersion 2.8.5.201 -Force
-			Install-Module -Force OpenSSHUtils
-			Set-Service -Name ssh-agent -StartupType ‘Automatic’
 			Set-Service -Name sshd -StartupType ‘Automatic’
-			Start-Service ssh-agent
 			Start-Service sshd
 			$pubKeyConf = (Get-Content -path C:\ProgramData\ssh\sshd_config) -replace '#PubkeyAuthentication yes','PubkeyAuthentication yes'
 			$pubKeyConf | Set-Content -Path C:\ProgramData\ssh\sshd_config
  			$passwordConf = (Get-Content -path C:\ProgramData\ssh\sshd_config) -replace '#PasswordAuthentication yes','PasswordAuthentication yes'
 			$passwordConf | Set-Content -Path C:\ProgramData\ssh\sshd_config
-			$authFileConf = (Get-Content -path C:\ProgramData\ssh\sshd_config) -replace 'AuthorizedKeysFile __PROGRAMDATA__/ssh/administrators_authorized_keys','#AuthorizedKeysFile __PROGRAMDATA__/ssh/administrators_authorized_keys'
-			$authFileConf | Set-Content -Path C:\ProgramData\ssh\sshd_config
-			$pubKeyLocationConf = (Get-Content -path C:\ProgramData\ssh\sshd_config) -replace 'Match Group administrators','#Match Group administrators'
-			$pubKeyLocationConf | Set-Content -Path C:\ProgramData\ssh\sshd_config
+			$authorizedKeyFilePath = "$env:ProgramData\ssh\administrators_authorized_keys"
+			New-Item -Force $authorizedKeyFilePath
+			echo "` + string(pubKeyBytes[:]) + `"| Out-File $authorizedKeyFilePath -Encoding ascii
+			$acl = Get-Acl C:\ProgramData\ssh\administrators_authorized_keys
+			$acl.SetAccessRuleProtection($true, $false)
+			$administratorsRule = New-Object system.security.accesscontrol.filesystemaccessrule("Administrators","FullControl","Allow")
+			$systemRule = New-Object system.security.accesscontrol.filesystemaccessrule("SYSTEM","FullControl","Allow")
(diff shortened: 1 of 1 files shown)


pkg/controller/secrets/secrets.go
```diff
@@ -65,22 +65,22 @@ func GenerateUserData(privateKey []byte) (*core.Secret, error) {
...
 			Install-PackageProvider -Name NuGet -MinimumVersion 2.8.5.201 -Force
-			Install-Module -Force OpenSSHUtils
-			Set-Service -Name ssh-agent -StartupType ‘Automatic’
 			Set-Service -Name sshd -StartupType ‘Automatic’
-			Start-Service ssh-agent
 			Start-Service sshd
 			$pubKeyConf = (Get-Content -path C:\ProgramData\ssh\sshd_config) -replace '#PubkeyAuthentication yes','PubkeyAuthentication yes'
 			$pubKeyConf | Set-Content -Path C:\ProgramData\ssh\sshd_config
- 			$passwordConf = (Get-Content -path C:\ProgramData\ssh\sshd_config) -replace '#PasswordAuthentication yes','PasswordAuthentication yes'
+ 			$passwordConf = (Get-Content -path C:\ProgramData\ssh\sshd_config) -replace '#PasswordAuthentication yes','PasswordAuthentication no'
```
sebsoto, line 72: Whats going on here? Was this incorrect before?

pkg/controller/secrets/secrets.go
```diff
@@ -65,22 +65,22 @@ func GenerateUserData(privateKey []byte) (*core.Secret, error) {
 			$containerLogsPort = "10250"
 			New-NetFirewallRule -DisplayName $firewallRuleName -Direction Inbound -Action Allow -Protocol TCP -LocalPort $containerLogsPort -EdgeTraversalPolicy Allow
 			Install-PackageProvider -Name NuGet -MinimumVersion 2.8.5.201 -Force
```
aravindhp, line 67: We were installing this package provider so that we could install `OpenSSHUtils` and we should be able to remove this.

pkg/controller/secrets/secrets.go
aravindhp, line 72: Are we sure we want to disable password authentication in this PR? That sort of config change should be a separate PR.

aravindhp (changes requested):
Thanks for working on this, @PratikMahajan. Please address my comments.

pkg/controller/secrets/secrets.go
mansikulkarni96: why do we want to set the password authentication to no? wouldnt just leaving that line commented as it is by default work?

PratikMahajan: I agree with @aravindhp that we should handle this change in different PR. 
also just setting `PasswordAuthentication no` wont work and we need to set `ChallengeResponseAuthentication no` as well to completely disable password auth.

PratikMahajan: i'm not sure if just keeping it commented out will disable password auth

PratikMahajan: cant say that it was incorrect. But as a part of configuring ssh with pubkey, it would be preferable is we just use pubkey and disable password.

pkg/controller/secrets/secrets.go
```diff
@@ -64,23 +64,24 @@ func GenerateUserData(privateKey []byte) (*core.Secret, error) {
...
 			Start-Service sshd
 			$pubKeyConf = (Get-Content -path C:\ProgramData\ssh\sshd_config) -replace '#PubkeyAuthentication yes','PubkeyAuthentication yes'
 			$pubKeyConf | Set-Content -Path C:\ProgramData\ssh\sshd_config
  			$passwordConf = (Get-Content -path C:\ProgramData\ssh\sshd_config) -replace '#PasswordAuthentication yes','PasswordAuthentication yes'
 			$passwordConf | Set-Content -Path C:\ProgramData\ssh\sshd_config
-			$authFileConf = (Get-Content -path C:\ProgramData\ssh\sshd_config) -replace 'AuthorizedKeysFile __PROGRAMDATA__/ssh/administrators_authorized_keys','#AuthorizedKeysFile __PROGRAMDATA__/ssh/administrators_authorized_keys'
-			$authFileConf | Set-Content -Path C:\ProgramData\ssh\sshd_config
-			$pubKeyLocationConf = (Get-Content -path C:\ProgramData\ssh\sshd_config) -replace 'Match Group administrators','#Match Group administrators'
-			$pubKeyLocationConf | Set-Content -Path C:\ProgramData\ssh\sshd_config
 			Restart-Service sshd
```
aravindhp, line 73: You are now restarting sshd twice. Please remove.

aravindhp (changes requested):

pkg/controller/secrets/secrets.go
PratikMahajan: removed

aravindhp (approved):
/approve

pkg/controller/secrets/secrets.go
usrbinkat: Would argue that alignment with CoreOS would require also having passwd auth disabled by default:

```
core@voyager:~$ sudo grep PasswordAuthentication /etc/ssh/sshd_config | grep no
PasswordAuthentication no
core@voyager:~$ sudo grep VERSION /etc/os-release
VERSION="46.82.202010091720-0"
VERSION_ID="4.6"
OPENSHIFT_VERSION="4.6"
RHEL_VERSION="8.2"
REDHAT_BUGZILLA_PRODUCT_VERSION="4.6"
REDHAT_SUPPORT_PRODUCT_VERSION="4.6"
OSTREE_VERSION='46.82.202010091720-0'
```

aravindhp: Agreed. However a config change should be made in a separate PR. We prefer not making too many changes in the same PR.

PR #153, opened 2020-10-30
WINC-460: [ci] Define build_root_image within repo

sebsoto:
This commit allows us to change the build_root_image with commits to
this repo. Previously we would hit an issue with breaking builds due to
bumping the required go version within this repository and then having
to update the used go version within the openshift/release repo.

https://docs.ci.openshift.org/docs/architecture/ci-operator/#build-root-image


aravindhp (approved):
/approve

PR #154, opened 2020-10-30
[build] Remove "COPY . ."  from Dockerfiles

aravindhp:
Remove `COPY . .` so that we can take advantage of image caching.


PratikMahajan (approved):
/lgtm

PR #156, opened 2020-11-02
[wmco] Fix error logging when transfering files

sebsoto:
This commit makes it so that the log.Error() function is used properly
when transferring files over ssh


VaishnaviHire:
Thanks @sebsoto for fixing this

/lgtm

aravindhp (approved):
/approve

PR #147, opened 2020-10-28
WINC-416: Improve logging and events

aravindhp:
- [wmco] Improve logging
- [wmco] Improve event messages

PR #157, opened 2020-11-02
[wmco] Refactor Windows service start/creation

saifshaikh48:
This change was made to reduce code duplication in the creation and start of Windows services, as well as when checking if a particular file exists on the VM. This makes it easier when adding new services into Windows VM configuration.

### DIFF
--- a/pkg/controller/windowsmachine/windows/windows.go
+++ b/pkg/controller/windowsmachine/windows/windows.go
@@ -199,19 +199,8 @@ func (vm *windows) ConfigureHybridOverlay(nodeName string) error {
 		return errors.Wrapf(err, "error creating %s service object", hybridOverlayServiceName)
 	}
 
-	serviceExists, err := vm.serviceExists(hybridOverlayServiceName)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/controller/windowsmachine/windows/windows.go
```diff
@@ -146,6 +148,15 @@ func (vm *windows) CopyFile(filePath, remoteDir string) error {
 	return nil
 }
 
+func (vm *windows) FileExists(path string) (bool, error) {
+	out, err := vm.Run("Test-Path "+path, true)
+	if err != nil {
+		return false, errors.Wrapf(err, "error checking if file %s exists on Windows VM %s. output: %s",
+			path, vm.ID(), out)
```
aravindhp, line 155: No need to return `out` in the error. I will be logging that error in #147 and that should suffice.

aravindhp (changes requested):

LorbusChris:
/approve
I suggest we merge this asap so I can easily base further windows_exporter work on top.

aravindhp (approved):
/approve

PR #159, opened 2020-11-03
WINC-527: [wmco] run WMCO on control plane

PratikMahajan:
WMCO has escalated permissions and we need to run it on 
control-plane hosts. 

This reverts commit 0c33029.


aravindhp (approved):
/approve

PR #160, opened 2020-11-03
[ci] Fix WMCO version injection in Dockerfile

aravindhp:
PR #154 tried to be smart by copying just enough git metadata for version reporting. This however was resulting in a non-fatal failure, `fatal: bad object HEAD`. Fix this by being less smart and more thorough.


mansikulkarni96:
LGTM

PratikMahajan (approved):
/lgtm

PR #161, opened 2020-11-04
[ci] Remove microdnf update

mansikulkarni96:
Do not update the packages for base image, since it might lead to failures with conflicting packages. If there are updates to the image, they should be picked up with the image itself.


aravindhp (approved):
/approve

aravindhp:
/hold cancel

Removing the hold so that bot will retest automatically. If the tests fail, it won't merge anyhow.

PR #162, opened 2020-11-04
[hack] Add private key input to olm.sh

aravindhp:
Using the `-k private-key-path` input with the `run` option will result in the cloud-private-key secret being created. This improves the development workflow.

### DIFF
--- a/hack/olm.sh
+++ b/hack/olm.sh
@@ -8,6 +8,7 @@
 #    $1      Action                   run/cleanup the operator installation
 #    -i      Ignore image cache       builds the operator image without using local image build cache
 #    -c=     Operator Image           container url and tag for the operator image
+#    -k=     Private key file         path to the private key file
 
 
 # container tool to use with operator-sdk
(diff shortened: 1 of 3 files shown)


hack/olm.sh
```diff
@@ -26,10 +27,12 @@ fi
 shift # shift position of the positional parameters for getopts
 
 # Options
-while getopts ":ic:" opt; do
+PRIVATE_KEY=""
+while getopts ":ic:k:" opt; do
     case "$opt" in
 	i) noCache="--image-build-args=\"--no-cache\"";;
 	c) OPERATOR_IMAGE="$OPTARG";;
+  k) PRIVATE_KEY="$OPTARG";;
```
LorbusChris, line 35: indentation

LorbusChris (changes requested):
very nice improvement! one nit

hack/olm.sh
aravindhp: What indentation do I need to fix? :-) Sorry not able to figure that out.

sebsoto: I think it should be in line with i) and c)

aravindhp: OK, some is not following tabs vs spaces consistently. This file is a mix of mainly spaces and some tabs. I use a tab here and fix tab and spaces in a separate PR.

aravindhp: Fixed. PTAL.

PR #165, opened 2020-11-09
WINC-460: Switch to using go1.15 

sebsoto:
This PR switches to building with go1.15.
The kubernetes node binaries we are using are also being updated, as the update was blocked by being on go1.14

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -5,12 +5,12 @@ LABEL stage=build
 RUN microdnf -y install rsync make git tar findutils diffutils
 RUN mkdir /build/
 
-# Install go 1.14
+# Install go 1.15
 RUN mkdir /build/go
 WORKDIR /build/go
-RUN curl https://golang.org/dl/go1.14.7.linux-amd64.tar.gz -L -o go.tar.gz
+RUN curl https://golang.org/dl/go1.15.4.linux-amd64.tar.gz -L -o go.tar.gz
 # Check sha256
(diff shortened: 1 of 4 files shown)


build/Dockerfile
```diff
@@ -5,12 +5,12 @@ LABEL stage=build
 RUN microdnf -y install rsync make git tar findutils diffutils
 RUN mkdir /build/
 
-# Install go 1.14
+# Install go 1.15
```
aravindhp, line 8: Please talk to @mansikulkarni96 about the 1.15 image we are using for the release and see if it can be used here.

aravindhp (changes requested):

build/Dockerfile
sebsoto: I think it is usable in CI, but an issue that it introduces is it will add a gpg requirement to run the e2e from contributor's dev machines. I think testing with the same builder image we are releasing with is important and we should explore that in https://issues.redhat.com/browse/WINC-477
I added a comment to look into that, if you are fine waiting until that ticket is worked on.

aravindhp: We cannot explore using a builder image in [WINC-477](https://issues.redhat.com/browse/WINC-477). I would rather have us use the same compiler across dev, CI and release. 

> an issue that it introduces is it will add a gpg requirement to run the e2e from contributor's dev machines.

Do you mean the gpg pakage? Shouldn't that already be present?

PR #166, opened 2020-11-09
Sync bundle.Dockerfile with WMCO directory structure

mansikulkarni96:
Sync bundle.Dockerfile with WMCO current directory structure, we had updated this file to match the WMCO bundle dist-git directory structure, the dist-git layout will be updated to match with WMCO and reuse this Dockerfile.

### DIFF
--- a/bundle.Dockerfile
+++ b/bundle.Dockerfile
@@ -1,8 +1,12 @@
 FROM scratch
 
-# These directories map to the layout of the directories in the WMCO dist-git repository.
-COPY manifests /manifests/
-COPY metadata/annotations.yaml /metadata/annotations.yaml
+# This block are standard Red Hat container labels
+LABEL name="openshift4/windows-machine-config-operator-bundle" \
+    License="ASL 2.0" \
+    io.k8s.display-name="Windows Machine Config Operator bundle" \
+    io.k8s.description="Windows Machine Config Operator's OLM bundle image" \
+    summary="Windows Machine Config Operator's OLM bundle image" \
+    maintainer="Team Windows Containers <team-winc@redhat.com>"
 
 # These are three labels needed to control how the pipeline should handle this container image
 # This first label tells the pipeline that this is a bundle image and should be
@@ -34,11 +38,6 @@ LABEL version="v0.0.0"
 # This label maps to the brew build target
 LABEL com.redhat.component="windows-machine-config-operator-bundle-container"
 
-# This last block are standard Red Hat container labels
-LABEL \
-    name="openshift4/windows-machine-config-operator-bundle" \
-    License="ASL 2.0" \
-    io.k8s.display-name="Windows Machine Config Operator bundle" \
-    io.k8s.description="Windows Machine Config Operator's OLM bundle image" \
-    summary="Windows Machine Config Operator's OLM bundle image" \
-    maintainer="Team Windows Containers <team-winc@redhat.com>"
+# COPY manifests and metadata directories
+COPY deploy/olm-catalog/windows-machine-config-operator/manifests /manifests/
+COPY deploy/olm-catalog/windows-machine-config-operator/metadata /metadata/


aravindhp (approved):
/approve

VaishnaviHire:
Tried this,

When the `bundle create` command is run for the this `bundle.Dockerfile`, following three labels get added under `LABEL \` on line 37
```
LABEL \
LABEL operators.operatorframework.io.metrics.mediatype.v1=metrics+v1
LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v0.19.4
LABEL operators.operatorframework.io.metrics.project_layout=go
```
resulting in parsing error -

```
Error response from daemon: Dockerfile parse error line 42: unknown instruction: NAME="OPENSHIFT4/WINDOWS-MACHINE-CONFIG-OPERATOR-BUNDLE"
```

aravindhp:
/approve cancel

With the revert, the `bundle create` command needs to work.

LorbusChris:
Does it work if we remove the newline here https://github.com/openshift/windows-machine-config-operator/blob/master/bundle.Dockerfile#L38-L39 and make it:
```
LABEL name="openshift4/windows-machine-config-operator-bundle" \
    License="ASL 2.0" \
    io.k8s.display-name="Windows Machine Config Operator bundle" \
    io.k8s.description="Windows Machine Config Operator's OLM bundle image" \
    summary="Windows Machine Config Operator's OLM bundle image" \
    maintainer="Team Windows Containers <team-winc@redhat.com>"
```

mansikulkarni96:
There are a few changes that need to be made for operator-sdk v0.19.4,
1. LABEL field has to be specified on each of the labels, having a single header will not work,
2. character '\' is not accepted as a valid character at end of any label.
Making these changes will fix all issues, adding them to the commit.

bundle.Dockerfile
```diff
@@ -28,17 +27,17 @@ LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1 \
...
-
-# This last block are standard Red Hat container labels
+# This block are standard Red Hat container labels
 LABEL \
     name="openshift4/windows-machine-config-operator-bundle" \
     License="ASL 2.0" \
     io.k8s.display-name="Windows Machine Config Operator bundle" \
     io.k8s.description="Windows Machine Config Operator's OLM bundle image" \
     summary="Windows Machine Config Operator's OLM bundle image" \
     maintainer="Team Windows Containers <team-winc@redhat.com>"
```
LorbusChris, line 37: since these are almost something like Header labels (with Name, License etc), I'd suggest moving them all the way to the top

bundle.Dockerfile
```diff
@@ -28,17 +27,17 @@ LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1 \
...
-# This label gets replaced by render_templates, so it is safer to keep it on its own
-LABEL version="v0.0.0"
-
-# This label maps to the brew build target
-LABEL com.redhat.component="windows-machine-config-operator-bundle-container"
-
-# This last block are standard Red Hat container labels
+# This block are standard Red Hat container labels
 LABEL \
     name="openshift4/windows-machine-config-operator-bundle" \
```
aravindhp, line 32: Can we combine this? i.e.`LABEL name="openshift4/windows-machine-config-operator-bundle" \`

bundle.Dockerfile
LorbusChris: (still below the FROM directive though)

bundle.Dockerfile
mansikulkarni96: I guess we can we had taken this from a document Confluence document that specifies required labels.

mansikulkarni96: updated.

bundle.Dockerfile
mansikulkarni96: done

bundle.Dockerfile
```diff
@@ -1,8 +1,15 @@
...
-COPY manifests /manifests/
-COPY metadata/annotations.yaml /metadata/annotations.yaml
+# This block are standard Red Hat container labels
+LABEL name="openshift4/windows-machine-config-operator-bundle" \
+    License="ASL 2.0" \
+    io.k8s.display-name="Windows Machine Config Operator bundle" \
+    io.k8s.description="Windows Machine Config Operator's OLM bundle image" \
+    summary="Windows Machine Config Operator's OLM bundle image" \
+    maintainer="Team Windows Containers <team-winc@redhat.com>"
+
```
aravindhp, line 10: Please move the labels to one spot. I would rather not have the labels spread around.

aravindhp (changes requested):

bundle.Dockerfile
aravindhp: Yes, I had copied this from the confluence doc and I did not want to change things around at that point. Now that we are more confident around the build no reason to blindly follow their formatting unless it breaks something.

bundle.Dockerfile
aravindhp: I am fine with that as long as all the labels are in one spot. i.e
```
FROM...

LABEL..

LABEL...

COPY...

#etc
```

aravindhp (approved):
/approve

PR #167, opened 2020-11-10
WINC-513: Ensure all transfered files have the correct hash

sebsoto:
This commit makes it so that we check the sha256 checksum of files that
were previously transfered to Windows nodes. If the checksum does not
match what we expect it to, the correct file will be transfered over.

### DIFF
--- a/pkg/controller/windowsmachine/windows/windows.go
+++ b/pkg/controller/windowsmachine/windows/windows.go
@@ -6,11 +6,13 @@ import (
 	"strings"
 	"time"
 
-	"github.com/openshift/windows-machine-config-operator/pkg/controller/retry"
-	wkl "github.com/openshift/windows-machine-config-operator/pkg/controller/wellknownlocations"
 	"github.com/pkg/errors"
 	"golang.org/x/crypto/ssh"
+	"k8s.io/apimachinery/pkg/util/wait"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+	"github.com/openshift/windows-machine-config-operator/pkg/controller/payload"
+	"github.com/openshift/windows-machine-config-operator/pkg/controller/retry"
 )
 
 const (
@@ -40,6 +42,7 @@ const (
 	kubeProxyPath = k8sDir + "kube-proxy.exe"
 	// hybridOverlayPath is the location of the hybrid-overlay-node exe
 	hybridOverlayPath = k8sDir + "hybrid-overlay-node.exe"
+
 	// hybridOverlayServiceName is the name of the hybrid-overlay-node Windows service
 	hybridOverlayServiceName = "hybrid-overlay-node"
 	// hybridOverlayConfigurationTime is the approximate time taken for the hybrid-overlay to complete reconfiguring
@@ -66,14 +69,48 @@ const (
 
 var log = logf.Log.WithName("windows")
 
+// filesToTransfer is a map of what files should be copied to the Windows VM and where they should be copied to
+var filesToTransfer map[*payload.FileInfo]string
+
+// getFilesToTransfer returns the properly populated filesToTransfer map
+func getFilesToTransfer() (map[*payload.FileInfo]string, error) {
+	if filesToTransfer != nil {
+		return filesToTransfer, nil
+	}
+	srcDestPairs := map[string]string{
+		payload.IgnoreWgetPowerShellPath: remoteDir,
+		payload.WmcbPath:                 k8sDir,
+		payload.HybridOverlayPath:        k8sDir,
+		payload.HNSPSModule:              remoteDir,
+		payload.WindowsExporterPath:      k8sDir,
+		payload.FlannelCNIPluginPath:     cniDir,
+		payload.WinBridgeCNIPlugin:       cniDir,
+		payload.HostLocalCNIPlugin:       cniDir,
+		payload.WinOverlayCNIPlugin:      cniDir,
+		payload.KubeProxyPath:            k8sDir,
+		payload.KubeletPath:              k8sDir,
+	}
+	files := make(map[*payload.FileInfo]string)
+	for src, dest := range srcDestPairs {
+		f, err := payload.NewFileInfo(src)
+		if err != nil {
+			return nil, errors.Wrapf(err, "could not create FileInfo object for file %s", src)
+		}
+		files[f] = dest
+	}
+	filesToTransfer = files
+	return filesToTransfer, nil
+}
+
 // Windows contains all the  methods needed to configure a Windows VM to become a worker node
 type Windows interface {
 	// ID returns the cloud provider ID of the VM
 	ID() string
-	// CopyFile copies the given file to the remote directory in the Windows VM. The remote directory is created if it
-	// does not exist
-	CopyFile(string, string) error
-	// FileExists ensures a specific file exists at the given path on the Windows VM
+	// EnsureFile ensures the given file exists within the specified directory on the Windows VM. The file will be copied
+	// to the Windows VM if it is not present or if it has the incorrect contents. The remote directory is created if it
+	// does not exist.
+	EnsureFile(*payload.FileInfo, string) error
+	// FileExists returns true if a specific file exists at the given path on the Windows VM
 	FileExists(string) (bool, error)
 	// Run executes the given command remotely on the Windows VM over a ssh connection and returns the combined output
 	// of stdout and stderr. If the bool is set, it implies that the cmd is to be execute in PowerShell. This function
@@ -150,10 +187,28 @@ func (vm *windows) ID() string {
 	return vm.id
 }
 
-func (vm *windows) CopyFile(filePath, remoteDir string) error {
-	log.V(1).Info("copy", "local file", filePath, "remote dir", remoteDir)
-	if err := vm.interact.transfer(filePath, remoteDir); err != nil {
-		return errors.Wrapf(err, "unable to transfer %s to remote dir %s", filePath, remoteDir)
+func (vm *windows) EnsureFile(file *payload.FileInfo, remoteDir string) error {
+	// Only copy the file to the Windows VM if it does not already exist wth the desired content
+	remotePath := remoteDir + "\\" + filepath.Base(file.Path)
+	fileExists, err := vm.FileExists(remotePath)
+	if err != nil {
+		return errors.Wrapf(err, "error checking if file '%s' exists on the Windows VM", remotePath)
+	}
+	if fileExists {
+		remoteFile, err := vm.newFileInfo(remotePath)
+		if err != nil {
+			return errors.Wrapf(err, "error getting info on file '%s' on the Windows VM", remotePath)
+		}
+		if file.SHA256 == remoteFile.SHA256 {
+			// The file already exists with the expected content, do nothing
+			log.V(1).Info("file already exists on VM with expected content", "file", file.Path)
+			return nil
+		}
+	}
+
+	log.V(1).Info("copy", "local file", file.Path, "remote dir", remoteDir)
+	if err := vm.interact.transfer(file.Path, remoteDir); err != nil {
+		return errors.Wrapf(err, "unable to transfer %s to remote dir %s", file.Path, remoteDir)
 	}
 	return nil
 }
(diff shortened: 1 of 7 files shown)


pkg/controller/windowsmachine/windows/windows.go
```diff
@@ -64,6 +69,53 @@ const (
...
+		f, err := NewFileInfo(src)
+		if err != nil {
+			log.Error(err, "could not create FileInfo object", "file", src)
+			os.Exit(1)
+		}
+		filesToTransfer[f] = dest
+	}
+
+}
+
```
aravindhp: I suggest moving this into a separate package. Maybe combine it with `wellknownlocations.go`  and change the package name to something like `files`.

pkg/controller/windowsmachine/windows/windows.go
```diff
@@ -64,6 +68,63 @@ const (
...
+	}
+	filesToTransfer = make(map[*FileInfo]string)
+	srcDestPairs := map[string]string{
+		wkl.IgnoreWgetPowerShellPath: remoteDir,
+		wkl.WmcbPath:                 k8sDir,
+		wkl.HybridOverlayPath:        k8sDir,
+		wkl.HNSPSModule:              remoteDir,
+		wkl.FlannelCNIPluginPath:     cniDir,
+		wkl.WinBridgeCNIPlugin:       cniDir,
+		wkl.WindowsExporterPath:      k8sDir,
```
sebsoto, line 111: I need to change the windows exporter be a special case that is not required

VaishnaviHire (commented):
Thanks @sebsoto for working on this. PTAL at my comment

pkg/controller/windowsmachine/windows/windows.go
```diff
@@ -424,15 +499,62 @@ func (vm *windows) createService(svc *service) error {
...
 	}
+
+	// Wait until the service stop is completed
+	err = wait.Poll(retry.Interval, retry.Timeout, func() (bool, error) {
+		serviceRunning, err := vm.isRunning(svc.name)
+		if err != nil {
+			log.V(1).Error(err, "unable to check if Windows service is running", "service", svc.name)
+			return false, nil
+		}
+		if !serviceRunning {
```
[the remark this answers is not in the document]
sebsoto: Almost, but we can return `!serviceRunning` so ill do that

pkg/controller/windowsmachine/windows/windows.go
sebsoto: After looking again, `WindowsExporterPath` is listed as a required file and so we should require it to be transferred properly

pkg/controller/windowsmachine/windows/windows.go
```diff
@@ -516,6 +577,20 @@ func (vm *windows) getSourceVIP() (string, error) {
...
 
+// newRemoteFileInfo returns a pointer to a FileInfo object created from the specified file on the Windows VM
+func (vm *windows) newRemoteFileInfo(path string) (*files.FileInfo, error) {
+	// Get-FileHash returns an object with multiple properties, we are interested in the `Hash` property
+	command := "$out = Get-FileHash " + path + " -Algorithm SHA256; $out.Hash"
+	out, err := vm.Run(command, true)
+	if err != nil {
+		return nil, errors.Wrap(err, "error getting file hash")
+	}
+	// The returned hash will have newline chars along with being in all caps, doing ToLower() to
```
aravindhp, line 588: I suggest: `The returned hash will be in caps with newline characters...`

test/e2e/reconfigure_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	nc "github.com/openshift/windows-machine-config-operator/pkg/controller/windowsmachine/nodeconfig"
+)
+
+// reconfigurationTest tests that the correct behavior occurs when a previously configured Machine is configured
+// again. In practice this should not happen, however it simulates a scenario where a Machine was almost completely
+// configured, an error occurred, and the Machine is requeued. This is a scenario that should be expected to be ran into
+// often enough.
```
aravindhp, line 20: `This is a scenario that should be expected to be ran into often enough.`

Is this true? I would rather say `This is a scenario that is expected to happen often enough to warrant this test..` But what confuses me is the previous sentence state this should not happen in practice.

test/e2e/reconfigure_test.go
```diff
@@ -0,0 +1,40 @@
...
+func reconfigurationTest(t *testing.T) {
+	testCtx, err := NewTestContext(t)
+	require.NoError(t, err)
+
+	nodes, err := testCtx.kubeclient.CoreV1().Nodes().List(context.TODO(),
+		metav1.ListOptions{LabelSelector: nc.WindowsOSLabel})
+	require.NoError(t, err)
+
+	// For each Windows node remove the version annotation
+	for _, node := range nodes.Items {
```
aravindhp, line 30: Do we need to do this for all nodes or will one suffice?

pkg/controller/windowsmachine/windows/windows.go
```diff
@@ -399,15 +416,59 @@ func (vm *windows) createService(svc *service) error {
...
 	if svc == nil {
 		return errors.New("service object should not be nil")
 	}
+	// This running successfully means that the stop has initiated, not necessarily completed
 	out, err := vm.Run("sc.exe stop "+svc.name, false)
 	if err != nil {
 		return errors.Wrapf(err, "failed to stop %s service with output: %s", svc.name, out)
 	}
+
+	// Wait until the service stop is completed
```
aravindhp, line 459: `...service has stopped.`

pkg/controller/windowsmachine/windows/windows.go
```diff
@@ -516,6 +577,20 @@ func (vm *windows) getSourceVIP() (string, error) {
 	return sourceVIP, nil
 }
 
+// newRemoteFileInfo returns a pointer to a FileInfo object created from the specified file on the Windows VM
+func (vm *windows) newRemoteFileInfo(path string) (*files.FileInfo, error) {
```
aravindhp, line 581: `vm` implies remote. I suggest naming this `newFileInfo()`. Though I am wondering if there is a Go type embedding friendly way of doing this that is not immediately obvious to me. Maybe @LorbusChris knows.

aravindhp (changes requested):
@sebsoto Thanks working on this. Please address my comments.

```
This commit makes it so that we check the sha256 checksum of files that
were previously transfered to Windows nodes. If the checksum does not
match what we expect it to, the correct file will be transfered over.
```
The above commit message implies you only perform the hash check on files that previously not transferred. Please sync this with what is actually being done.

pkg/controller/files/files.go
```diff
@@ -75,3 +83,54 @@ const (
...
+		HostLocalCNIPlugin:       RemoteCNIDir,
+		WinOverlayCNIPlugin:      RemoteCNIDir,
+		KubeProxyPath:            K8sDir,
+		KubeletPath:              K8sDir,
+	}
+	files := make(map[*FileInfo]string)
+	for src, dest := range srcDestPairs {
+		f, err := NewFileInfo(src)
+		if err != nil {
+			return nil, errors.Wrapf(err, "could not create FileInfo object for file %s", src)
```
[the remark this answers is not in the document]
sebsoto: I dont understand. I used an intermediary variable to avoid having to free the global. The global won't be set until we are sure the data is valid.

pkg/controller/windowsmachine/windows/windows.go
```diff
@@ -46,7 +48,7 @@ type Windows interface {
 	ID() string
 	// CopyFile copies the given file to the remote directory in the Windows VM. The remote directory is created if it
 	// does not exist
-	CopyFile(string, string) error
+	CopyFile(*files.FileInfo, string) error
```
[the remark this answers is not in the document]
sebsoto: I feel as if the `CopyFile` name and description makes sense as this is an `Windows` interface signature. Skipping the actual copying if its unnecessary seems more like an implementation detail to me within the `windows` method. `CopyFile` makes more sense to me, but I can change it if we want to explicit about what is happening.

test/e2e/reconfigure_test.go
sebsoto: I reworded it to be more readable

pkg/controller/files/files.go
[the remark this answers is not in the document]
aravindhp: You are right. I was overthinking this.

pkg/controller/windowsmachine/windows/windows.go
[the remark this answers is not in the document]
aravindhp: I have to disagree here. `CopyFile` as it name states should always copy the file over whether if a file exists or not or if there is a checksum mismatch or not. You are actually changing the semantics of how `CopyFile` works with this change. Ideally `EnsureFile` should be added to the interface which should in turn call `CopyFile` only when needed. A change like this would be totally unacceptable if this was an exported library with other external callers.

pkg/controller/windowsmachine/windows/windows.go
```diff
@@ -40,6 +43,42 @@ const (
...
+		if err != nil {
+			return nil, errors.Wrapf(err, "could not create FileInfo object for file %s", src)
+		}
+		files[f] = dest
+	}
+	filesToTransfer = files
+	return filesToTransfer, nil
+}
+
+const (
```
aravindhp, line 81: Why are you creating two sets of `const`? Please keep them as one and move the code you are introducing after `var log = logf.Log.WithName("windows")`

aravindhp (changes requested):

aravindhp (approved):
/approve

pkg/controller/windowsmachine/windows/windows.go
```diff
@@ -40,6 +42,7 @@ const (
 	kubeProxyPath = k8sDir + "kube-proxy.exe"
 	// hybridOverlayPath is the location of the hybrid-overlay-node exe
 	hybridOverlayPath = k8sDir + "hybrid-overlay-node.exe"
+
```
VaishnaviHire, line 45: extra line

sebsoto: I put the extra line there to seperate paths and variables

aravindhp: Please add comment then otherwise folks are going to mix this up in the future.

pkg/controller/windowsmachine/windows/windows.go
```diff
@@ -424,15 +478,59 @@ func (vm *windows) createService(svc *service) error {
...
 	out, err := vm.Run("sc.exe stop "+svc.name, false)
 	if err != nil {
 		return errors.Wrapf(err, "failed to stop %s service with output: %s", svc.name, out)
 	}
+
+	// Wait until the service has stopped
+	err = wait.Poll(retry.Interval, retry.Timeout, func() (bool, error) {
+		serviceRunning, err := vm.isRunning(svc.name)
+		if err != nil {
+			log.V(1).Error(err, "unable to check if Windows service is running", "service", svc.name)
```
mansikulkarni96, line 525: error log is of the highest importance and I am not sure why we would want to use V(1) for it, if we want to have debug log we can use log.V(1).Info

mansikulkarni96 (commented):
Thanks for working on this @sebsoto , PTAL at my comments.

pkg/controller/windowsmachine/windows/windows.go
```diff
@@ -66,14 +69,48 @@ const (
...
 	ID() string
-	// CopyFile copies the given file to the remote directory in the Windows VM. The remote directory is created if it
-	// does not exist
-	CopyFile(string, string) error
-	// FileExists ensures a specific file exists at the given path on the Windows VM
+	// EnsureFile ensures the given file exists within the specified directory on the Windows VM. The file will be copied
+	// to the Windows VM if it is not present or if it has the incorrect contents. The remote directory is created if it
+	// does not exist.
+	EnsureFile(*payload.FileInfo, string) error
+	// FileExists returns true if a specific file exists at the given path on the Windows VM
```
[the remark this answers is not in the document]
sebsoto: I wanted to make this change because the comment for this function made it seem like it was doing a very similar thing to the new `EnsureFile` method

pkg/controller/windowsmachine/windows/windows.go
aravindhp: @mansikulkarni96 are you saying this `log.V(1).Error()` will show up on a release build? I think @sebsoto did not think this needs to be printed in release builds.

PR #172, opened 2020-11-12
[build] Add `microdnf clean all` to Dockerfiles

LorbusChris:
This avoids problems when files that cannot be archived with tar, like sockets, are present in the yum cache (container images are tar archives).

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -2,7 +2,7 @@
 # to install it.
 FROM registry.access.redhat.com/ubi8/ubi-minimal as build
 LABEL stage=build
-RUN microdnf -y install rsync make git tar findutils diffutils
+RUN microdnf -y install rsync make git tar findutils diffutils && microdnf clean all
 RUN mkdir /build/
 
 # Install go 1.14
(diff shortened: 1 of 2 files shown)


build/Dockerfile
```diff
@@ -2,7 +2,7 @@
 # to install it.
 FROM registry.access.redhat.com/ubi8/ubi-minimal as build
 LABEL stage=build
-RUN microdnf -y install rsync make git tar findutils diffutils
+RUN microdnf -y install rsync make git tar findutils diffutils && microdnf clean all
```
aravindhp, line 5: Can we make this a separate RUN command?

aravindhp (changes requested):
Please mention why we are making this change in the commit and PR message.

build/Dockerfile
LorbusChris: I don't think that's desirable as the yum cache would then still land in the image build cache (as it's cached after each directive, i.e. each `RUN`): https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run

LorbusChris: Doing this within the same RUN directive is also the common practice within OpenShift, e.g. here https://github.com/openshift/machine-config-operator/blob/master/Dockerfile.rhel7#L13

aravindhp (approved):
/approve

PR #183, opened 2020-11-17
WINC-519: skip operator deployment in e2e tests

mansikulkarni96:
This PR checks if the OO_INDEX ENV var is set and the operator deployment is already rolled out to skip deployment in WMCO e2e tests.
 $OO_INDEX points to the WMCO bundle index image created through https://docs.ci.openshift.org/docs/how-tos/testing-operator-sdk-operators/ workflow in CI. 
This is a step towards moving to the _Testing Operators Built With The Operator SDK and Deployed Through OLM_ workflow.

### DIFF
--- a/hack/run-ci-e2e-test.sh
+++ b/hack/run-ci-e2e-test.sh
@@ -86,12 +86,26 @@ fi
 # https://steps.svc.ci.openshift.org/help/ci-operator#release
 OPERATOR_IMAGE=${OPERATOR_IMAGE:-${IMAGE_FORMAT//\/stable:\$\{component\}//stable:windows-machine-config-operator-test}}
 
-# Setup and run the operator
-if ! run_WMCO $OSDK; then
-  # Try to get the WMCO logs if possible
-  get_WMCO_logs
-  cleanup_WMCO $OSDK
-  exit 1
+# OO_INDEX points to the WMCO bundle index image created through https://docs.ci.openshift.org/docs/how-tos/testing-operator-sdk-operators/ workflow in CI.
+# Setup and run the operator only if $OO_INDEX is not set in CI, and the operator deployment is not available.
+OO_INDEX=${OO_INDEX:-}
+
+# SKIP_OPERATOR_RUN_CLEANUP will skip operator deployment and cleanup if set to true
+SKIP_OPERATOR_RUN_CLEANUP=${SKIP_OPERATOR_RUN_CLEANUP:-"false"}
+
+# TODO: Remove the deployment status check once optional operator testing workflow is implemented
+if [ ! -z "$OO_INDEX" ] && oc rollout status deployment windows-machine-config-operator -n openshift-windows-machine-config-operator; then
+   SKIP_OPERATOR_RUN_CLEANUP="true"
+fi
+
+if ! $SKIP_OPERATOR_RUN_CLEANUP; then
+  # Setup and run the operator
+  if ! run_WMCO $OSDK; then
+    # Try to get the WMCO logs if possible
+    get_WMCO_logs
+    cleanup_WMCO $OSDK
+    exit 1
+  fi
 fi
 
 # The bool flags in golang does not respect key value pattern. They follow -flag=x pattern.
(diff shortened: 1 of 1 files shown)


hack/run-ci-e2e-test.sh
```diff
@@ -86,12 +86,17 @@ fi
...
-# Setup and run the operator
-if ! run_WMCO $OSDK; then
-  # Try to get the WMCO logs if possible
-  get_WMCO_logs
-  cleanup_WMCO $OSDK
-  exit 1
+# OO_INDEX points to the WMCO bundle index image created through https://docs.ci.openshift.org/docs/how-tos/testing-operator-sdk-operators/ workflow in CI.
+# Setup and run the operator only if $OO_INDEX is not set in CI, and the operator deployment is not available.
+OO_INDEX=${OO_INDEX:-}
+if [ -z "$OO_INDEX" ] && ! oc rollout status deployment windows-machine-config-operator -n openshift-windows-machine-config-operator; then
```
sebsoto, line 92: Need to skip cleanup as well https://github.com/openshift/windows-machine-config-operator/pull/183/files#diff-9c914ef97321839b726df636cf533083ac1893c90dbe13db77382e043d1f1ba4R128

mansikulkarni96: the new workflow does not mention anything about handling cleanup, are you suggesting this because our cleanup is using packagemanifest command as well?

sebsoto: Yes, but in general if we arent deploying the operator we shouldnt be removing it.

sebsoto: Unless we have a specific need to do so

hack/run-ci-e2e-test.sh
```diff
@@ -86,12 +86,25 @@ fi
...
+# SKIP_OPERATOR_RUN_CLEANUP will skip operator deployment and cleanup if set to true
+SKIP_OPERATOR_RUN_CLEANUP=${SKIP_OPERATOR_RUN_CLEANUP:-"false"}
+
+if [ -z "$OO_INDEX" ] && ! oc rollout status deployment windows-machine-config-operator -n openshift-windows-machine-config-operator; then
+ SKIP_OPERATOR_RUN_CLEANUP="true"
+fi
+
+if ! $SKIP_OPERATOR_RUN_CLEANUP; then
+  # Setup and run the operator
+  if ! run_WMCO $OSDK; then
```
LorbusChris, line 102: nit: This could be put on the same line, as one `if`
I'm still ok with it, though.

LorbusChris (commented):
/lgtm

VaishnaviHire:
LGTM

hack/run-ci-e2e-test.sh
```diff
@@ -86,12 +86,25 @@ fi
...
-  # Try to get the WMCO logs if possible
-  get_WMCO_logs
-  cleanup_WMCO $OSDK
-  exit 1
+# OO_INDEX points to the WMCO bundle index image created through https://docs.ci.openshift.org/docs/how-tos/testing-operator-sdk-operators/ workflow in CI.
+# Setup and run the operator only if $OO_INDEX is not set in CI, and the operator deployment is not available.
+OO_INDEX=${OO_INDEX:-}
+
+# SKIP_OPERATOR_RUN_CLEANUP will skip operator deployment and cleanup if set to true
+SKIP_OPERATOR_RUN_CLEANUP=${SKIP_OPERATOR_RUN_CLEANUP:-"false"}
```
aravindhp, line 94: Do we need another variable to skip the cleanup? Can't you reuse `OO_INDEX`?

hack/run-ci-e2e-test.sh
```diff
@@ -86,12 +86,25 @@ fi
...
+
+if [ -z "$OO_INDEX" ] && ! oc rollout status deployment windows-machine-config-operator -n openshift-windows-machine-config-operator; then
```
aravindhp, line 96: What's the reason for checking `oc rollout status deployment windows-machine-config-operator -n openshift-windows-machine-config-operator`?

aravindhp (commented):
Thanks for working on this, @mansikulkarni96. I have a couple of questions.

hack/run-ci-e2e-test.sh
mansikulkarni96: yes we can just added that so we don't have to check a long if statement in two places.

hack/run-ci-e2e-test.sh
mansikulkarni96: to make sure we only skip deployment when the operator is actually running, added this as an additional guard to skip our manual deployment

aravindhp: This is so that we can keep calling `olm.sh run` even when the operator is running?

mansikulkarni96: not really, olm.sh calls run_wmco() independently, but my main reason was to keep this guard until we have confidence in this process, we can remove it once everything is in good shape.

LorbusChris: please add a TODO comment for this ^

aravindhp (approved):
/approve

PR #186, opened 2020-11-20
WINC-547: [wmco] Don't log when a service is not found

aravindhp:
Use the command "sc.exe qe" and the error code "1060" as hints to not log the error. We are doing this as it was spamming the log with no informational benefit.

### DIFF
--- a/pkg/controller/windowsmachine/windows/windows.go
+++ b/pkg/controller/windowsmachine/windows/windows.go
@@ -65,6 +65,12 @@ const (
 	// remotePowerShellCmdPrefix holds the PowerShell prefix that needs to be prefixed  for every remote PowerShell
 	// command executed on the remote Windows VM
 	remotePowerShellCmdPrefix = "powershell.exe -NonInteractive -ExecutionPolicy Bypass "
+	// serviceQueryCmd is the Windows command used to query a service
+	serviceQueryCmd = "sc.exe qc "
+	// serviceNotFound is part of the error message returned when a service does not exist. 1060 is an error code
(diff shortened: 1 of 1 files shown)


VaishnaviHire:
LGTM

pkg/controller/windowsmachine/windows/windows.go
```diff
@@ -228,7 +234,10 @@ func (vm *windows) Run(cmd string, psCmd bool) (string, error) {
 
 	out, err := vm.interact.run(cmd)
 	if err != nil {
-		log.Error(err, "error running", "cmd", cmd, "out", out)
+		// Hack to not print the error log for "sc.exe qc" returning 1060 for non existent services.
+		if !(strings.HasPrefix(cmd, serviceQueryCmd) && strings.HasSuffix(err.Error(), serviceNotFound)) {
+			log.Error(err, "error running", "cmd", cmd, "out", out)
+		}
```
sebsoto, line 240: I'm worried about putting special cases like this in Run(), in the future we might have more cases like that. Did you consider not logging within `Run()` but logging at the callsite if the error is actually an error?

aravindhp: Yes I did. This means we have to keep ensuring folks log at the call site for new code. I prefer this for now and we can revisit if we keep adding exceptions.

mansikulkarni96:
LGTM, realized we are now looking at the memory issues with tests in our CI as well.

pkg/controller/windowsmachine/windows/windows.go
sebsoto: Sounds good to me

PR #191, opened 2020-11-24
WINC-519: [community-4.6] skip operator deployment in e2e tests

LorbusChris:
This commit checks if the OO_INDEX env
var is set and the opeartor deployment
is rolled out to skip deployment in
WMCO e2e tests.

Backport of #183


aravindhp:
Given we cannot follow the CI model anytime soon we should not backport this change but rather revert #183.

PR #192, opened 2020-11-24
WINC-543: [hack] Revert oo_index change and add retry logic to 'operator-sdk run packagemanifests'

LorbusChris:


LorbusChris:
@aravindhp I believe this is good to go, please review and lgtm.
I'll open PRs to backport https://github.com/openshift/windows-machine-config-operator/pull/192/commits/6223d9ebb25d1ce68e45755991fcd1c5267ee41d to community-4.6 and release-4.6 once this merges.

aravindhp (approved):
/approve
/lgtm

hack/common.sh
```diff
@@ -35,11 +35,19 @@ OSDK_WMCO_management() {
...
-  # https://github.com/operator-framework/operator-sdk/issues/2938
-  if ! $OSDK_PATH $COMMAND packagemanifests --olm-namespace openshift-operator-lifecycle-manager --operator-namespace openshift-windows-machine-config-operator \
-  --operator-version 0.0.0 ; then
-    echo operator-sdk $1 failed
+  local retries=5
+  for ((i=0; i<retries; i++)); do
+    $OSDK_PATH $COMMAND packagemanifests \
+      --olm-namespace openshift-operator-lifecycle-manager \
+      --operator-namespace openshift-windows-machine-config-operator \
+      --operator-version 0.0.0 \
```
aravindhp, line 43: I don't think you can rerun `run packagemanifests` because I suspect it's not really failing in the first try. So I think it would be safer to do a clean between runs. Sorry I should have caught it on first review.

LorbusChris: That makes sense. I've opened #195 for this, and cherry-picked that to #193 and #194.

PR #193, opened 2020-11-25
WINC-453: [release-4.6] [hack] Extend OSDK run timeout to 5min

LorbusChris:
The tests have been failing a lot because
`operator-sdk run packagemanifests` has a default timeout of 2min which
was often missed. This commit extends the timeout to 5min to ensure the
command can return properly even when running on slower infrastructure.

With version 0.19.4 of the operator-sdk, the `run` command should now
return correct values, hence the condition and comment to explicitly
check for that have been removed.

Backport of #195

### DIFF
--- a/hack/common.sh
+++ b/hack/common.sh
@@ -35,12 +35,11 @@ OSDK_WMCO_management() {
   local COMMAND=$1
   local OSDK_PATH=$2
 
-  # Currently this fails even on successes, adding this check to ignore the failure
-  # https://github.com/operator-framework/operator-sdk/issues/2938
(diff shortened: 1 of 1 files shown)


aravindhp (approved):
/lgtm

aravindhp:
Please see https://github.com/openshift/windows-machine-config-operator/pull/192#pullrequestreview-538102599

LorbusChris:
couldn't pull images from quay due to an outage - should be resolved now
/retest

LorbusChris:
got OOMKilled
/retest

aravindhp (approved):
/lgtm

PR #196, opened 2020-11-30
Build promu binary from vendor, remove submodule

LorbusChris:
- Adds promu as a build dependency by adding it to `tools.go`.
- Builds promu binary from vendor
- Remove promu submodule


build/build.sh
```diff
@@ -29,3 +29,7 @@ goflags=${GOFLAGS:-}
 
 
 CGO_ENABLED=0 GO111MODULE=on GOOS=linux go build ${GOFLAGS} -ldflags="-X 'github.com/openshift/windows-machine-config-operator/version.Version=${VERSION}'" -o ${BIN_DIR}/${BIN_NAME} ${MAIN_PACKAGE}
+
+# Build promu utility tool from vendor dir. Promu is used for building prometheus projects like windows_exporter
+echo "building promu..."
+CGO_ENABLED=0 GO111MODULE=on GOOS=linux go build ${GOFLAGS} -o ${BIN_DIR}/promu "github.com/prometheus/promu"
```
aravindhp, line 35: Shouldn't we be using `github.com/openshift/prometheus-promu`?

LorbusChris: That'd have to go into go.mod as a replace directive. Is there any reason for creating the fork other than needing it for promu as a submodule? If not, I suggest we drop maintenance of the fork entirely. If there are however, I'll change this. Please advise.

aravindhp: That's a good point @LorbusChris. We should no longer need the fork.

go.mod
```diff
@@ -38,14 +38,15 @@ replace (
 )
 
 require (
-	github.com/aws/aws-sdk-go v1.25.48
+	github.com/aws/aws-sdk-go v1.27.0
```
aravindhp, line 41: Is this change needed?

build/build.sh
aravindhp: I am not 100% convinced  on doing the build here. Wouldn't it be better to do the build as part of the container build? Because I could make the same argument about building everything else (kubelet, kube-proxy etc) here also.

go.mod
LorbusChris: `go mod tidy` did that..things like this will happen unless the version is pinned with a replace directive. In general though, Go modules are pretty good though at not introducing breakages with these unwarranted updates.

build/build.sh
LorbusChris: Main reason for adding it here was that we only copy the repo (and with it the vendor dir) over to the container image in the CI Dockerfile.
I think the best fix would indeed be to _always_ build everything within a container though, even in the non-CI Dockerfile (currently that copies the already built binaries into the container which isn't ideal, in case one forgets to run `make build` before...)

aravindhp: >I think the best fix would indeed be to always build everything within a container though, even in the non-CI Dockerfile

We are going to do that when we move to SDK 1.0. So I suggest we move to building promu in the Dockerfile also.

PR #202, opened 2020-12-03
WINC-485: [wmco] Delete Machines when SSH auth fails

sebsoto:
This commit prevents an issue where a user changes the private key used
to configure Windows Machines, after unconfigured Machines have been
created with the original userdata secret. When those unconfigured
Machines are reconciled they will never be able to be SSH'd into and
will infinitely be requeued.

This issue is being fixed by deleting Machines that have had an
authentication error so that they can be re-provisioned with userdata
corresponding with the new private key.

### DIFF
--- a/pkg/controller/windowsmachine/windows/connectivity.go
+++ b/pkg/controller/windowsmachine/windows/connectivity.go
@@ -5,15 +5,33 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
 	"github.com/pkg/sftp"
 	"golang.org/x/crypto/ssh"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	"github.com/openshift/windows-machine-config-operator/pkg/controller/retry"
 )
 
 // sshPort is the default SSH port
 const sshPort = "22"
+
+// AuthErr occurs when our authentication into the VM is rejected
+type AuthErr struct {
+	err string
+}
+
+func (e *AuthErr) Error() string {
+	return fmt.Sprintf("SSH authentication failed: %s", e.err)
+}
+
+// newAuthErr returns a new AuthErr
+func newAuthErr(err error) *AuthErr {
+	return &AuthErr{err: err.Error()}
+}
 
 type connectivity interface {
 	// run executes the given command on the remote system
@@ -64,16 +82,19 @@ func (c *sshConnectivity) init() error {
 	}
 	var err error
 	var sshClient *ssh.Client
-	// Retry if we are unable to create a client as the VM could still be executing the steps in its user data. We
-	// cannot reuse the entries in the retry package as they are too granular.
-	for retries := 0; retries < 5; retries++ {
+	// Retry if we are unable to create a client as the VM could still be executing the steps in its user data
+	err = wait.PollImmediate(time.Minute, retry.Timeout, func() (bool, error) {
 		sshClient, err = ssh.Dial("tcp", c.ipAddress+":"+sshPort, config)
 		if err == nil {
-			break
+			return true, nil
 		}
 		log.V(1).Info("SSH dial", "IP Address", c.ipAddress, "error", err)
-		time.Sleep(1 * time.Minute)
-	}
+		if strings.Contains(err.Error(), "unable to authenticate") {
+			// Authentication failure is a special case that must be handled differently
+			return false, newAuthErr(err)
+		}
+		return false, nil
+	})
 	if err != nil {
 		return errors.Wrapf(err, "unable to connect to Windows VM %s", c.ipAddress)
 	}
--- a/pkg/controller/windowsmachine/windowsmachine_controller.go
+++ b/pkg/controller/windowsmachine/windowsmachine_controller.go
@@ -32,6 +32,7 @@ import (
 	"github.com/openshift/windows-machine-config-operator/pkg/controller/signer"
 	"github.com/openshift/windows-machine-config-operator/pkg/controller/windowsmachine/metrics"
 	"github.com/openshift/windows-machine-config-operator/pkg/controller/windowsmachine/nodeconfig"
+	"github.com/openshift/windows-machine-config-operator/pkg/controller/windowsmachine/windows"
 	"github.com/openshift/windows-machine-config-operator/version"
 )
 
@@ -309,20 +310,7 @@ func (r *ReconcileWindowsMachine) Reconcile(request reconcile.Request) (reconcil
 						machine.Name, maxUnhealthyCount)
 					return reconcile.Result{Requeue: true}, nil
 				}
-				if !machine.GetDeletionTimestamp().IsZero() {
-					// Delete already initiated
-					return reconcile.Result{}, nil
-				}
-
-				if err := r.client.Delete(context.TODO(), machine); err != nil {
-					r.recorder.Eventf(machine, core.EventTypeWarning, "MachineDeletionFailed",
-						"Machine %v deletion failed: %v", machine.Name, err)
-					return reconcile.Result{}, err
-				}
-				log.Info("machine has been remediated by deletion", "name", machine.GetName())
-				r.recorder.Eventf(machine, core.EventTypeNormal, "MachineDeleted",
-					"Machine %v has been remediated by deleting the Machine object", machine.Name)
-				return reconcile.Result{}, nil
+				return reconcile.Result{}, r.deleteMachine(machine)
 			}
 			log.Info("machine has current version", "name", machine.GetName(),
 				"version", node.Annotations[nodeconfig.VersionAnnotation])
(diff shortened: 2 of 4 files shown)


pkg/controller/windowsmachine/windows/connectivity.go
```diff
@@ -71,6 +86,10 @@ func (c *sshConnectivity) init() error {
 		if err == nil {
 			break
 		}
+		if strings.Contains(err.Error(), "unable to authenticate") {
+			// Authentication failure is a special case that must be handled differently
+			return newAuthErr(err)
+		}
 		log.V(1).Info("SSH dial", "IP Address", c.ipAddress, "error", err)
```
VaishnaviHire, line 93: This log message should be before the condition

pkg/controller/windowsmachine/windows/connectivity.go
```diff
@@ -71,6 +86,10 @@ func (c *sshConnectivity) init() error {
 		if err == nil {
 			break
 		}
```
VaishnaviHire, line 88: Should we add `wait.Poll()` instead of `for loop`?

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -310,20 +310,7 @@ func (r *ReconcileWindowsMachine) Reconcile(request reconcile.Request) (reconcil
...
-				if err := r.client.Delete(context.TODO(), machine); err != nil {
-					r.recorder.Eventf(machine, core.EventTypeWarning, "MachineDeletionFailed",
-						"Machine %v deletion failed: %v", machine.Name, err)
-					return reconcile.Result{}, err
-				}
-				log.Info("machine has been remediated by deletion", "name", machine.GetName())
-				r.recorder.Eventf(machine, core.EventTypeNormal, "MachineDeleted",
-					"Machine %v has been remediated by deleting the Machine object", machine.Name)
-				return reconcile.Result{}, nil
+				return reconcile.Result{}, r.deleteMachine(machine)
```
VaishnaviHire, line 313: Do we not need to validate if we have the expected private key when the machine is in `Running` phase ?

VaishnaviHire (commented):
Thanks @sebsoto for working on this. Please address my comments

pkg/controller/windowsmachine/windowsmachine_controller.go
VaishnaviHire: Sorry, I see you are handling only `unconfigured` machines in this commit.

test/e2e/create_test.go
```diff
@@ -67,11 +70,25 @@ func testWindowsNodeCreation(t *testing.T) {
...
 		}
+		if test.isWindows {
+			// We need to cover the case where a user changes the private key secret before the WMCO has a chance to
+			// configure the Machine. In order to simulate that case we need to wait for the MachineSet to be fully
+			// provisioned and then change the key. Depending on timing and configuration flakes this will either cause
+			// the all Machines, or all Machines after the first configured Machines to hit this scenario.
+			err := testCtx.waitForWindowsMachines(int(test.replicas), "Provisioned")
+			require.NoError(t, err, "error waiting for Windows Machines to be provisioned")
+			err = testCtx.createPrivateKeySecret(true)
+			require.NoError(t, err, "error replacing private key")
```
mansikulkarni96, line 89: "error replacing private key secret"

test/e2e/create_test.go
```diff
@@ -94,6 +111,35 @@ func (tc *testContext) createWindowsMachineSet(replicas int32, windowsLabel bool
 			Timeout: cleanupTimeout, RetryInterval: cleanupRetryInterval})
 }
 
+// waitForMachines waits for the a certain amount of Windows Machines to reach a certain phase
```
mansikulkarni96, line 114: waitForWindowsMachines*  waits for a* certain amount

test/e2e/create_test.go
```diff
@@ -94,6 +111,35 @@ func (tc *testContext) createWindowsMachineSet(replicas int32, windowsLabel bool
...
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				log.Printf("waiting for %d Windows Machines", machineCount)
+				return false, nil
+			}
+			log.Printf("machine object listing failed: %v", err)
+			return false, nil
+		}
+		if len(machines.Items) != machineCount {
+			log.Printf("waiting for %d/%d Windows Machines", len(machines.Items), machineCount)
```
mansikulkarni96, line 131: should this be waiting for  -> machineCount - len(machines.Items), machineCount

test/e2e/create_test.go
```diff
@@ -169,25 +215,44 @@ func (tc *testContext) waitForWindowsNodes(nodeCount int32, waitForAnnotations,
...
+		if err != nil {
+			return errors.Wrapf(err, "unable to read private key data from file %s", gc.privateKeyPath)
 		}
 	} else {
-		// Secret already exists, delete it
-		if err := secretsClient.Delete(context.TODO(), secrets.PrivateKeySecret, metav1.DeleteOptions{}); err != nil {
-			return errors.Wrap(err, "unable to delete existing private key secret")
+		keyData, err = generatePrivateKey()
+		if err != nil {
+			return errors.Wrap(err, "error generating private key")
```
mansikulkarni96, line 252: error generating private key secret*
maybe we should mention it everywhere as private key secret, WDYT?

mansikulkarni96 (commented):
Thanks for working on this @sebsoto . PTAL at my comments

test/e2e/create_test.go
sebsoto: Thats actually just returning a private key, not a secret

aravindhp (changes requested):

pkg/controller/windowsmachine/windows/connectivity.go
```diff
@@ -64,16 +82,19 @@ func (c *sshConnectivity) init() error {
...
+	err = wait.PollImmediate(time.Minute, retry.Timeout, func() (bool, error) {
 		sshClient, err = ssh.Dial("tcp", c.ipAddress+":"+sshPort, config)
 		if err == nil {
-			break
+			return true, nil
 		}
 		log.V(1).Info("SSH dial", "IP Address", c.ipAddress, "error", err)
-		time.Sleep(1 * time.Minute)
-	}
+		if strings.Contains(err.Error(), "unable to authenticate") {
```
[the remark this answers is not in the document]
sebsoto: I was not able to find one.
The error message we are looking for is constructed [here](https://github.com/golang/crypto/blob/eec23a3978ad/ssh/client_auth.go#L77).

test/e2e/create_test.go
```diff
@@ -67,11 +70,25 @@ func testWindowsNodeCreation(t *testing.T) {
...
 
 	for _, test := range testCases {
 		if err := testCtx.createWindowsMachineSet(test.replicas, test.isWindows); err != nil {
 			t.Fatalf("failed to create Windows MachineSet %v ", err)
 		}
+		if test.isWindows {
+			// We need to cover the case where a user changes the private key secret before the WMCO has a chance to
+			// configure the Machine. In order to simulate that case we need to wait for the MachineSet to be fully
+			// provisioned and then change the key. Depending on timing and configuration flakes this will either cause
+			// the all Machines, or all Machines after the first configured Machines to hit this scenario.
```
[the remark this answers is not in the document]
sebsoto: > Are you hoping that this happens after the Machine is Provisioned and before it is fully prepped? Won't this potentially result in flakes?

Yes, I am hoping to catch the second Machine which is why I wait for the Machine to enter provision state. Most likely the first one will be provisioned normally, as the key most likely not be changed before the first machine is queued. This can result in non-consistent test runs, such as if the first Machine has a reconcile error and needs to be re-queued, causing it to be deleted and recreated. But there shouldnt be a large negative impact on the test suite.

> Moreover where are you testing if the Machine got deleted in the scenario that the key is changed?

This PR doesn't deal with ensuring all the Machines were configured with the same key, it just solves the case where a Machine is stuck in Provisioned state due to a mismatched key. The fact that the Machine is configured is proof that the mismatched Machine was deleted. 

The next PR I have will deal with deleting configured Machines after the provided private key is changed.

aravindhp (approved):
/approve

with the caveat that a [comment](https://github.com/openshift/windows-machine-config-operator/pull/202#discussion_r552015902) is added

ravisantoshgudimetla (approved):
/lgtm

Thank you for working on this @sebsoto. Hopefully we can add a remediation logic in secret controller in future.
When private key changes, before deleting the Windows VMs, we can login using old private key and modify the ssh key and start using it for configuring.

aravindhp:
> Hopefully we can add a remediation logic in secret controller in future.
When private key changes, before deleting the Windows VMs, we can login using old private key and modify the ssh key and start using it for configuring.

@ravisantoshgudimetla that's a good idea. Please add a story to track this.

ravisantoshgudimetla (approved):
/lgtm

once more. If you find time please find out if this is a flake or one off event.

PR #203, opened 2020-12-03
[build] Fix version and remove operator-sdk build 

PratikMahajan:
this commit removes the usage of `operator-sdk build`
command to build the operator image.
We are now building the operator binary in dockerfile
similar to what we are doing in dockerfile.ci and
using `make build` which uses `build/build.sh` to set
the version for operator binary. Removed other
references of setting version in `hack/olm.sh`

Also included building the WMCO binary using
`make build` while launching the tests through local
machine.
We use this binary for validating the version
annotation on the operator.
This change is needed as `operator-sdk build`
used to generate the operator binary before building
the container image. Now that we are using
`podman build`, we need to explicitly use
`make build` to generate operator binary for e2e test.
Only e2e tests launched through local machine needs
this change. We use different `wmcoPath` while running
the test through CI based on how we generate the
container image.

Removing `operator-sdk build` dependency is important
because `operator-sdk build` is being depreciated in
operator-sdk version 1.x.

Also, this enables us to trigger automatic builds on
quay

### DIFF
--- a/build/Dockerfile.ci
+++ b/build/Dockerfile.ci
@@ -14,25 +14,10 @@ RUN curl -L -s https://mirror.openshift.com/pub/openshift-v4/clients/ocp/4.6.4/o
     && tar -xzf openshift-origin-client-tools.tar.gz \
     && rm -rf ./{openshift*,README.md}
 
-# Build WMCO
 # The source here corresponds to the code in the PR and is placed here by the CI infrastructure.
 WORKDIR /build/windows-machine-config-operator/
 # Copy .git metadata so that we can generate the version for the WMCO binary
 COPY .git .git
-# Copy files and directories needed to build the WMCO binary
-COPY build build
-COPY cmd cmd
-COPY deploy deploy
-COPY hack hack
-COPY pkg pkg
-COPY test test
-COPY vendor vendor
-COPY version version
-COPY go.mod go.mod
-COPY go.sum go.sum
-COPY Makefile Makefile
-COPY tools.go tools.go
-RUN make build
 
 # Build WMCB
 WORKDIR /build/windows-machine-config-operator/windows-machine-config-bootstrapper/
@@ -67,6 +52,24 @@ WORKDIR /build/windows-machine-config-operator/containernetworking-plugins/
 COPY containernetworking-plugins/ .
 ENV CGO_ENABLED=0
 RUN ./build_windows.sh
+
+# Build WMCO
+WORKDIR /build/windows-machine-config-operator
+# Copy files and directories needed to build the WMCO binary
+COPY build build
+COPY cmd cmd
+COPY deploy deploy
+COPY hack hack
+COPY pkg pkg
+COPY test test
+COPY vendor vendor
+COPY version version
+COPY go.mod go.mod
+COPY go.sum go.sum
+COPY Makefile Makefile
+COPY tools.go tools.go
+COPY .gitignore .gitignore
+RUN make build
 
 # Build the operator image with following payload structure
 # /payload/
--- a/hack/common.sh
+++ b/hack/common.sh
@@ -49,8 +49,7 @@ build_WMCO() {
       error-exit "OPERATOR_IMAGE not set"
   fi
 
-  $OSDK build "$OPERATOR_IMAGE" --image-builder $CONTAINER_TOOL $noCache \
-    --go-build-args "-ldflags -X=github.com/openshift/windows-machine-config-operator/version.Version=${VERSION}"
+  $CONTAINER_TOOL build . -t "$OPERATOR_IMAGE" -f build/Dockerfile $noCache
   if [ $? -ne 0 ] ; then
       error-exit "failed to build operator image"
   fi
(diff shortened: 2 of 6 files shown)


hack/common.sh
```diff
@@ -50,8 +50,7 @@ build_WMCO() {
       error-exit "OPERATOR_IMAGE not set"
   fi
 
-  $OSDK build "$OPERATOR_IMAGE" --image-builder $CONTAINER_TOOL $noCache \
-    --go-build-args "-ldflags -X=github.com/openshift/windows-machine-config-operator/version.Version=${VERSION}"
+  $CONTAINER_TOOL build . -t "$OPERATOR_IMAGE" -f build/Dockerfile $noCache
```
aravindhp, line 53: This is not deprecating the SDK  build but removing it. I am fine with it as long as rest of the team agrees but you need to frame the PR differently.

hack/common.sh
```diff
@@ -132,7 +131,7 @@ get_version() {
   GIT_COMMIT=$(git rev-parse --short HEAD)
   VERSION="${OPERATOR_VERSION}+${GIT_COMMIT}"
 
-  if [ -n "$(git status --porcelain)" ]; then
+  if [ -n "$(git status version tools.go go.mod go.sum vendor build cmd hack pkg --porcelain)" ]; then
```
aravindhp, line 134: Missing `Makefile` here.

hack/common.sh
aravindhp, line 53: Does this mean that `operator-sdk build` will not work now? If that is the case, you also need to update [HACKING.md](https://github.com/openshift/windows-machine-config-operator/blob/master/docs/HACKING.md). If `operator-sdk build` works then you don't need to update the doc.

hack/olm.sh
```diff
@@ -42,7 +42,6 @@ source $WMCO_ROOT/hack/common.sh
 
 cd $WMCO_ROOT
 OSDK=$(get_operator_sdk)
-VERSION=$(get_version)
```
aravindhp, line 45: Why is this change needed? A question like this implies that it warrants a mention in the commit message.

aravindhp (changes requested):
Thanks for working on this, @PratikMahajan. Please address my comments.

hack/common.sh
PratikMahajan: Ah!! Thanks for catching !

hack/common.sh
PratikMahajan: it should not break `operator-sdk build` command and it will work. But now that we are removing it, i think we should remove it from doc as well.

hack/olm.sh
PratikMahajan: changed the commit message

hack/common.sh
aravindhp: I can change the way the operator is built by modifying the `Makefile`. It needs to be present here. Moreover you don't get anything by **not** adding it here.

aravindhp (changes requested):

aravindhp (approved):
/approve

hack/common.sh
```diff
@@ -132,7 +131,7 @@ get_version() {
   GIT_COMMIT=$(git rev-parse --short HEAD)
   VERSION="${OPERATOR_VERSION}+${GIT_COMMIT}"
 
-  if [ -n "$(git status --porcelain)" ]; then
+  if [ -n "$(git status version tools.go go.mod go.sum vendor Makefile build cmd hack pkg --porcelain)" ]; then
```
sebsoto, line 134: I dont understand why we are being so specific here, it will make it a pain to update in the future. Can you call out in the code why this needs to be done?

sebsoto (changes requested):

hack/common.sh
PratikMahajan: Added comment in `dockerfile` and `common.sh`

hack/common.sh
aravindhp: Just realized you have not updated the doc.

PratikMahajan:
/hold 
until we fix the tests

build/Dockerfile.ci
```diff
@@ -68,6 +54,24 @@ COPY containernetworking-plugins/ .
 ENV CGO_ENABLED=0
 RUN ./build_windows.sh
 
+# Build WMCO
+WORKDIR /build/windows-machine-config-operator
+# Copy files and directories needed to build the WMCO binary
+COPY build build
```
mansikulkarni96, line 60: is this the only folder that is required to be transferred at the end to solve the dirty tag issue, or do we need something else too?

build/Dockerfile.ci
```diff
@@ -14,25 +14,11 @@ RUN curl -L -s https://mirror.openshift.com/pub/openshift-v4/clients/ocp/4.6.4/o
     && tar -xzf openshift-origin-client-tools.tar.gz \
     && rm -rf ./{openshift*,README.md}
 
-# Build WMCO
 # The source here corresponds to the code in the PR and is placed here by the CI infrastructure.
-WORKDIR /build/windows-machine-config-operator/
+RUN mkdir /build/windows-machine-config-operator
```
aravindhp, line 18: This is not needed

aravindhp (changes requested):

build/Dockerfile.ci
PratikMahajan: We need to transfer .gitignore and move building the WMCO operator after building all submodules. That will remove the dirty tag.

aravindhp (approved):
/approve

PR #211, opened 2020-12-08
[release-4.6][deploy] Fix channel name and description

aravindhp:
- Remove upgrade verbiage from the description in the CSV which only applies to the community operator
- The channel name needs to match the name in the Dockerfile label in the release Dockerfile.


mansikulkarni96:
Thanks for adding this @aravindhp 
/lgtm

PR #212, opened 2020-12-08
[community-4.6] Use go1.15 and update submodules

sebsoto:
This PR:
* Updates to building with go1.15
* Updates submodules to tips of release-4.6 branches

### DIFF
--- a/.ci-operator.yaml
+++ b/.ci-operator.yaml
@@ -2,4 +2,4 @@
 build_root_image:
   namespace: openshift
   name: release
-  tag: golang-1.14
+  tag: golang-1.15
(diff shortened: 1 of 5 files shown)


aravindhp (changes requested):
@sebsoto please follow the pattern introduced in [[submodule][wmcb] Update to b57094033](https://github.com/openshift/windows-machine-config-operator/pull/204). If you are updating multiple sub-modules in one commit, then please do:
```
[submodule] Update

- [$name] Update to $hashlink
...
```
where `$hashlink` is the github url to the commit in repo that we are updating to. This way we can easily track where we stand with sub-modules.

aravindhp (approved):

/approve

PR #215, opened 2020-12-08
[deploy] Update channel in annotations.yaml

mansikulkarni96:
Sync the channel name in annotations.yaml to match the channel set in bundle.Dockerfile that is "preview"


aravindhp (approved):
Thanks for doing this, @mansikulkarni96 

/approve

PR #216, opened 2020-12-09
[test] fix failing upgrade and delete tests

PratikMahajan:
the `defer deleteService` in network test is not able to delete the
service in time. Sometimes it succeeds in deleting the service and
sometimes it takes a lot of time.
we are creating service of same name in upgrade test. In case the
service is not deleted, it throws an error.
This is causing our tests to fail.

The fix includes changing the service name. We are doing
this by using `generateName` field while creating the service

we are frequently coming across timeout while deleting the nodes
the nodes are usually deleted within a minute but sometimes they
time out.
this PR increases the timeout from 5 minutes to 10 minutes

### DIFF
--- a/test/e2e/create_test.go
+++ b/test/e2e/create_test.go
@@ -97,26 +97,30 @@ func (tc *testContext) createWindowsMachineSet(replicas int32, windowsLabel bool
 // waitForWindowsNode waits until there exists nodeCount Windows nodes with the correct set of annotations.
 // if waitForAnnotations = false, the function will return when the node object is first seen and not wait until
 // the expected annotations are present.
-// if expectError = true, the function will wait for duration of 5 minutes for the nodes as the error would be thrown
-// immediately, else we will wait for the duration given by nodeCreationTime variable which is 20 minutes increasing
-// the overall wait time in test suite
+// if expectError = true, the function will wait for duration of 10 minutes if we are deleting all nodes i.e. 0 nodesCount
+// else 5 minutes for the nodes as the error would be thrown immediately, else we will wait for the duration given by
+// nodeCreationTime variable which is 20 minutes increasing the overall wait time in test suite
 func (tc *testContext) waitForWindowsNodes(nodeCount int32, waitForAnnotations, expectError, checkVersion bool) error {
 	var nodes *v1.NodeList
 	annotations := []string{nodeconfig.HybridOverlaySubnet, nodeconfig.HybridOverlayMac, nodeconfig.VersionAnnotation}
 	var creationTime time.Duration
 	startTime := time.Now()
 	if expectError {
-		// The time we expect to wait, if the windowsLabel is
-		// not used while creating nodes.
-		creationTime = time.Minute * 5
+		if nodeCount == 0 {
+			creationTime = time.Minute * 10
+		} else {
+			// The time we expect to wait, if the windowsLabel is
+			// not used while creating nodes.
+			creationTime = time.Minute * 5
+		}
 	} else {
 		creationTime = nodeCreationTime
 	}
 
-	// As per testing, each windows VM is taking roughly 12 minutes to be shown up in the cluster, so to be on safe
-	// side, let's make it as 20 minutes per node. The value comes from nodeCreationTime variable.  If we are testing a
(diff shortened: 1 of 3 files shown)


aravindhp:
/retest

The deletion test is timing out which we will have to look at separately.

test/e2e/network_test.go
```diff
@@ -215,7 +215,7 @@ func retryGET(url string) (*http.Response, error) {
 func (tc *testContext) createService(name string, serviceType v1.ServiceType, selector metav1.LabelSelector) (*v1.Service, error) {
 	svcSpec := &v1.Service{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: name,
+			GenerateName: name + "-",
```
aravindhp, line 218: Is it enough if we do it for the `Service`? I remember seeing this issue with [Jobs](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/175/pull-ci-openshift-windows-machine-config-operator-master-azure-e2e-operator/1332051863398256640) too. So we might want to do this across the board for all objects we are creating in the e2e tests.

aravindhp (changes requested):
Thanks for working on this, @PratikMahajan. Please address my comment.

test/e2e/network_test.go
PratikMahajan: Done !

test/e2e/create_test.go
```diff
@@ -108,15 +108,15 @@ func (tc *testContext) waitForWindowsNodes(nodeCount int32, waitForAnnotations,
 	if expectError {
 		// The time we expect to wait, if the windowsLabel is
 		// not used while creating nodes.
-		creationTime = time.Minute * 5
+		creationTime = time.Minute * 10
```
aravindhp, line 111: Is this correct fix? This increases the time we will wait if we create a Machine without the `windowsLabel`. What does it have to do with deletion?

aravindhp, line 111: In `testWindowsNodeDeletion` the comment for `testCtx.waitForWindowsNodes()` call says `As per testing, each windows VM is taking roughly 12 minutes to be shown up in the cluster`. This is just plain misleading and should be fixed as part of your PR.

Now in waitForWindowsNodes() , I think you need to do:
```suggestion
		if nodeCount == 0 {
			creationTime = time.Minute * 10
		} else {
			// The time we expect to wait, if the windowsLabel is
			// not used while creating nodes.
			creationTime = time.Minute * 5
		}

```

test/e2e/create_test.go
```diff
@@ -108,15 +108,15 @@ func (tc *testContext) waitForWindowsNodes(nodeCount int32, waitForAnnotations,
...
 	} else {
 		creationTime = nodeCreationTime
 	}
 
 	// As per testing, each windows VM is taking roughly 12 minutes to be shown up in the cluster, so to be on safe
 	// side, let's make it as 20 minutes per node. The value comes from nodeCreationTime variable.  If we are testing a
 	// scale down from n nodes to 0, then we should not take the number of nodes into account. If we are testing node
-	// creation without applying Windows label, we should throw error within 5 mins.
+	// creation without applying Windows label, we should throw error within 10 mins.
```
aravindhp, line 119: This comment should stay the same for the incorrect Windows label and you should add a line for the deletion case.

aravindhp (changes requested):

test/e2e/create_test.go
PratikMahajan: Agreed

aravindhp (approved):
/approve

aravindhp:
New error seen here: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/216/pull-ci-openshift-windows-machine-config-operator-master-aws-e2e-operator-with-custom-vxlan-port/1338159023312080896

This needs to be investigated.

aravindhp:
`time="2020-12-13T23:29:02Z" level=fatal msg="Failed to run operator: error waiting for CSV to install: etcdserver: leader changed` is another error that we are seeing consistently. This requires investigation also. 

/test all

PR #221, opened 2020-12-10
[release-4.6][build] Fix version and remove operator-sdk build

PratikMahajan:
This is a backport of https://github.com/openshift/windows-machine-config-operator/pull/203

this PR removes the usage of `operator-sdk build`
command to build the operator image.
We are now building the operator binary in dockerfile
similar to what we are doing in dockerfile.ci and
using `make build` which uses `build/build.sh` to set
the version for operator binary. Removed other
references of setting version in `hack/olm.sh`

Also included building the WMCO binary using
`make build` while launching the tests through local
machine.
We use this binary for validating the version
annotation on the operator.
This change is needed as `operator-sdk build`
used to generate the operator binary before building
the container image. Now that we are using
`podman build`, we need to explicitly use
`make build` to generate operator binary for e2e test.
Only e2e tests launched through local machine needs
this change. We use different `wmcoPath` while running
the test through CI based on how we generate the
container image.

Removing `operator-sdk build` dependency is important
because `operator-sdk build` is being depreciated in
operator-sdk version 1.x.

Also, this enables us to trigger automatic builds on
quay

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -28,6 +28,23 @@ COPY containernetworking-plugins/ .
 ENV CGO_ENABLED=0
 RUN ./build_windows.sh
 
+WORKDIR /build/windows-machine-config-operator/
+# Copy files and directories needed to build the WMCO binary
+# `make build` uses `get_version()` in `hack/common.sh` to determine the version of binary created.
+# Any new file added here should be reflected in `hack/common.sh` if it dirties the git working tree.
+COPY version version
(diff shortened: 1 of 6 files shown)


VaishnaviHire:
LGTM

aravindhp:
/override ci/prow/aws-e2e-operator-with-custom-vxlan-port

All tests had passed but something caused them all to get triggered. Merging given we need to push the code downstream.

PR #222, opened 2020-12-11
WINC-500: controller: Ignore machine events with invalid machine objects

LorbusChris:


pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -214,6 +218,34 @@ func isWindowsMachine(labels map[string]string) bool {
...
 
+func isValidMachine(object runtime.Object) bool {
+	machine := &mapi.Machine{}
+
+	// If for some reason this function called on an object which is not a Node, return
+	if kind := object.GetObjectKind().GroupVersionKind(); kind.Kind != machine.Kind {
+		return false
+	}
+
+	obj := object.DeepCopyObject()
```
LorbusChris, line 229: note to self: don't deepcopy twice

controllers/windowsmachine/windowsmachine_controller.go
```diff
@@ -116,10 +116,10 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		// We need the create event to account for Machines that are in provisioned state but were created
 		// before WMCO started running
 		CreateFunc: func(e event.CreateEvent) bool {
-			return isWindowsMachine(e.Meta.GetLabels())
+			return isValidMachine(e.Object.DeepCopyObject()) && isWindowsMachine(e.Meta.GetLabels())
```
aravindhp, line 119: Is it safe to de-reference `e` without a check?

aravindhp (changes requested):
Thanks for working on this, @LorbusChris. Please address my comments.

Please add a couple of lines to the commit message calling out why we are doing this i.e. reduce number of events being sent to the controller and reduce error logging.

controllers/windowsmachine/windowsmachine_controller.go
sebsoto, line 119: Sorry for the basic question but why does the object need to be DeepCopied? Why can't you just use it directly?

controllers/windowsmachine/windowsmachine_controller.go
```diff
@@ -218,6 +222,33 @@ func isWindowsMachine(labels map[string]string) bool {
...
+
+	var ok bool
+	machine, ok = obj.(*mapi.Machine)
+	log.Info(fmt.Sprintf("%+v\n", machine))
+	if !ok {
+		log.Info("unable to typecast object to machine:\n%s", fmt.Sprintf("%+v\n", obj))
+		return false
+	}
+	if machine.Status.Phase == nil {
+		log.Info("machine object has no phase associated with it:\n%s", fmt.Sprintf("%+v\n", machine.Name))
```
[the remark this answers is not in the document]
LorbusChris: so use `log.Debug()`?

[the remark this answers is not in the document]
aravindhp: I don't `log.Debug()` exists. We have been using in `log.V(1).Info()` instead.

controllers/windowsmachine/windowsmachine_controller.go
LorbusChris: I wanted to be overly cautious. Might not actually be needed here since the func shouldn't mutate the object

controllers/windowsmachine/windowsmachine_controller.go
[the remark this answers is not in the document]
mansikulkarni96: additionally the log message should be of the format log.V(1).Info("machine object has no phase associated with it" , "name", machine.Name) else it will throw odd number of key value pairs error.

controllers/windowsmachine/windowsmachine_controller.go
```diff
@@ -192,7 +192,11 @@ func (m *nodeToMachineMapper) Map(object handler.MapObject) []reconcile.Request
 		log.Error(err, "could not get a list of machines")
 	}
 	for _, machine := range machines.Items {
-		if machine.Status.NodeRef != nil && machine.Status.NodeRef.UID == object.Meta.GetUID() {
+		ok := machine.Status.Phase != nil &&
+			len(machine.Status.Addresses) > 0 &&
+			machine.Status.NodeRef != nil &&
+			machine.Status.NodeRef.UID == object.Meta.GetUID()
```
aravindhp, line 198: Why can't `isValidMachine()` be called here? I understand the signature would have to change, but I am wondering if the main guts of the function applies here.

controllers/windowsmachine/windowsmachine_controller.go
```diff
@@ -192,7 +192,11 @@ func (m *nodeToMachineMapper) Map(object handler.MapObject) []reconcile.Request
...
+		if ok {
```
aravindhp, line 199:
```suggestion
              if !ok {
                  return nil
              }
              return []reconcile.Request{...
```
Less indentation this way.
```

**Review by aravindhp — CHANGES_REQUESTED** [review]

File: controllers/windowsmachine/windowsmachine_controller.go (hunk 2 above)
Comment (sebsoto) [reply controllers/windowsmachine/windowsmachine_controller.go:119]: I'd rather not do it if its not necessary, since the predicates will be called on every machine object change. I'm not sure if its needed either

**Comment by LorbusChris** [comment]
Thanks for the reviews everybody! Will update shortly

File: controllers/windowsmachine/windowsmachine_controller.go [hunk 6]
```diff
@@ -218,6 +222,33 @@ func isWindowsMachine(labels map[string]string) bool {
...
+	// If for some reason this function called on an object which is not a Node, return
+	if kind := obj.GetObjectKind().GroupVersionKind(); kind.Kind != machine.Kind {
+		return false
+	}
+
+	var ok bool
+	machine, ok = obj.(*mapi.Machine)
+	log.Info(fmt.Sprintf("%+v\n", machine))
+	if !ok {
+		log.Info("unable to typecast object to machine:\n%s", fmt.Sprintf("%+v\n", obj))
```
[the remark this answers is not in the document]
LorbusChris: @aravindhp `log.Error` takes an error object as input, which I haven't added to the func.
Should I change the func to return `(bool, error)`?

controllers/windowsmachine/windowsmachine_controller.go
LorbusChris: That way the function would return at the first machine in the list that is `!ok`, instead of continuing to range through them and only returning nil when none of the machine objects is ok.

controllers/windowsmachine/windowsmachine_controller.go
LorbusChris: `isValidMachine()` is it is now takes a `runtime.Object` parameter -- the `machine` obj passed here doesn't implement that.

controllers/windowsmachine/windowsmachine_controller.go
LorbusChris: @aravindhp I've changed this to not do a deep copy here. Could you elaborate if this what you meant?

controllers/windowsmachine/windowsmachine_controller.go
```diff
@@ -218,6 +222,32 @@ func isWindowsMachine(labels map[string]string) bool {
 	return false
 }
 
+func isValidMachine(obj runtime.Object) bool {
+	machine := &mapi.Machine{}
+
+	// If for some reason this function called on an object which is not a Node, return
+	if kind := obj.GetObjectKind().GroupVersionKind(); kind.Kind != machine.Kind {
```
sebsoto, line 229: Comment doesn't seem to match up with code, should be `not a Machine` not node, right?

sebsoto (changes requested):

controllers/windowsmachine/windowsmachine_controller.go
aravindhp: My question was if a `nil` check on `e` was required.

controllers/windowsmachine/windowsmachine_controller.go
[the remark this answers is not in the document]
aravindhp: Can't you do `log.Error(errors.New("blah"), "blah")`?

controllers/windowsmachine/windowsmachine_controller.go
LorbusChris: I've added a nil check in the function :+1:

aravindhp (approved):
Thanks for doing this, @LorbusChris 

/approve

aravindhp:
/retitle WINC-500: controller: Ignore machine events with invalid machine objects

@LorbusChris assuming this is no longer a WIP

LorbusChris:
bootstrap timeout
/retest

aravindhp:
/hold

vSphere cluster bring up is not going past the bootstrapping phase with networking errors.

PR #223, opened 2020-12-11
WIP: [test] Increase webserver replica count to 3

aravindhp:
This is to give us a signal when [Bugzilla 1905950](https://bugzilla.redhat.com/show_bug.cgi?id=1905950)  is fixed and to prevent similar issues in the future.


aravindhp:
Need to fix affinity before this can work.

PR #225, opened 2020-12-11
[wmco][e2e] Add test to check Node Status

VaishnaviHire:
This commit ensures that we test Node Status as a part of our e2e tests.
When the operator reconciles successfully, the Node can be in NotReady
state. This is an unexpected outcome and should be tested as a part of
waitForWindowsNodes function. The changes include checking if the Node
has condition type Ready and if it has a value 'true' for a successful
test.


aravindhp (approved):
Thanks for working on this, @VaishnaviHire 

/approve

PR #227, opened 2020-12-14
[hack] Update WMCO version to 1.0.2

mansikulkarni96:
This PR updates the WMCO version to 1.0.2 in hack/common.sh script.

### DIFF
--- a/hack/common.sh
+++ b/hack/common.sh
@@ -127,7 +127,7 @@ cleanup_WMCO() {
 
 # returns the operator version in `Version+GitHash` format
 get_version() {
-  OPERATOR_VERSION=1.0.1
+  OPERATOR_VERSION=1.0.2
   GIT_COMMIT=$(git rev-parse --short HEAD)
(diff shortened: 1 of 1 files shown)


aravindhp (approved):
/approve

VaishnaviHire:
According to the doc `4.5` index is included as a part of this change. Just want to make sure if that is expected

aravindhp:
/hold
Until #230 merges

hack/common.sh
```diff
@@ -127,7 +127,7 @@ cleanup_WMCO() {
 
 # returns the operator version in `Version+GitHash` format
 get_version() {
-  OPERATOR_VERSION=1.0.1
+  OPERATOR_VERSION=1.0.2
```
aravindhp, line 130: At this point we should just update the version to 2.0.0 in master.

aravindhp (changes requested):

PR #229, opened 2020-12-14
[release-4.6][test] fix failing upgrade and delete tests

PratikMahajan:
the defer deleteService in network test is not able to delete the
service in time. Sometimes it succeeds in deleting the service and
sometimes it takes a lot of time.
we are creating service of same name in upgrade test. In case the
service is not deleted, it throws an error.
This is causing our tests to fail.

The fix includes changing the service name. We are doing
this by using generateName field while creating the service

we are frequently coming across timeout while deleting the nodes
the nodes are usually deleted within a minute but sometimes they
time out.
this PR increases the timeout from 5 minutes to 10 minutes

Backport of #216


aravindhp (approved):
/approve

PR #236, opened 2021-01-06
WINC-557: [test] Use generateName instead of Name

PratikMahajan:
This commit carries over the changes from #216,
i.e using generateName for deployment and job objects.

This change also requires changing the way we use win-webserver in the
upgrade test.
In the upgrade test, we are getting the win-webserver deployment object
created in the network test using a get method. We are getting this
deployment using the deployment name.
As we will be using generateName across the board, we need to change the
way we get the windows web server deployment.
As this is a web server deployment, we can create a new one for
upgrade tests just before we start the test and the linux curler which
ensures that the webserver is functioning across an upgrade, and not use
the win-webserver created in network test. Everything
should be destroyed once we tear down the cluster.

### DIFF
--- a/test/e2e/network_test.go
+++ b/test/e2e/network_test.go
@@ -356,7 +356,7 @@ func (tc *testContext) createWindowsServerDeployment(name string, command []stri
 	containerUserName := "ContainerAdministrator"
 	deployment := &appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: name + "-deployment",
+			GenerateName: name + "-deployment-",
 		},
 		Spec: appsv1.DeploymentSpec{
 			Replicas: &replicaCount,
@@ -493,7 +493,7 @@ func (tc *testContext) createJob(name, image string, command []string, selector 
 	jobsClient := tc.kubeclient.BatchV1().Jobs(tc.workloadNamespace)
 	job := &batchv1.Job{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: name + "-job",
+			GenerateName: name + "-job-",
 		},
 		Spec: batchv1.JobSpec{
 			Template: v1.PodTemplateSpec{
--- a/test/e2e/upgrade_test.go
+++ b/test/e2e/upgrade_test.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/require"
+	batchv1 "k8s.io/api/batch/v1"
 	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -36,7 +37,7 @@ func upgradeTestSuite(t *testing.T) {
 	require.NoError(t, err)
 
 	// test if Windows workloads are running by creating a Job that curls the workloads continuously.
-	err = testCtx.deployWindowsWorkloadTester()
+	testerJob, err := testCtx.deployWindowsWorkloadAndTester()
 	require.NoError(t, err, "error testing Windows workloads")
 
 	// apply configuration steps before running the upgrade tests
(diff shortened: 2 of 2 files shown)


test/e2e/network_test.go
```diff
@@ -290,8 +290,8 @@ func (tc *testContext) deleteNamespace(name string) error {
 	return tc.kubeclient.CoreV1().Namespaces().Delete(context.TODO(), name, metav1.DeleteOptions{})
 }
 
-// deployWindowsWebServer creates a deployment with a single Windows Server pod, listening on port 80
-func (tc *testContext) deployWindowsWebServer(name string, affinity *v1.Affinity) (*appsv1.Deployment, error) {
+// DeployWindowsWebServer creates a deployment with a single Windows Server pod, listening on port 80
```
LorbusChris, line 293: does this func have to be exported? Looks like the new call in `test/e2e/upgrade_test.go` is still from within the same package

PratikMahajan: Correct! 
Thanks for catching that !

test/e2e/upgrade_test.go
```diff
@@ -176,11 +173,10 @@ func (tc *testContext) scaleWMCODeployment(desiredReplicas int32) error {
 // deployWindowsWorkloadTester tests if the Windows Webserver deployment is available.
 // This is achieved by creating a Job object that continuously curls the webserver every 5 seconds.
 func (tc *testContext) deployWindowsWorkloadTester() error {
-	// Get Windows Webserver deployment
-	deployment, err := tc.kubeclient.AppsV1().Deployments(tc.workloadNamespace).Get(context.TODO(), "win-webserver-deployment",
-		metav1.GetOptions{})
+	// Create a Windows Webserver deployment
+	deployment, err := tc.deployWindowsWebServer("win-webserver", nil)
```
aravindhp, line 177: From my reading this breaks the upgrade test. I don't think we are actually testing the upgrade scenario if we create the deployment here. How are we ensuring the workload is functioning across a WMCO "upgrade".

aravindhp: @VaishnaviHire please confirm.

PratikMahajan: [We deploy the workload before scaling down the operator and configuring it for upgrade](https://github.com/openshift/windows-machine-config-operator/blob/602f4f140a57582eb0fc5b08c29795fc6e922cbe/test/e2e/upgrade_test.go#L38-L44)

aravindhp: Ah I see. Sorry I missed that. What you then need to do is not create the workload in `deployWindowsWorkloadTester()`. Or change the name of the function to something else, maybe `deployWindowsWorkloadAndTester()`. 

I am also not a huge fan of not cleaning up resources given your `Everything should be destroyed once we tear down the cluster.`. While you are correct it is not a good pattern to follow. There could be situations where a dangling resource could prevent a cluster from being torn down gracefully.

test/e2e/network_test.go
```diff
@@ -356,7 +356,10 @@ func (tc *testContext) createWindowsServerDeployment(name string, command []stri
 	containerUserName := "ContainerAdministrator"
 	deployment := &appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: name + "-deployment",
+			GenerateName: name + "-deployment-",
+			Labels: map[string]string{
+				"app": name + "-deployment",
```
aravindhp, line 361: Why do we need this label?

sebsoto:
lgtm except for confusion about https://github.com/openshift/windows-machine-config-operator/pull/236#discussion_r555393746

aravindhp (approved):
/approve

aravindhp:
> lgtm except for confusion about [#236 (comment)](https://github.com/openshift/windows-machine-config-operator/pull/236#discussion_r555393746)

What is the confusion?

PR #238, opened 2021-01-08
[hack] Add script to collect Windows network trace

sebsoto:
Captures .etl traces which can be converted into a .pcapng file, to be
consumed by either Wireshark or other programs.

### DIFF
--- a/docs/TROUBLESHOOTING.md
+++ b/docs/TROUBLESHOOTING.md
@@ -85,6 +85,18 @@ Alternatively, the Docker event logs can be viewed using SSH with the following 
   PS C:\Users\username> Get-EventLog -LogName Application -Source Docker
   ```
 
+## How to collect a packet trace on Windows nodes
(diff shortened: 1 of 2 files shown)


aravindhp (approved):
Thanks for doing this, @sebsoto. 

/approve

docs/TROUBLESHOOTING.md
```diff
@@ -85,6 +85,17 @@ Alternatively, the Docker event logs can be viewed using SSH with the following
   PS C:\Users\username> Get-EventLog -LogName Application -Source Docker
   ```
```
VaishnaviHire, line 87: Can you add the following here -
`Requires SSH [bastion](https://github.com/eparis/ssh-bastion) deployed on cluster`

VaishnaviHire (commented):
Thanks @sebsoto . 
Mostly LGTM.  PTAL at my suggestion

PR #239, opened 2021-01-12
Add pubkey annotation to Windows nodes

sebsoto:
If the private key used to configure nodes changes, Machines configured
with the old key will be deleted, so that they can be replaced with new
Machines.

### DIFF
--- a/pkg/controller/secret/secret_controller.go
+++ b/pkg/controller/secret/secret_controller.go
@@ -2,6 +2,8 @@ package secret
 
 import (
 	"context"
+	"fmt"
+	"strings"
 
 	"github.com/pkg/errors"
 	core "k8s.io/api/core/v1"
@@ -22,6 +24,8 @@ import (
 
 	"github.com/openshift/windows-machine-config-operator/pkg/cluster"
 	"github.com/openshift/windows-machine-config-operator/pkg/controller/secrets"
+	"github.com/openshift/windows-machine-config-operator/pkg/controller/signer"
+	"github.com/openshift/windows-machine-config-operator/pkg/controller/windowsmachine/nodeconfig"
 )
 
 const (
@@ -169,12 +173,41 @@ func (r *ReconcileSecret) Reconcile(request reconcile.Request) (reconcile.Result
 		// valid userData secret already exists
 		return reconcile.Result{}, nil
 	} else {
-		// secret is updated
+		// userdata secret data does not match what is expected
+		// Mark nodes configured with the previous private key for deletion
+		signer, err := signer.Create(privateKey)
+		if err != nil {
+			return reconcile.Result{}, errors.Wrap(err, "error creating signer from private key")
+		}
+		nodes := &core.NodeList{}
+		err = r.client.List(context.TODO(), nodes, client.HasLabels{nodeconfig.WindowsOSLabel})
+		if err != nil {
+			return reconcile.Result{}, errors.Wrapf(err, "error getting node list")
+		}
+		expectedPubKeyAnno := nodeconfig.CreatePubKeyHashAnnotation(signer.PublicKey())
+		escapedPubKeyAnnotation := strings.Replace(nodeconfig.PubKeyHashAnnotation, "/", "~1", -1)
+		patchData := fmt.Sprintf(`[{"op":"add","path":"/metadata/annotations/%s","value":""}]`, escapedPubKeyAnnotation)
+		for _, node := range nodes.Items {
+			existingPubKeyAnno := node.Annotations[nodeconfig.PubKeyHashAnnotation]
+			if existingPubKeyAnno == expectedPubKeyAnno {
+				continue
+			}
+			node.Annotations[nodeconfig.PubKeyHashAnnotation] = ""
+			err = r.client.Patch(context.TODO(), &node, client.RawPatch(kubeTypes.JSONPatchType, []byte(patchData)))
+			if err != nil {
+				return reconcile.Result{}, errors.Wrapf(err, "error clearing public key annotation on node %s",
+					node.GetName())
+			}
+			log.V(1).Info("patched node object", "node", node.GetName(), "patch", patchData)
+		}
+
+		// Set userdata to expected value
 		log.Info("updating secret", "name", userDataSecret)
 		err = r.client.Update(context.TODO(), validUserData)
 		if err != nil {
 			return reconcile.Result{}, err
 		}
+
 		// Secret updated successfully
 		return reconcile.Result{}, nil
 	}
--- a/pkg/controller/windowsmachine/nodeconfig/nodeconfig.go
+++ b/pkg/controller/windowsmachine/nodeconfig/nodeconfig.go
@@ -2,6 +2,7 @@ package nodeconfig
 
 import (
 	"context"
+	"crypto/sha256"
 	"fmt"
 	"net/url"
 	"strings"
@@ -35,6 +36,8 @@ const (
 	WorkerLabel = "node-role.kubernetes.io/worker"
 	// VersionAnnotation indicates the version of WMCO that configured the node
 	VersionAnnotation = "windowsmachineconfig.openshift.io/version"
+	// PubKeyHashAnnotation corresponds to the public key present on the VM
+	PubKeyHashAnnotation = "windowsmachineconfig.openshift.io/pub-key-hash"
 )
 
 // nodeConfig holds the information to make the given VM a kubernetes node. As of now, it holds the information
@@ -48,6 +51,8 @@ type nodeConfig struct {
 	node *core.Node
 	// network holds the network information specific to the node
 	network *network
+	// publicKeyHash is the hash of the public key present on the VM
+	publicKeyHash string
 	// clusterServiceCIDR holds the service CIDR for cluster
 	clusterServiceCIDR string
 }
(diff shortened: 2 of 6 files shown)


pkg/controller/secret/secret_controller.go
```diff
@@ -175,6 +176,20 @@ func (r *ReconcileSecret) Reconcile(request reconcile.Request) (reconcile.Result
...
+
+		// Mark nodes configured with the previous userData for deletion
+		nodes := &core.NodeList{}
+		err := r.client.List(context.TODO(), nodes, client.HasLabels{nodeconfig.WindowsOSLabel})
+		if err != nil {
+			return reconcile.Result{}, errors.Wrapf(err, "error getting node list")
+		}
+		for _, node := range nodes.Items {
+			node.Annotations[nodeconfig.PubKeyHashAnnotation] = ""
+			err := r.client.Update(context.TODO(), &node)
```
VaishnaviHire, line 188: We could `patch` the node here

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -295,8 +306,10 @@ func (r *ReconcileWindowsMachine) Reconcile(request reconcile.Request) (reconcil
 		}
 
 		if _, present := node.Annotations[nodeconfig.VersionAnnotation]; present {
```
VaishnaviHire, line 308: Shouldn`t we also be checking for `PubKeyHashAnnotation` here ?

VaishnaviHire (commented):

pkg/controller/windowsmachine/windowsmachine_controller.go
sebsoto: Thats not necessary, as this line is a check to see if the node is marked as fully configured or not

pkg/controller/windowsmachine/nodeconfig/nodeconfig.go
```diff
@@ -46,6 +49,8 @@ type nodeConfig struct {
 	node *v1.Node
 	// network holds the network information specific to the node
 	network *network
+	// signer is used for authenticating against the VM
+	signer ssh.Signer
```
aravindhp, line 53: You just need the hash here and don't need to pass `signer` unless I am missing something.

pkg/controller/windowsmachine/nodeconfig/nodeconfig.go
```diff
@@ -201,6 +207,19 @@ func (nc *nodeConfig) addVersionAnnotation() {
...
 }
 
+// addPubKeyHashAnnotation adds the public key annotation to nc.node
+func (nc *nodeConfig) addPubKeyHashAnnotation() {
+	nc.node.Annotations[PubKeyHashAnnotation] = CreatePubKeyHashAnnotation(nc.signer.PublicKey())
+}
+
+// CreatePubKeyHashAnnotation returns a formatted string which can be used for a public key annotation on a node.
+// The annotation is the sha256 of the public key
+func CreatePubKeyHashAnnotation(key ssh.PublicKey) string {
```
aravindhp, line 217: Please bunch helper functions together

test/e2e/validation_test.go
```diff
@@ -44,6 +48,12 @@ func testNodeMetadata(t *testing.T) {
 				assert.Equal(t, operatorVersion, node.Annotations[nc.VersionAnnotation],
 					"WMCO version annotation mismatch")
 			})
+			t.Run("PubKey Annotation", func(t *testing.T) {
```
aravindhp, line 51: `Public Key Annotation`

aravindhp (changes requested):

test/e2e/secrets_test.go
```diff
@@ -0,0 +1,188 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kubeTypes "k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	"github.com/openshift/windows-machine-config-operator/pkg/controller/secrets"
+)
+
+// getExpectedPublicKey returns the public key associated with the private key within the cloud-private-key secret
+func getExpectedPublicKey() (ssh.PublicKey, error) {
```
ravisantoshgudimetla, line 30: nit: Having a struct for Pulickey,PrivateKey and corresponding hash in nodeConfig would have allowed you to reuse it and reduced the calls to CreatePubKeyHashAnnotation

ravisantoshgudimetla (commented):

Mostly LGTM except for some small comments.

pkg/controller/windowsmachine/windowsmachine_controller.go
sebsoto: I wanted to keep the node deletion in one spot so that we arent duplicating code. I also think its better if the Machine controller handles the Machine deletion logic

pkg/controller/secret/secret_controller.go
```diff
@@ -169,12 +173,41 @@ func (r *ReconcileSecret) Reconcile(request reconcile.Request) (reconcile.Result
...
+		signer, err := signer.Create(privateKey)
+		if err != nil {
+			return reconcile.Result{}, errors.Wrap(err, "error creating signer from private key")
+		}
+		nodes := &core.NodeList{}
+		err = r.client.List(context.TODO(), nodes, client.HasLabels{nodeconfig.WindowsOSLabel})
+		if err != nil {
+			return reconcile.Result{}, errors.Wrapf(err, "error getting node list")
+		}
+		expectedPubKeyAnno := nodeconfig.CreatePubKeyHashAnnotation(signer.PublicKey())
```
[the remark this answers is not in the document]
sebsoto: This is only being ran when the userdata/privatekey has been changed so the hash will be different than what it was.

test/e2e/create_test.go
```diff
@@ -42,64 +43,45 @@ func creationTestSuite(t *testing.T) {
...
-			err = testCtx.createPrivateKeySecret(true)
-			require.NoError(t, err, "error replacing private key secret")
-
-		}
-		err := testCtx.waitForWindowsNodes(test.replicas, true, !test.isWindows, false)
+	t.Run("Windows Machines without the Windows label are not configured", func(t *testing.T) {
+		ms, err := testCtx.createWindowsMachineSet(1, false)
+		require.NoError(t, err, "failed to create Windows MachineSet")
+		defer framework.Global.Client.Delete(context.TODO(), ms)
+		err = testCtx.waitForWindowsNodes(1, true, false, false)
```
VaishnaviHire, line 57: If we are expecting an error here, shouldn`t this be testCtx.waitForWindowsNodes(1, true, true, false)

pkg/controller/windowsmachine/windowsmachine_controller.go
```diff
@@ -302,8 +313,10 @@ func (r *ReconcileWindowsMachine) Reconcile(request reconcile.Request) (reconcil
...
-			// version annotation doesn't match the current operator version
-			if node.Annotations[nodeconfig.VersionAnnotation] != version.Get() {
+			// If either the version annotation doesn't match the current operator version, or the private key used
+			// to configure the machine is out of date, the machine should be deleted
+			if node.Annotations[nodeconfig.VersionAnnotation] != version.Get() ||
+				node.Annotations[nodeconfig.PubKeyHashAnnotation] != nodeconfig.CreatePubKeyHashAnnotation(r.signer.PublicKey()) {
```
VaishnaviHire, line 319: This log [message](https://github.com/openshift/windows-machine-config-operator/pull/239/commits/5f997795192e75e5b59b117fdc2692e45722db1e#diff-f6cf75653deff7ed72be94f059acc292d15113ed7c9f9ce0fdc5e842fa71aa72R324) is no longer true and needs to be updated

test/e2e/create_test.go
sebsoto: That option reduces the amount of time the function waits for the Windows node. Its not right to use it in this case we want to wait as long as we usually do to ensure that no node is configured. That option really should only be used when deleting nodes.

pkg/controller/windowsmachine/windowsmachine_controller.go
sebsoto: > Aren't we already in machine controller?

Sorry yeah, I misread githubs formatting of the comments. 

I'm not fully sure what you are asking, are you saying that 
```

	// Update the signer with the current privateKey
	r.signer, err = signer.Create(privateKey)
	if err != nil {
		return reconcile.Result{}, errors.Wrap(err, "error creating signer")
	}
	// validate userData secret
	if err := r.validateUserData(privateKey); err != nil {
		return reconcile.Result{}, errors.Wrapf(err, "error validating userData secret")
	}
```
Should be done where it was before?

sebsoto: If so I have just done so

aravindhp (approved):
/approve

pkg/controller/secret/secret_controller.go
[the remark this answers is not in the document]
ravisantoshgudimetla: nit: I thought we'd run it on every reconcile even if there is version change. Agreed it may be short circuited but why to do it if we know it hasn't changed.

pkg/controller/windowsmachine/windowsmachine_controller.go
ravisantoshgudimetla: > Should be done where it was before?

Yes but I think that partly solves the problem.

Let me ask in a different way, what is the chance the version annotation is not present on the node object and publicKey annotation has changed.

sebsoto: The version annotation not being present means that the node has not been fully configured. We have to check that the version annotation is set before checking the public key hash

pkg/controller/secret/secret_controller.go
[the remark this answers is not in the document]
sebsoto: There are two ways to get to this, either the private key has changed or the userdata has changed. The userdata will only be changed if either the private key has changed or someone is tampering with the secret. I dont think we should consider tampering with the secret something that will happen in a normal cluster, and I'd prefer to keep the code here simpler.

ravisantoshgudimetla (approved):

/hold cancel

PR #242, opened 2021-01-14
[community-4.6][test] Use generateName instead of Name

PratikMahajan:
This is a backport of #236 

This commit ensures that we use generateName instead of Name                                        
for deployment and job objects.  

This change also requires changing the way we use win-webserver in the
upgrade test.
In the upgrade test, we are getting the win-webserver deployment object
created in the network test using a get method. We are getting this
deployment using the deployment name.
As we will be using generateName across the board, we need to change the
way we get the windows web server deployment.
As this is a web server deployment, we can create a new one for
upgrade tests just before we start the test and the linux curler which
ensures that the webserver is functioning across an upgrade, and not use
the win-webserver created in network test.


ravisantoshgudimetla (approved):
Thank you for working on this @PratikMahajan 

/approve

PR #246, opened 2021-01-15
WINC-528: [community-4.6][wmco] add metrics to display node graphs

VaishnaviHire:
This commit ensures that we create a PrometheusRule object that
defines Windows specific rules. The queries defined utilze the
metrics generated from Windows exporter. Going forward, we need
to add required queries to the 'windows.rules' group. The queries
added also result in displaying node graphs for Windows nodes in
OpenShift console.The queries are tested by calling prometheus
api and querying every record created.

(cherry picked from commit 9ef5a2aba84d391e6bed0c8cd79b200c02eab522)


aravindhp (approved):
/approve

PR #249, opened 2021-01-15
[community-4.6] Delete Machines when SSH auth fails

sebsoto:
Backport of https://github.com/openshift/windows-machine-config-operator/pull/202

This commit prevents an issue where a user changes the private key used
to configure Windows Machines, after unconfigured Machines have been
created with the original userdata secret. When those unconfigured
Machines are reconciled they will never be able to be SSH'd into and
will infinitely be requeued.

This issue is being fixed by deleting Machines that have had an
authentication error so that they can be re-provisioned with userdata
corresponding with the new private key.


aravindhp (approved):
/approve

PR #252, opened 2021-01-18
Bug 1905280: [wmco] Delete existing service monitor object on WMCO restart

alinaryan:
This change checks for an existing service monitor object on a WMCO restart.
If a service monitor object does exist, it is deleted and a new one is created.
We are deleting to ensure that the SM always exists with the correct spec. Otherwise,
metrics may exhibit unexpected behavior if created by a previous version of WMCO.

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -53,6 +53,7 @@ rules:
   - update
   - list
   - patch
+  - delete
 # deployment/finalizers permissions needed for the metrics server
 - apiGroups:
   - apps
--- a/deploy/olm-catalog/windows-machine-config-operator/manifests/windows-machine-config-operator.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/windows-machine-config-operator/manifests/windows-machine-config-operator.clusterserviceversion.yaml
@@ -314,6 +314,7 @@ spec:
           - update
           - list
           - patch
+          - delete
         - apiGroups:
           - apps
           resourceNames:
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -10,6 +10,7 @@ import (
 	"github.com/pkg/errors"
 	monclient "github.com/prometheus-operator/prometheus-operator/pkg/client/versioned/typed/monitoring/v1"
 	"k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
@@ -86,6 +87,20 @@ func Add(ctx context.Context, cfg *rest.Config, namespace string) error {
 
 	// the name for the metrics resources is set during creation of metrics service and is equivalent to the service name
 	windowsMetricsResource = service.GetName()
+
+	// Create a monitoring client to interact with the ServiceMonitor object
+	mclient, err := monclient.NewForConfig(cfg)
+	if err != nil {
+		return errors.Wrap(err, "could not create monitoring client")
+	}
+
+	// In the case of an operator restart, a previous SM object will be deleted and a new one will
+	// be created. We are deleting to ensure that the SM always exists with the correct spec. Otherwise,
+	// metrics may exhibit unexpected behavior if created by a previous version of WMCO.
+	err = mclient.ServiceMonitors(namespace).Delete(context.TODO(), windowsMetricsResource, metav1.DeleteOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
+		return errors.Wrap(err, "could not delete existing ServiceMonitor object")
+	}
 
 	// CreateServiceMonitors will automatically create the prometheus-operator ServiceMonitor resources
 	// necessary to configure Prometheus to scrape metrics from this operator.


pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -84,18 +85,30 @@ func Add(ctx context.Context, cfg *rest.Config, namespace string) error {
...
-		// ErrServiceMonitorNotPresent, which can be used to safely skip ServiceMonitor creation.
-		if err == metrics.ErrServiceMonitorNotPresent {
-			metricsEnabled = false
-			return errors.Wrap(err, "install prometheus-operator in your cluster to create ServiceMonitor objects")
+		return errors.Wrap(err, "could not create monitoring client")
+	}
 
+	// If ServiceMonitor object does not exist, create one
+	// else use existing ServiceMonitor object
+	_, err = mclient.ServiceMonitors(namespace).Get(context.TODO(), windowsMetricsEndpoints, metav1.GetOptions{})
```
sebsoto, line 96: You cant just check if the service monitor exists, we need to ensure that it exists with the correct spec. Consider the case where the service monitor is created by a previous version of WMCO. The expected value of the service monitor back then might differ than the current version. In that case metrics may exhibit unexpected behavior.

sebsoto (changes requested):
Good work @alinaryan PTAL at my comments

pkg/controller/windowsmachine/metrics/metrics.go
alinaryan: Got it. So instead could I do something like "If ServiceMonitor object does not exist, create one. Else delete current one/create a new one"?

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -84,18 +85,30 @@ func Add(ctx context.Context, cfg *rest.Config, namespace string) error {
...
+			// CreateServiceMonitors will automatically create the prometheus-operator ServiceMonitor resources
+			// necessary to configure Prometheus to scrape metrics from this operator.
+			_, err = metrics.CreateServiceMonitors(cfg, namespace, services)
+			if err != nil {
+				log.Error(err, "could not create ServiceMonitor object")
+				// If this operator is deployed to a cluster without the prometheus-operator running, it will return
+				// ErrServiceMonitorNotPresent, which can be used to safely skip ServiceMonitor creation.
+				if err == metrics.ErrServiceMonitorNotPresent {
+					metricsEnabled = false
+					return errors.Wrap(err, "install prometheus-operator in your cluster to create ServiceMonitor objects")
```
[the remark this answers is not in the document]
alinaryan: Would you recommend creating a separate function for this check before metrics are handled or setting `metricsEnabled = false` for every function in Add() that returns an error?

pkg/controller/windowsmachine/metrics/metrics.go
sebsoto: Yep it could be handled that way

pkg/controller/windowsmachine/metrics/metrics.go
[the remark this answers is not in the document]
sebsoto: Sorry I didnt respond to this earlier, the first option is definitely the better way to go, but its adds the extra requirement that `Add()` needs to be retried if it fails.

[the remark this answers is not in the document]
alinaryan: Got it. Thanks! I'll create a separate bug or commit for this after I make the other changes.

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -84,19 +85,37 @@ func Add(ctx context.Context, cfg *rest.Config, namespace string) error {
...
+		return errors.Wrap(err, "could not create monitoring client")
+	}
 
+	// If ServiceMonitor object does not exist, create one
+	// else update the current one.
+	existingSM, err := mclient.ServiceMonitors(namespace).Get(context.TODO(), windowsMetricsEndpoints, metav1.GetOptions{})
+	if existingSM != nil {
+		_ = mclient.ServiceMonitors(namespace).Delete(context.TODO(), windowsMetricsEndpoints, metav1.DeleteOptions{})
+		services := []*v1.Service{service}
+		_, err = metrics.CreateServiceMonitors(cfg, namespace, services)
```
mansikulkarni96, line 100: should we return an error here ? something like `could not create ServiceMonitor object`

pkg/controller/windowsmachine/metrics/metrics.go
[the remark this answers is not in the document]
VaishnaviHire: This checks if `err` returned by the `metrics.CreateServiceMonitors` is the one defined by [service-monitor.go](https://github.com/operator-framework/operator-sdk/blob/v0.19.x/pkg/metrics/service-monitor.go#L30). We would not be able to call the function before we create the `metrics service`.

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -84,19 +85,36 @@ func Add(ctx context.Context, cfg *rest.Config, namespace string) error {
...
+				log.Error(err, "could not create ServiceMonitor object")
+				// If this operator is deployed to a cluster without the prometheus-operator running, it will return
+				// ErrServiceMonitorNotPresent, which can be used to safely skip ServiceMonitor creation.
+				if err == metrics.ErrServiceMonitorNotPresent {
+					metricsEnabled = false
+					return errors.Wrap(err, "install prometheus-operator in your cluster to create ServiceMonitor objects")
+				}
+			}
 		}
+		return errors.Wrap(err, "could not get ServiceMonitor object")
```
aravindhp, line 117:
```suggestion
	}
	// The ServiceMonitor object was either not found or was deleted
	if (err != nil && apierrors.IsNotFound(err)) || existingSM != nil {
		services := []*v1.Service{service}
		// CreateServiceMonitors will automatically create the prometheus-operator ServiceMonitor resources
		// necessary to configure Prometheus to scrape metrics from this operator.
		_, err = metrics.CreateServiceMonitors(cfg, namespace, services)
		if err != nil {
			log.Error(err, "could not create ServiceMonitor object")
			// If this operator is deployed to a cluster without the prometheus-operator running, it will return
			// ErrServiceMonitorNotPresent, which can be used to safely skip ServiceMonitor creation.
			if err == metrics.ErrServiceMonitorNotPresent {
				metricsEnabled = false
				return errors.Wrap(err, "install prometheus-operator in your cluster to create ServiceMonitor objects")
			}
			return errors.Wrap(err, "could not create ServiceMonitor object")
		}
	}
	if err != nil && !apierrors.IsNotFound(err) {
		return errors.Wrap(err, "could not get ServiceMonitor object")
	}
```
An completely untested suggestion that might fix my other comments :-) WDYT?

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -84,19 +85,36 @@ func Add(ctx context.Context, cfg *rest.Config, namespace string) error {
...
-		log.Error(err, "could not create ServiceMonitor object")
-		// If this operator is deployed to a cluster without the prometheus-operator running, it will return
-		// ErrServiceMonitorNotPresent, which can be used to safely skip ServiceMonitor creation.
-		if err == metrics.ErrServiceMonitorNotPresent {
-			metricsEnabled = false
-			return errors.Wrap(err, "install prometheus-operator in your cluster to create ServiceMonitor objects")
+		return errors.Wrap(err, "could not create monitoring client")
+	}
 
+	// Check for existing SM object. If one exists, delete. Else create.
```
aravindhp, line 94: Please mention why we are deleting.

aravindhp (changes requested):
Thanks for working on this, @alinaryan. Please address my comments

pkg/controller/windowsmachine/metrics/metrics.go
alinaryan: Thanks Aravindh! I'll test these changes. In line 103, I think  existingSM would need to be `existingSM == nil` because in the case that existingSM was deleted it would be empty. Also do we need `could not create ServiceMonitor object`error message twice (lines 109 and 116)? Or is it fine just to have it on line 116?

VaishnaviHire:
@alinaryan  you would also need to run following command to reflect changes in the `role.yam` into the csv

```
operator-sdk generate csvs
```

pkg/controller/windowsmachine/metrics/metrics.go
aravindhp: Why would the variable `existingSM` be `nil` because the object was deleted? The variable `existingSM` will still hold the object data. It is just that it will be stale. 

I am fine not having the log (line 109) and just reflecting it when returning the error (line 116).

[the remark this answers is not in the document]
alinaryan: Fixed this in latest commit

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -84,19 +85,36 @@ func Add(ctx context.Context, cfg *rest.Config, namespace string) error {
...
+	existingSM, err := mclient.ServiceMonitors(namespace).Get(context.TODO(), windowsMetricsEndpoints, metav1.GetOptions{})
+	if existingSM != nil {
+		deleteErr := mclient.ServiceMonitors(namespace).Delete(context.TODO(), windowsMetricsEndpoints, metav1.DeleteOptions{})
```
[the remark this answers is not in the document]
alinaryan: Fixed this in latest commit

pkg/controller/windowsmachine/metrics/metrics.go
alinaryan: I found my previous approach to be faulty. Please see my latest commit.

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -84,18 +85,31 @@ func Add(ctx context.Context, cfg *rest.Config, namespace string) error {
...
-		if err == metrics.ErrServiceMonitorNotPresent {
-			metricsEnabled = false
-			return errors.Wrap(err, "install prometheus-operator in your cluster to create ServiceMonitor objects")
+		return errors.Wrap(err, "could not create monitoring client")
+	}
 
+	// In the case of an operator restart, a previous SM object will be deleted and a
+	// new one will be created with the current spec. If a SM object does not yet exist,
+	// one will be created.
+	deleteErr := mclient.ServiceMonitors(namespace).Delete(context.TODO(), windowsMetricsEndpoints, metav1.DeleteOptions{})
```
VaishnaviHire, line 97: Instead of calling `createServiceMonitor` twice and creating a separate function, you could check for error at the start
```suggestion

	deleteErr := mclient.ServiceMonitors(namespace).Delete(context.TODO(), windowsMetricsEndpoints, metav1.DeleteOptions{})
	if deleteErr != nil  && !apierrors.IsNotFound(err) {
		return errors.Wrap(err, "could not delete existing ServiceMonitor object")
	}  
	
	
	// CreateServiceMonitors will automatically create the prometheus-operator ServiceMonitor resources
	// necessary to configure Prometheus to scrape metrics from this operator.
	services := []*v1.Service{service}
	_, err = metrics.CreateServiceMonitors(cfg, namespace, services)
	if err != nil {
		log.Error(err, "could not create ServiceMonitor object")
		// If this operator is deployed to a cluster without the prometheus-operator running, it will return
		// ErrServiceMonitorNotPresent, which can be used to safely skip ServiceMonitor creation.
		if err == metrics.ErrServiceMonitorNotPresent {
			metricsEnabled = false
			return errors.Wrap(err, "install prometheus-operator in your cluster to create ServiceMonitor objects")

		}
	}
```

deploy/role.yaml
```diff
@@ -53,6 +53,7 @@ rules:
   - update
   - list
   - patch
+  - delete
```
VaishnaviHire, line 56: csv needs to be updated

VaishnaviHire (commented):
Thanks @alinaryan for the PR. PTAL at my comments

pkg/controller/windowsmachine/metrics/metrics.go
alinaryan: Ok thanks! This looks like it will work if the operator is restarting but what about when it is the first time the operator is being deployed? A SM object won't yet exist so deleteErr will be != nil, will it just jump to SM creation?

VaishnaviHire: It should, since we are moving to `creation`  if Service Monitor object is deleted successfully or is not found

alinaryan: Ok testing it now

alinaryan: I tested this approach but it fails to create a SM. I ran `oc -n openshift-windows-machine-config-operator get servicemonitor` and it returned `No resources found in openshift-windows-machine-config-operator namespace.`

alinaryan: These changes worked! I just added them to the PR

pkg/controller/windowsmachine/metrics/metrics.go
alinaryan: There was an error where existingSM != nil was always true and a SM was not able to be created.

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -84,6 +85,20 @@ func Add(ctx context.Context, cfg *rest.Config, namespace string) error {
 		return errors.Wrap(err, "could not create metrics Service")
 	}
 
+	// Create a monitoring client to hold the ServiceMonitor object
```
aravindhp, line 88: Rather than `hold` I would say `interact with`

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -84,6 +85,20 @@ func Add(ctx context.Context, cfg *rest.Config, namespace string) error {
...
+	mclient, err := monclient.NewForConfig(cfg)
+	if err != nil {
+		return errors.Wrap(err, "could not create monitoring client")
+	}
+
+	// In the case of an operator restart, a previous SM object will be deleted and a
+	// new one will be created with the current spec. If a SM object does not yet exist,
+	// one will be created.
```
aravindhp, line 96: Please mention [why](https://github.com/openshift/windows-machine-config-operator/pull/252#discussion_r560380125) we are deleting it. This should be called out in the commit message also.

aravindhp (approved):
/approve

with the caveat of addressing the comments.

Nice work @alinaryan @VaishnaviHire. This is much cleaner and simpler 👍🏾

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -84,6 +85,20 @@ func Add(ctx context.Context, cfg *rest.Config, namespace string) error {
...
+	deleteErr := mclient.ServiceMonitors(namespace).Delete(context.TODO(), windowsMetricsEndpoints, metav1.DeleteOptions{})
+	if deleteErr != nil && !apierrors.IsNotFound(deleteErr) {
+		return errors.Wrap(err, "could not delete existing ServiceMonitor object")
```
sebsoto, line 99: instead of using `deleteErr` just use `err`
you are returning wrapped `err` anyway :wink:

sebsoto (changes requested):

pkg/controller/windowsmachine/metrics/metrics.go
sebsoto: Only reason to use another error variable besides `err` is if you want to preserve a previous error for some reason.

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -95,7 +110,6 @@ func Add(ctx context.Context, cfg *rest.Config, namespace string) error {
 		if err == metrics.ErrServiceMonitorNotPresent {
 			metricsEnabled = false
 			return errors.Wrap(err, "install prometheus-operator in your cluster to create ServiceMonitor objects")
-
```
VaishnaviHire, line 98: This change can be reverted

alinaryan: Do you mean delete the newline? I don't see one when I view the file https://github.com/openshift/windows-machine-config-operator/blob/8d2b8cfea355e4de75e408890a8429c917972798/pkg/controller/windowsmachine/metrics/metrics.go#L113

alinaryan: or remove the check for ErrServiceMonitorNotPresent and setting metricsEnabled = false?

VaishnaviHire: A line is removed. It can be added back.

aravindhp:
/hold

until we figure out why  `vsphere-e2e-operator` is failing

aravindhp:
/hold
Didn't see the merge conflict

PR #254, opened 2021-01-19
Explicitly set GOBIN path for promu installation

mansikulkarni96:
This PR creates windows_exporter directory before-hand and explicitly sets the GOBIN path to specify where the promu binary should be installed, promu installed in the windows_exporter directory is then utilized for its installation.
This change was necessary because Cachito integration on CpaaS side changes the GOPATH during image build and if the path is not specified promu is installed to some location from which it is not accessible to build windows_exporter executable.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -18,12 +18,13 @@ RUN make windows
 # Build promu utility tool, needed to build the windows_exporter.exe metrics binary
 WORKDIR /build/windows-machine-config-operator/promu/
 COPY promu/ .
-RUN go install .
+# Explicitly set the $GOBIN path for promu installation
+RUN GOBIN=/build/windows-machine-config-operator/windows_exporter/ go install .
 
 # Build windows_exporter
 WORKDIR /build/windows-machine-config-operator/windows_exporter/
 COPY windows_exporter/ .
-RUN GOOS=windows promu build -v
+RUN GOOS=windows ./promu build -v
 
 # Build Kubernetes node binaries
 WORKDIR /build/windows-machine-config-operator/kubernetes/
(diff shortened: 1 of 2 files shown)


build/Dockerfile
```diff
@@ -18,12 +18,15 @@ RUN make windows
 # Build promu utility tool, needed to build the windows_exporter.exe metrics binary
 WORKDIR /build/windows-machine-config-operator/promu/
 COPY promu/ .
-RUN go install .
+
+# Create a directory for windows_exporter and set the $GOBIN path for promu installation
+RUN mkdir /build/windows-machine-config-operator/windows_exporter/
```
PratikMahajan, line 23: the WORKDIR below(L27) basically does `mkdir` and `cd`. 
So this step can be skipped.

build/Dockerfile
```diff
@@ -18,12 +18,15 @@ RUN make windows
...
+RUN export GOBIN=/build/windows-machine-config-operator/windows_exporter/ && go install .
```
PratikMahajan, line 24: You can remove `export` if `GOBIN` needs to be set only for this command.

build/Dockerfile
PratikMahajan: `go install .` below should create the directory you're passing as `GOBIN`. Is it facing any permission issues on cachito side?

build/Dockerfile
```diff
@@ -18,12 +18,13 @@ RUN make windows
 # Build promu utility tool, needed to build the windows_exporter.exe metrics binary
 WORKDIR /build/windows-machine-config-operator/promu/
 COPY promu/ .
-RUN go install .
+# Explicitly set the $GOBIN path for promu installation
+RUN GOBIN=/build/windows-machine-config-operator/windows_exporter/ && go install .
```
PratikMahajan, line 22:
```suggestion
RUN GOBIN=/build/windows-machine-config-operator/windows_exporter/ go install .
```

PratikMahajan (changes requested):

build/Dockerfile
mansikulkarni96: thanks for the catch

aravindhp (approved):
/approve

PR #258, opened 2021-01-20
Switch build base image to Red Hat hosted image

sebsoto:
This commit switches from a dockerhub hosted image to a Red Hat hosted
image. This will fix an issue with dockerhub rate limiting when building
the community operator.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,4 +1,4 @@
-FROM docker.io/openshift/origin-release:golang-1.15 as build
+FROM registry.svc.ci.openshift.org/origin/builder:golang-1.15 as build
 LABEL stage=build
 
 WORKDIR /build/windows-machine-config-operator/


ravisantoshgudimetla (approved):
/approve

build/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM docker.io/openshift/origin-release:golang-1.15 as build
+FROM registry.svc.ci.openshift.org/origin/builder:golang-1.15 as build
```
VaishnaviHire, line 1: Is there any difference between this and `registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.7` image ?

sebsoto: I'm not familiar with the other image, this is the equivalent of the previous image

VaishnaviHire: Its the image used in [WMCB](https://github.com/openshift/windows-machine-config-bootstrapper/blob/master/Dockerfile.tools#L1)

PR #260, opened 2021-01-20
[hack] Fix user issue in packet trace script

sebsoto:
This commit fixes an issue that would prevent the packet trace script
from running against any node without the user capi.


aravindhp (approved):
/approve

ravisantoshgudimetla (approved):
/lgtm

PR #261, opened 2021-01-20
[hack] Modify ami_date for machineset.sh script

mansikulkarni96:
This PR modifies the  ami_date for AWS machineset creation script from 2020.09.09 to 2020-10-14, since
the earlier is no longer available in AWS regions.

### DIFF
--- a/hack/machineset.sh
+++ b/hack/machineset.sh
@@ -78,9 +78,12 @@ get_aws_ms() {
   local provider=$4
 
   # get the AMI id for the Windows VM
-  ami_date="2020.09.09"
+  ami_date="2020.10.14"
   ami_id=$(aws ec2 describe-images --filters Name=name,Values=Windows_Server-2019-English-Full-ContainersLatest-${ami_date} --region ${region} --query 'Images[*].[ImageId]' --output text)
-
+  if [ -z "$ami_id" ]; then
+        error-exit "unable to find AMI ID for $ami_date"
+  fi
+  
   cat <<EOF
 $(get_spec $infraID $az $provider)
       providerSpec:


aravindhp (approved):

/approve

hack/machineset.sh
```diff
@@ -78,7 +78,7 @@ get_aws_ms() {
   local provider=$4
 
   # get the AMI id for the Windows VM
-  ami_date="2020.09.09"
+  ami_date="2020.10.14"
```
aravindhp, line 81: Please add a check after `ami_id` is populated to throw an error if `ami_id` is empty,

aravindhp (changes requested):

hack/machineset.sh
mansikulkarni96: updated.

hack/machineset.sh
```diff
@@ -78,9 +78,12 @@ get_aws_ms() {
...
   ami_id=$(aws ec2 describe-images --filters Name=name,Values=Windows_Server-2019-English-Full-ContainersLatest-${ami_date} --region ${region} --query 'Images[*].[ImageId]' --output text)
-
+  if [ -z "$ami_id" ]; then
+        error-exit "ami_id not set"
```
aravindhp, line 84: A more useful error would be `unable to find AMI ID for $amd_date`. WDYT?

mansikulkarni96: sounds good to me.

aravindhp (approved):
Thanks for fixing this @mansikulkarni96 

/approve

PR #262, opened 2021-01-25
WINC-565: [wmco] Relax k8s version strict dependency

VaishnaviHire:
This commit ensures that we no longer are tied to exact version
 OpenShift. The changes allow WMCO to run on N and N+1 versions of
 OpenShift. This is required to perform seamless upgrade between
N to N+1 versions.

### DIFF
--- a/pkg/cluster/config.go
+++ b/pkg/cluster/config.go
@@ -4,12 +4,14 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"strconv"
 	"strings"
 
 	oconfig "github.com/openshift/api/config/v1"
 	configclient "github.com/openshift/client-go/config/clientset/versioned"
 	operatorv1 "github.com/openshift/client-go/operator/clientset/versioned/typed/operator/v1"
 	"github.com/pkg/errors"
+	"golang.org/x/mod/semver"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/rest"
 )
@@ -18,7 +20,7 @@ const (
 	ovnKubernetesNetwork = "OVNKubernetes"
 	// baseK8sVersion specifies the base k8s version supported by the operator. (For eg. All versions in the format
 	// 1.20.x are supported for baseK8sVersion 1.20)
-	baseK8sVersion = "1.20"
+	baseK8sVersion = "v1.20"
 )
 
 // Network interface contains methods to interact with cluster network objects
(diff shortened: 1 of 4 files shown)


pkg/cluster/config_test.go
```diff
@@ -124,9 +124,12 @@ func TestIsValidKubernetesVersion(t *testing.T) {
...
 		errorMessage string
 	}{
-		{"cluster version lower than supported version ", "v1.17.1", "Unsupported server version: v1.17.1. Supported version is v1.20.x"},
+		{"cluster version lower than supported version ", "v1.17.1", "Unsupported server version: " +
+			"v1.17.1. Supported versions are v1.20.x to v1.21.x"},
 		{"cluster version equals supported version", "v1.20.0", ""},
-		{"cluster version greater than supported version", "v1.21.0", "Unsupported server version: v1.21.0. Supported version is v1.20.x"},
+		{"cluster version equals supported version", "v1.21.4", ""},
+		{"cluster version greater than supported version ", "v1.22.2", "Unsupported server version: " +
+			"v1.22.2. Supported versions are v1.20.x to v1.21.x"},
```
sebsoto, line 132: I dont see a reason to check the exact error message, to me all it adds is extra work when bumping the expected kube version. Instead we should check if there was an error or not, since that is what we care about. Since you are changing the function/test I think it would be a good time to make this change.

pkg/cluster/config.go
```diff
@@ -107,23 +109,28 @@ func NewConfig(restConfig *rest.Config) (Config, error) {
...
-	k8sVersion := strings.TrimLeft(versionInfo.GitVersion, "v")
-	clusterBaseVersion := strings.Join(strings.SplitN(k8sVersion, ".", 3)[:2], ".")
-
-	if strings.Compare(clusterBaseVersion, baseK8sVersion) != 0 {
-		return errors.Errorf("Unsupported server version: v%v. Supported version is v%v.x", k8sVersion,
-			baseK8sVersion)
+	// split the version in the form Major.Minor. For e.g v1.18.0-rc.1 -> v1.18
+	clusterBaseVersion := semver.MajorMinor(versionInfo.GitVersion)
+	// Convert base version to float and add 1 to Minor version
+	baseVersion, err := strconv.ParseFloat(strings.TrimPrefix(baseK8sVersion, "v"), 64)
```
sebsoto, line 124: You need to handle the error here

pkg/cluster/config.go
```diff
@@ -4,12 +4,14 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"strconv"
 	"strings"
 
 	oconfig "github.com/openshift/api/config/v1"
 	configclient "github.com/openshift/client-go/config/clientset/versioned"
 	operatorv1 "github.com/openshift/client-go/operator/clientset/versioned/typed/operator/v1"
 	"github.com/pkg/errors"
```
sebsoto, line 13: We should really consider using the [stdlib errors package](https://golang.org/pkg/errors/) instead of this one

sebsoto (changes requested):

pkg/cluster/config.go
sebsoto: not a problem for you to solve of course :smile:

VaishnaviHire: I could do it in a separate commit

pkg/cluster/config.go
```diff
@@ -107,23 +109,28 @@ func NewConfig(restConfig *rest.Config) (Config, error) {
 	}, nil
 }
 
-// validateK8sVersion checks for valid k8s version in the cluster. It returns an error for all versions not equal
-// to supported major version. This is being done this way, and not by directly getting the cluster version, as OpenShift CI
-// returns version in the format 0.0.x and not the actual version attached to its clusters.
+// validateK8sVersion checks for valid k8s version in the cluster. It returns an error for all versions that are not in
+// range of given base version(x.y.z) and x.y+1.z version. This is being done this way, and not by directly getting
+// the cluster version, as OpenShift CI returns version in the format 0.0.x and not the actual version attached to its
+// clusters.
```
aravindhp, line 115: > OpenShift CI returns version in the format 0.0.x and not the actual version attached to its clusters.

This is no longer true. I suggest we remove this from the comment. On that note: should we start using OpenShift versions now?

pkg/cluster/config.go
sebsoto: The wrapping part is just `fmt.Errorf("%w: ...", err, ...))`

But the reasoning would be code purity and reducing the amount of imports. No functional improvements as far as I'm aware. In the future language improvements might make that case though.

aravindhp (changes requested):

pkg/cluster/config.go
VaishnaviHire: I will remove the comment here. 

As discussed in the standup:
The consensus was to start using `OpenShift versions` since we are releasing our operator for particular OpenShift versions. There is a [story](https://issues.redhat.com/browse/WINC-518)  regarding this, so the changes can be made when we decide to undertake it.

pkg/cluster/config.go
VaishnaviHire: This needs to be done as part of separate PR since it include changes in almost every file in `/pkg` and `/test` modules

pkg/cluster/config.go
```diff
@@ -107,23 +109,29 @@ func NewConfig(restConfig *rest.Config) (Config, error) {
...
 	}
-	// split the version in the form Major.Minor. For e.g v1.18.0-rc.1 -> 1.18
-	k8sVersion := strings.TrimLeft(versionInfo.GitVersion, "v")
-	clusterBaseVersion := strings.Join(strings.SplitN(k8sVersion, ".", 3)[:2], ".")
+	// split the version in the form Major.Minor. For e.g v1.18.0-rc.1 -> v1.18
+	clusterBaseVersion := semver.MajorMinor(versionInfo.GitVersion)
+	// Convert base version to float and add 1 to Minor version
+	baseVersion, err := strconv.ParseFloat(strings.TrimPrefix(baseK8sVersion, "v"), 64)
+	if err != nil {
+		return errors.Wrap(err, "error converting kube MajorMinor version to float")
```
aravindhp, line 124: `major.minor`. Moreover it will help if you mention the `baseK8sVersion` in the error. i.e. `error converting k8s version 1.20 to float`.

aravindhp (changes requested):

aravindhp (approved):
/approve

nit: we should stick to `k8s` and not `K8s`.

aravindhp (approved):
/approve

nit: use `k8s` rather than `K8s`.

mansikulkarni96:
LGTM

PR #263, opened 2021-01-26
WINC-565: [release-4.6] [wmco] Relax k8s version strict dependency

VaishnaviHire:
This commit ensures that we no longer are tied to exact version
OpenShift. The changes allow WMCO to run on N and N+1 versions of
OpenShift. This is required to perform seamless upgrade between
4.6 to 4.7 OpenShift version.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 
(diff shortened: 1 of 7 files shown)


aravindhp (approved):
/approve

nit: we should stick to k8s and not K8s.

ravisantoshgudimetla (approved):
/lgtm

Thank you for working on this @VaishnaviHire

PR #264, opened 2021-01-27
[community-4.7] Explicitly set GOBIN path for promu installation

mansikulkarni96:
This commit explicitly sets the GOBIN path
to specify where the promu binary should be
installed, promu installed in the windows_exporter
directory is then utilized for its installation.
This change was necessary because cachito changes
the GOPATH during image build and if the path is
not specified promu is installed to some location
from which it is not accessible to build
windows_exporter executable.

(cherry picked from commit 6f65fced396dcfb17c73a97aebd0119740c85cb6)


aravindhp (approved):
/approve

PR #266, opened 2021-01-27
[deploy] Add skipRange to csv for 2.0.0 release

mansikulkarni96:
This PR adds the skipRange option and value `">=1.0.0 <2.0.0" ` to CSV for the 2.0.0 release. Since the 2.0.0 operator
is not in the same index as 1.x operator, we need to use 'skipRange' instead of the 'replaces' field for the upgrade process.


aravindhp (approved):
/approve

PR #268, opened 2021-01-28
[hack] Add script to update submodules

sebsoto:


hack/update-submodule.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
```
aravindhp, line 1: Please add `set -euo pipefail`. We should be doing this for all our scripts.

hack/update-submodule.sh
```diff
@@ -0,0 +1,55 @@
...
+    echo "Must be ran in repo root directory"
+    echo ""
+    echo "Options:"
+    echo "-h   Shows usage text"
+    echo "Example:"
+    echo "update_submodules.sh master release-4.6"
+}
+
+function update_branch_submodules() {
+    submodules=(kubernetes ovn-kubernetes windows-machine-config-bootstrapper windows_exporter containernetworking-plugins)
```
aravindhp, line 17: Can't you figure this out? Otherwise we would have to keep updating this script.
`git config --file .gitmodules --get-regexp path | awk '{ print $2 }'`

hack/update-submodule.sh
```diff
@@ -0,0 +1,55 @@
...
+        echo "--- Updating $submodule ---"
+        cd $top_dir
+        cd $submodule \
+        && origin_url=$(git remote get-url origin) \
+        && git fetch origin > /dev/null \
+        && git checkout origin/$branch \
+        && short_head=$(git rev-parse --short HEAD) \
+        && long_head=$(git rev-parse HEAD) \
+        && cd .. \
+        && git add $submodule \
```
aravindhp, line 35: Can't we just do `git submodule update --remote --merge/rebasse` or something along those lines? This seems overly complicated.

aravindhp (changes requested):
Thanks for doing this @sebsoto. Please address my comments.

hack/update-submodule.sh
aravindhp: I also don't understand the need for `&&` instead of individual commands.

mansikulkarni96:
LGTM

hack/update-submodule.sh
```diff
@@ -0,0 +1,54 @@
...
+
+function update_branch_submodules() {
+    branch=$1
+    working_branch_name="$branch-SubmoduleUpdate$(date +%m-%d)"
+    echo Using branch $working_branch_name
+    git branch -D $working_branch_name || true
+    git checkout $branch -b $working_branch_name
+    git submodule update --remote
+
+    # The submodules have been updated at this point, now a commit needs to be generated for each of them
```
aravindhp, line 26: Should we trying to update all of them at the same time? How is this going to work with `windows_exporter`?

aravindhp (changes requested):
Thanks for working, on this @sebsoto.

hack/update-submodule.sh
```diff
@@ -0,0 +1,54 @@
...
+    echo ""
+    echo "Options:"
+    echo "-h   Shows usage text"
+    echo "Example:"
+    echo "update_submodules.sh master release-4.6"
+}
+
+function update_branch_submodules() {
+    branch=$1
+    working_branch_name="$branch-SubmoduleUpdate$(date +%m-%d)"
```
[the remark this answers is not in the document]
sebsoto: That is creating a unique name of a branch in the format `master-SubmoduleUpdate03-25`

hack/update-submodule.sh
```diff
@@ -0,0 +1,54 @@
...
+    echo Using branch $working_branch_name
+    git branch -D $working_branch_name || true
```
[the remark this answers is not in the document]
sebsoto: It's deleting the branch if it already exists, say someone was updating the submodules twice in one day for whatever reason. Because the name is very specific there should never be a case where someone has a branch with said name that they care about for any reason other than updating the submodules

Its not a pipe, but an `or`, this is because if the branch does not exist the command will fail causing the script to exit. This needs to be done because of the addition of ` set -euo pipefail`

hack/update-submodule.sh
sebsoto: I dont see a problem with updating them all at the same time.

I'm not sure what the issue is with `windows_exporter`? Each submodule should have its proper remote set up in the config.

hack/update-submodule.sh
```diff
@@ -0,0 +1,54 @@
...
+    # The submodules have been updated at this point, now a commit needs to be generated for each of them
+    git config --file .gitmodules --get-regexp path | awk '{ print $2 }' | while read submodule; do
+        cd $submodule
+        origin_url=$(git remote get-url origin)
+        short_head=$(git rev-parse --short HEAD)
+        long_head=$(git rev-parse HEAD)
+        cd ..
+        git add $submodule
+        # Commit changes if there are any
+        git commit -m "[submodule][$submodule] Bump to $short_head" -m "Bump to $origin_url/commit/$long_head" -m "This commit was generated using hack/update_submodules.sh" || true
```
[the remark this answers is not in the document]
sebsoto: If there is nothing to commit then `git commit` will exit 1,
` # Commit changes if there are any`  was meant to express that this won't always make a commit.

hack/update-submodule.sh
```diff
@@ -0,0 +1,54 @@
...
+    echo "Options:"
+    echo "-h   Shows usage text"
+    echo "Example:"
+    echo "update_submodules.sh master release-4.6"
+}
+
+function update_branch_submodules() {
+    branch=$1
+    working_branch_name="$branch-SubmoduleUpdate$(date +%m-%d)"
+    echo Using branch $working_branch_name
```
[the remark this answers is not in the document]
sebsoto: The new branch is created from the branch you want updated: `git checkout $branch -b $working_branch_name`, this means the branch in question's `.gitmodules` file has the correct contents and the submodules will be updated from the correct remotes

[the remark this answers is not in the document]
aravindhp: I don't follow why you are creating a branch here. Aren't you just checking out an existing branch. I feel I am missing something basic you are doing in your workflow.

PS: I had to explicitly run `git config -f .gitmodules submodule.$submodule.branch $branch` to switch branches in `.gitmodules` or update the file by hand.

hack/update-submodule.sh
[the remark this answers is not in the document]
aravindhp: >It's deleting the branch if it already exists, say someone was updating the submodules twice in one day for whatever reason. Because the name is very specific there should never be a case where someone has a branch with said name that they care about for any reason other than updating the submodules

I again don't follow the need to delete. Won't we just update the branch in that case?

>Its not a pipe, but an or, this is because if the branch does not exist the command will fail causing the script to exit. This needs to be done because of the addition of set -euo pipefail

Sorry I missed the second `|`.
What if the command fails because of other reasons? It will then be a false positive and can lead to a bad commit.
Also shouldn't we exit if the branch does not exist?

hack/update-submodule.sh
aravindhp: `windows_exporter` only has a `master` branch today. So what does the script do today?

In general I feel there should updating one submodule or all submodules should be an option. Take the instance we just want to update one submodule because of a bug fix and don't want the other submodules updated.

hack/update-submodule.sh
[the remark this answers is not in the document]
aravindhp: What if the command fails because of other reasons? It will then be a false positive.

hack/update-submodule.sh
```diff
@@ -0,0 +1,92 @@
+#!/bin/bash
+
+# Update submodules HEAD to remote branches
+
+set -euo pipefail
+
+function help() {
+    echo "Usage: update_submodules.sh [OPTIONS]... branch_name..."
```
aravindhp, line 8: Please remove `...`. They can be misleading.

hack/update-submodule.sh
```diff
@@ -0,0 +1,92 @@
...
+    echo "Update submodules HEAD to remote branches"
+    echo "Must be ran in repo root directory"
```
aravindhp, line 10: Nit: `Must be executed in...`. `ran` sounds a bit strange.

hack/update-submodule.sh
```diff
@@ -0,0 +1,92 @@
...
+    local modulelist=$3
+
+    if git branch -D $new_branch; then
+	echo "Deleted branch $new_branch"
+    fi
+    git checkout $base_branch -b $new_branch
+
+    # Generate a commit updating each submodule
+    for submodule in $modulelist; do
+        git submodule update --remote $submodule
```
aravindhp, line 48: Did you confirm that `.gitmodules` gets updated if you are changing branches for the submodules?

hack/update-submodule.sh
```diff
@@ -0,0 +1,92 @@
...
+        generate_submodule_commit $submodule
+    done
+}
+
+modulelist=""
+while getopts "m:h?" opt; do
+    case "$opt" in
+    m) modulelist=$OPTARG;;
+    h|\?) help; exit 0;;
+    esac
```
aravindhp, line 58: Please be consistent about indenting. In some places you use 2, in other 4 and then 8. I think the majority of the usage in other scripts has been 2. So lets stick to that.

hack/update-submodule.sh
```diff
@@ -0,0 +1,92 @@
...
+done
+shift $((OPTIND -1))
+
+# If the -m option wasnt given, update all modules
+if [ -z "$modulelist" ]; then
+        modulelist=$(git config --file .gitmodules --get-regexp path | awk '{ print $2 }')
```
aravindhp, line 64: 8 spaces here...

hack/update-submodule.sh
```diff
@@ -0,0 +1,92 @@
...
+fi
+
+# Return if the working branch is dirty
+if ! git diff --quiet; then
+  echo "branch dirty, exiting to not overwrite work"
```
aravindhp, line 69: vs 2

hack/update-submodule.sh
```diff
@@ -0,0 +1,92 @@
...
+  exit 1
+fi
+
+# For each branch that the user gives (master, release-4.7), create a new branch based off it, with commits updating the submodules
+initial_branch=$(git branch --show-current)
+new_branch_list=()
+for branch in "$@"; do
+    new_branch="$branch-submodule-update-$(date +%m-%d)"
```
aravindhp, line 77: vs 4

hack/update-submodule.sh
```diff
@@ -0,0 +1,92 @@
...
+        modulelist=$(git config --file .gitmodules --get-regexp path | awk '{ print $2 }')
+fi
+
+# Return if the working branch is dirty
+if ! git diff --quiet; then
+  echo "branch dirty, exiting to not overwrite work"
+  exit 1
+fi
+
+# For each branch that the user gives (master, release-4.7), create a new branch based off it, with commits updating the submodules
```
aravindhp, line 73: Please stick to 120 cols

hack/update-submodule.sh
```diff
@@ -0,0 +1,92 @@
...
+# Return to initial branch
+git checkout $initial_branch
+echo "****"
+echo "New branches created:"
+echo "    [${new_branch_list[@]}]"
+echo ""
+echo "For each branch, you may push to your fork and create a PR against openshift/windows-machine-config-operator"
+echo "example:"
+echo "# assumes you have your fork set as the remote 'origin'"
+echo "git push origin ${new_branch_list[0]}"
```
aravindhp, line 91: Will this push to all the branches?

hack/update-submodule.sh
```diff
@@ -0,0 +1,92 @@
...
+function generate_submodule_commit() {
+    local submodule=$1
+    cd $submodule
+    origin_url=$(git remote get-url origin)
+    short_head=$(git rev-parse --short HEAD)
+    long_head=$(git rev-parse HEAD)
+    cd ..
+    git add $submodule
+    # Commit changes if there are any
+    if git commit -m "[submodule][$submodule] Update to $short_head" -m "Bump to $origin_url/commit/$long_head" -m "This commit was generated using hack/update_submodules.sh"; then
```
aravindhp, line 30: Please use `update` instead of `bump`

hack/update-submodule.sh
```diff
@@ -0,0 +1,92 @@
...
+    echo "Example:"
+    echo "update_submodules.sh master release-4.6"
+    echo "update_submodules.sh master -m \"kubernetes ovn-kubernetes\""
+}
+
+# Stages changes to the given submodule and makes a commit including those changes
+function generate_submodule_commit() {
+    local submodule=$1
+    cd $submodule
+    origin_url=$(git remote get-url origin)
```
aravindhp, line 24: How about `submodule_url` to reduce confusion?

aravindhp (changes requested):

hack/update-submodule.sh
sebsoto: Yes i did

hack/update-submodule.sh
sebsoto: No this is just giving an example for one of the brances

hack/update-submodule.sh
aravindhp: I am not sure how that happens. Say the `kubernetes` submodule tracks `release-4.8` and I now want it to track `release-4.9`, how do I go about doing that?

aravindhp (approved):
/approve

Thanks for patiently working through this, @sebsoto. This is going to make updating submodules really simple.

aravindhp (changes requested):

aravindhp (approved):
/approve

hack/update_submodules.sh
```diff
@@ -0,0 +1,119 @@
...
+
+set -euo pipefail
+
+function help() {
+  echo "Usage: update_submodules.sh [OPTIONS] branch_name"
+  echo "Update submodules HEAD to remote branches"
+  echo "Must be executed in repo root directory"
+  echo ""
+  echo "Options:"
+  echo "-m   Specific modules to update, separated by spaces. Defaults to all submodules"
```
mansikulkarni96, line 13: specify*

mansikulkarni96 (commented):
Thanks for adding this @sebsoto lgtm except a nit

hack/update_submodules.sh
sebsoto: I dont really agree with that, what I put is grammatically correct and describes the value given to the arguments, as opposed to the action of using the flag.

aravindhp:
override ci/prow/azure-e2e-operator

PR #269, opened 2021-01-28
[master] Bump submodules

sebsoto:
Bumps kubernetes and ovn-kubernetes submodules


aravindhp (approved):
/approve

Is there a reason for the difference in hash length across commits?

sebsoto:
@aravindhp https://git-scm.com/docs/git-rev-parse#Documentation/git-rev-parse.txt---shortlength

The minimum length is 4, the default is the effective value of the core.abbrev configuration variable (see git-config[1]). This is based on the submodule's config

mansikulkarni96:
Thanks for doing this @sebsoto 
/lgtm

PR #274, opened 2021-02-03
[wmco] Remove constant for metrics resources name 

VaishnaviHire:
This commit removes the constant used as name for  metrics resources.
The name for metrics resources is set by 'metrics' package of operator-sdk
and is equivalent to the name of the metrics service. The changes include
retrieving the name set by the 'metrics' package and using it for accessing
the resources.

### DIFF
--- a/pkg/controller/windowsmachine/metrics/metrics.go
+++ b/pkg/controller/windowsmachine/metrics/metrics.go
@@ -25,11 +25,11 @@ var (
 	log = logf.Log.WithName("metrics")
 	// metricsEnabled specifies if metrics are enabled in the current cluster
 	metricsEnabled = true
+	// windowsMetricsResource is the name of an object created for Windows metrics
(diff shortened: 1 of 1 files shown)


sebsoto (approved):

pkg/controller/windowsmachine/metrics/metrics.go
```diff
@@ -84,6 +84,9 @@ func Add(ctx context.Context, cfg *rest.Config, namespace string) error {
 		return errors.Wrap(err, "could not create metrics Service")
 	}
 
+	// the name for the metrics resources is set during creation of metrics service and is equivalent to the service name
+	windowsMetricsResource = service.Name
```
sebsoto, line 88: Does service have a `GetName()` method that "should" be used?

aravindhp (approved):
Thanks for fixing this, @VaishnaviHire 
/lgtm

PR #279, opened 2021-02-05
WINC-551: Cluster autoscaling with Windows machines

saifshaikh48:
This PR adds a README section documenting how to enable
autoscaling of Windows MachineSets managed by WMCO.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,15 @@ WMCO is not responsible for Windows operating system updates. The cluster admini
 creating the VMs and hence, the cluster administrator is responsible for providing an updated image. The cluster 
 administrator can provide an updated image by changing the image in the MachineSet spec.
 
+## Enabled features
+
+### Autoscaling Windows nodes
+Cluster autoscaling is supported for Windows instances. 
+
+- Define and deploy a [ClusterAutoscaler](https://docs.openshift.com/container-platform/latest/machine_management/applying-autoscaling.html#configuring-clusterautoscaler).
+- Create a Windows node through a MachineSet (see spec in [Usage section](https://github.com/openshift/windows-machine-config-operator#usage)).
+- Define and deploy a [MachineAutoscaler](https://docs.openshift.com/container-platform/latest/machine_management/applying-autoscaling.html#configuring-machineautoscaler), referencing a Windows MachineSet.
+
 ## Development
 
 See [HACKING.md](docs/HACKING.md).


aravindhp:
/cherry-pick community-4.7

Just to check if the e2e tests are passing.

README.md
```diff
@@ -146,6 +146,34 @@ WMCO is not responsible for Windows operating system updates. The cluster admini
...
+## Applying autoscaling to WMCO-managed Windows Machines
+
+There are two components to managing resources using autoscaling: a ClusterAutoscaler and MachineAutoscalers.
+The cluster autoscaler is capable of automatically adjust the size of the cluster based on node resource usage
+and overall deployment needs, within specific cluster constraints. Machine autoscalers modify machine sets
+to scale up and down the number of machines.
+Multiple MachineAutoscalers can be deployed to manage multiple MachineSets.
+Each machine autoscaler object should be mapped to a single machine set.
+
+- Define and deploy a [ClusterAutoscaler](https://docs.openshift.com/container-platform/4.6/machine_management/applying-autoscaling.html#configuring-clusterautoscaler).
```
jrvaldes, line 158: for my own education; does it only applies to version `4.6`  ? 

Otherwise, `latest` may help. See:  https://docs.openshift.com/container-platform/latest/machine_management/applying-autoscaling.html#configuring-clusterautoscaler

saifshaikh48: Nope, you are correct -- also applies to 4.7/4.8. I'll update to latest.

README.md
```diff
@@ -146,6 +146,34 @@ WMCO is not responsible for Windows operating system updates. The cluster admini
...
+- Ensure a MachineSet resource is deployed for Windows machines (see spec in [Usage section](https://github.com/openshift/windows-machine-config-operator#usage)).
+- Create a MachineAutoscaler resource, referencing a MachineSet resource. Replace *\<windows-machineset-name\>*
```
mansikulkarni96, line 160: <windows-machineset-name\> does not appear to be in code format in the rich-diff.

README.md
```diff
@@ -146,6 +146,34 @@ WMCO is not responsible for Windows operating system updates. The cluster admini
...
+The cluster autoscaler is capable of automatically adjusting the size of the cluster based on node resource usage
+and overall deployment needs, within specified constraints. Machine autoscalers modify machine sets
+to scale up and down the number of machines.
+Multiple MachineAutoscalers can be deployed to manage multiple MachineSets.
+Each machine autoscaler object should be mapped to a single machine set.
+
+- Define and deploy a [ClusterAutoscaler](https://docs.openshift.com/container-platform/latest/machine_management/applying-autoscaling.html#configuring-clusterautoscaler).
+- Ensure a MachineSet resource is deployed for Windows machines (see spec in [Usage section](https://github.com/openshift/windows-machine-config-operator#usage)).
+- Create a MachineAutoscaler resource, referencing a MachineSet resource. Replace *\<windows-machineset-name\>*
+  in the template below with the name of Windows MachineSet that was created, then deploy.
```
mansikulkarni96, line 161: maybe rephrase to specifics -> *name of Windows MachineSet that was created in the previous step, then deploy the MachineAutoscaler resource.

mansikulkarni96 (commented):
LGTM, just some nits to consider.

README.md
```diff
@@ -146,6 +146,34 @@ WMCO is not responsible for Windows operating system updates. The cluster admini
 creating the VMs and hence, the cluster administrator is responsible for providing an updated image. The cluster 
 administrator can provide an updated image by changing the image in the MachineSet spec.
 
+## Applying autoscaling to WMCO-managed Windows Machines
```
sebsoto, line 149: How does this process differ from Linux Machines?

If it doesn't it might make more sense to just point to the proper docs so we have less documentation to maintain

README.md
saifshaikh48: Alright, I will change from _italicized_ to `code format`

README.md
saifshaikh48: Great point... there is no difference between them. In that case would we want a short section that just points to the official docs, or delete this PR altogether?

sebsoto: I think that comes down to the reason this PR is being created. Is this something that we feel we need to convey to customers? Basically, is this a common question that we are getting? If so lets have a short section pointing to the official docs. If not? The section might not be needed. But what about the case where we have Windows only customers who are using only using this readme/the csv for MachineSet configuration? A short section makes sense. All that considered I think having a short section is better choice.

The other thing we have to think about is whether or not we are properly covering this in our tests. What is the underlying mechanism that autoscaling needs? Is it `oc adm top`? If so maybe the better way to approach this is adding an e2e test for interacting with a Windows node with `oc adm top`. If we arent testing this can we really promote this functionality?

aravindhp:
>If so we might need to add an e2e test for interacting with a Windows node with oc adm top. If we arent testing this can we really promote this functionality?

While I am fine adding a test to see if `oc adm top` works, it is not tenable to for us to test very k8s feature out there in our e2e tests.

README.md
saifshaikh48: Moved forward with a short section calling out our support + OpenShift docs.

Also, clarified the underlying mechanism [on the story](https://issues.redhat.com/browse/WINC-551?focusedCommentId=16205716&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16205716).

test/e2e/metrics_test.go
```diff
@@ -287,3 +289,25 @@ func (tc *testContext) ensureMonitoringIsEnabled() error {
...
+	for _, winNode := range gc.nodes {
+		t.Run(winNode.Name, func(t *testing.T) {
+			cmd := exec.Command("oc", "adm", "top", "node", winNode.Name)
+			out, err := cmd.Output()
+			require.NoErrorf(t, err, "failed to get result of `%s`: %s", cmd, out)
+
+			// parse output for unreadable values. When metrics are not configured properly, values
+			// will be <unknown> as per https://bugzilla.redhat.com/show_bug.cgi?id=1920903
+			require.NotContains(t, string(out), "<unknown>")
+			assert.NoError(t, err, "could not retreive resource usage statistics for node %s", winNode.Name)
```
sebsoto, line 310: Looks like you're checking the same error twice

Instead try:
```
require.NoErrorf(t, err, "whichever error msg you think fits better")
assert.NotContains(t, string(out), "<unknown>")

```

sebsoto (changes requested):
Thanks for making these changes @saifshaikh48

test/e2e/metrics_test.go
sebsoto: Duplicate check was:
`require.NoErrorf(t, err, "failed to get result of `%s`: %s", cmd, out)`

aravindhp:
/retitle WINC-551: Cluster autoscaling with Windows machines

test/e2e/metrics_test.go
```diff
@@ -30,6 +31,7 @@ func testMetrics(t *testing.T) {
 	t.Run("Windows Exporter configuration validation", testWindowsExporter)
 	t.Run("Prometheus configuration validation", testPrometheus)
 	t.Run("Windows Prometheus rules validation", testWindowsPrometheusRules)
+	t.Run("Windows Node resource usage statistics validation", testNodeResourceUsage)
```
aravindhp, line 34: I suggest making this `Windows Node resource usage info validation` given how you have described the test

test/e2e/metrics_test.go
```diff
@@ -287,3 +289,24 @@ func (tc *testContext) ensureMonitoringIsEnabled() error {
...
+func testNodeResourceUsage(t *testing.T) {
+	_, err := NewTestContext()
+	require.NoError(t, err)
+
+	// Need at least one Windows node to run these tests, throwing error if this condition is not met
+	require.Greater(t, len(gc.nodes), 0, "test requires at least one Windows node to run")
+
+	for _, winNode := range gc.nodes {
+		t.Run(winNode.Name, func(t *testing.T) {
+			cmd := exec.Command("oc", "adm", "top", "node", winNode.Name)
```
aravindhp, line 303: Should we make this an API call?

test/e2e/metrics_test.go
```diff
@@ -7,6 +7,7 @@ import (
 	"fmt"
```
aravindhp, line 7: Given the test has nothing to do with cluster autoscaling, this should be a separate PR

test/e2e/metrics_test.go
```diff
@@ -287,3 +289,24 @@ func (tc *testContext) ensureMonitoringIsEnabled() error {
...
+
+	for _, winNode := range gc.nodes {
+		t.Run(winNode.Name, func(t *testing.T) {
+			cmd := exec.Command("oc", "adm", "top", "node", winNode.Name)
+			out, err := cmd.Output()
+			require.NoError(t, err, "could not retreive resource usage statistics for node %s", winNode.Name)
+
+			// parse output for unreadable values. When metrics are not configured properly, values
+			// will be <unknown> as per https://bugzilla.redhat.com/show_bug.cgi?id=1920903
+			assert.NotContains(t, string(out), "<unknown>")
```
aravindhp, line 309: Is this the only error output that is possible?

aravindhp (changes requested):
Thanks for working on this @saifshaikh48. Please address my comments.

README.md
```diff
@@ -131,6 +131,14 @@ Usage:
 ./hack/machineset.sh apply/delete    # to create/delete MachineSet directly on cluster
 ```
 
+## Applying autoscaling to WMCO-managed Windows instances
+
+Cluster autoscaling is supported for Windows instances. 
+
+- Define and deploy a [ClusterAutoscaler](https://docs.openshift.com/container-platform/latest/machine_management/applying-autoscaling.html#configuring-clusterautoscaler).
+- Ensure a MachineSet resource is deployed for Windows machines (see spec in [Usage section](https://github.com/openshift/windows-machine-config-operator#usage)).
```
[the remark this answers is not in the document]
saifshaikh48: I wanted to emphasize that autoscaling requires a MachineSet be present. I will change this to `Create a Windows node through a MachineSet` and that way it will stay relevant when BYOH support is introduced.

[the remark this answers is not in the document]
aravindhp: I am fine with that.

test/e2e/metrics_test.go
saifshaikh48: This may be the way forward if we want to avoid 
> test very k8s feature out there in our e2e tests

Updated this test case in [a new PR](https://github.com/openshift/windows-machine-config-operator/pull/453). Please let me know if it is better.

aravindhp: Please remove this commit from this PR so that we can get it merged.

test/e2e/metrics_test.go
saifshaikh48: This is the only error output I (and QE) has seen. I think 0% usage would also be worrisome as _some_ resources are always used by kubelet etc.

test/e2e/metrics_test.go
saifshaikh48: On second thought, the test may be robust if we also included this logic/asserts as `oc adm top node` checks current resource usage stats and the API call checks allocatable resource stats. Both are useful to be checked in my opinion. Thoughts?

aravindhp: Like I said before, please remove this commit from this PR and move this discussion to #453

mansikulkarni96:
/lgtm

aravindhp (approved):
Thanks for working on this
/approve

README.md
```diff
@@ -146,6 +146,15 @@ WMCO is not responsible for Windows operating system updates. The cluster admini
 creating the VMs and hence, the cluster administrator is responsible for providing an updated image. The cluster 
 administrator can provide an updated image by changing the image in the MachineSet spec.
 
+## Enabled features
+
+### Autoscaling Windows nodes
```
aravindhp, line 151: @EricPonvelle is there a way to call out that this feature will be enabled for Windows nodes in our 2.0.1 release? Do you release note it? Or add it to the [Cluster Autoscaling section](https://docs.openshift.com/container-platform/4.7/machine_management/applying-autoscaling.html)? Anything works for us as long as it is called out.

/cc @Anandnatraj

README.md
```diff
@@ -146,6 +146,15 @@ WMCO is not responsible for Windows operating system updates. The cluster admini
...
+Cluster autoscaling is supported for Windows instances. 
+
+- Define and deploy a [ClusterAutoscaler](https://docs.openshift.com/container-platform/latest/machine_management/applying-autoscaling.html#configuring-clusterautoscaler).
+- Create a Windows node through a MachineSet (see spec in [Usage section](https://github.com/openshift/windows-machine-config-operator#usage)).
+- Define and deploy a [MachineAutoscaler](https://docs.openshift.com/container-platform/latest/machine_management/applying-autoscaling.html#configuring-machineautoscaler), referencing a Windows MachineSet.
```
EricPonvelle, line 156: With this items, should I include these sections directly in the OpenShift docs version of this PR? Just want to make sure I have a full grasp of how to use this material correctly.

EricPonvelle (commented):
lgtm - I added a few comments, more for my clarification.

I was also curious about what versions of OCP that these changes apply. Thanks for this!

README.md
aravindhp: I don't think so given that there is nothing different for Windows here when compared to Linux. These steps are more for anyone trying to use the operator directly from the repo.

@saifshaikh48 please correct me if I am wrong.

PR #289, opened 2021-02-08
WINC-574: [community-4.7][ci] Introduce retries for WMCO deployment

sebsoto:
This commit makes it so the e2e hack script retries multiple times in
case there is an transient issue when deploying the WMCO. Recently we
have been seeing an issue with Azure clusters going through an etcd
leader election soon after install. This change is being made as a
reaction to that, so that we retry WMCO deployment instead of having to
retry the entire test, including cluster provision.

(cherry picked from commit 1f840a658ff62e5252245d825ae8c5b14ec9e00d)


aravindhp (approved):
/approve

PR #290, opened 2021-02-09
[docs] Remove AWS describe-image command 

alinaryan:
This commit removes the AWS describe-image command 
from the docs as we will be changing the script in a follow-up 
PR to grab the latest tag for getting the ami since the date is 
constantly being changed by AWS. It also updates the
ami_date in the hackscript from 2020.10.14 to 2021.01.13 
to be used in the meantime, since the earlier date is no longer 
available in AWS regions.

### DIFF
--- a/docs/machineset-aws.md
+++ b/docs/machineset-aws.md
@@ -1,14 +1,7 @@
 # Creating an AWS Windows MachineSet
 
 _\<windows_container_ami\>_ should be replaced with the AMI ID of a Windows image with a container run-time installed. 
- You must use Windows Server 2019 with a version 10.0.17763.1457 or earlier. Run the following command to list AWS image info:
-  ```shell script
-  $ aws ec2 describe-images \
-    --filters Name=name,Values=Windows_Server-2019-English-Full-ContainersLatest-2020.09.09 \
-    --region <region> \
-    --query 'Images[*].[ImageId]' \
-    --output=json | jq .[0][0]
-```
+ You must use Windows Server 2019 with a version 10.0.17763.1457 or earlier.
                                                                            
 _\<infrastructureID\>_ should be replaced with the output of:
 ```shell script
--- a/hack/machineset.sh
+++ b/hack/machineset.sh
@@ -78,7 +78,7 @@ get_aws_ms() {
   local provider=$4
 
   # get the AMI id for the Windows VM
-  ami_date="2020.10.14"
+  ami_date="2021.01.13"
   ami_id=$(aws ec2 describe-images --filters Name=name,Values=Windows_Server-2019-English-Full-ContainersLatest-${ami_date} --region ${region} --query 'Images[*].[ImageId]' --output text)
   if [ -z "$ami_id" ]; then
         error-exit "unable to find AMI ID for $ami_date"


hack/machineset.sh
```diff
@@ -78,7 +78,7 @@ get_aws_ms() {
   local provider=$4
 
   # get the AMI id for the Windows VM
-  ami_date="2020.10.14"
+  ami_date="2021.01.13"
```
aravindhp, line 81: We are going to have to constantly update this. We wanted to hard code this for predictability but the fact that AWS will constantly change this, I suggest making this script smarter i.e. grab the latest `Windows_Server-2019-English-Full-ContainersLatest` image.

docs/machineset-aws.md
```diff
@@ -4,7 +4,7 @@ _\<windows_container_ami\>_ should be replaced with the AMI ID of a Windows imag
  You must use Windows Server 2019 with a version 10.0.17763.1457 or earlier. Run the following command to list AWS image info:
   ```shell script
   $ aws ec2 describe-images \
-    --filters Name=name,Values=Windows_Server-2019-English-Full-ContainersLatest-2020.09.09 \
+    --filters Name=name,Values=Windows_Server-2019-English-Full-ContainersLatest-2021.01.13 \
```
aravindhp, line 7: This means that we will be moving away from `10.0.17763.1457`. But it looks like we cannot pin to that kernel version anymore :-( @openshift/openshift-team-windows-containers how should we proceed?

@codyhoag are we mentioning this command anywhere in the OpenShift docs? Or was it just in the release notes?

aravindhp (commented):
Thanks for working on this, @alinaryan. PTAL at my comments.

VaishnaviHire (commented):
Thanks @alinaryan for the fix

hack/machineset.sh
aravindhp: Forgot to mention that this can be done in a follow up PR.

docs/machineset-aws.md
VaishnaviHire: Should we test newer kernel versions for the issue and see if it still persists?

aravindhp: Yes we should. But it should be of lower priority as of now.

codyhoag: @aravindhp this is only mentioned in the release notes as part of the [known issue](https://docs.openshift.com/container-platform/4.6/windows_containers/windows-containers-release-notes.html#wmco-1-0-1-known-issues).

aravindhp: @codyhoag I assume we don't need to retroactively keep updating that. So we should be fine from a doc perspective.

hack/machineset.sh
alinaryan: Ok, should I create a ticket for that and then create a follow up PR? And how should I proceed with this current PR?

docs/machineset-aws.md
aravindhp: @alinaryan please remove `Run the following command to list AWS image info:` and the shell script section.

hack/machineset.sh
aravindhp: >Ok, should I create a ticket for that and then create a follow up PR? 

Correct

>And how should I proceed with this current PR?

We can merge after you fix the doc section as mentioned in my previous comment.

aravindhp (approved):
Thanks for fixing this, @alinaryan 

/approve

PR #294, opened 2021-02-10
[community-4.6][wmco] Remove constant for metrics resources name

VaishnaviHire:
This commit removes the constant used as name for  metrics resources.
The name for metrics resources is set by 'metrics' package of operator-sdk
and is equivalent to the name of the metrics service. The changes include
retrieving the name set by the 'metrics' package and using it for accessing
the resources.

(cherry picked from commit 831e42db1a24264fd28bb6f38e780e576ea3d6ce)

### DIFF
--- a/pkg/controller/windowsmachine/metrics/metrics.go
+++ b/pkg/controller/windowsmachine/metrics/metrics.go
@@ -25,11 +25,11 @@ var (
 	log = logf.Log.WithName("metrics")
 	// metricsEnabled specifies if metrics are enabled in the current cluster
 	metricsEnabled = true
+	// windowsMetricsResource is the name of an object created for Windows metrics
+	windowsMetricsResource = ""
 )
 
 const (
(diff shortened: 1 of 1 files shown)


sebsoto:
/hold
I dont know if this should be merged into community-4.6.
I feel like only bug fixed should be merged there, but I'm not sure

aravindhp (approved):
/approve

VaishnaviHire:
> /hold
> I dont know if this should be merged into community-4.6.
> I feel like only bug fixed should be merged there, but I'm not sure

Since this is a fix, I feel it should be backported to community-4.6

aravindhp:
/hold cancel

Given we added the monitoring feature to community-4.6, we should back port this. It is a fix given hard coding the name can give rise to subtle problems if the metrics package changes and we decide to move to that. Plus this will make other backports easier.

PR #295, opened 2021-02-10
[release-4.7][test] Introduce CloudProvider.GetType()

aravindhp:
And use it to only run the `Changing the private key ensures all Windows nodes use the new private key` test on AWS. This cannot be run against a vSphere cluster because this random key is not part of the vSphere template image. Moreover this test is platform agnostic so is not needed to be run for every supported platform.

(cherry picked from commit 34cb58f1828825e54af9381730a3b00a28fe1453)

### DIFF
--- a/test/e2e/create_test.go
+++ b/test/e2e/create_test.go
@@ -7,6 +7,7 @@ import (
 	"testing"
 	"time"
 
+	config "github.com/openshift/api/config/v1"
 	mapi "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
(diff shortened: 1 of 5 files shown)


aravindhp:
/retitle [release-4.7][test] Introduce CloudProvider.GetType()

sebsoto:
a/pprove

aravindhp:
/retitle [release4.7][test] Fix e2e tests for vSphere

aravindhp:
/retitle [release-4.7][test] Fix e2e tests for vSphere

PR #296, opened 2021-02-10
[community-4.7] Add pubkey annotation to Windows nodes

sebsoto:
If the private key used to configure nodes changes, Machines configured
with the old key will be deleted, so that they can be replaced with new
Machines.

Backport of https://github.com/openshift/windows-machine-config-operator/pull/239


aravindhp (approved):
/approve

aravindhp:
/hold
 OKD 4.7 is broken

PR #302, opened 2021-02-12
 WINC-501: [wmco] Reorganize pkg directory for OSDK 1.0 

sebsoto:
This PR reorganizes the pkg directory to be more like what is
used in projects created by operator-sdk v1.0+. In 1.0, projects are
created with a `controllers/` directory with each controller existing
flat in that directory. All controllers are part of the same
`controllers` package.

This is being done in accordance with the directions given here:
https://master.sdk.operatorframework.io/docs/building-operators/golang/migration/#what-was-changed


aravindhp (approved):
/approve

aravindhp (changes requested):
/approve cancel

Moving directories out of `pkg` goes against [Go directory layout standards](https://github.com/golang-standards/project-layout#go-directories).

aravindhp (approved):
/approve

PR #312, opened 2021-02-17
Bug 1929579: [wmco] Fix issue with dereferencing nil signer

sebsoto:
This commit fixes an issue where a the `signer` of the Windows Machine
reconciler could be used when nil. This was not being caught in our e2e
tests because if the signer's value was set in a previous reconcile that
stale value would be used.

This is being fixed by moving where the signer is set to before its
usage, where it should have been in the first place.


aravindhp (approved):
/approve

mansikulkarni96:
/lgtm 
thanks for the quick fix!

PR #314, opened 2021-02-17
WINC-501: [ci] Remove operator-sdk framework from e2e suite

sebsoto:
This commit is being made in preparation to the switch to operator-sdk
1.x. In 1.x the test framework our tests used does not exist, so we must
stop using it. Because the framework is no longer used, the tests should
now be run using `go test`, instead of the operator-sdk binary.


test/e2e/clusterinfo/openshift.go
```diff
@@ -7,17 +7,23 @@ import (
...
+	routeClient "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
 	mapiClient "github.com/openshift/machine-api-operator/pkg/generated/clientset/versioned/typed/machine/v1beta1"
+	monitoringClient "github.com/prometheus-operator/prometheus-operator/pkg/client/versioned/typed/monitoring/v1"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kubeclient "k8s.io/client-go/kubernetes"
 	ctrlruntimecfg "sigs.k8s.io/controller-runtime/pkg/client/config"
 )
 
 // OpenShift is an Client struct which will be used for all OpenShift related functions to interact with the existing
 // Cluster.
```
aravindhp, line 19: Given you are in the area, please fix the comment:
`OpenShift contains clients for interacting with the OpenShift API`

test/e2e/create_test.go
```diff
@@ -103,19 +102,20 @@ func (tc *testContext) createWindowsMachineSet(replicas int32, windowsLabel bool
 	if err != nil {
 		return nil, err
 	}
-	return machineSet, framework.Global.Client.Create(context.TODO(), machineSet,
-		&framework.CleanupOptions{TestContext: tc.osdkTestCtx,
-			Timeout: cleanupTimeout, RetryInterval: cleanupRetryInterval})
+	return tc.client.MachineClient.MachineSets(clusterinfo.MachineAPINamespace).Create(context.TODO(), machineSet, metav1.CreateOptions{})
```
aravindhp, line 105: The `client` is repeated in the call sites here. I suggest removing `Client` from `Machine`. Applies to all clients.

test/e2e/delete_test.go
```diff
@@ -42,39 +35,34 @@ func testWindowsNodeDeletion(t *testing.T) {
...
 
 	// Cleanup all the MachineSets created by us.
-	for _, machineSetName := range e2eMachineSetNames {
-		machineSet := &mapi.MachineSet{}
-		err = framework.Global.Client.Get(context.TODO(), types.NamespacedName{Name: machineSetName,
-			Namespace: "openshift-machine-api"}, machineSet)
-		assert.NoError(t, framework.Global.Client.Delete(context.TODO(), machineSet))
+	for _, machineSet := range machineSetList.Items {
+		err = testCtx.deleteMachineSet(&machineSet)
+		assert.NoError(t, err, "error deleting Machine")
```
aravindhp, line 55: `error deleting MachineSet`

I don't see any point using the err variable here.

test/e2e/network_test.go
```diff
@@ -48,9 +48,8 @@ const (
 
 // testEastWestNetworking deploys Windows and Linux pods, and tests that the pods can communicate
 func testEastWestNetworking(t *testing.T) {
-	testCtx, err := NewTestContext(t)
+	testCtx, err := NewTestContext()
 	require.NoError(t, err)
-	defer testCtx.cleanup()
```
aravindhp, line 53: General question: have you verified that we are deleting all the resources we are creating? This becomes more important when we want to run e2e tests multiple times locally.

hack/run-ci-e2e-test.sh
```diff
@@ -115,7 +115,7 @@ done
 # -flag x is allowed for non-boolean flags only(https://golang.org/pkg/flag/)
 
 # Test that the operator is running when the private key secret is not present
-go test ./test/e2e/... -run=TestWMCO/operator_deployed_without_private_key_secret -v -args --node-count=$NODE_COUNT --private-key-path=$KUBE_SSH_KEY_PATH $WMCO_PATH_OPTION
+go test ./test/e2e/... -run=TestWMCO/operator_deployed_without_private_key_secret -v -timeout=5m -args --node-count=$NODE_COUNT --private-key-path=$KUBE_SSH_KEY_PATH $WMCO_PATH_OPTION
```
aravindhp, line 118: General comment: these times seem overly aggressive. Lets not introduce flakes because of this.

aravindhp (changes requested):
Thanks for working on this, @sebsoto. 

We keep using `[ci]` and `[test]` when modifying test code. I suggest we pick one. Lets go with the majority used in prior commits.

Also add the command you executed when updating the vendor directory to the commit message.

hack/run-ci-e2e-test.sh
aravindhp: I also think this should not be part of this PR.

test/e2e/delete_test.go
```diff
@@ -42,39 +34,33 @@ func testWindowsNodeDeletion(t *testing.T) {
...
+	require.NoError(t, err, "Windows node deletion failed")
 
 	// Cleanup all the MachineSets created by us.
-	for _, machineSetName := range e2eMachineSetNames {
-		machineSet := &mapi.MachineSet{}
-		err = framework.Global.Client.Get(context.TODO(), types.NamespacedName{Name: machineSetName,
-			Namespace: "openshift-machine-api"}, machineSet)
-		assert.NoError(t, framework.Global.Client.Delete(context.TODO(), machineSet))
+	for _, machineSet := range machineSetList.Items {
+		testCtx.deleteMachineSet(&machineSet)
```
mansikulkarni96, line 53: don't we need to assert for error here?

test/e2e/clusterinfo/openshift.go
```diff
@@ -7,17 +7,22 @@ import (
 	config "github.com/openshift/api/config/v1"
 	configClient "github.com/openshift/client-go/config/clientset/versioned"
 	operatorClient "github.com/openshift/client-go/operator/clientset/versioned/typed/operator/v1"
+	routeClient "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
 	mapiClient "github.com/openshift/machine-api-operator/pkg/generated/clientset/versioned/typed/machine/v1beta1"
+	monitoringClient "github.com/prometheus-operator/prometheus-operator/pkg/client/versioned/typed/monitoring/v1"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kubeclient "k8s.io/client-go/kubernetes"
```
aravindhp, line 14: I would rather we use `k8sclient` given `k8s` seems to be more prevalent than `kube`.

test/e2e/clusterinfo/openshift.go
```diff
@@ -7,17 +7,22 @@ import (
...
 type OpenShift struct {
-	ConfigClient   *configClient.Clientset
-	OperatorClient operatorClient.OperatorV1Interface
-	MachineClient  mapiClient.MachineV1beta1Interface
+	Config     *configClient.Clientset
+	Operator   operatorClient.OperatorV1Interface
+	Machine    mapiClient.MachineV1beta1Interface
+	Monitoring monitoringClient.MonitoringV1Interface
+	Route      routeClient.RouteV1Interface
+	Kube       kubeclient.Interface
```
aravindhp, line 25: `K8s` please

test/e2e/network_test.go
aravindhp: Ping on this

aravindhp (changes requested):
Mostly ready except for some naming.

test/e2e/delete_test.go
aravindhp: We should not fail the test because clean up failed. I suggest we log the error here.

aravindhp: Though such behavior change should not be done in this PR. This PR should be mechanical with no change in behavior or policy unless required for the SDK 1.0 migration.

sebsoto: I misread your previous suggestion @aravindhp https://github.com/openshift/windows-machine-config-operator/pull/314#discussion_r579431579
Thought you were saying that we shouldnt capture the error. I'll correct it

test/e2e/network_test.go
sebsoto: I think we are, we also make sure to delete the namespace the resources are created in. That really should be moved out of the `delete` tests, and into the main `TestWMCO` function, as its going to cause issues rerunning the tests locally, as if the test attempts to create the namespace and it already exists its going to throw an error.

sebsoto: I did have to add https://github.com/openshift/windows-machine-config-operator/pull/314/commits/746abd1de5fcd2a55a27cdf41516a9a7161a3382#diff-84675b994e150b2fe4128b3f0a171854d1060ba909b9ff0aeb8833c8d9f4dae9R163

aravindhp: OK, just making sure we are not leaving stale resources around mainly for local testing like you mentioned.

test/e2e/delete_test.go
```diff
@@ -42,39 +35,33 @@ func testWindowsNodeDeletion(t *testing.T) {
...
-	}
+	require.NoError(t, err, "Windows node deletion failed")
 
 	// Cleanup all the MachineSets created by us.
-	for _, machineSetName := range e2eMachineSetNames {
-		machineSet := &mapi.MachineSet{}
-		err = framework.Global.Client.Get(context.TODO(), types.NamespacedName{Name: machineSetName,
-			Namespace: "openshift-machine-api"}, machineSet)
-		assert.NoError(t, framework.Global.Client.Delete(context.TODO(), machineSet))
+	for _, machineSet := range machineSetList.Items {
```
aravindhp, line 53: This would delete all the `MachineSets` not just the ones created by us.

aravindhp (changes requested):

aravindhp (approved):
/approve

test/e2e/delete_test.go
```diff
@@ -22,59 +20,49 @@ func deletionTestSuite(t *testing.T) {
...
+	require.NoError(t, err, "Windows node deletion failed")
 
 	// Cleanup all the MachineSets created by us.
-	for _, machineSetName := range e2eMachineSetNames {
-		machineSet := &mapi.MachineSet{}
-		err = framework.Global.Client.Get(context.TODO(), types.NamespacedName{Name: machineSetName,
-			Namespace: "openshift-machine-api"}, machineSet)
-		assert.NoError(t, framework.Global.Client.Delete(context.TODO(), machineSet))
+	for _, machineSet := range e2eMachineSets {
+		assert.NoError(t, testCtx.deleteMachineSet(machineSet), "error deleting MachineSet")
```
mansikulkarni96, line 55: did we decide to log the error instead of failing here?

mansikulkarni96: resolved, we want to log such errors ideally, but it shouldn't be part of this PR.

mansikulkarni96:
/lgtm

PR #323, opened 2021-03-01
Bug 1930791: [wmco] Stop adding pub-key-hash label to all nodes

sebsoto:
This commit fixes an issue where the public key hash annotation was being
added to not all nodes, not just Windows nodes. This commit also adds
logic to remove the annotation from Linux nodes when the operator is
installed.

As part of this fix, Windows node selection in the secret controller
now uses the label "kubernetes.io/os=windows". There is a mix of this
label and "node.openshift.io/os_id=Windows" in the node selection logic
throughout WMCO, and we should move towards just using one. Future work
should be done to only use the "kubernetes.io/os" label for node
selection throughout WMCO.

### DIFF
--- a/controllers/secret/secret_controller.go
+++ b/controllers/secret/secret_controller.go
@@ -59,6 +59,10 @@ func newReconciler(mgr manager.Manager) (reconcile.Reconciler, error) {
 	}
 
 	reconciler := &ReconcileSecret{client: client, scheme: mgr.GetScheme()}
+	if err = reconciler.removeInvalidAnnotationsFromLinuxNodes(); err != nil {
+		log.Error(err, "unable to clean up annotations on Linux nodes")
+	}
+
 	return reconciler, nil
 }
 
@@ -180,7 +184,7 @@ func (r *ReconcileSecret) Reconcile(request reconcile.Request) (reconcile.Result
 			return reconcile.Result{}, errors.Wrap(err, "error creating signer from private key")
 		}
 		nodes := &core.NodeList{}
-		err = r.client.List(context.TODO(), nodes, client.HasLabels{nodeconfig.WindowsOSLabel})
+		err = r.client.List(context.TODO(), nodes, client.MatchingLabels{core.LabelOSStable: "windows"})
 		if err != nil {
 			return reconcile.Result{}, errors.Wrapf(err, "error getting node list")
 		}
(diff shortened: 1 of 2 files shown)


controllers/secret/secret_controller.go
```diff
@@ -180,7 +184,13 @@ func (r *ReconcileSecret) Reconcile(request reconcile.Request) (reconcile.Result
 			return reconcile.Result{}, errors.Wrap(err, "error creating signer from private key")
 		}
 		nodes := &core.NodeList{}
-		err = r.client.List(context.TODO(), nodes, client.HasLabels{nodeconfig.WindowsOSLabel})
```
aravindhp, line 183: Can't you do the equivalent of  `tc.client.K8s.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{LabelSelector: nodeconfig.WindowsOSLabel})` that we have in our e2e tests? i.e use the label selector.

controllers/secret/secret_controller.go
```diff
@@ -213,6 +223,28 @@ func (r *ReconcileSecret) Reconcile(request reconcile.Request) (reconcile.Result
 	}
 }
 
+// removeDepreciatedAnnotations corrects annotations applied by previous versions of WMCO.
```
aravindhp, line 226: I suggest renaming this to `removeInvalidAnnotationsFromLinuxNodes` or something to that effect. Please mention the annotation that you are deleting. I don't see any reason to future proof given this is just a helper.

controllers/secret/secret_controller.go
```diff
@@ -213,6 +223,28 @@ func (r *ReconcileSecret) Reconcile(request reconcile.Request) (reconcile.Result
...
+func (r *ReconcileSecret) removeDepreciatedAnnotations() error {
+	nodes := &core.NodeList{}
+	err := r.client.List(context.TODO(), nodes)
```
aravindhp, line 229: Shouldn't you just list Linux nodes?

controllers/secret/secret_controller.go
```diff
@@ -213,6 +223,28 @@ func (r *ReconcileSecret) Reconcile(request reconcile.Request) (reconcile.Result
...
+	if err != nil {
+		return errors.Wrapf(err, "error getting node list")
+	}
+	escapedPubKeyAnnotation := strings.Replace(nodeconfig.PubKeyHashAnnotation, "/", "~1", -1)
```
aravindhp, line 233: Please comment why this is needed.

controllers/secret/secret_controller.go
```diff
@@ -213,6 +223,28 @@ func (r *ReconcileSecret) Reconcile(request reconcile.Request) (reconcile.Result
...
+	patchData := fmt.Sprintf(`[{"op":"remove","path":"/metadata/annotations/%s"}]`, escapedPubKeyAnnotation)
+	for _, node := range nodes.Items {
+		// The public key hash was accidentally added to Linux nodes in WMCO 2.0, this needs to be removed
+		if _, present := node.Annotations[nodeconfig.PubKeyHashAnnotation]; present == true &&
+			node.Labels[core.LabelOSStable] != "windows" {
```
aravindhp, line 238: If you only list Linux nodes using the label selector you won't have to do this check.

aravindhp (changes requested):

VaishnaviHire (commented):
@sebsoto Thanks for the fix. PTAL at my comments:

Commit message needs to be updated to `annotation was being added to all nodes, `

controllers/secret/secret_controller.go
```diff
@@ -180,7 +184,13 @@ func (r *ReconcileSecret) Reconcile(request reconcile.Request) (reconcile.Result
 			return reconcile.Result{}, errors.Wrap(err, "error creating signer from private key")
 		}
 		nodes := &core.NodeList{}
-		err = r.client.List(context.TODO(), nodes, client.HasLabels{nodeconfig.WindowsOSLabel})
+		// The controller-runtime client requires a map for label matching, so the Windows identifying label var needs
+		// to be split into a key/value pair.
+		windowsLabelKeyValuePair := strings.SplitN(nodeconfig.WindowsOSLabel, "=", 2)
```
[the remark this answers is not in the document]
sebsoto: I was considering making this change but was on the fence about it. Since we are using it in the node mapper I'm happy to use it here.

controllers/secret/secret_controller.go
sebsoto: I ended up taking @VaishnaviHire suggestion and using the core.LabelOSStable label which makes this simpler

aravindhp: While I am fine with the change, it has again been made silently. Please call this out in the commit message. Also add a follow up story to replace `nodeconfig.WindowsOSLabel` with `core.LabelOSStable`

controllers/secret/secret_controller.go
```diff
@@ -59,6 +59,10 @@ func newReconciler(mgr manager.Manager) (reconcile.Reconciler, error) {
 	}
 
 	reconciler := &ReconcileSecret{client: client, scheme: mgr.GetScheme()}
+	if err = reconciler.removeInvalidAnnotationsFromLinuxNodes(); err != nil {
+		log.Error(err, "unable to clean up annotations from previous WMCO versions")
```
aravindhp, line 63: `...annotations on Linux nodes...`

aravindhp (approved):
/approve

with the caveat that you address the minor comments.

aravindhp:
/hold
so that you can address the comments.

PR #328, opened 2021-03-03
[community-4.7] WINC-501: [ci] Remove operator-sdk framework from e2e suite

sebsoto:
This commit is being made in preparation to the switch to operator-sdk
1.x. In 1.x the test framework our tests used does not exist, so we must
stop using it. Because the framework is no longer used, the tests should
now be run using go test, instead of the operator-sdk binary.

Backport of https://github.com/openshift/windows-machine-config-operator/pull/314


aravindhp (approved):
/approve

PR #330, opened 2021-03-04
WINC-585: Remove operator-sdk metrics package dependency

mansikulkarni96:
This PR removes the dependency on operator-sdk metrics package in order to make it easier to transition to operator-sdk v1.x since the metrics package is no longer required. 
Co-authored by @VaishnaviHire


pkg/metrics/metrics.go
```diff
@@ -110,6 +110,11 @@ func Validate(ctx context.Context, cfg *rest.Config, namespace string) error {
 		return errors.Wrap(err, "could not get metrics Service monitor")
 	}
 
+	// Remove any existing stale resources from previous operator version.
+	if err := removeInvalidResources(ctx, cfg, namespace); err != nil {
```
VaishnaviHire, line 114: `occlient` and `mclient` can be passed as arguments for this function

pkg/metrics/metrics.go
```diff
@@ -71,71 +68,46 @@ func NewPrometheusNodeConfig(clientset *kubernetes.Clientset) (*PrometheusNodeCo
...
 	}
-
-	// In the case of an operator restart, a previous SM object will be deleted and a new one will
-	// be created. We are deleting to ensure that the SM always exists with the correct spec. Otherwise,
-	// metrics may exhibit unexpected behavior if created by a previous version of WMCO.
-	err = mclient.ServiceMonitors(namespace).Delete(context.TODO(), windowsMetricsResource, metav1.DeleteOptions{})
-	if err != nil && !apierrors.IsNotFound(err) {
-		return errors.Wrap(err, "could not delete existing ServiceMonitor object")
+	if wincNamespace.Labels["openshift.io/cluster-monitoring"] != "true" {
+		metricsEnabled = false
```
sebsoto, line 84: If metrics is not enabled shouldnt we return here?

pkg/metrics/metrics.go
```diff
@@ -71,71 +68,46 @@ func NewPrometheusNodeConfig(clientset *kubernetes.Clientset) (*PrometheusNodeCo
...
-		// If this operator is deployed to a cluster without the prometheus-operator running, it will return
-		// ErrServiceMonitorNotPresent, which can be used to safely skip ServiceMonitor creation.
-		if err == metrics.ErrServiceMonitorNotPresent {
-			metricsEnabled = false
-			return errors.Wrap(err, "install prometheus-operator in your cluster to create ServiceMonitor objects")
-
-		}
+	// Check if metrics service exists
+	serviceList, err := oclient.CoreV1().Services(namespace).List(ctx,
+		metav1.ListOptions{LabelSelector: "name=windows-machine-config-operator"})
```
sebsoto: Where is `windows-machine-config-operator` coming from?
Instead of using a string here lets use a variable/constant if possible.

aravindhp:
/retitle WINC-585: Remove operator-sdk metrics package dependency

pkg/metrics/metrics.go
```diff
@@ -71,71 +68,46 @@ func NewPrometheusNodeConfig(clientset *kubernetes.Clientset) (*PrometheusNodeCo
...
+	if err != nil || len(serviceList.Items) == 0 {
```
sebsoto, line 90: Should this be handled in the same way? If theres an issue talking to the kube API server should metricsEnabled really be set to false?

pkg/metrics/metrics.go
```diff
@@ -71,71 +68,46 @@ func NewPrometheusNodeConfig(clientset *kubernetes.Clientset) (*PrometheusNodeCo
...
-func Add(ctx context.Context, cfg *rest.Config, namespace string) error {
-	// Add to the below struct any other metrics ports you want to expose.
-	servicePorts := []v1.ServicePort{
-		{Port: Port, Name: PortName, Protocol: v1.ProtocolTCP, TargetPort: intstr.IntOrString{Type: intstr.Int, IntVal: Port}},
-	}
-
-	// Create Service object to expose the metrics port(s).
-	service, err := metrics.CreateMetricsService(ctx, cfg, servicePorts)
+// Validate will verify if the Services and Service Monitors are created in the operator namespace
+func Validate(ctx context.Context, cfg *rest.Config, namespace string) error {
```
aravindhp, line 72: Why are you validating this here? Shouldn't this be an e2e test?

vendor/modules.txt
```diff
@@ -55,11 +55,6 @@ github.com/aws/aws-sdk-go/service/sts/stsiface
 github.com/beorn7/perks/quantile
 # github.com/cespare/xxhash/v2 v2.1.1
 github.com/cespare/xxhash/v2
-# github.com/coreos/prometheus-operator v0.38.1-0.20200424145508-7e176fda06cc => github.com/coreos/prometheus-operator v0.38.1-0.20200424145508-7e176fda06cc
```
aravindhp, line 58: Please add the command used to update the vendor directory.

pkg/metrics/metrics.go
```diff
@@ -250,3 +255,38 @@ func updateServiceMonitors(cfg *rest.Config, namespace string) error {
 	}
 	return nil
 }
+
+// removeInvalidResources deletes any stale resources that could be left behind during the
+// upgrade process due to the renaming of resources between current and previous operator version
+func removeInvalidResources(ctx context.Context, cfg *rest.Config, namespace string) error {
```
aravindhp, line 268: Rather than invalid  I suggest using stale i.e. `removeStaleResources`

pkg/metrics/metrics.go
```diff
@@ -250,3 +255,38 @@ func updateServiceMonitors(cfg *rest.Config, namespace string) error {
...
+	invalidResourceName := "windows-machine-config-operator-metrics"
```
aravindhp, line 269: Same here: `staleResourceName`

deploy/role.yaml
```diff
@@ -25,6 +25,7 @@ rules:
   - get
   - update
   - patch
+  - list
```
sebsoto, line 28: Are create/delete/patch/update needed anymore?

pkg/metrics/metrics.go
```diff
@@ -71,71 +68,46 @@ func NewPrometheusNodeConfig(clientset *kubernetes.Clientset) (*PrometheusNodeCo
...
+		return errors.Wrap(err, "could not get metrics Service")
 	}
 
-	// The ServiceMonitor created by the operator-sdk metrics package doesn't have fields required to display
-	// node graphs for Windows. Update the Service monitor with the required fields.
-	err = updateServiceMonitors(cfg, namespace)
-	if err != nil {
-		return errors.Wrap(err, "error updating service monitor")
+	// the name for the metrics resources is set during creation of metrics service and is equivalent to the service name
+	windowsMetricsResource = serviceList.Items[0].Name
```
sebsoto, line 96: Is it always going to be the 0th service? Why?

sebsoto: Use `GetName()` instead of `Name`

pkg/metrics/metrics.go
```diff
@@ -71,71 +68,46 @@ func NewPrometheusNodeConfig(clientset *kubernetes.Clientset) (*PrometheusNodeCo
...
+	if windowsMetricsResource == "" {
+		return errors.New("could not get metrics resource name")
```
sebsoto, line 98: Theres a lack of `metricsEnabled = false` here. Is this intentional?
It seems like almost every error return has `metricsEnabled = false`, that leads me to believe this should be handled at the function call site.

pkg/metrics/metrics.go
```diff
@@ -250,3 +255,38 @@ func updateServiceMonitors(cfg *rest.Config, namespace string) error {
 	}
 	return nil
 }
+
+// removeInvalidResources deletes any stale resources that could be left behind during the
+// upgrade process due to the renaming of resources between current and previous operator version
+func removeInvalidResources(ctx context.Context, cfg *rest.Config, namespace string) error {
+	invalidResourceName := "windows-machine-config-operator-metrics"
+
+	oclient, err := k8sclient.NewForConfig(cfg)
```
aravindhp, line 271: These clients seem to be getting created in multiple places. Maybe this calls for a configuration metrics object?

pkg/metrics/metrics.go
```diff
@@ -71,71 +68,46 @@ func NewPrometheusNodeConfig(clientset *kubernetes.Clientset) (*PrometheusNodeCo
...
+	if windowsMetricsResource == "" {
+		return errors.New("could not get metrics resource name")
 	}
 
-	oclient, err := k8sclient.NewForConfig(cfg)
+	// Create a monitoring client to interact with the ServiceMonitor object
+	mclient, err := monclient.NewForConfig(cfg)
 	if err != nil {
-		return errors.Wrap(err, "could not create config clientset")
+		return errors.Wrap(err, "could not create monitoring client")
```
sebsoto, line 104: Theres a lack of metricsEnabled = false here. Is this intentional?

pkg/metrics/metrics.go
```diff
@@ -71,71 +68,51 @@ func NewPrometheusNodeConfig(clientset *kubernetes.Clientset) (*PrometheusNodeCo
...
-	// When a selector is present in a headless service i.e. spec.ClusterIP=None, Kubernetes manages the
-	// list of endpoints reverting all the changes made by the operator. Remove selector from Metrics Service to avoid
-	// reverting changes in the Endpoints object.
-	patchData := fmt.Sprintf(`{"spec":{"selector": null }}`)
-	service, err = oclient.CoreV1().Services(namespace).Patch(ctx, service.Name, types.MergePatchType,
-		[]byte(patchData), metav1.PatchOptions{})
-	if err != nil {
-		return errors.Wrap(err, "could not remove selector from metrics service")
+	// Remove any existing stale resources from previous operator version.
+	if err := removeInvalidResources(ctx, cfg, namespace); err != nil {
```
VaishnaviHire, line 114: `oclient` and `mclient` can be passed as arguments for this function

pkg/metrics/metrics.go
```diff
@@ -208,6 +184,37 @@ func (pc *PrometheusNodeConfig) Configure() error {
...
+	}
+	// create new Endpoint
+	newEndpoint := &v1.Endpoints{
+		TypeMeta: metav1.TypeMeta{
+			Kind: "Endpoints",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      windowsMetricsResource,
+			Namespace: pc.namespace,
+			Labels:    map[string]string{"app.kubernetes.io/name": windowsMetricsResource},
```
VaishnaviHire, line 202: this label needs to be changed to `name`

pkg/metrics/metrics.go
```diff
@@ -86,7 +86,7 @@ func Validate(ctx context.Context, cfg *rest.Config, namespace string) error {
 
 	// Check if metrics service exists
 	serviceList, err := oclient.CoreV1().Services(namespace).List(ctx,
-		metav1.ListOptions{LabelSelector: "name=windows-machine-config-operator"})
+		metav1.ListOptions{LabelSelector: "name=windows-exporter"})
```
sebsoto, line 89: Use a variable/constant here to make it clear whats going on. The comment at definition of the constant should mention that the name is set in the service manifest yaml

VaishnaviHire (commented):
Thanks @mansikulkarni96 for working on this. PTAL at my comments

pkg/metrics/metrics.go
```diff
@@ -185,6 +184,37 @@ func (pc *PrometheusNodeConfig) Configure() error {
 	return nil
 }
 
+// retrieveMetricsEndpoint returns an existing Endpoint object or creates a new one in the operator namespace
+// if it does not already exist.
+func (pc *PrometheusNodeConfig) retrieveMetricsEndpoint() (*v1.Endpoints, error) {
```
aravindhp, line 189: This functions is combining getting the endpoints or creating them. They should be separated out. Moreover, please explain why you need to create this here when it was not the case before.

pkg/metrics/metrics.go
```diff
@@ -110,6 +110,11 @@ func Validate(ctx context.Context, cfg *rest.Config, namespace string) error {
 		return errors.Wrap(err, "could not get metrics Service monitor")
 	}
 
+	// Remove any existing stale resources from previous operator version.
+	if err := removeInvalidResources(ctx, cfg, namespace); err != nil {
+		return errors.Wrap(err, "could not remove invalid resources")
+	}
```
sebsoto: This seems like it would be better suited at the start of the function instead of the end.
Even if it doesnt matter functionally now, it gives us a clean slate to work with

pkg/metrics/metrics.go
mansikulkarni96: missed this change, thanks for pointing out!

pkg/metrics/metrics.go
```diff
@@ -185,6 +184,37 @@ func (pc *PrometheusNodeConfig) Configure() error {
...
+
+	if apierrors.IsNotFound(err) {
+		// Create Endpoint object since it does not exist
+		endpoint, err := pc.k8sclientset.CoreV1().Endpoints(pc.namespace).Create(context.TODO(), newEndpoint, metav1.CreateOptions{})
+		if err != nil {
+			return nil, errors.Wrap(err, "error creating Metrics Endpoint")
+		}
+		return endpoint, nil
+	}
+	return nil, errors.Wrap(err, "could not retrieve Metrics Endpoint")
```
aravindhp, line 215: The error here could be with getting or creating the object. Retrieve is another word for get and can cause confusion.

test/e2e/metrics_test.go
```diff
@@ -73,18 +73,18 @@ func testPrometheus(t *testing.T) {
 
 	// check that service exists
 	_, err = testCtx.client.K8s.CoreV1().Services("openshift-windows-machine-config-operator").Get(context.TODO(),
-		"windows-machine-config-operator-metrics", metav1.GetOptions{})
+		"windows-exporter", metav1.GetOptions{})
```
sebsoto, line 76: Reuse the constant you will define in the metrics package here.

aravindhp (changes requested):
Thanks for working on this @VaishnaviHire @mansikulkarni96. Please address my comments.

pkg/metrics/metrics.go
```diff
@@ -185,6 +184,37 @@ func (pc *PrometheusNodeConfig) Configure() error {
...
+			Labels:    map[string]string{"app.kubernetes.io/name": windowsMetricsResource},
+		},
+		Subsets: nil,
+	}
+
+	if apierrors.IsNotFound(err) {
+		// Create Endpoint object since it does not exist
+		endpoint, err := pc.k8sclientset.CoreV1().Endpoints(pc.namespace).Create(context.TODO(), newEndpoint, metav1.CreateOptions{})
+		if err != nil {
+			return nil, errors.Wrap(err, "error creating Metrics Endpoint")
```
sebsoto, line 211: This can be better organized to reduce indentation and return early on error:

```

	existingEndpoint, err := pc.k8sclientset.CoreV1().Endpoints(pc.namespace).Get(context.TODO(), windowsMetricsResource, metav1.GetOptions{})
	if err == nil {
		return existingEndpoint, nil
	} else if err != apierrors.IsNotFound(err) {
		return nil, errors.Wrap(err, "could not retrieve Metrics Endpoint")
	}
	<... rest of code ...>

```

sebsoto (changes requested):

deploy/role.yaml
mansikulkarni96: edited the roles, create delete get are needed

pkg/metrics/metrics.go
mansikulkarni96: we already have an e2e test for this, it was just to make sure the resources exist before doing anything but it is not required to do all the validations here hence I removed it.

pkg/metrics/metrics.go
mansikulkarni96: I have changed some of the implementation to simply get the resource name from a constant instead of querying and listing the services.

pkg/metrics/metrics.go
mansikulkarni96: this is now done only when the monitoring label is not present

pkg/metrics/metrics.go
mansikulkarni96: updated to separate out.

pkg/metrics/metrics.go
mansikulkarni96: created a struct around them

vendor/modules.txt
mansikulkarni96: done

cmd/manager/main.go
```diff
@@ -152,9 +152,9 @@ func main() {
 		os.Exit(1)
 	}
 
-	// Add the Metrics Service and Service Monitor
-	if err := winmetrics.Add(ctx, cfg, namespace); err != nil {
-		log.Error(err, "failed to add Metrics Service and Service Monitor")
+	// Validate will verify if cluster monitoring is enabled in the operator namespace
+	if err := winmetrics.Validate(ctx, cfg, namespace); err != nil {
+		log.Error(err, "failed to validate if Metrics Service and Service Monitor are created")
```
aravindhp, line 157: How about `error validating metric setup`?

We should be generating an event here so that the user is aware..

deploy/olm-catalog/windows-machine-config-operator/manifests/service-monitor.yaml
```diff
@@ -0,0 +1,23 @@
+# Prometheus Monitor Service (Metrics)
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    name: windows-machine-config-operator
+  name: windows-machine-config-operator-metrics
```
aravindhp: Shouldn't this be `windows-exporter`? The confusion here is that you are introducing this here as `windows-machine-config-operator-metrics` and you then have a constant in the same commit that says `windows-exporter` and then you rename `windows-machine-config-operator-metrics` to `windows-exporter` in another commit.

deploy/olm-catalog/windows-machine-config-operator/manifests/service.yaml
```diff
@@ -0,0 +1,12 @@
+kind: Service
+apiVersion: v1
+metadata:
+  name: windows-machine-config-operator-metrics
+  labels:
+    name: windows-machine-config-operator
```
aravindhp: Same comment here.

pkg/metrics/metrics.go
```diff
@@ -37,6 +32,9 @@ const (
 	Host = "0.0.0.0"
 	// Port is the port number on which windows-exporter is exposed.
 	Port int32 = 9182
+	// WindowsMetricsResource is the name for objects created for Prometheus monitoring
+	// by current operator version. Its name is defined through the bundle manifests
+	WindowsMetricsResource = "windows-exporter"
```
aravindhp, line 37: This should be introduced in the commit where you are renaming.

pkg/metrics/metrics.go
```diff
@@ -69,14 +80,36 @@ func NewPrometheusNodeConfig(clientset *kubernetes.Clientset) (*PrometheusNodeCo
...
+	mclient, err := monclient.NewForConfig(cfg)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not create monitoring clientset")
+	}
+	return &metricsConfig{oclient, mclient}, nil
+}
+
 // Validate will verify if cluster monitoring is enabled in the operator namespace
+// and delete any stale metrics resources from previous operator version
 func Validate(ctx context.Context, cfg *rest.Config, namespace string) error {
```
aravindhp, line 101: This should be part of the `metricsConfig` struct. Please place this along with `removeStaleResources`.

pkg/metrics/metrics.go
```diff
@@ -224,3 +257,26 @@ func updateServiceMonitors(cfg *rest.Config, namespace string) error {
 	}
 	return nil
 }
+
+// removeStaleResources deletes any stale resources that could be left behind during the
+// upgrade process due to the renaming of resources between current and previous operator version
+func (mc *metricsConfig) removeStaleResources(ctx context.Context, cfg *rest.Config, namespace string) error {
```
aravindhp, line 268: You need to call out that you are deleting the old resource in the commit message.

pkg/metrics/metrics.go
```diff
@@ -69,14 +80,36 @@ func NewPrometheusNodeConfig(clientset *kubernetes.Clientset) (*PrometheusNodeCo
...
-	oclient, err := k8sclient.NewForConfig(cfg)
+	metricsConfig, err := newMetricsConfig(cfg)
 	if err != nil {
-		return errors.Wrap(err, "could not create config clientset")
+		return errors.Wrap(err, "unable to initialize metrics config")
+	}
+
+	// Remove any existing stale resources from previous operator version.
+	if err := metricsConfig.removeStaleResources(ctx, cfg, namespace); err != nil {
+		return errors.Wrap(err, "could not remove stale metrics resources")
```
aravindhp, line 109: You should consider separation of concerns here. This shouldn't be part of validation.

pkg/metrics/metrics.go
```diff
@@ -168,11 +168,16 @@ func (pc *PrometheusNodeConfig) Configure() error {
 		return errors.Wrap(err, "could not get Windows nodes")
 	}
 
-	// get Metrics Endpoints object
-	endpoints, err := pc.k8sclientset.CoreV1().Endpoints(pc.namespace).Get(context.TODO(),
-		WindowsMetricsResource, metav1.GetOptions{})
+	endpoints, err := pc.k8sclientset.CoreV1().Endpoints(pc.namespace).Get(context.TODO(), WindowsMetricsResource, metav1.GetOptions{})
```
aravindhp: Why are you making this change in this commit?

pkg/metrics/metrics.go
```diff
@@ -187,6 +192,29 @@ func (pc *PrometheusNodeConfig) Configure() error {
 	return nil
 }
 
+// createMetricsEndpoint creates an endpoint object in the operator namespace.
+// WMCO is no longer creating a service with selector hence no Endpoint object is created
+// and WMCO needs to create it.
```
aravindhp, line 197: I don't follow this statement. This too is being created by WMCO.

pkg/metrics/metrics.go
```diff
@@ -187,6 +192,29 @@ func (pc *PrometheusNodeConfig) Configure() error {
...
+func (pc *PrometheusNodeConfig) createMetricsEndpoint() (*v1.Endpoints, error) {
```
aravindhp, line 198: I am confused now :-) Shouldn't this be created by OLM?

pkg/metrics/metrics.go
```diff
@@ -235,29 +263,6 @@ func isEndpointsValid(nodes *v1.NodeList, endpoints *v1.Endpoints) bool {
 	return true
 }
 
-// updateServiceMonitors patches the metrics Service Monitor to include required fields to display node graphs on the
-// OpenShift console. Console graph queries require metrics endpoint target name to be node name, however
-// windows_exporter returns node IP. We replace the target name by adding `replace` action field to the ServiceMonitor
-// object that replaces node IP to node name as the metrics endpoint target.
-func updateServiceMonitors(cfg *rest.Config, namespace string) error {
```
aravindhp, line 263: Where is this being done now?

aravindhp (changes requested):

pkg/metrics/metrics.go
VaishnaviHire: This cannot be part of OLM manifest. I was getting `invalidResource` error when trying to deploy the `Endpoints` resource.

pkg/metrics/metrics.go
VaishnaviHire: The emphasis here is on `selector`. Previously the `service` had selector field by default  that was responsible for creating the `Endpoints` object and we were removing in the later step.

Since we can now create our own resource definition, we create a service without `selector`.  See [Service without selector](https://kubernetes.io/docs/concepts/services-networking/service/#services-without-selectors) for reference.

pkg/metrics/metrics.go
aravindhp: Please add this as a comment.

pkg/metrics/metrics.go
aravindhp: Please clarify this in the comment. Saying `it` makes it unclear which object you are referring to.

pkg/metrics/metrics.go
mansikulkarni96: This is being done now in the manifest for serviceMonitor in manifests/serviceMonitor.yaml

pkg/metrics/metrics.go
```diff
@@ -48,6 +44,16 @@ type PrometheusNodeConfig struct {
 	namespace string
 }
 
+// MetricsConfig holds the information required to interact with metrics objects
+type MetricsConfig struct {
```
aravindhp, line 51: Is there a reason for the `Metrics` prefix? I suggest going with `Config` given the name of the package is `metrics`. We should always ensure there is no stuttering in our APIs. `metrics.MetricsConfig` vs `metrics.Config`.

pkg/metrics/metrics.go
```diff
@@ -71,74 +77,20 @@ func NewPrometheusNodeConfig(clientset *kubernetes.Clientset) (*PrometheusNodeCo
...
-	// node graphs for Windows. Update the Service monitor with the required fields.
-	err = updateServiceMonitors(cfg, namespace)
-	if err != nil {
-		return errors.Wrap(err, "error updating service monitor")
-	}
-
 	oclient, err := k8sclient.NewForConfig(cfg)
 	if err != nil {
-		return errors.Wrap(err, "could not create config clientset")
+		return nil, errors.Wrap(err, "could not create config clientset")
```
aravindhp, line 90: `error creating config client`

pkg/metrics/metrics.go
VaishnaviHire, line 51: Given that we have `MetricsConfig` struct, do we need `PrometheusNodeConfig` ?

pkg/metrics/metrics.go
```diff
@@ -71,74 +77,20 @@ func NewPrometheusNodeConfig(clientset *kubernetes.Clientset) (*PrometheusNodeCo
...
-	// When a selector is present in a headless service i.e. spec.ClusterIP=None, Kubernetes manages the
-	// list of endpoints reverting all the changes made by the operator. Remove selector from Metrics Service to avoid
-	// reverting changes in the Endpoints object.
-	patchData := fmt.Sprintf(`{"spec":{"selector": null }}`)
-	service, err = oclient.CoreV1().Services(namespace).Patch(ctx, service.Name, types.MergePatchType,
-		[]byte(patchData), metav1.PatchOptions{})
+	mclient, err := monclient.NewForConfig(cfg)
 	if err != nil {
-		return errors.Wrap(err, "could not remove selector from metrics service")
+		return nil, errors.Wrap(err, "could not create monitoring clientset")
```
aravindhp, line 94: `error creating...`

cmd/manager/main.go
```diff
@@ -152,9 +152,23 @@ func main() {
...
+
+	// validate if cluster monitoring is enabled in the operator namespace
+	if err := metricsConfig.Validate(ctx); err != nil {
+		log.Error(err, "error validating cluster monitoring label")
+		os.Exit(1)
+	}
+
+	// Setup will do metrics setup like create Endpoint object
+	if err := metricsConfig.Setup(); err != nil {
+		log.Error(err, "failed to do metrics setup")
```
aravindhp: `error setting up metrics`

pkg/metrics/metrics.go
```diff
@@ -232,6 +241,29 @@ func (mc *MetricsConfig) Validate(ctx context.Context) error {
 	return nil
 }
 
+// removeStaleResources deletes any stale resources that could be left behind during the
+// upgrade process due to the renaming of resources between current and previous operator version
+func (mc *MetricsConfig) removeStaleResources(ctx context.Context) error {
+	if _, err := mc.CoreV1().Services(mc.namespace).Get(ctx, staleResourceName, metav1.GetOptions{}); err == nil {
+		if err = mc.CoreV1().Services(mc.namespace).Delete(ctx, staleResourceName, metav1.DeleteOptions{}); err != nil {
+			return errors.Wrap(err, "could not delete stale service object")
```
aravindhp, line 249: Please make a best effort to clean up all stale resources rather than return on the first error encountered. I also think this function should not return an error bur rather log and send events.

cmd/manager/main.go
```diff
@@ -152,9 +152,23 @@ func main() {
...
+		os.Exit(1)
+	}
+
+	// validate if cluster monitoring is enabled in the operator namespace
+	if err := metricsConfig.Validate(ctx); err != nil {
+		log.Error(err, "error validating cluster monitoring label")
+		os.Exit(1)
+	}
+
+	// Setup will do metrics setup like create Endpoint object
```
aravindhp, line 168: ` Setup configures the metric resources`

pkg/metrics/metrics.go
```diff
@@ -256,25 +208,52 @@ func isEndpointsValid(nodes *v1.NodeList, endpoints *v1.Endpoints) bool {
...
+	return nil
+}
+
+// createMetricsEndpoint creates an endpoint object in the operator namespace.
+// WMCO is no longer creating a service with a selector therefore no Endpoint
+// object is created and WMCO needs to create the Endpoint object.
+// We cannot create endpoints as a part of manifests deployment as there are
+// no Windows node IP's available to be added to the list of addresses in
+// subsets during operator deployment.
+func (mc *MetricsConfig) createMetricsEndpoint() error {
```
aravindhp, line 273: `createEndpoint`

pkg/metrics/metrics.go
```diff
@@ -32,7 +32,10 @@ const (
 	Port int32 = 9182
 	// WindowsMetricsResource is the name for objects created for Prometheus monitoring
 	// by current operator version. Its name is defined through the bundle manifests
-	WindowsMetricsResource = "windows-machine-config-operator-metrics"
+	WindowsMetricsResource = "windows-exporter"
+	// staleResourceName is the name for metrics objects created for Prometheus monitoring
+	// through previous versions of the operator
+	staleResourceName = "windows-machine-config-operator-metrics"
```
VaishnaviHire, line 38: The stale resource will be `<operator-name>- metrics`, where operator name is `k8sutil.GetOperatorName()`. This is required because the `community` operator name is different.

cmd/manager/main.go
```diff
@@ -152,9 +152,23 @@ func main() {
 		os.Exit(1)
 	}
 
-	// Add the Metrics Service and Service Monitor
-	if err := winmetrics.Add(ctx, cfg, namespace); err != nil {
-		log.Error(err, "failed to add Metrics Service and Service Monitor")
+	// Get metrics configuration
+	metricsConfig, err := winmetrics.NewMetricsConfig(cfg, namespace)
```
aravindhp, line 156: Nit: I think we can get rid of `winmetrics` naming. We had to use that because of the SDK `metrics` pkg which has now been removed.

pkg/metrics/metrics.go
```diff
@@ -71,74 +77,20 @@ func NewPrometheusNodeConfig(clientset *kubernetes.Clientset) (*PrometheusNodeCo
...
-		log.Error(err, "could not create ServiceMonitor object")
-		// If this operator is deployed to a cluster without the prometheus-operator running, it will return
-		// ErrServiceMonitorNotPresent, which can be used to safely skip ServiceMonitor creation.
-		if err == metrics.ErrServiceMonitorNotPresent {
-			metricsEnabled = false
-			return errors.Wrap(err, "install prometheus-operator in your cluster to create ServiceMonitor objects")
-
-		}
+// NewMetricsConfig creates a new instance for MetricsConfig  to be used by the caller.
+func NewMetricsConfig(cfg *rest.Config, namespace string) (*MetricsConfig, error) {
```
aravindhp, line 84: Now this will become `NewConfig`

pkg/metrics/metrics.go
```diff
@@ -256,25 +208,52 @@ func isEndpointsValid(nodes *v1.NodeList, endpoints *v1.Endpoints) bool {
...
+		return errors.Wrap(err, "error create Metrics Endpoint")
+	}
+	return nil
+}
 
-	mclient, err := monclient.NewForConfig(cfg)
+// Validate will verify if cluster monitoring is enabled in the operator namespace
+func (mc *MetricsConfig) Validate(ctx context.Context) error {
+	// validate if metrics label is added to namespace
+	wincNamespace, err := mc.CoreV1().Namespaces().Get(ctx, mc.namespace, metav1.GetOptions{})
```
aravindhp, line 233: `wmcoNamespace` would be more appropriate.

pkg/metrics/metrics.go
```diff
@@ -256,25 +208,52 @@ func isEndpointsValid(nodes *v1.NodeList, endpoints *v1.Endpoints) bool {
...
-// object that replaces node IP to node name as the metrics endpoint target.
-func updateServiceMonitors(cfg *rest.Config, namespace string) error {
-
-	patchData := fmt.Sprintf("[{\"op\": \"replace\", \"path\": \"/spec/endpoints/0\", "+
-		"\"value\":{\"path\": \"/%s\",\"port\": \"%s\",\"relabelings\": [{\"action\": \"replace\", \"regex\": \"(.*)\", "+
-		"\"replacement\": \"$1\", \"sourceLabels\": [\"__meta_kubernetes_endpoint_address_target_name\"],"+
-		"\"targetLabel\": \"instance\"}]}}]", PortName, PortName)
+// Setup will take care of all the required configuration steps
+// for Prometheus monitoring like creating metrics Endpoint object.
+func (mc *MetricsConfig) Setup() error {
```
aravindhp: How about using `Configure` to be consistent with rest of the code?

aravindhp (changes requested):
Thanks for addressing the first round of comments, @mansikulkarni96. You are definitely moving in the correct direction. PTAL at the new comments.

cmd/manager/main.go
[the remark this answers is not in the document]
aravindhp: I suggest that we keep them separate. `Setup()` is cleaning up stale resources in the next commit.

Should we be configuring metrics if validation fails?

pkg/metrics/metrics.go
aravindhp: Nice catch, @VaishnaviHire

cmd/manager/main.go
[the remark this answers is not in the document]
mansikulkarni96: +1 on this, main reason is to separate concerns.

pkg/metrics/metrics.go
mansikulkarni96: I think that can be done, given they are holding very similar information at this point, but removing it would have many changes which are not in conjunction with this PR.

cmd/manager/main.go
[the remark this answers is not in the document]
VaishnaviHire: We should not. Reasoning :  Failure in validation suggest we are unable to access `service` or `service-monitor` object. Endpoints object references port defined by these objects for displaying graphs.

pkg/metrics/metrics.go
VaishnaviHire: Not sure. The only additional change I think is creating MetricsConfig() object [here](https://github.com/openshift/windows-machine-config-operator/blob/master/controllers/windowsmachine/windowsmachine_controller.go#L89) instead of `PrometheusNodeConfig` .

mansikulkarni96: It seems feasible, I was thinking just in terms of renaming changes but that could be another commit as well if we want to do this.

pkg/metrics/metrics.go
```diff
@@ -256,25 +208,52 @@ func isEndpointsValid(nodes *v1.NodeList, endpoints *v1.Endpoints) bool {
...
+	}
+	return nil
+}
+
+// createMetricsEndpoint creates an endpoint object in the operator namespace.
+// WMCO is no longer creating a service with a selector therefore no Endpoint
+// object is created and WMCO needs to create the Endpoint object.
+// We cannot create endpoints as a part of manifests deployment as there are
+// no Windows node IP's available to be added to the list of addresses in
+// subsets during operator deployment.
```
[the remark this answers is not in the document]
mansikulkarni96: The reason is Endpoints kind is not yet supported by OLM bundle
https://github.com/operator-framework/operator-lifecycle-manager/issues/1996
https://github.com/operator-framework/operator-registry/blob/92a4c92168862d509841cef6fa949a1b6dec1ec9/pkg/lib/bundle/supported_resources.go#L27
Included this in the commit message and comment.

cmd/manager/main.go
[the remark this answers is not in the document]
aravindhp: What happens if the label is added after the operator is installed? That is the scenario we need to think through.

pkg/metrics/metrics.go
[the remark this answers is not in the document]
mansikulkarni96: sorry for the confusion, my thinking earlier was that the error was due to empty subsets, after digging more into this I found the kind is not supported by OLM and I have updates the comments to reflect that.

pkg/metrics/metrics.go
```diff
@@ -256,25 +215,60 @@ func isEndpointsValid(nodes *v1.NodeList, endpoints *v1.Endpoints) bool {
...
+	// validate if metrics label is added to namespace
+	wmcoNamespace, err := c.CoreV1().Namespaces().Get(ctx, c.namespace, metav1.GetOptions{})
 	if err != nil {
-		return errors.Wrap(err, "error creating monitoring client")
+		return errors.Wrap(err, "error getting operator namespace")
+	}
+	if wmcoNamespace.Labels["openshift.io/cluster-monitoring"] != "true" {
+		metricsEnabled = false
+		c.Recorder.Eventf(wmcoNamespace, core.EventTypeWarning, "labelValidationFailed",
+			"Cluster monitoring openshift.io/cluster-monitoring label is not enabled in %s namespace", c.namespace)
```
VaishnaviHire, line 245: This needs to be `openshift.io/cluster-monitoring=true`

VaishnaviHire (commented):
Thanks @mansikulkarni96 for working on this.

LGTM, except one minor request.

pkg/metrics/metrics.go
```diff
@@ -256,25 +215,60 @@ func isEndpointsValid(nodes *v1.NodeList, endpoints *v1.Endpoints) bool {
...
+		return errors.Errorf("monitoring not enabled in %s namespace", c.namespace)
```
aravindhp, line 246: Why is this an error? It seems the function is more about checking if metrics is enabled or not and should return `true` or `false` in addition to an error.

pkg/metrics/metrics.go
```diff
@@ -256,25 +215,60 @@ func isEndpointsValid(nodes *v1.NodeList, endpoints *v1.Endpoints) bool {
...
+func (c *Config) Configure(ctx context.Context) error {
+	// validate if cluster monitoring is enabled in the operator namespace
+	if err := c.validate(ctx); err != nil {
+		log.Error(err, "error validating cluster monitoring label")
+		return nil
+	}
+	// Create Metrics Endpoint object in the operator namespace
+	if err := c.createEndpoint(); err != nil {
+		return errors.Wrap(err, "error creating metrics Endpoint")
+	}
```
aravindhp, line 230: I thought we decided not to create the endpoint if metrics was not enabled. Or am I mistaken?

pkg/metrics/metrics.go
```diff
@@ -248,6 +248,42 @@ func (c *Config) validate(ctx context.Context) error {
...
+		}
+	}
+
+	if serviceMonitor, err := c.ServiceMonitors(c.namespace).Get(ctx, staleResourceName, metav1.GetOptions{}); err == nil {
+		if err = c.ServiceMonitors(c.namespace).Delete(ctx, staleResourceName, metav1.DeleteOptions{}); err != nil {
+			c.Recorder.Eventf(serviceMonitor, core.EventTypeWarning, "serviceMonitorDeletionFailed",
+				"ServiceMonitor %s deletion failure", serviceMonitor.GetName())
+			log.Error(err, "error deleting stale serviceMonitor object")
+		}
+	}
```
aravindhp, line 284: What if the resource is present but the `Get` fails? Won't we be silently ignoring it? Furthermore, not attempting to delete the resource.

aravindhp (changes requested):

pkg/metrics/metrics.go
mansikulkarni96: @aravindhp we are not creating endpoints here if the validation fails, according to the implementation here if there is an error we are logging and returning nil, it could be done in another way where we check for the metricsEnabled bool value and proceed.

pkg/metrics/metrics.go
mansikulkarni96: can be implemented as you suggested, the current implementation looks for an error for returning

pkg/metrics/metrics.go
mansikulkarni96: I ll add a check for not found error.

pkg/metrics/metrics.go
aravindhp: My point is that you cannot assume that all errors returned by `validate()` is because metrics is disabled. What if there is an error getting the operator namespace? That cannot be treated as metrics being disabled. So either a bool needs to be returned or a specific error that indicates that metrics is disabled.

pkg/metrics/metrics.go
aravindhp: Please see previous comment.

aravindhp (approved):
Nice work, @mansikulkarni96 @VaishnaviHire 

/approve

pkg/metrics/metrics.go
```diff
@@ -48,6 +48,18 @@ type PrometheusNodeConfig struct {
...
+// Config holds the information required to interact with metrics objects
+type Config struct {
+	// a handle that allows us to interact with the Kubernetes API.
+	*kubernetes.Clientset
+	// a handle that allows us to interact with the Monitoring API.
+	*monclient.MonitoringV1Client
+	// namespace is the namespace in which metrics objects are created
+	namespace string
+	// recorder to generate events
+	Recorder record.EventRecorder
```
sebsoto, line 61: Does this need to be public?

pkg/metrics/metrics.go
```diff
@@ -256,25 +215,62 @@ func isEndpointsValid(nodes *v1.NodeList, endpoints *v1.Endpoints) bool {
...
+		return errors.Wrap(err, "error validating cluster monitoring label")
+	}
+	// Create Metrics Endpoint object only if monitoring is enabled
+	if enabled {
+		if err := c.createEndpoint(); err != nil {
+			return errors.Wrap(err, "error creating metrics Endpoint")
+		}
+	}
+	return nil
+}
```
sebsoto, line 235: nit: You can reduce indentation by handling the not enabled case first
Either approach is fine since the enabled case is fairly short in this case
```

	enabled, err := c.validate(ctx)
	if err != nil {
		return errors.Wrap(err, "error validating cluster monitoring label")
	}
	// Create Metrics Endpoint object only if monitoring is enabled
	if !enabled {
		return nil
	}
	if err := c.createEndpoint(); err != nil {
		return errors.Wrap(err, "error creating metrics Endpoint")
	}
	return nil
}

```

pkg/metrics/metrics.go
```diff
@@ -251,6 +252,60 @@ func (c *Config) validate(ctx context.Context) (bool, error) {
...
+	wmcoNamespace, err := c.CoreV1().Namespaces().Get(ctx, c.namespace, metav1.GetOptions{})
+	if err != nil {
+		log.Error(err, "error getting operator namespace")
+	}
+
+	// staleResourceName is the metrics object name created for Prometheus monitoring by previous operator versions
+	staleResourceName := wmcoNamespace.GetName() + "-metrics"
+
+	// remove stale service objects
+	if _, err := c.CoreV1().Services(c.namespace).Get(ctx, staleResourceName, metav1.GetOptions{}); err != nil && !apierrors.IsNotFound(err) {
```
mansikulkarni96, line 267: err is not being created again here which causes issue when using err nil check for deletion, adding a fix for this.

aravindhp: Won't `err` always get overwritten in the subsequent calls?

aravindhp: NVM, I see the problem.

pkg/metrics/metrics.go
```diff
@@ -256,25 +216,131 @@ func isEndpointsValid(nodes *v1.NodeList, endpoints *v1.Endpoints) bool {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      WindowsMetricsResource,
+			Namespace: c.namespace,
+			Labels:    map[string]string{"name": WindowsMetricsResource},
+		},
+		Subsets: nil,
+	}
+	_, err := c.CoreV1().Endpoints(c.namespace).Get(context.TODO(),
+		WindowsMetricsResource, metav1.GetOptions{})
+	if err != nil && apierrors.IsNotFound(err) {
```
aravindhp, line 337: Don't we need to delete and recreate the endpoint to ensure we are not using a stale `windows-exporter` resource created by a previous version 2.0 version of the operator?

mansikulkarni96:
aws-e2e-operator test passed successfully [logs](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/330/pull-ci-openshift-windows-machine-config-operator-master-aws-e2e-operator/1371932346520965120/artifacts/aws-e2e-operator/windows-e2e-operator-test/build-log.txt)
with upgrade test timeout increased by 10mins, reverting this change and retesting to verify the need for this change.

mansikulkarni96:
timed out on during deletionTestSuite [logs](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/330/pull-ci-openshift-windows-machine-config-operator-master-aws-e2e-operator/1371980380990607360/artifacts/aws-e2e-operator/gather-extra/artifacts/pods/openshift-windows-machine-config-operator_windows-machine-config-operator-74df5f75db-zs2j2_windows-machine-config-operator.log)

mansikulkarni96:
test passed after separation of deletionTestSuite
/test aws-e2e-operator

PR #337, opened 2021-03-10
Bug 1928809: Bump libthrift to 0.14.1

aravindhp:
This addresses [CVE-2020-13949](https://access.redhat.com/security/cve/CVE-2020-13949)

Commands run:
```shell
go mod tidy
go mod vendor
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.15
 replace github.com/docker/docker => github.com/moby/moby v0.7.3-0.20190826074503-38ab9da00309 // Required by Helm
 
 replace (
+	github.com/apache/thrift => github.com/apache/thrift v0.14.1 // For fixing https://bugzilla.redhat.com/show_bug.cgi?id=1928809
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -5,6 +5,7 @@ go 1.15
 replace github.com/docker/docker => github.com/moby/moby v0.7.3-0.20190826074503-38ab9da00309 // Required by Helm
 
 replace (
+	github.com/apache/thrift => github.com/apache/thrift v0.14.1 // For fixing https://bugzilla.redhat.com/show_bug.cgi?id=1928809
```
sebsoto, line 8: Whats the reasoning behind putting this in the replace directive vs the require directive?

aravindhp: `go get -u github.com/apache/thrift@v0.14.1` will add it to `require`. However the next `go mod tidy` will remove it as we don't directly use it. LMK if I am missing something.

aravindhp:
/close

Upon closer inspection we are not vendoring in libthrift, so I don't see how we are affected by the vulnerability in the library.

PR #351, opened 2021-03-18
[test] Separate out the network tests

aravindhp:
Move the network tests out from the creation tests. This is in preparation for separating out the upgrade and reconfiguration tests into a separate CI job.

### DIFF
--- a/hack/run-ci-e2e-test.sh
+++ b/hack/run-ci-e2e-test.sh
@@ -120,10 +120,14 @@ go test ./test/e2e/... -run=TestWMCO/operator_deployed_without_private_key_secre
 
 # Run the creation tests of the Windows VMs
 printf "\n####### Testing creation #######\n" >> "$ARTIFACT_DIR"/wmco.log
-go test ./test/e2e/... -run=TestWMCO/create -v -timeout=120m -args --node-count=$NODE_COUNT --private-key-path=$KUBE_SSH_KEY_PATH $WMCO_PATH_OPTION
+go test ./test/e2e/... -run=TestWMCO/create -v -timeout=90m -args --node-count=$NODE_COUNT --private-key-path=$KUBE_SSH_KEY_PATH $WMCO_PATH_OPTION
 # Get logs for the creation tests
 printf "\n####### WMCO logs for creation tests #######\n" >> "$ARTIFACT_DIR"/wmco.log
 get_WMCO_logs
+
+# Run the network tests
+printf "\n####### Testing network #######\n" >> "$ARTIFACT_DIR"/wmco.log
+go test ./test/e2e/... -run=TestWMCO/network -v -timeout=20m -args --node-count=$NODE_COUNT --private-key-path=$KUBE_SSH_KEY_PATH $WMCO_PATH_OPTION
 
 # Run the upgrade tests and skip deletion of the Windows VMs
 printf "\n####### Testing upgrade #######\n" >> "$ARTIFACT_DIR"/wmco.log
--- a/test/e2e/network_test.go
+++ b/test/e2e/network_test.go
@@ -15,6 +15,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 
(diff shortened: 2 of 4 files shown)


aravindhp:
/retitle [test] Separate out the network tests

hack/run-ci-e2e-test.sh
```diff
@@ -120,11 +120,15 @@ go test ./test/e2e/... -run=TestWMCO/operator_deployed_without_private_key_secre
...
 printf "\n####### Testing creation #######\n" >> "$ARTIFACT_DIR"/wmco.log
-go test ./test/e2e/... -run=TestWMCO/create -v -timeout=120m -args --node-count=$NODE_COUNT --private-key-path=$KUBE_SSH_KEY_PATH $WMCO_PATH_OPTION
+go test ./test/e2e/... -run=TestWMCO/create -v -timeout=90m -args --node-count=$NODE_COUNT --private-key-path=$KUBE_SSH_KEY_PATH $WMCO_PATH_OPTION
 # Get logs for the creation tests
 printf "\n####### WMCO logs for creation tests #######\n" >> "$ARTIFACT_DIR"/wmco.log
 get_WMCO_logs
 
+# Run the network tests
+printf "\n####### Testing network #######\n" >> "$ARTIFACT_DIR"/wmco.log
+go test ./test/e2e/... -run=TestWMCO/network -v -timeout=20m -args --node-count=$NODE_COUNT --private-key-path=$KUBE_SSH_KEY_PATH $WMCO_PATH_OPTION
```
mansikulkarni96, line 130: I see the timeout being decreased by 10 minutes in total is that on purpose, if yes is there an evidence for doing so?

test/e2e/network_test.go
```diff
@@ -23,9 +23,11 @@ import (
 
 // testNetwork runs all the cluster and node network tests
 func testNetwork(t *testing.T) {
+	// Populate the global test context
 	testCtx, err := NewTestContext()
-	require.NoError(t, err)
```
mansikulkarni96, line 27: shouldn't we check for error?

mansikulkarni96 (commented):
Thanks for working on this @aravindhp , PTAL at my comments.

hack/run-ci-e2e-test.sh
aravindhp: Yes. On an average the networking e2e tests is taking around 10 minutes to complete across all providers.

test/e2e/network_test.go
aravindhp: Good catch.

test/e2e/network_test.go
```diff
@@ -23,9 +23,12 @@ import (
 
 // testNetwork runs all the cluster and node network tests
 func testNetwork(t *testing.T) {
+	// Populate the global test context
 	testCtx, err := NewTestContext()
 	require.NoError(t, err)
-	require.NoError(t, testCtx.createNamespace(testCtx.workloadNamespace), "error creating test namespace")
+	err = testCtx.waitForWindowsNodes(gc.numberOfNodes, true, false, false)
+	assert.NoError(t, err, "Windows node creation failed")
```
sebsoto, line 30: no creation happening here, `timed out waiting for Windows nodes` is more proper

test/e2e/metrics_test.go
```diff
@@ -26,6 +26,10 @@ const (
 )
 
 func testMetrics(t *testing.T) {
+	testCtx, err := NewTestContext()
+	require.NoError(t, err)
+	require.NoError(t, testCtx.createNamespace(testCtx.workloadNamespace), "error creating test namespace")
```
sebsoto, line 31: Why is this change being made? Is it because the metrics tests now run before the networking tests when run in CI? This will cause issues if the network tests are run independently. Instead the workload namespace creation/deletion can be handled in the common `TestWMCO` function.

sebsoto (changes requested):

test/e2e/metrics_test.go
aravindhp: >Is it because the metrics tests now run before the networking tests when run in CI?

Yes

> Instead the workload namespace creation/deletion can be handled in the common TestWMCO function.

Moving the creation is not straightforward as `testCtx.createNamespace` would need to become `testCtx.EnsureNamespace`. Given how rarely we run the network tests independently I decided to put this off. But I can make the change as part of this PR.

test/e2e/network_test.go
mansikulkarni96: this comments need to be addressed, LGTM apart from it.

test/e2e/wmco_test.go
```diff
@@ -31,10 +31,15 @@ func TestWMCO(t *testing.T) {
 		wmcoPath = "/usr/local/bin/windows-machine-config-operator"
 	}
 
+	testCtx, err := NewTestContext()
+	require.NoError(t, err)
+	require.NoError(t, testCtx.ensureNamespace(testCtx.workloadNamespace), "error creating test namespace")
```
sebsoto, line 36: Doesnt this need to be cleaned up?

sebsoto (approved):

sebsoto:
/hold
Approved but holding due to my comment

test/e2e/wmco_test.go
aravindhp: That [happens](https://github.com/openshift/windows-machine-config-operator/blob/master/test/e2e/delete_test.go#L69) in the deletion tests

aravindhp:
/hold cancel
/retest

https://github.com/openshift/installer/pull/4779 has merged

PR #356, opened 2021-03-23
[build] Move to Go 1.16

aravindhp:
As of https://github.com/openshift/ocp-build-data/pull/858 OCP will be built with Go 1.16, so move the WMCO builds to match that change.

### DIFF
--- a/.ci-operator.yaml
+++ b/.ci-operator.yaml
@@ -2,4 +2,4 @@
 build_root_image:
   namespace: openshift
   name: release
-  tag: golang-1.15
+  tag: golang-1.16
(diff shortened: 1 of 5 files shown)


mansikulkarni96:
Not sure if it is the cause that image pull is failing but registry.svc.ci is being deprecated and we should move to pulling from registry.ci instead.

aravindhp:
> Not sure if it is the cause that image pull is failing but registry.svc.ci is being deprecated and we should move to pulling from registry.ci instead.

Good catch @mansikulkarni96. I thought the image had still not been propagated as Luke mentioned on aos-devel.

mansikulkarni96:
Don't we need to bump to go1.16 in the go.mod file?

aravindhp:
> Don't we need to bump to go1.16 in the go.mod file?

Yes we do. I am now thinking we need a script for bumping golang versions 😆

mansikulkarni96:
/lgtm

PR #360, opened 2021-03-24
[wmco] Move controller loggers into Reconcilers

sebsoto:
The loggers for each controller were moved into their respective
reconciler struct. This was done in preperation of the controllers
being moved into the same package.
The loggers for each controller shared the same name and needed to be
made unique in order to do the future move.
Having these loggers be part of each reconciler solves the
uniqueness problem, and is the way newly generated controllers handle
logging in OSDK1.x projects.
In addition the relevant structs in the nodeconfig and windows packages
had loggers added to them. This makes it so that we are using the same
pattern throughout the code, as well as having the major benefit of our
log messages having more information regarding the controller
instantiating the nodeconfig object. This will be helpful in the future
if we are not just configuring VMs sourced from Machine objects.

Since this work required a change to the Windows Machine controller unit tests,
I have added an additional commit which ensures that those specific unit tests are
being run with the rest of the unit tests.


controllers/windowsmachine/windowsmachine_controller.go
```diff
@@ -425,7 +426,7 @@ func (r *ReconcileWindowsMachine) deleteMachine(machine *mapi.Machine) error {
 // addWorkerNode configures the given Windows VM, adding it as a node object to the cluster
 func (r *ReconcileWindowsMachine) addWorkerNode(ipAddress, instanceID, machineName string, platform oconfig.PlatformType) error {
 	nc, err := nodeconfig.NewNodeConfig(r.k8sclientset, ipAddress, instanceID, machineName, r.clusterServiceCIDR,
-		r.vxlanPort, r.signer, platform)
+		r.vxlanPort, r.signer, platform, r.log)
```
aravindhp, line 429: This is a general question that applies everywhere: should we be re-using the logger here? Shouldn't this be an internal field that the caller shouldn't be exposed to? IOW should we passing loggers around or be instantiating them in the New* functions.

aravindhp (changes requested):
Thanks for splitting this into a separate PR @sebsoto. This is so much easier to review. Please address my comments.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -83,10 +85,6 @@ func discoverKubeAPIServerEndpoint() (string, error) {
 // NewNodeConfig creates a new instance of nodeConfig to be used by the caller.
 func NewNodeConfig(clientset *kubernetes.Clientset, ipAddress, instanceID, machineName, clusterServiceCIDR,
 	vxlanPort string, signer ssh.Signer, platform oconfig.PlatformType) (*nodeConfig, error) {
-	// Update the logger name with the VM's cloud ID. Ideally this should be the Machine name but is not available at
-	// this point.
```
aravindhp, line 87: Any reason to remove this comment?

aravindhp (approved):
/approve

pkg/nodeconfig/nodeconfig.go
sebsoto: I was thinking its not going to be accurate in the future when we are not just dealing with Machine objects.
I may have been jumping this gun with this change. Reverted it, it can be changed when those future changes are made.

mansikulkarni96:
thanks for working on this @sebsoto 
/lgtm

aravindhp:
/hold

@sebsoto please re-run `go mod tidy` and `go mod vendor` with Go 1.16.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/openshift/windows-machine-config-operator
 
-go 1.15
+go 1.16
```
mansikulkarni96, line 3: not sure why this change? Shouldn't it be already present?

controllers/secret/secret_controller.go
```diff
@@ -58,9 +57,10 @@ func newReconciler(mgr manager.Manager) (reconcile.Reconciler, error) {
 		return nil, err
 	}
 
-	reconciler := &ReconcileSecret{client: client, scheme: mgr.GetScheme()}
+	logger := logf.Log.WithName(ControllerName)
```
aravindhp, line 60: Please call this `log`

controllers/secret/secret_controller.go
```diff
@@ -77,7 +77,8 @@ func add(mgr manager.Manager, r reconcile.Reconciler, watchNamespace string) err
 	// Check that the private key exists, if it doesn't, log a warning
 	_, err = secrets.GetPrivateKey(kubeTypes.NamespacedName{Namespace: watchNamespace, Name: secrets.PrivateKeySecret}, mgr.GetClient())
 	if err != nil {
-		log.Error(err, "Unable to retrieve private key, please ensure it is created")
+		logger := logf.Log.WithName(ControllerName)
```
aravindhp, line 80: Please call this `log`

aravindhp (approved):
/approve

PR #362, opened 2021-03-25
[hack] Allow specifying tests

aravindhp:
- Given we don't want to run the upgrade tests on all providers, introduce a `-t` option to run-ci-e2e-test.sh that can be used to  specify the test. The allowed options are:
  - `all` all the tests are run
  - `basic` creation, network and deletion tests are run
  - `upgrade` creation, upgrade, reconfiguration and deletion tests are run
- Remove stale `-k` option from the script
- Add new `run-ci-e2e-upgrade-test` Makefile endpoint which will be used  in the new upgrade CI job. Next step would be to add the new CI job that uses this new endpoint. At which point we will change the `run-ci-e2e-test` endpoint to only run the `basic` tests.

### DIFF
--- a/docs/HACKING.md
+++ b/docs/HACKING.md
@@ -43,6 +43,7 @@ hack/olm.sh cleanup -c "<OPERATOR_IMAGE>"
 We need to set up all the environment variables required in [Development workflow](#development-workflow) as well as: 
 ```shell script
 export OPERATOR_IMAGE=<registry url for remote WMCO image>
+export OPENSHIFT_CI=false
 ```
 Once the above variables are set, run the following script:
 ```shell script
@@ -55,13 +56,15 @@ Additional flags that can be passed to `hack/run-ci-e2e-test.sh` are
 - `-b` gives an alternative path to the WMCO binary. This option overridden in OpenShift CI.
        When building the operator locally, the WMCO binary is created as part of the operator image build process and
        can be found at `build/_output/bin/windows-machine-config-operator`, this is the default value of this option.
+- `-t` specify the test to run. All tests are run if the option is not used. The allowed options are:
+  - `all` all the tests are run
+  - `basic` creation, network and deletion tests are run
+  - `upgrade` creation, upgrade, reconfiguration and deletion tests are run
 
-       
 Example command to spin up 2 Windows nodes and retain them after test run:
 ```shell script
 hack/run-ci-e2e-test.sh -s -n 2      
 ```
-
 Please note that you do not need to run `hack/olm.sh run` before `hack/run-ci-e2e-test.sh`.
 
 ## Bundling the Windows Machine Config Operator
--- a/hack/run-ci-e2e-test.sh
+++ b/hack/run-ci-e2e-test.sh
@@ -35,7 +35,8 @@ OSDK_WMCO_test() {
   fi
 }
 
-while getopts ":n:k:b:s" opt; do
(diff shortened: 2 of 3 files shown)


hack/run-ci-e2e-test.sh
```diff
@@ -115,8 +123,10 @@ done
 # -flag x is allowed for non-boolean flags only(https://golang.org/pkg/flag/)
 
 # Test that the operator is running when the private key secret is not present
-printf "\n####### Testing operator deployed without private key secret #######\n" >> "$ARTIFACT_DIR"/wmco.log
-go test ./test/e2e/... -run=TestWMCO/operator_deployed_without_private_key_secret -v -args --node-count=$NODE_COUNT --private-key-path=$KUBE_SSH_KEY_PATH $WMCO_PATH_OPTION
+if [[ "$TEST" = "all" || "$TEST" = "basic" ]]; then
+  printf "\n####### Testing operator deployed without private key secret #######\n" >> "$ARTIFACT_DIR"/wmco.log
+  go test ./test/e2e/... -run=TestWMCO/operator_deployed_without_private_key_secret -v -args --node-count=$NODE_COUNT --private-key-path=$KUBE_SSH_KEY_PATH $WMCO_PATH_OPTION
+fi
```
sebsoto, line 129: This test is just checking if the operator is deployed. It should complete almost instantly. I'm not sure that there is a reason we wouldnt want it to run

docs/HACKING.md
```diff
@@ -55,13 +56,15 @@ Additional flags that can be passed to `hack/run-ci-e2e-test.sh` are
...
+  - `all` all the tests are run
+  - `basic` creation, network and deletion tests are run
+  - `upgrade` creation, upgrade, reconfiguration and deletion tests are run
 
-       
 Example command to spin up 2 Windows nodes and retain them after test run:
 ```shell script
 hack/run-ci-e2e-test.sh -s -n 2      
 ```
-
```
mansikulkarni96, line 64: can add this line back

hack/run-ci-e2e-test.sh
```diff
@@ -35,7 +35,8 @@ OSDK_WMCO_test() {
   fi
 }
 
-while getopts ":n:k:b:s" opt; do
+TEST="all"
+while getopts ":n:b:st:" opt; do
```
mansikulkarni96, line 39: should there be a : between s and t?

mansikulkarni96 (commented):
PTAL at my comments

hack/run-ci-e2e-test.sh
aravindhp: OK, the name of the test made think this was a special case. I will update it to run always.

hack/run-ci-e2e-test.sh
aravindhp: Not from my reading of `man getopts`. Please double check and let me know.

docs/HACKING.md
aravindhp: I removed it on purpose as I was editing that area. Is there a reason you want that extra line?

mansikulkarni96: doesn't seem like we need it from the rich diff, lgtm

aravindhp:
/hold

I just realized if I want to use the operator bundle workflow in CI, we should account for not using the SDK to deploy in the script.

PR #372, opened 2021-03-31
[test] Explicitly apply monitoring label

aravindhp:
When the upgrade test is launched as a separate job, https://github.com/openshift/release/pull/17227, the monitoring label does not get applied to the operator namespace. Handle that difference by explicitly applying the label.

### DIFF
--- a/test/e2e/metrics_test.go
+++ b/test/e2e/metrics_test.go
@@ -17,6 +17,7 @@ import (
 	"github.com/stretchr/testify/require"
 	core "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


test/e2e/wmco_test.go
```diff
@@ -35,6 +35,10 @@ func TestWMCO(t *testing.T) {
 	require.NoError(t, err)
 	require.NoError(t, testCtx.ensureNamespace(testCtx.workloadNamespace), "error creating test namespace")
 
+	// When the upgrade test is run from CI, the namespace that gets created does not have the required monitoring
+	// label, so we apply that here.
+	require.NoError(t, testCtx.applyMonitoringLabelToOperatorNamespace(), "error applying monitoring label")
```
sebsoto, line 40: I feel like we should check if the cluster has monitoring enabled as well.

test/e2e/metrics_test.go
```diff
@@ -256,3 +257,11 @@ func (tc *testContext) getPrometheusToken() (string, error) {
 	}
 	return string(token), nil
 }
+
+// applyMonitoringLabelToOperatorNamespace adds the "openshift.io/cluster-monitoring:"true"" label to the
+// openshift-windows-machine-config-operator namespace
+func (tc *testContext) applyMonitoringLabelToOperatorNamespace() error {
+	_, err := tc.client.K8s.CoreV1().Namespaces().Patch(context.TODO(), resourceNamespace, types.MergePatchType,
```
sebsoto, line 264: Please use tc.namespace

mansikulkarni96:
thanks for adding this @aravindhp 
/lgtm

PR #376, opened 2021-04-01
controller: Hardcode machine.Kind as "Machine"

LorbusChris:
Fixes a bug where the condition is entered even though the object is in
fact of type Machine.

### DIFF
--- a/controllers/windowsmachine_controller.go
+++ b/controllers/windowsmachine_controller.go
@@ -239,7 +239,7 @@ func (r *ReconcileWindowsMachine) isValidMachine(obj runtime.Object) bool {
 	}
 
 	// If for some reason this function is called on an object which is not a Machine, return false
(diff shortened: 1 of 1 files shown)


controllers/windowsmachine_controller.go
```diff
@@ -239,7 +239,7 @@ func (r *ReconcileWindowsMachine) isValidMachine(obj runtime.Object) bool {
 	}
 
 	// If for some reason this function is called on an object which is not a Machine, return false
-	if kind := obj.GetObjectKind().GroupVersionKind(); kind.Kind != machine.Kind {
+	if kind := obj.GetObjectKind().GroupVersionKind(); kind.Kind != "Machine" {
```
aravindhp, line 242: Please move the `machine` initialization closer to the usage.

aravindhp (changes requested):
Thanks for fixing this, @LorbusChris. Just a minor request.

PR #377, opened 2021-04-01
[test] Restart WMCO after monitoring label is applied

aravindhp:
While #372 ensured that the `openshift.io/cluster-monitoring` is applied, that was not sufficient as WMCO is already running at that point and assumes monitoring is not enabled. Work around this by restarting WMCO after the label is applied.

This is needed for https://github.com/openshift/release/pull/17227/.

### DIFF
--- a/test/e2e/metrics_test.go
+++ b/test/e2e/metrics_test.go
@@ -258,10 +258,32 @@ func (tc *testContext) getPrometheusToken() (string, error) {
 	return string(token), nil
 }
 
-// applyMonitoringLabelToOperatorNamespace adds the "openshift.io/cluster-monitoring:"true"" label to the
-// openshift-windows-machine-config-operator namespace
-func (tc *testContext) applyMonitoringLabelToOperatorNamespace() error {
-	_, err := tc.client.K8s.CoreV1().Namespaces().Patch(context.TODO(), tc.namespace, types.MergePatchType,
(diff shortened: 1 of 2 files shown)


test/e2e/metrics_test.go
```diff
@@ -258,10 +258,33 @@ func (tc *testContext) getPrometheusToken() (string, error) {
...
+		if err = tc.scaleWMCODeployment(0); err != nil {
+			return err
+		}
+		// Scale up the WMCO deployment to 1, so that WMCO is aware that monitoring is enabled
+		if err = tc.scaleWMCODeployment(1); err != nil {
+			return err
+		}
+
+	}
+
```
sebsoto, line 288: nit: extra newlines

sebsoto (changes requested):

test/e2e/metrics_test.go
aravindhp: I only see a single new line and that pattern is present in many parts of the code.

aravindhp: I typically try to add news lines across functional blocks in functions. So the `return nil` is independent of the previous block where I am applying the label and restarting the operator.

sebsoto: what about the extra newline between the two curly braces?

aravindhp: That should go. Fixed.

PR #390, opened 2021-04-09
[community-4.7][wmco] Update operator version to 2.0.1

sebsoto:


aravindhp (commented):
Can we use `update` instead of `bump`?

aravindhp (approved):
/lgtm

PR #394, opened 2021-04-13
[wmco] Remove apis/ directory and package

sebsoto:
The schema adding logic was already moved into main.go, the extra
schema adding is superflous.
This as a whole is uneeded as we have no CRDs being introduced by the
operator. The directory will be regenerated in the case a CRD is added.


aravindhp (approved):

/approve

sebsoto:
/hold
while network test issues persist

PR #399, opened 2021-04-14
WINC-500: [wmco] Improve isValidMachine()

aravindhp:
Introduce `getInternalIPAddress()` that returns the internal IPv4 address of a Machine and use it in all the call sites. This has the good side effect of improving `isValidMachine()` as it was not doing all the necessary checks for IP addresses in Machine objects.

### DIFF
--- a/controllers/windowsmachine_controller.go
+++ b/controllers/windowsmachine_controller.go
@@ -3,6 +3,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"net"
 	"strings"
 
 	"github.com/go-logr/logr"
@@ -215,22 +216,24 @@ func (r *WindowsMachineReconciler) isValidMachine(obj client.Object) bool {
 
 	// If this function is called on an object that equals nil, return false
 	if obj == nil {
-		r.log.Error(errors.New("expected machine object to not equal nil"), "invalid Machine", "object", obj)
+		r.log.Error(errors.New("Machine object cannot be nil"), "invalid Machine", "object", obj)
 		return false
 	}
 
 	var ok bool
 	machine, ok = obj.(*mapi.Machine)
 	if !ok {
-		r.log.Error(errors.New("unable to typecast object to machine"), "invalid Machine", "object", obj)
+		r.log.Error(errors.New("unable to typecast object to Machine"), "invalid Machine", "object", obj)
 		return false
 	}
 	if machine.Status.Phase == nil {
-		r.log.V(1).Info("machine object has no phase associated with it", "name", machine.Name)
+		r.log.V(1).Info("Machine has no phase associated with it", "name", machine.Name)
 		return false
 	}
-	if len(machine.Status.Addresses) == 0 {
-		r.log.V(1).Info("machine object has no address associated with it", "name", machine.Name)
+
+	_, err := getInternalIPAddress(machine.Status.Addresses)
+	if err != nil {
+		r.log.V(1).Info("invalid Machine", "name", machine.Name, "error", err)
 		return false
 	}
 
@@ -342,19 +345,9 @@ func (r *WindowsMachineReconciler) Reconcile(ctx context.Context, request ctrl.R
 	}
 
 	// Get the IP address associated with the Windows machine, if not error out to requeue again
-	if len(machine.Status.Addresses) == 0 {
-		return ctrl.Result{}, errors.Errorf("machine %s doesn't have any ip addresses defined",
-			machine.Name)
-	}
-	ipAddress := ""
-	for _, address := range machine.Status.Addresses {
-		if address.Type == core.NodeInternalIP {
-			ipAddress = address.Address
-		}
-	}
-	if len(ipAddress) == 0 {
-		return ctrl.Result{}, errors.Errorf("no internal ip address associated with machine %s",
-			machine.Name)
+	ipAddress, err := getInternalIPAddress(machine.Status.Addresses)
+	if err != nil {
+		return ctrl.Result{}, errors.Wrapf(err, "invalid machine %s", machine.Name)
 	}
 
 	// Get the instance ID associated with the Windows machine.
@@ -518,4 +511,19 @@ func (r *WindowsMachineReconciler) isWindowsMachineHealthy(machine *mapi.Machine
 	}
 
 	return true
+}
+
+// getInternalIPAddress returns the internal IP address of the Machine
+func getInternalIPAddress(addresses []core.NodeAddress) (string, error) {
+	// Get the IP address associated with the Windows machine, if not error out to requeue again
+	if len(addresses) == 0 {
+		return "", errors.New("no IP addresses defined")
+	}
+	for _, address := range addresses {
+		// Only return the IPv4 address
+		if address.Type == core.NodeInternalIP && net.ParseIP(address.Address).To4() != nil {
+			return address.Address, nil
+		}
+	}
+	return "", errors.New("no internal IP address associated")
 }
--- a/controllers/windowsmachine_controller_test.go
+++ b/controllers/windowsmachine_controller_test.go
@@ -23,15 +23,6 @@ func TestIsValidMachine(t *testing.T) {
 	invalidMachine3 := mapi.Machine{}
 	invalidMachine3.Name = "invalid_2"
 	invalidMachine3.Status.Phase = strToPtr("running")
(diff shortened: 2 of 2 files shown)


controllers/windowsmachine_controller.go
```diff
@@ -515,3 +508,21 @@ func (r *WindowsMachineReconciler) isWindowsMachineHealthy(machine *mapi.Machine
...
+func getInternalIPAddress(addresses []core.NodeAddress) (string, error) {
+	// Get the IP address associated with the Windows machine, if not error out to requeue again
+	if len(addresses) == 0 {
+		return "", errors.New("no IP addresses defined")
+	}
+	ipAddress := ""
+	for _, address := range addresses {
+		if address.Type == core.NodeInternalIP {
+			ipAddress = address.Address
+		}
```
LorbusChris, line 522:
```suggestion
		if address.Type == core.NodeInternalIP || address.Type == core.NodeHostName {
			ipAddress = address.Address
		}
```
@aravindhp don't we support the NodeHostName address type here too? (If I'm not mistaken some platforms set the InternalP, while others set HostName)

LorbusChris (commented):
Thanks @aravindhp, the changes lgtm, but one question

controllers/windowsmachine_controller.go
aravindhp: If we do, why is this not present in the original code? I am bit hesitant to change behavior in this PR.

LorbusChris: If the case never occurs where ONLY hostname (and no InternalIP) is set, this can be disregarded.
@jcpowermac can you confirm this never happens on vSphere?

controllers/windowsmachine_controller_test.go
```diff
@@ -23,15 +23,6 @@ func TestIsValidMachine(t *testing.T) {
...
 	invalidMachine3.Status.Phase = strToPtr("running")
-	validMachine1 := mapi.Machine{}
-	validMachine1.Name = "valid_1"
-	validMachine1.Status.Phase = strToPtr("running")
-	validMachine1.Status.Addresses = []core.NodeAddress{
-		{
-			Type:    "Hostname",
-			Address: "valid1.acme.com",
-		},
-	}
```
LorbusChris, line 34: While it might make sense to remove this test, this is actually a valid Machine object per the spec, so maybe we should really account for this programmatically.

controllers/windowsmachine_controller.go
aravindhp: And if it does happen then we should introduce checking for hostname in a separate PR as it is a clear change in WMCO behavior.

controllers/windowsmachine_controller_test.go
aravindhp: While I agree with you, I prefer opening a separate PR where we account for hostname and then reintroduce this unit test.

controllers/windowsmachine_controller.go
```diff
@@ -338,19 +340,10 @@ func (r *WindowsMachineReconciler) Reconcile(ctx context.Context, request ctrl.R
...
-			ipAddress = address.Address
-		}
-	}
-	if len(ipAddress) == 0 {
-		return ctrl.Result{}, errors.Errorf("no internal ip address associated with machine %s",
-			machine.Name)
+	ipAddress, err := getInternalIPAddress(machine.Status.Addresses)
+	if err != nil {
+		r.log.V(1).Info("invalid machine", "error", err)
+		return ctrl.Result{}, err
```
sebsoto, line 346: This is going to result in a debug logging of the error + context, as well as an info logging of just the error.
It makes more sense to me to just return
```suggestion
		return ctrl.Result{}, errors.Wrap(err, "invalid machine")
```

controllers/windowsmachine_controller.go
```diff
@@ -515,3 +508,21 @@ func (r *WindowsMachineReconciler) isWindowsMachineHealthy(machine *mapi.Machine
...
+// getInternalIPAddress returns the internal IP address of the Machine
+func getInternalIPAddress(addresses []core.NodeAddress) (string, error) {
+	// Get the IP address associated with the Windows machine, if not error out to requeue again
+	if len(addresses) == 0 {
+		return "", errors.New("no IP addresses defined")
+	}
+	ipAddress := ""
+	for _, address := range addresses {
+		if address.Type == core.NodeInternalIP {
+			ipAddress = address.Address
```
sebsoto, line 521: Can do a break here to return early once the correct address is found

sebsoto (changes requested):

controllers/windowsmachine_controller.go
aravindhp: I rewrote the function. PTAL.

controllers/windowsmachine_controller.go
aravindhp: Done except I added the Machine name to the returned error.

controllers/windowsmachine_controller.go
```diff
@@ -515,3 +507,17 @@ func (r *WindowsMachineReconciler) isWindowsMachineHealthy(machine *mapi.Machine
...
+	// Get the IP address associated with the Windows machine, if not error out to requeue again
+	if len(addresses) == 0 {
+		return "", errors.New("no IP addresses defined")
+	}
+	for _, address := range addresses {
+		if address.Type == core.NodeInternalIP {
+			return address.Address, nil
+		}
+	}
+	return "", errors.Errorf("no internal IP address associated")
```
mansikulkarni96, line 522: can use errors.new here if there isn't a format string to be specified.

aravindhp: Done.

aravindhp:
/hold

Getting an IPv6 internal address:
```
2021-04-17T13:22:52.110Z	DEBUG	VM 422ca56c-0b60-39c7-6eb0-5a69f5535b71	SSH dial	{"IP Address": "fe80::ac77:2493:9af7:4ff8", "error": "dial tcp: address fe80::ac77:2493:9af7:4ff8:22: too many colons in address"}
```

aravindhp:
Continue seeing `no route to host` in the [machine controller log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/399/pull-ci-openshift-windows-machine-config-operator-master-vsphere-e2e-operator/1383382188786978816/artifacts/vsphere-e2e-operator/gather-extra/artifacts/pods/openshift-machine-api_machine-api-controllers-698fcf9697-mc2sk_machine-controller.log). Not sure if this causing the internal IP address from being updated.

aravindhp:
On vSphere we are seeing `Machines` with both IPv4 and IPv6 `address` fields:
```yaml
  addresses: 
  - address: fe80::7dd2:7826:65e0:c043
    type: InternalIP
  - address: 172.31.251.29
    type: InternalIP
  - address: winworker-nwqqg
    type: InternalDNS
```
Updated the PR to only return the IPv4 `address`.

LorbusChris:
thanks for this fix @aravindhp 
/lgtm

test timed out
/retest

controllers/windowsmachine_controller.go
```diff
@@ -519,3 +512,18 @@ func (r *WindowsMachineReconciler) isWindowsMachineHealthy(machine *mapi.Machine
...
 }
+
+// getInternalIPAddress returns the internal IP address of the Machine
+func getInternalIPAddress(addresses []core.NodeAddress) (string, error) {
+	// Get the IP address associated with the Windows machine, if not error out to requeue again
+	if len(addresses) == 0 {
+		return "", errors.New("no IP addresses defined")
+	}
+	for _, address := range addresses {
+		// Only return the IPv4 address
```
sebsoto, line 523: Is it possible for a cluster to only have ipv6 addresses?

aravindhp: It is possible. But addressing that is out of the scope of this PR. We might need to holistically look at IPv6.

aravindhp:
/hold

Wait for https://github.com/openshift/windows-machine-config-operator/pull/408 to pass

LorbusChris:
#408 has passed tests and merged
/hold cancel

/retest

aravindhp:
/hold

until we figure out why the metrics tests is failing during destroy

PR #407, opened 2021-04-19
WINC-541: [submodule][windows-machine-config-bootstrapper] Update to 0d5d2f4

### LINKED ISSUE #205: Prevent large windows container image pull timeouts with kubelet configuration
Many Windows Container images are huge (upwards of 5GB) and this can cause trouble with the windows node kubelet.  I've observed with [this image](https://quay.io/repository/mhildenb/hplussports-win?tab=tags) running on (an admittedly small) Standard_D2s_v3 windows node configured by the WMCO.  The issue is that the container is so big that it can't be unpacked before the (default 1 minute) image pull progress deadline is exceeded.  This causes the pull to fail and be restarted.  The effect is that the container never runs.

I have confirmed that this issue can be remedied by increasing the --image-pull-progress-deadline of the windows node's Kubelet to something like 15m.  That is based on [this advice from google](https://cloud.google.com/kubernetes-engine/docs/how-to/creating-a-cluster-windows#image_pull_errors).

In my demo I used this [TaskRun](https://github.com/hatmarch/k8-for-windows-devs/blob/master/install/kube/tekton/taskrun/run-increase-pull-deadline.yaml) using [this task](https://github.com/hatmarch/k8-for-windows-devs/blob/master/install/kube/tekton/tasks/powershell-command.yaml) to run [this Powershell script](https://github.com/hatmarch/k8-for-windows-devs/blob/master/install/windows-nodes/scripts/image-pull-timeout.ps1) on the node after the WMCO finished its configuration.

Obviously, it would be better if the WMCO ansible playbook set the image timeout as part of the node setup.

aravindhp:
Update to https://github.com/openshift/windows-machine-config-bootstrapper/commit/0d5d2f4fce523dfb6978b78c64705c9a123836ea

This commit was generated using hack/update_submodules.sh

Fixes #205


aravindhp:
/hold

Wait for https://github.com/openshift/windows-machine-config-operator/pull/408 to pass

aravindhp:
/hold

until we figure out why the metrics tests is failing during destroy

PR #411, opened 2021-04-22
[test] Wait for machines to be deleted

mansikulkarni96:
This PR introduces a wait time for the machines to be deleted after deleteMachineSet() has been called in deletionTestSuite()
to make sure we are proceeding to test prometheus endpoints after all the machines have been deleted.

### DIFF
--- a/test/e2e/create_test.go
+++ b/test/e2e/create_test.go
@@ -107,14 +107,27 @@ func (tc *testContext) deleteMachineSet(ms *mapi.MachineSet) error {
 }
 
 // waitForWindowsMachines waits for a certain amount of Windows Machines to reach a certain phase
+// if machineCount = 0, it implies we are only waiting for Machines to be deleted and the phase is
+// ignored in this case.
 func (tc *testContext) waitForWindowsMachines(machineCount int, phase string) error {
-	machineCreationTimeLimit := time.Minute * 5
+	if machineCount == 0 && phase != "" {
+		return errors.New("expected phase to be to be an empty string if machineCount is 0")
+	}
+
+	var machines *mapi.MachineList
+	machineStateTimeLimit := time.Minute * 5
+	startTime := time.Now()
 	// Increasing the time limit due to https://bugzilla.redhat.com/show_bug.cgi?id=1936556
 	if tc.CloudProvider.GetType() == config.VSpherePlatformType {
-		machineCreationTimeLimit = time.Minute * 20
+		// When deleting Machines, set the time limit to 10 minutes
+		if machineCount == 0 {
+			machineStateTimeLimit = time.Minute * 10
+		} else {
+			machineStateTimeLimit = time.Minute * 20
+		}
 	}
-	return wait.Poll(retryInterval, machineCreationTimeLimit, func() (done bool, err error) {
-		machines, err := tc.client.Machine.Machines(clusterinfo.MachineAPINamespace).List(context.TODO(), metav1.ListOptions{
+	err := wait.Poll(retryInterval, machineStateTimeLimit, func() (done bool, err error) {
+		machines, err = tc.client.Machine.Machines(clusterinfo.MachineAPINamespace).List(context.TODO(), metav1.ListOptions{
 			LabelSelector: controllers.MachineOSLabel + "=Windows"})
 		if err != nil {
 			if apierrors.IsNotFound(err) {
@@ -128,6 +141,10 @@ func (tc *testContext) waitForWindowsMachines(machineCount int, phase string) er
 			log.Printf("waiting for %d/%d Windows Machines", machineCount-len(machines.Items), machineCount)
 			return false, nil
 		}
+		// A phase of "" skips the phase check
+		if phase == "" {
+			return true, nil
+		}
 		for _, machine := range machines.Items {
 			if machine.Status.Phase == nil || *machine.Status.Phase != phase {
 				return false, nil
@@ -135,6 +152,14 @@ func (tc *testContext) waitForWindowsMachines(machineCount int, phase string) er
 		}
 		return true, nil
 	})
+	if phase == "" {
+		phase = "deleted"
+	}
+
+	// Log the time elapsed while waiting for creation of the Machines
+	endTime := time.Now()
+	log.Printf("%v time is required for %d Machines to reach phase %s", endTime.Sub(startTime), len(machines.Items), phase)
+	return err
 }
 
 // waitForWindowsNode waits until there exists nodeCount Windows nodes with the correct set of annotations.
--- a/test/e2e/delete_test.go
+++ b/test/e2e/delete_test.go
@@ -54,6 +54,8 @@ func testWindowsNodeDeletion(t *testing.T) {
 	for _, machineSet := range e2eMachineSets {
 		assert.NoError(t, testCtx.deleteMachineSet(machineSet), "error deleting MachineSet")
 	}
+	// Phase is ignored during deletion, in this case we are just waiting for Machines to be deleted.
+	require.NoError(t, testCtx.waitForWindowsMachines(int(gc.numberOfNodes), ""), "Windows machine deletion failed")
 
 	// Test if prometheus configuration is updated to have no node entries in the endpoints object
 	testPrometheus(t)


test/e2e/create_test.go
```diff
@@ -246,3 +246,25 @@ func (tc *testContext) waitForWindowsNodes(nodeCount int32, waitForAnnotations,
 
 	return err
 }
+
+// waitForMachineDeletion waits for a certain amount of time for the Windows Machines to be deleted
+func (tc *testContext) waitForMachineDeletion() error {
```
sebsoto, line 251: This function has overlapping functionality with the [waitForWindowsMachines](https://github.com/openshift/windows-machine-config-operator/blob/1f2bf433fe2e812df22cbc4879d3b0437f8e5788/test/e2e/create_test.go#L110) function

Instead of introducing this new function it would be better to make use of the existing one.

sebsoto (changes requested):

test/e2e/create_test.go
```diff
@@ -246,3 +246,25 @@ func (tc *testContext) waitForWindowsNodes(nodeCount int32, waitForAnnotations,
...
+	machineDeletionTimeLimit := time.Minute * 5
```
jrvaldes, line 252: should `machineDeletionTimeLimit` be a constant?

test/e2e/create_test.go
mansikulkarni96: my main reasoning to not reuse that function due to two reasons: it has some vsphere platform specific time limits additionally when a machine is being deleted it could quickly change phases and waitForWindowsMachines waits for the machine to be in a particular phase which could lead to potential misses.

jrvaldes (approved):

test/e2e/create_test.go
mansikulkarni96: @jrvaldes it could be a constant preferred to keep it is a variable here as we might have to change the value depending on the various platforms in future.

test/e2e/create_test.go
sebsoto: Thanks for the reasoning, I understand that thought process

The vsphere platform specific time limits were put in place to deal with the root cause of this issue, the machine controller being slow on vsphere. I think it is actually beneficially to have that on the deletion side as well.

> it could quickly change phases 

That is true but calling `waitForWindowsMachines(0)`  will wait until there are 0 Machines, and skip the phase checking. This function is essentially the same as `waitForWindowsNodes` which is being used the same way in the deletion test, specifying that we wait for `0` nodes

mansikulkarni96: I think I can make it work for both as you mentioned just not sure if we should go for a time limit of 20 minutes as in case of creation, might seem like an overkill. We could stick to 10 for vsphere and 5 for rest WDYT?

sebsoto: I'm not sure that might not be enough. @aravindhp might have more insight.

test/e2e/create_test.go
```diff
@@ -128,6 +133,10 @@ func (tc *testContext) waitForWindowsMachines(machineCount int, phase string) er
 			log.Printf("waiting for %d/%d Windows Machines", machineCount-len(machines.Items), machineCount)
 			return false, nil
 		}
+		// When testing Machine deletion, skipping the phase matching as it could quickly change during deletion
+		if phase == "" {
```
aravindhp, line 137: Please update the function doc to say that `phase` being an empty implies deletion

aravindhp (changes requested):

aravindhp (changes requested):
We are getting a good CI signal with this change. Lets also get some timing numbers. Please add similar code like we have in  [waitForWindowsNodes()](https://github.com/openshift/windows-machine-config-operator/blob/master/test/e2e/create_test.go#L150) to `waitForWindowsMachine`. Please mention this in the commit message.

test/e2e/create_test.go
sebsoto: I dont agree with the comment in the code and we should be cautious with the wording around passing in a an empty phase.

If there are 3 machines in the running state and we want to delete a single one, and wait for that to occur, the function call would look like: `tc.waitForWindowsMachines(2, "Running")`, The reason you wouldnt want to enter `""` as the phase, is that in that case we would be cautious against too many machines being deleted, so we'd want to make sure the remaining Machines are still in the `Running` phase, and not transitioned to the deletion phase.

The phase should be empty whenever we truly don't care about the phase, such as in the case in this commit, where the amount of Machines is expected to be zero, and therefore cannot have a phase.

My point here is that `When testing Machine deletion, skipping the phase matching...` is only accurate when scaling down to zero Machines. Instead I'd prefer wording like: `A phase of "" skips the phase check and just waits until the correct amount of Windows Machines exist`

aravindhp: In general I agree with @sebsoto however...

>If there are 3 machines in the running state and we want to delete a single one, and wait for that to occur, the function call would look like: tc.waitForWindowsMachines(2, "Running"),

The function the way it is written will not do what you are saying. Not sure if we need this sort of functionality at the moment.

I think the easiest way forward would be to say that if `machineCount` is 0, then `phase` is ignored and that implies we are waiting for Machines to be deleted.

sebsoto: > the function the way it is written will not do what you are saying

I'm not able to see what you mean, the function waits until len(machines.Items) == machineCount and then for each machine that was found it verifies the phase is correct.

> I think the easiest way forward would be to say that if machineCount is 0, then phase is ignored and that implies we are waiting for Machines to be deleted.

I agree with that, that functionality was present in the function already so documenting that makes sense.

aravindhp: >I'm not able to see what you mean, the function waits until len(machines.Items) == machineCount and then for each machine that was found it verifies the phase is correct.

Take your example , `tc.waitForWindowsMachines(2, "Running")` when there are 3 `Machines` and one Machine is not in the `Running` phase but say in the `Provisioned` phase.
```go
machines, err := tc.client.Machine.Machines(clusterinfo.MachineAPINamespace).List(context.TODO(), metav1.ListOptions{
			LabelSelector: controllers.MachineOSLabel + "=Windows"})
```
Will always return 3. Your example assumes that the 3rd Machine is going to be deleted but that needn't be the case. The unwritten assumption in the function is that we are actually dealing with `Machines` from a `MachineSet` and we sort of expect them to change state as a unit. If we start going outside the boundaries of that assumption, the function will not work as expected.

test/e2e/create_test.go
```diff
@@ -107,13 +107,21 @@ func (tc *testContext) deleteMachineSet(ms *mapi.MachineSet) error {
 }
 
 // waitForWindowsMachines waits for a certain amount of Windows Machines to reach a certain phase
+// if machineCount = 0, it implies we are only waiting for Machines to be deleted and the phase is
+// ignored in this case.
 func (tc *testContext) waitForWindowsMachines(machineCount int, phase string) error {
-	machineCreationTimeLimit := time.Minute * 5
+	machineStateTimeLimit := time.Minute * 5
```
aravindhp, line 113: Please return an error if `machineCount == 0 && phase != ""`

test/e2e/create_test.go
```diff
@@ -128,13 +136,21 @@ func (tc *testContext) waitForWindowsMachines(machineCount int, phase string) er
...
 		for _, machine := range machines.Items {
 			if machine.Status.Phase == nil || *machine.Status.Phase != phase {
 				return false, nil
 			}
 		}
 		return true, nil
 	})
+	// Log the time elapsed while waiting for creation of the Machines
+	endTime := time.Now()
+	log.Printf("%v time is required to configure %v Machines", endTime.Sub(startTime), machineCount)
```
aravindhp, line 152:
```suggestion
if phase == "" {
   phase = "deleted"
}
log.Printf("%v time is required for %d Machines to reach phase %s", endTime.Sub(startTime), machineCount, phase)

**Review by aravindhp — CHANGES_REQUESTED** [review]

File: test/e2e/create_test.go [hunk 6]
```diff
@@ -107,14 +107,27 @@ func (tc *testContext) deleteMachineSet(ms *mapi.MachineSet) error {
...
+		if machineCount == 0 {
+			machineStateTimeLimit = time.Minute * 10
+		} else {
+			machineStateTimeLimit = time.Minute * 20
+		}
 	}
-	return wait.Poll(retryInterval, machineCreationTimeLimit, func() (done bool, err error) {
-		machines, err := tc.client.Machine.Machines(clusterinfo.MachineAPINamespace).List(context.TODO(), metav1.ListOptions{
+	err := wait.Poll(retryInterval, machineStateTimeLimit, func() (done bool, err error) {
+		machines, err = tc.client.Machine.Machines(clusterinfo.MachineAPINamespace).List(context.TODO(), metav1.ListOptions{
```
aravindhp, line 130: I suggest not re-using the `err` variable here. Think of this code as a separate function. Or am I missing something?

test/e2e/create_test.go
```diff
@@ -107,14 +107,27 @@ func (tc *testContext) deleteMachineSet(ms *mapi.MachineSet) error {
...
 // waitForWindowsMachines waits for a certain amount of Windows Machines to reach a certain phase
+// if machineCount = 0, it implies we are only waiting for Machines to be deleted and the phase is
+// ignored in this case.
 func (tc *testContext) waitForWindowsMachines(machineCount int, phase string) error {
-	machineCreationTimeLimit := time.Minute * 5
+	if machineCount == 0 && phase != "" {
+		return errors.New("expected phase to be to be an empty string if machineCount is 0")
+	}
+
+	var machines *mapi.MachineList
```
aravindhp, line 117: This won't be needed if you don't reuse the `err` variable below.

mansikulkarni96: this is needed to print the machines.Items in the log at the end of the function as the machines variable declared in poll function has a limited scope.

test/e2e/create_test.go
mansikulkarni96: It is using the err variable from func() (done bool, err error), redeclaring it gives a redeclared variable error.

test/e2e/create_test.go
aravindhp: Shouldn't you be using `machineCount` in the print statement?

mansikulkarni96: I think that will give an inaccurate time log, consider the case where it timed out on reaching the desired number of Machines. In this case the log will print the time with the original machineCount(desired) passed by the caller instead of the current number of Machines that took the time printed.

aravindhp: But what we care about is reaching the desired number of Machines and not the current number of Machines.

mansikulkarni96: I agree with your statement, that is what we care about. My only concern here is the accuracy of the log, the time printed in the log statement would not correspond to the time it took to create those Machines. 
Going by the statement "%v time is required for %v Machines to reach phase %s" , if a timeout occurs and only n machines are created/deleted by that time we will get an accurate indication of how much time was taken to do that. It might not be a huge concern in this case but just a thought.

aravindhp: But won't the [log](https://github.com/openshift/windows-machine-config-operator/blob/master/test/e2e/create_test.go#L128) within the loop clarify this?

aravindhp: Anyhow we are just bikeshedding now 😄 I am fine with whatever approach you decide on.

aravindhp (approved):

test/e2e/create_test.go
```diff
@@ -128,13 +141,25 @@ func (tc *testContext) waitForWindowsMachines(machineCount int, phase string) er
...
 		}
 		return true, nil
 	})
+	if phase == "" {
+		phase = "deleted"
+	}
+
+	// Log the time elapsed while waiting for creation of the Machines
+	endTime := time.Now()
+	log.Printf("%v time is required for %v Machines to reach phase %s", endTime.Sub(startTime), machines.Items, phase)
```
aravindhp, line 161: You need to do `len(machine.Items)` here and replace the %v with a %d.

aravindhp (changes requested):

test/e2e/create_test.go
mansikulkarni96: Thanks for the catch @aravindhp, that was a silly one from my side! PTAL when you get a chance.

aravindhp (approved):
/approve
/lgtm

Thanks for fixing the log @mansikulkarni96.

PR #412, opened 2021-04-23
[build] Improve development build process

aravindhp:
Introduce Dockerfile.base that builds only the submodules. This base image is used in Dockerfile.wmco which results in only the WMCO being built.

Now that we have simplified the image build process normalize on using `$OPERATOR_IMAGE` as the source of truth for the WMCO image.

### DIFF
--- a/docs/HACKING.md
+++ b/docs/HACKING.md
@@ -3,15 +3,22 @@
 # Hacking on the WMCO
 
 ## Build
-To build the operator image, execute:
+To build and push the operator image, execute:
 ```shell script
-podman build . -t quay.io/<insert username>/wmco:$VERSION_TAG -f build/Dockerfile
+export OPERATOR_IMAGE=quay.io/<insert username>/wmco:$VERSION_TAG
+hack/olm.sh build
 ```
 
-The operator image needs to be pushed to a remote repository:
+Building the full WMCO image is a time consuming process given we are building all the submodules. If you are just
+changing WMCO and not the submodules, you can separately build the submodules as a base image and then just build the
+WMCO  image from this base image:
 ```shell script
-podman push quay.io/<insert username>/wmco:$VERSION_TAG
+make base-img
+make wmco-img IMG=$OPERATOR_IMAGE
 ```
+`make base-img` will not push the base image as a push is not needed. `make wmco-img` will make and push the
+`$OPERATOR_IMAGE`
+
 ## Development workflow
 With the Operator image built and pushed to a remote repository, the operator can be deployed either directly,
 or through OLM.
@@ -23,7 +30,7 @@ export KUBECONFIG=<path to kubeconfig>
 
 ### Deploying the operator directly
 ```shell script
-make deploy IMG=<quay.io/<insert username>/wmco:$VERSION_TAG>
+make deploy IMG=$OPERATOR_IMAGE
 ```
 The above command will directly deploy all the resources required for the operator to run, as well as the WMCO pod.
 To remove the installed resources:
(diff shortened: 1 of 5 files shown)


jrvaldes (approved):
/lgtm

docs/HACKING.md
```diff
@@ -12,6 +12,17 @@ The operator image needs to be pushed to a remote repository:
...
 podman push quay.io/<insert username>/wmco:$VERSION_TAG
 ```
+
+Building the full WMCO image is a time consuming process given we are building all the submodules. If you are just
+changing WMCO and not the modules, you can separately build the submodules as a base image and then just build the WMCO
+image from this base image:
+```shell
+podman build . -t wmco-base:latest -f build/Dockerfile.base
+podman build . -t quay.io/<insert username>/wmco:$VERSION_TAG -f build/Dockerfile.wmco
+podman push quay.io/<insert username>/wmco:$VERSION_TAG
```
sebsoto, line 22: It would be nice if a make target was added that did this.
So then you could just do something like this:
`make dev-build IMG=quay.io/<insert username>/wmco:$VERSION_TAG`

sebsoto:
```
.PHONY: dev-build
dev-build:
        podman build . -t wmco-base:latest -f build/Dockerfile.base
        podman build . -t $(IMG) -f build/Dockerfile.wmco
        podman push $(IMG)
```

sebsoto (changes requested):

docs/HACKING.md
aravindhp: The main reason for the split Dockerfiles was so that you don't keep rebuilding the submodules which is exactly what this `Makefile` endpoint you are asking for is doing. IOW the endpoint is as good as doing `podman build . -t $(IMG) -f build/Dockerfile; podman push $(IMG)`. So I don't see how it helps us. So if you want `Makefile` endpoints it would be `dev-base-build`, `dev-wmco-build`, `dev-img-push`. Alternatively looking more at the new `Makefile`, I am wondering if `make run` with a local `/payload` directory is good enough. At which point, what we should do is add a `payload` endpoint.

aravindhp: >I am wondering if make run with a local /payload directory is good enough. At which point, what we should do is add a payload endpoint.

Just remembered that this won't work for us as WMCO needs to run on the private subnet.

mansikulkarni96:
changes LGTM, please remove hold when you are ready.

docs/HACKING.md
```diff
@@ -3,15 +3,22 @@
 # Hacking on the WMCO
 
 ## Build
-To build the operator image, execute:
+To build and pushthe operator image, execute:
```
sebsoto, line 6: `push the`

hack/olm.sh
```diff
@@ -51,8 +53,6 @@ case "$ACTION" in
 
     ;;
     run)
-  build_WMCO $OSDK
-
```
sebsoto, line 55: Thank you for changing this :smile:

sebsoto (changes requested):

docs/HACKING.md
aravindhp: Fixed

docs/HACKING.md
```diff
@@ -3,15 +3,22 @@
 # Hacking on the WMCO
 
 ## Build
-To build the operator image, execute:
+To build and push the operator image, execute:
 ```shell script
-podman build . -t quay.io/<insert username>/wmco:$VERSION_TAG -f build/Dockerfile
+export OPERATOR_IMAGE=quay.io/<insert username>/wmco:$VERSION_TAG
```
jrvaldes, line 8: Since `$VERSION_TAG` is a user defined field like `<insert username>`, I feel it should be written as `<version tag>`

jrvaldes (approved):
/lgtm

mansikulkarni96:
LGTM
/hold for addressing @jrvaldes comment.

docs/HACKING.md
aravindhp: I am not introducing the `$VERSION_TAG` in this PR but is existing usage. Changes like this warrants a separate PR given we are using `$*_TAG` in multiple places and should be changed across the board.

PR #418, opened 2021-04-26
WINC-597: Use controller runtime client where possible

sebsoto:
Replaces usage of the client-go clientset with the client.Client interface from controller runtime where possible in both nodeconfig and the Windows Machine controller

### DIFF
--- a/pkg/nodeconfig/nodeconfig.go
+++ b/pkg/nodeconfig/nodeconfig.go
@@ -15,8 +15,8 @@ import (
 	core "k8s.io/api/core/v1"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
-	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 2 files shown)


mansikulkarni96:
Thanks for working on this @sebsoto , the changes LGTM although metrics tests are failing.

aravindhp:
/hold

until @sebsoto figures out why the metrics tests are failing

pkg/nodeconfig/nodeconfig.go
```diff
@@ -44,8 +44,8 @@ const (
 // nodeConfig holds the information to make the given VM a kubernetes node. As of now, it holds the information
 // related to kubeclient and the windowsVM.
 type nodeConfig struct {
-	// k8sclientset holds the information related to kubernetes clientset
-	k8sclientset *kubernetes.Clientset
```
aravindhp, line 48: Looks like we will need the kubernetes clientset for using the [drain helper](https://pkg.go.dev/k8s.io/kubectl/pkg/drain#Helper) which will be used to cordon the node. Given that I suggest not removing this.

PR #419, opened 2021-04-26
[hack] Explicitly apply monitoring label

mansikulkarni96:
This PR explicitly applies the monitoring label to be used when deploying the operator through OLM using the hack/olm.sh script to enable metrics configuration. This change needs to be made to align with the changes made as part of OSDK migration.
Prior to the migration this label was applied through the deploy/namespace.yaml which has now been replaced by a section of config/manager/manager.yaml.

### DIFF
--- a/hack/common.sh
+++ b/hack/common.sh
@@ -114,6 +114,11 @@ run_WMCO() {
       fi
   fi
 
+  # Add the "openshift.io/cluster-monitoring:"true"" label to the operator namespace to enable monitoring
+  if ! oc label ns $WMCO_DEPLOY_NAMESPACE openshift.io/cluster-monitoring=true --overwrite; then
+      return 1
+  fi
+
   if [ -n "$PRIVATE_KEY" ]; then
       if ! oc get secret cloud-private-key -n openshift-windows-machine-config-operator; then
           echo "Creating private-key secret"


sebsoto:
Why does this need to be done if the label is present in the config file?
```
apiVersion: v1
kind: Namespace
metadata:
  labels:
    openshift.io/cluster-monitoring: "true"
    control-plane: controller-manager
  name: system
```

hack/common.sh
```diff
@@ -114,6 +114,11 @@ run_WMCO() {
       fi
   fi
 
+  # Add the "openshift.io/cluster-monitoring:"true"" label to the operator namespace to enable monitoring
+  if ! oc patch ns $WMCO_DEPLOY_NAMESPACE -p '{"metadata":{"labels":{"openshift.io/cluster-monitoring":"true"}}}'; then
```
aravindhp, line 118: I suggest using `oc label ns` rather than patching it given the intent is to only add a label.

aravindhp (changes requested):

hack/common.sh
jrvaldes, line 118: Will `openshift-windows-machine-config-operator` be a valid value for  `$WMCO_DEPLOY_NAMESPACE` ?

mansikulkarni96: yes since it is the operator namespace we are creating in the step above and using it here.

jrvaldes: So, just wondering; When to use the `$WMCO_DEPLOY_NAMESPACE` variable or the hard-coded value `openshift-windows-machine-config-operator`?

Refer to line #123, #135, etc
image

mansikulkarni96: that's a good point @jrvaldes, I would think we could have reused the variable declared [here](https://github.com/openshift/windows-machine-config-operator/blob/aff550fdba1d688d63d4d4f9df1773e3a3eb9e35/hack/common.sh#L5) when creating the ns as well. @sebsoto is there are reason to use the hard coded value or we can stick to reusing the ENV var in the lines pointed out?

jrvaldes: In addition,  about [line #112](https://github.com/openshift/windows-machine-config-operator/blob/aff550fdba1d688d63d4d4f9df1773e3a3eb9e35/hack/common.sh#L112), I feel the `WMCO_DEPLOY_NAMESPACE` value is getting ignored. 

image

aravindhp: I want to call out that fixing the usage of `$WMCO_DEPLOY_NAMESPACE` consistently is outside the scope of the PR. @jrvaldes feel free to open a PR that fixes this and we can move this discussion over there.

jrvaldes: Agree

jrvaldes (approved):

hack/common.sh
sebsoto: > is there are reason to use the hard coded value or we can stick to reusing the var in the lines pointed out?

No, we should stick to using $WMCO_DEPLOY_NAMESPACE wherever possible

aravindhp (approved):

/approve

PR #420, opened 2021-04-27
[hack] fix consistent use of parameter

jrvaldes:
This PR align the usage of the WMCO_DEPLOY_NAMESPACE parameter
within the `hack/common.sh` script.

See: https://github.com/openshift/windows-machine-config-operator/pull/419


hack/common.sh
```diff
@@ -2,7 +2,9 @@
 
 # Location of the manifests file
 MANIFEST_LOC=bundle/
-WMCO_DEPLOY_NAMESPACE=openshift-windows-machine-config-operator
+
+# define namespace
+declare -r WMCO_DEPLOY_NAMESPACE=openshift-windows-machine-config-operator
```
sebsoto, line 7: https://tldp.org/LDP/abs/html/declareref.html

For anyone else who hadn't seen this before

sebsoto: I'm thinking we should also use this definition here:
https://github.com/openshift/windows-machine-config-operator/blob/master/hack/run-ci-e2e-test.sh#L16
https://github.com/openshift/windows-machine-config-operator/blob/master/hack/run-ci-e2e-test.sh#L32

This e2e `source`'s hack/common.sh so the variable should be available

aravindhp (changes requested):
Please follow a rebase workflow as mentioned in the contribution guide. Your PR should not have any merge commits.

PR #421, opened 2021-04-28
Generalize nodeconfig to work with VMs not sourced from Machines

sebsoto:
Changes that need to be made in order for Bring Your Own Host (BYOH) to be possible.

### DIFF
--- a/controllers/windowsmachine_controller.go
+++ b/controllers/windowsmachine_controller.go
@@ -371,7 +371,7 @@ func (r *WindowsMachineReconciler) Reconcile(ctx context.Context, request ctrl.R
 
 	log.Info("processing")
 	// Make the Machine a Windows Worker node
-	if err := r.addWorkerNode(ipAddress, instanceID, machine.Name, r.platform); err != nil {
+	if err := r.addWorkerNode(ipAddress, instanceID, machine.Name); err != nil {
 		var authErr *windows.AuthErr
 		if errors.As(err, &authErr) {
 			// SSH authentication errors with the Machine are non recoverable, stemming from a mismatch with the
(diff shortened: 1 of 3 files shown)


pkg/windows/windows.go
```diff
@@ -428,14 +419,14 @@ func (vm *windows) ensureHostName() error {
...
+		return false, errors.Wrapf(err, "error getting the host name, with stdout %s", out)
 	}
-	return !strings.Contains(out, vm.hostName), nil
+	return !strings.Contains(out, vm.machineName), nil
 }
 
 // changeHostName changes the hostName of the Windows VM to match with the machine name
 func (vm *windows) changeHostName() error {
-	changeHostNameCommand := "Rename-Computer -NewName " + vm.hostName + " -Force -Restart"
+	changeHostNameCommand := "Rename-Computer -NewName " + vm.machineName + " -Force -Restart"
```
jrvaldes, line 429: Consider building the resulting `string` with a string formatting approach,  `fmt.Sprintf()` maybe. 

Should we entertain a default locale for  `vm.machineName`?

jrvaldes (approved):
/lgtm

controllers/windowsmachine_controller.go
```diff
@@ -424,7 +436,7 @@ func (r *WindowsMachineReconciler) addWorkerNode(ipAddress, instanceID, machineN
 		return errors.Wrapf(err, "failed to configure Windows VM %s", instanceID)
 	}
 
-	r.log.Info("Windows VM has been configured as a worker node", "ID", nc.ID())
+	r.log.Info("Windows VM has been configured as a worker node", "ID", instanceID)
```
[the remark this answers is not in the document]
sebsoto: It will be a key value pair

pkg/windows/windows.go
```diff
@@ -157,13 +157,15 @@ type windows struct {
...
+	// machineName is the name of the Machine to be configured. This will only be set if the source of the VM is a
+	// Machine. In the case of BYOH, this field will be empty. The name of the Machine must be the same as the
+	// hostname of the associated VM. This is currently not true in the case of vSphere VMs provisioned by MAPI.
+	// In case of Linux, ignition was handling it. As we don't have an equivalent of ignition in Windows, we are
+	// WMCO must correct this by changing the VM's hostname.
 	// TODO: Remove this once we figure out how to do this via guestInfo in vSphere
 	// 		https://bugzilla.redhat.com/show_bug.cgi?id=1876987
-	hostName string
-	log      logr.Logger
+	machineName string
```
aravindhp, line 165: I prefer leaving this as `hostName` and stating that if `hostName` is not set it implies that WMCO does not have to explicitly set the host name for vSphere instances. It is possible we might need to set the host name even in cases where the provider is `None`.

I suggest thinking more about how we can abstract the existing `machineName` input i.e. `NodeConfig` should not have logic to figure out whether the request came from a `Machine` or `ConfigMap` or special casing for that matter.

controllers/windowsmachine_controller.go
```diff
@@ -76,6 +76,8 @@ type WindowsMachineReconciler struct {
 	// 		 in vSphere
 	//		 https://bugzilla.redhat.com/show_bug.cgi?id=1876987
 	platform oconfig.PlatformType
+	// username is the user that will be logged into on each VM in order to configure the Machine.
+	username string
```
aravindhp, line 80: Why is `username` a property of the reconciler while IP address is not? `username` should be treated the same way. This should just be an input to `addWorkerNode`.

aravindhp (changes requested):
@sebsoto thanks for an early look at your direction. PTAL at my comments.

controllers/windowsmachine_controller.go
sebsoto: I made it a property of the reconciler as username will be the same for all Machines reconciled, so it didnt make sense to me to figure out the username based on the platform, every time a Machine is added. IP address will be different for each Machine, so it is not comparable.

aravindhp: Please think from a BYOH perspective where username can be different per instance. What will you do in that scenario?

controllers/windowsmachine_controller.go
```diff
@@ -371,7 +371,7 @@ func (r *WindowsMachineReconciler) Reconcile(ctx context.Context, request ctrl.R
 
 	log.Info("processing")
 	// Make the Machine a Windows Worker node
-	if err := r.addWorkerNode(ipAddress, instanceID, machine.Name, r.platform); err != nil {
+	if err := r.addWorkerNode(ipAddress, instanceID, machine.Name); err != nil {
```
aravindhp, line 374: You are also changing `addWorkerNode` but not calling that out.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -87,7 +86,7 @@ func discoverKubeAPIServerEndpoint() (string, error) {
 // NewNodeConfig creates a new instance of nodeConfig to be used by the caller.
 // hostName having a value will result in the VM's hostname being changed to the given value.
 func NewNodeConfig(clientset *kubernetes.Clientset, ipAddress, hostname, clusterServiceCIDR,
-	vxlanPort string, signer ssh.Signer, platform oconfig.PlatformType) (*nodeConfig, error) {
+	vxlanPort, username string, signer ssh.Signer) (*nodeConfig, error) {
```
aravindhp, line 89: Every commit is changing the signature of `NewNodeConfig`. Not sure if this is a case of too many commits and they all should be squashed with a clear explanation in the commit message 😃

aravindhp (approved):
/approve

mansikulkarni96:
/lgtm

PR #422, opened 2021-04-28
[build] Remove generate requirement from building

sebsoto:
Removes the `generate` target as a prerequisite of the `build` target.
Generate runs `go get`, which cannot be done when building with CpaaS.


Makefile
```diff
@@ -88,7 +88,7 @@ GO_MOD_FLAGS ?= -mod=vendor
 endif
 
 .PHONY: build
-build: generate fmt vet
+build: fmt vet
```
aravindhp, line 91: Please add a comment to the endpoint and the commit message that `make generate` needs to be called when adding a new CRD and will not happen by running `make build`.

aravindhp (changes requested):

Makefile
aravindhp: Please add the generated `bin` directory to `.gitignore` as part of this PR.

sebsoto: Done

aravindhp (approved):

/approve

PR #424, opened 2021-04-28
Bug 1953692: Fix node being shown as Ready after failed config

aravindhp:
- Cordon the node at beginning of a re-configuration or after the kubelet has been configured
- Uncordon the node after successful configuration
- In case of a configuration error, stop the kubelet before requeuing so that the node is marked as NotReady
- Manually added `k8s.io/kubectl v0.21.0-rc.0` to go.mod as v0.18.0 was getting added by default
- Ran `go mod tidy && go mod vendor && go mod verify`
- Added `[hack] Simplify verify-vendor.sh` as a bug in the script was causing the lint test to fail and blocking the PR from merging


controllers/windowsmachine_controller.go
```diff
@@ -420,6 +420,9 @@ func (r *WindowsMachineReconciler) addWorkerNode(ipAddress, instanceID, machineN
 		return errors.Wrapf(err, "failed to configure Windows VM %s", instanceID)
 	}
 	if err := nc.Configure(); err != nil {
+		if err := nc.MarkNotReady(); err != nil {
+			r.log.Info("Unable to mark node as NotReady", "error", err)
+		}
```
sebsoto, line 425: I dont like this having to be called if Configure() fails.
nc.Configure() should be properly cleaning up after itself on failure, and not leave partially configured nodes dangling.

sebsoto: Ideas for how to deal with this below

pkg/nodeconfig/nodeconfig.go
```diff
@@ -142,13 +144,28 @@ func getClusterAddr(kubeAPIServerEndpoint string) (string, error) {
...
+			nc.log.Info("unable to cordon", "node", nc.node.GetName(), "error", err)
+		}
+	}
+
 	if err := nc.Windows.Configure(); err != nil {
 		return errors.Wrap(err, "configuring the Windows VM failed")
 	}
-	// populate node object in nodeConfig
-	if err := nc.setNode(); err != nil {
+
```
sebsoto, line 159: This is the point where a partially configured node is now present in the cluster. In order for Configure() to clean up properly before returning error, the rest of this function can be either be wrapped in an anonymous function, such as:
```
err := func() error {
	// all configuration steps after the Node object has been created
}()
if err != nil {
	// note: markNotReady is no longer exported
	if markErr := nc.markNotReady(); err != nil {
		nc.log.Info("Unable to mark node as NotReady", "error", markErr)
	}
	return err
}
return nil
```

Or alternatively it can be moved to a defined function.

pkg/windows/windows.go
```diff
@@ -135,6 +136,8 @@ type Windows interface {
 	ConfigureWindowsExporter() error
 	// ConfigureKubeProxy ensures that the kube-proxy service is running
 	ConfigureKubeProxy(string, string) error
+	// StopKubelet stop the kubelet service
```
sebsoto, line 139: `stop` -> `stops`

sebsoto (changes requested):

pkg/nodeconfig/nodeconfig.go
aravindhp: This looks cool. I will give it a shot.

pkg/windows/windows.go
```diff
@@ -408,6 +411,10 @@ func (vm *windows) ConfigureKubeProxy(nodeName, hostSubnet string) error {
 	return nil
 }
 
+func (vm *windows) StopKubelet() error {
+	return vm.ensureServiceNotRunning(&service{name: kubeletServiceName})
```
aravindhp, line 415: @mansikulkarni96 is it safe to do this or do I need to stop all services by calling `vm.ensureRequiredServicesStopped()`?

mansikulkarni96: I think you will need to call vm.ensureRequiredServicesStopped() before doing vm.ensureServiceNotRunning()  otherwise we will run into an error from WMCB : trying to stop a service that other services are dependent on.

aravindhp: `vm.ensureRequiredServicesStopped()` ensures that the kubelet is not running, so why would we need to call `vm.ensureServiceNotRunning(&service{name: kubeletServiceName})` in addition?

mansikulkarni96: that's right we don't need to do it in addition, edited my comment a moment ago.

aravindhp: Ah you said instead. Sorry I must have misread.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -142,33 +144,65 @@ func getClusterAddr(kubeAPIServerEndpoint string) (string, error) {
...
+		nc.addPubKeyHashAnnotation()
+		node, err := nc.k8sclientset.CoreV1().Nodes().Update(context.TODO(), nc.node, meta.UpdateOptions{})
+		if err != nil {
+			return errors.Wrap(err, "error updating node labels and annotations")
+		}
+		nc.node = node
+
+		// Uncordon the node now that it is fully configured
+		if err := drain.RunCordonOrUncordon(drainHelper, nc.node, false); err != nil {
+			return errors.Wrapf(err, "error uncording the node %s", nc.node.GetName())
```
mansikulkarni96, line 193: error uncordoning*

pkg/windows/windows.go
```diff
@@ -408,6 +410,10 @@ func (vm *windows) ConfigureKubeProxy(nodeName, hostSubnet string) error {
 	return nil
 }
 
+func (vm *windows) StopKubelet() error {
```
mansikulkarni96, line 413: do we need this function anymore?

mansikulkarni96 (commented):

pkg/windows/windows.go
aravindhp: No we don't. I missed deleting it.

mansikulkarni96:
changes LGTM, will add the label after lint job is fixed.

mansikulkarni96:
Thansk for working on this @aravindhp 
/lgtm

PR #425, opened 2021-04-29
Bug 1948037: Enable and rename metric windows_cpu_info

mansikulkarni96:
After the windows_exporter update to v0.16.0, it is possible to collect metric windows_cpu_info that is equivalent to node_cpu_info on the linux side. This PR enables cpu_info collector since it is not enabled by default.
Additionally renaming the metric windows_cpu_info to node_cpu_info  through the PrometheusRule object, this will help in retrieving the telemetry info for Windows nodes collected using the recording rule node_role_os_version_machine:cpu_capacity_cores:sum based on node_cpu_info.


aravindhp (approved):

/approve

PR #435, opened 2021-05-03
Make Machines created by the e2e suite identifiable

sebsoto:
* The special casing for AWS MachineSets was removed, AWS MachineSets
are now using the same naming convention as Azure and vSphere.
* Adds a label to all Machines and MachineSets created by the e2e, allowing for Windows Machines without the 'Windows" label to be found.


jrvaldes (approved):
/lgtm

test/e2e/delete_test.go
```diff
@@ -51,9 +36,8 @@ func testWindowsNodeDeletion(t *testing.T) {
 	require.NoError(t, err, "Windows node deletion failed")
 
 	// Cleanup all the MachineSets created by us.
-	for _, machineSet := range e2eMachineSets {
-		assert.NoError(t, testCtx.deleteMachineSet(machineSet), "error deleting MachineSet")
-	}
+	assert.NoError(t, testCtx.ensureMachineSetIsDeleted(machineSetWithLabel), "error deleting MachineSet")
+	assert.NoError(t, testCtx.ensureMachineSetIsDeleted(machineSetWithoutLabel), "error deleting MachineSet")
```
mansikulkarni96, line 40: can use a different error message here which will help to differentiate if the tests fail.

test/e2e/providers/aws/aws.go
```diff
@@ -380,7 +377,7 @@ func (a *awsProvider) GenerateMachineSet(withWindowsLabel bool, replicas int32)
 	// Set up the test machineSet
 	machineSet := &mapi.MachineSet{
 		ObjectMeta: meta.ObjectMeta{
-			Name:      machineSetName + rand.String(4),
+			Name:      name,
```
mansikulkarni96, line 380: would name be unique if we create more than 1 machineSet?

mansikulkarni96 (commented):

test/e2e/providers/aws/aws.go
sebsoto: If we need to create more than 1 machineset that name differentiation should be handled at the function call site. additionally none of the other cloud platforms had this randomness added to the machineset name

test/e2e/providers/aws/aws.go
```diff
@@ -325,14 +324,14 @@ func (a *awsProvider) GenerateMachineSet(withWindowsLabel bool, replicas int32)
...
+	machineSetName := "e2e-wmco"
 	matchLabels := map[string]string{
 		"machine.openshift.io/cluster-api-cluster": clusterName,
 	}
 
 	if withWindowsLabel {
-		matchLabels[windowsLabel] = "Windows"
-		machineSetName = machineSetName + "with-windows-label-"
+		matchLabels[clusterinfo.MachineOSIDLabel] = "Windows"
+		machineSetName = machineSetName + "w"
```
aravindhp, line 335: Given that you are standardizing... The end result `e2e-wmcow` looks strange. I would rather we come up with a better naming convention.
```go
machineSetName := e2e-
if withWindowsLabel {
    machineSetName += "wmco"
} else {
    machineSetName += "nolabel"
}
```
Or something along those lines.

test/e2e/clusterinfo/constants.go
```diff
@@ -1,10 +1,12 @@
...
-	MachineOSIDLabel    = "machine.openshift.io/os-id"
-	MachineSetLabel     = "machine.openshift.io/cluster-api-machineset"
-	MachineRoleLabel    = "machine.openshift.io/cluster-api-machine-role"
-	MachineTypeLabel    = "machine.openshift.io/cluster-api-machine-type"
+	MachineOSIDLabel = "machine.openshift.io/os-id"
+	MachineSetLabel  = "machine.openshift.io/cluster-api-machineset"
+	MachineRoleLabel = "machine.openshift.io/cluster-api-machine-role"
+	MachineTypeLabel = "machine.openshift.io/cluster-api-machine-type"
+	// MachineE2ELabel signifies that the Machine was created as part of the WMCO e2e tests
+	MachineE2ELabel     = "wmco-e2e"
```
aravindhp, line 9: I suggest being consistent and sticking with `e2e-wmco`.

test/e2e/delete_test.go
```diff
@@ -23,42 +22,45 @@ func testWindowsNodeDeletion(t *testing.T) {
...
-	err = testCtx.waitForWindowsNodes(gc.numberOfNodes, true, true, false)
-	require.NoError(t, err, "Windows node deletion failed")
 
-	// Cleanup all the MachineSets created by us.
-	for _, machineSet := range e2eMachineSets {
-		assert.NoError(t, testCtx.deleteMachineSet(machineSet), "error deleting MachineSet")
-	}
-	// Phase is ignored during deletion, in this case we are just waiting for Machines to be deleted.
-	require.NoError(t, testCtx.waitForWindowsMachines(int(gc.numberOfNodes), ""), "Windows machine deletion failed")
+	t.Run("Windows Nodes scale down with Machines", func(t *testing.T) {
```
aravindhp, line 37: Lets be less verbose in the naming of tests... `Nodes scale down`

test/e2e/delete_test.go
```diff
@@ -23,42 +22,45 @@ func testWindowsNodeDeletion(t *testing.T) {
...
+		// we are waiting 10 minutes for all windows machines to get deleted.
+		err = testCtx.waitForWindowsNodes(gc.numberOfNodes, true, true, false)
+		require.NoError(t, err, "Windows node deletion failed")
+	})
+
+	t.Run("Prometheus configuration is updated when Machines are deleted", func(t *testing.T) {
+		// Cleanup all the MachineSets created by us.
+		for _, machineSet := range e2eMachineSets.Items {
+			assert.NoError(t, testCtx.deleteMachineSet(&machineSet), "error deleting MachineSet")
+		}
```
aravindhp, line 57: Why are you doing this as part of this test? This should happen irrespective or am I missing something?

test/e2e/delete_test.go
```diff
@@ -23,42 +22,45 @@ func testWindowsNodeDeletion(t *testing.T) {
...
+		_, err = testCtx.client.Machine.MachineSets(clusterinfo.MachineAPINamespace).Update(context.TODO(),
+			windowsMachineSetWithLabel, meta.UpdateOptions{})
+		require.NoError(t, err, "error updating Windows MachineSet")
+
+		// we are waiting 10 minutes for all windows machines to get deleted.
+		err = testCtx.waitForWindowsNodes(gc.numberOfNodes, true, true, false)
+		require.NoError(t, err, "Windows node deletion failed")
+	})
+
+	t.Run("Prometheus configuration is updated when Machines are deleted", func(t *testing.T) {
```
aravindhp, line 53: `Prometheus configuration`

aravindhp (changes requested):
Thanks for working on this @sebsoto. Please address my comments.

test/e2e/providers/aws/aws.go
sebsoto: As there is a limit of 9 characters I was limited in doing that.
I ended up using `e2e-wm` for machines meant to be configured by the Windows Machine controller, let me know what you think

test/e2e/providers/aws/aws.go
```diff
@@ -325,14 +324,16 @@ func (a *awsProvider) GenerateMachineSet(withWindowsLabel bool, replicas int32)
...
 		"machine.openshift.io/cluster-api-cluster": clusterName,
 	}
 
 	if withWindowsLabel {
-		matchLabels[windowsLabel] = "Windows"
-		machineSetName = machineSetName + "with-windows-label-"
+		matchLabels[clusterinfo.MachineOSIDLabel] = "Windows"
+		// Designate MachineSets that set a Windows label on the Machine with
+		// "e2e-wm", to signify they should be configured by the Windows Machine controller
+		machineSetName += "-wm"
```
aravindhp, line 337: So you have copy pasted this code to three places and will have to do so as add more providers. Isn't this a hint that it needs to be modularized?

test/e2e/delete_test.go
```diff
@@ -23,42 +22,46 @@ func testWindowsNodeDeletion(t *testing.T) {
...
+		assert.NoError(t, testCtx.deleteMachineSet(&machineSet), "error deleting MachineSet")
 	}
-	// Phase is ignored during deletion, in this case we are just waiting for Machines to be deleted.
-	require.NoError(t, testCtx.waitForWindowsMachines(int(gc.numberOfNodes), ""), "Windows machine deletion failed")
 
-	// Test if prometheus configuration is updated to have no node entries in the endpoints object
-	testPrometheus(t)
+	t.Run("Prometheus configuration", func(t *testing.T) {
+		// Phase is ignored during deletion, in this case we are just waiting for Machines to be deleted.
+		require.NoError(t, testCtx.waitForWindowsMachines(int(gc.numberOfNodes), ""), "Windows machine deletion failed")
```
aravindhp, line 60: This again as nothing to do with the prometheus test, so why is it part of the test? There could be future scenarios where we could decide to skip the test but don't waiting for the machines to be deleted at the call site will result in issues.

aravindhp (changes requested):

test/e2e/delete_test.go
sebsoto: The connection to the prometheus is test is that Machines being fulled scaled down is a requirement of the test. The test is testing that the windows-exporter service has been updated to remove all the Windows node. 

I can move it out of the test in case of a future scenario like you described

aravindhp: So the issue arose from splitting things out in to sub-tests. Having sub-tests are useful and they typically imply that they can be called on their own if needed. Do we really need a sub-test for node scale down that does not check if the Prometheus configuration worked?

test/e2e/providers/aws/aws.go
```diff
@@ -335,8 +336,8 @@ func (a *awsProvider) GenerateMachineSet(withWindowsLabel bool, replicas int32)
 	matchLabels["machine.openshift.io/cluster-api-machineset"] = machineSetName + *subnet.AvailabilityZone
```
aravindhp, line 336: Nit: given you fixing the constant usage, I suggest fixing this too.

aravindhp (approved):
/approve

jrvaldes (approved):
/lgtm

PR #438, opened 2021-05-06
[hack] refactor consistent use of parameter

jrvaldes:
This PR align the usage of the WMCO_DEPLOY_NAMESPACE parameter
within the `hack/common.sh` script.

See: https://github.com/openshift/windows-machine-config-operator/pull/419#discussion_r621435105

### DIFF
--- a/hack/common.sh
+++ b/hack/common.sh
@@ -2,7 +2,9 @@
 
 # Location of the manifests file
 MANIFEST_LOC=bundle/
-WMCO_DEPLOY_NAMESPACE=openshift-windows-machine-config-operator
+
+# define namespace
+declare -r WMCO_DEPLOY_NAMESPACE=openshift-windows-machine-config-operator
 
 error-exit() {
     echo "Error: $*" >&2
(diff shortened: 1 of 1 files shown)


aravindhp (approved):

One suggestion would be link to comment, https://github.com/openshift/windows-machine-config-operator/pull/419#discussion_r621435105, rather than the PR in the commit and PR message.

sebsoto:
@jrvaldes 
Can you rework the git history into a singular commit?
Having 3 commits for a change like this is a bit much, we also want to avoid having merge commits that arent caused by the PR itself merging.

jrvaldes:
> Leaving this in a review so its more visible:
> 
> @jrvaldes
> Can you rework the git history into a singular commit?
> Having 3 commits for a change like this is a bit much, we also want to avoid having merge commits that arent caused by the PR itself merging.

The interactive rebase worked like a charm

@sebsoto  thanks for the feedback, this is complete. /PTAL

aravindhp (approved):
/approve

PR #444, opened 2021-05-10
[wmco] Remove error from GetServiceCIDR() returns

sebsoto:
An error does not need to be returned, so remove it from the signature
and correct its call sites.


aravindhp (approved):
/approve

saifshaikh48 (approved):
/lgtm

PR #406, opened 2021-04-15
WINC-501: Upgrade to Operator SDK 1.6.1

sebsoto:

PR #446, opened 2021-05-10
WINC-582: Enable adding BYOH instances

sebsoto:
This PR enables any Windows instance which meets the stated requirements to be added to an OpenShift cluster


docs/HACKING.md
```diff
@@ -81,7 +80,7 @@ Additional flags that can be passed to `hack/run-ci-e2e-test.sh` are
 
 Example command to spin up 2 Windows nodes and retain them after test run:
```
jrvaldes, line 81: Consider rewording the message, since the `NODE_COUNT` is no longer available.

aravindhp (changes requested):
Nice work, @sebsoto!

We should not be using `VM`. Please use `instance` or come up with a better term. In general we need to clean up the terminology. We use hosts, instances, VM interchangeably. We should pick a term and stick to that everywhere. In addition please rephrase `Enable adding Windows VMs not sourced from Machines to a cluster` 😆

controllers/configmap_controller.go
```diff
@@ -18,45 +18,243 @@ package controllers
...
+
+//+kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups="",resources=configmaps/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups="",resources=configmaps/finalizers,verbs=update
+
+const (
+	// BYOHAnnotation is an an anotation that should be applied to all Windows nodes not associated with a Machine.
+	BYOHAnnotation = "windowsmachineconfig.openshift.io/byoh"
+	// HostsConfigMap is the name of the ConfigMap where VMs to be configured should be described.
+	HostsConfigMap = "windows-instances"
```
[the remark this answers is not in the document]
sebsoto: As in we create it and the user adds their data?
Wouldnt that mean that the user would be forced to use `oc edit` instead of being able to apply a config file? That seems like a large downside to me

controllers/configmap_controller.go
```diff
@@ -18,45 +18,243 @@ package controllers
...
+	}
+	return hosts, nil
+
+}
+
+// ensureNodeExists ensures that the given host has an associated Node
+func (r *ConfigMapReconciler) ensureNodeExists(host *hostInfo, nodes *core.NodeList) error {
+	_, found := findNode(host.address, nodes)
+	if found {
+		// TODO: Remove and re-add the node if it is not 'Ready'
```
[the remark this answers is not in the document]
sebsoto: I'm actually not sure if this should be done. What do you think? MachineHealthChecks must not be default for a reason, some clusters must have reasons for nodes being not ready for a period of time.

controllers/configmap_controller.go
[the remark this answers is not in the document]
aravindhp: The user can do `oc apply -f windows-instances.yaml` and the object will get updated. This is how other OpenShift config objects work.

controllers/configmap_controller.go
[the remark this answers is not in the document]
aravindhp: Please bring this up at the next stand up.

test/e2e/create_test.go
```diff
@@ -44,50 +44,114 @@ func testWindowsNodeCreation(t *testing.T) {
...
-			// Replace the known private key with a randomly generated one.
-			err = testCtx.createPrivateKeySecret(false)
-			require.NoError(t, err, "error replacing private key secret")
+// testBYOHConfiguration tests that the ConfigMap controller can properly configure VMs
+func (tc *testContext) testBYOHConfiguration(t *testing.T) {
+	if gc.numberOfBYOHNodes == 0 {
+		t.Skip("BYOH testing disabled")
+	}
+	_, err := tc.createWindowsMachineSet(gc.numberOfBYOHNodes, false)
+	machines, err := tc.waitForWindowsMachines(int(gc.numberOfBYOHNodes), "Provisioned", false)
```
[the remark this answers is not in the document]
aravindhp: >The issue with that is that the successful creation of the MachineSet does not imply that all the Machines have been created,

Correct. That is what `waitForWindowsMachines()` does.

> or that the Machines originally created will be the ones that hit the Provisioned state. 

This seems to imply that the list of `Machines` should be an input to `waitForWindowsMachines()` instead of it trying to heuristically figure out the `Machines`. WDYT?

test/e2e/create_test.go
```diff
@@ -200,14 +271,26 @@ func (tc *testContext) waitForWindowsNodes(nodeCount int32, waitForAnnotations,
...
+		for _, node := range nodes.Items {
+			if node.Annotations[controllers.BYOHAnnotation] == "true" {
+				if isBYOH {
+					foundNodes = append(foundNodes, node)
+				}
+			} else {
+				if !isBYOH {
+					foundNodes = append(foundNodes, node)
+				}
+			}
```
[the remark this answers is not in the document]
aravindhp: I find the way it is written unreadable and confusing 😆 which is why I commented.

sebsoto:
I think the vSphere test is failing because vSphere machines can be created without an IP, and will get one later.  waitForWindowsMachines needs to wait for the IP to present

controllers/configmap_controller.go
```diff
@@ -18,45 +18,229 @@ package controllers
...
+		return errors.Wrap(err, "error configuring node")
+	}
+
+	return nil
+}
 
-	return ctrl.Result{}, nil
+// findNode returns a pointer to the node with an address matching the given address and a bool indicating if the node
+// was found or not.
+func findNode(address string, nodes *core.NodeList) (*core.Node, bool) {
```
mansikulkarni96, line 216: looks like the node returned here is not used anywhere, in that case the function can return only a bool

controllers/controllers.go
```diff
@@ -0,0 +1,60 @@
...
+	k8sclientset *kubernetes.Clientset
+	// clusterServiceCIDR holds the cluster network service CIDR
+	clusterServiceCIDR string
+	// watchNamespace is the namespace that should be watched for configmaps
+	watchNamespace string
+	// vxlanPort is the custom VXLAN port
+	vxlanPort string
+	// signer is a signer created from the user's private key
+	signer ssh.Signer
+	// prometheusConfig stores information required to configure Prometheus
```
mansikulkarni96, line 30: prometheusNodeConfig*

controllers/controllers.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// configureInstance adds the specified instance to the cluster. if hostname is not empty, the instance's hostname will be
+// changed to the passed in value. If annotations is not nil, the node will have the specified annotations applied to
+// it.
+func (r *instanceReconciler) configureInstance(info *instanceInfo, hostname string, annotations map[string]string) error {
+	nc, err := nodeconfig.NewNodeConfig(r.k8sclientset, info.address, hostname, r.clusterServiceCIDR,
+		r.vxlanPort, info.username, r.signer, annotations)
+	if err != nil {
+		return errors.Wrapf(err, "failed to configure Windows instance")
```
mansikulkarni96, line 43: error message can be specific to creating the config
errors.Wrap* should suffice

test/e2e/create_test.go
```diff
@@ -126,9 +191,15 @@ func (tc *testContext) waitForWindowsMachines(machineCount int, phase string) er
 			machineStateTimeLimit = time.Minute * 20
 		}
 	}
+
+	listOptions := metav1.ListOptions{LabelSelector: clusterinfo.MachineE2ELabel + "=true"}
+	if windowsLabel {
+		listOptions.LabelSelector += "," + controllers.MachineOSLabel + "=Windows"
+	} else {
+		listOptions.LabelSelector += "," + controllers.MachineOSLabel + "!=Windows"
```
saifshaikh48, line 199: I am not sure I completely understand this part. For my own learning --

1. BYOH nodes will not have the `machine.openshift.io/os-id=Windows` label, correct?
2. What sort of machines would be caught with a label selector of `!=Windows`? All non-Windows machines, including  Linux machines?

controllers/configmap_controller.go
```diff
@@ -18,45 +18,229 @@ package controllers
...
+	}, nil
+}
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
-// TODO(user): Modify the Reconcile function to compare the state specified by
-// the ConfigMap object against the actual cluster state, and then
-// perform operations to make the cluster state reflect the state specified by
-// the user.
-//
```
aravindhp: Is this being addressed in this PR?

controllers/configmap_controller.go
```diff
@@ -18,45 +18,229 @@ package controllers
...
+}
+
+// parseHosts gets the lists of hosts specified in the configmap's data
+func (r *ConfigMapReconciler) parseHosts(configMapData map[string]string) ([]*instanceInfo, error) {
+	hosts := make([]*instanceInfo, len(configMapData))
+	entryCount := 0
+	// Get information about the hosts from each entry. The expected key/value format for each entry is:
+	// <address>: username=<username>
+	for address, data := range configMapData {
+		if err := addressValid(address); err != nil {
```
aravindhp, line 127: `validateAddress` reads better here

controllers/configmap_controller.go
```diff
@@ -18,45 +18,229 @@ package controllers
...
+			return nil, errors.Wrapf(err, "invalid address %s", address)
+		}
+		splitData := strings.SplitN(data, "=", 2)
+		if len(splitData) == 0 || splitData[0] != "username" {
+			return hosts, fmt.Errorf("data for entry %s has an incorrect format", address)
```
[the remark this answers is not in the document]
aravindhp: Please reflect this in all sites

controllers/configmap_controller.go
```diff
@@ -18,45 +18,229 @@ package controllers
...
+
+// addressValid checks that the given address is either an ipv4 address, or resolves to any ip address
+func addressValid(address string) error {
+	// first check if address is an IP address
+	if parsedAddr := net.ParseIP(address); parsedAddr != nil {
+		if parsedAddr.To4() != nil {
+			return nil
+		}
+		// if the address parses into an IP but is not ipv4 it must be ipv6
+		return fmt.Errorf("ipv6 is not supported")
```
aravindhp, line 149: I would rather say `error parsing IPv4 address` and please mention the address in the error. Parsing error need not imply an IPv6 address.

controllers/configmap_controller.go
```diff
@@ -18,45 +18,229 @@ package controllers
...
+	if err != nil {
+		return errors.Wrapf(err, "error looking up DNS")
+	}
+	if len(addressList) == 0 {
+		return fmt.Errorf("DNS did not resolve to an address")
+	}
+	return nil
+}
+
+// addAndRemoveNodes corrects the discrepancy between the "expected" hosts slice, and the "actual" nodelist
```
aravindhp, line 162: Won't `reconcileConfigMap` or `reconcileInstances` or `reconcileNodes` be more appropriate here? `addAndRemoveNodes` implies you are always doing both while you could being doing one or the other.

controllers/configmap_controller.go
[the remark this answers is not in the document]
aravindhp: Shouldn't the two TODOs be merged into one?

controllers/configmap_controller.go
```diff
@@ -0,0 +1,246 @@
...
+		}
+
+		hosts[entryCount] = newInstanceInfo(address, splitData[1])
+		entryCount++
+	}
+	return hosts, nil
+}
+
+// addressValid checks that the given address is either an ipv4 address, or resolves to any ip address
+func addressValid(address string) error {
```
saifshaikh48, line 142: Should this function return a bool instead of using the error/nil as a defacto boolean? Just not sure if this is standard practice

controllers/configmap_controller.go
```diff
@@ -18,45 +18,243 @@ package controllers
...
+
+	return nil
+}
+
+// findNode returns a pointer to the node with an address matching the given address and a bool indicating if the node
+// was found or not.
+func findNode(address string, nodes *core.NodeList) (*core.Node, bool) {
+	for _, node := range nodes.Items {
+		for _, nodeAddress := range node.Status.Addresses {
+			if address == nodeAddress.Address {
```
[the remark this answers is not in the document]
aravindhp: `version` mismatch will be handled by the upgrade story. Will the annotation not being present also be done in that story?

controllers/configmap_controller_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+	testCases := []struct {
+		name        string
+		input       map[string]string
+		expectedOut []*instanceInfo
+		expectedErr bool
+	}{
+		{
+			name:        "invalid username",
+			input:       map[string]string{"localhost": "notusername=core"},
```
aravindhp, line 21: What about `map[string]string{"invalid": "invalid"}`

controllers/controllers.go
```diff
@@ -0,0 +1,60 @@
...
+	signer ssh.Signer
+	// prometheusConfig stores information required to configure Prometheus
+	prometheusNodeConfig *metrics.PrometheusNodeConfig
+	// recorder to generate events
+	recorder record.EventRecorder
+}
+
+// configureInstance adds the specified instance to the cluster. if hostname is not empty, the instance's hostname will be
+// changed to the passed in value. If annotations is not nil, the node will have the specified annotations applied to
+// it.
```
aravindhp, line 38: Reading this description makes me think that setting the hostname is a property of `instanceInfo` i.e. there would be a new field `setHostname bool` in `instanceInfo`. If that is set `instanceInfo.address` will be used to set the hostname. Or can `instanceInfo.address` not be equal to `hostname`? Either way `setHostname` or `hostname` belongs to `instanceInfo`. `newInstanceInfo()` can take `platform` as an input.

docs/HACKING.md
```diff
@@ -70,7 +70,8 @@ hack/run-ci-e2e-test.sh
 
 Additional flags that can be passed to `hack/run-ci-e2e-test.sh` are
 - `-s` to skip the deletion of Windows nodes that are created as part of test suite run
-- `-n` to represent the number of Windows nodes to be created for test run
+- `-m` to represent the number of Windows instances to be configured by the Windows Machine controller
+- `-c` to represent the number of Windows instances to be configured by the ConfigMap controller
```
aravindhp, line 74: Thanks for doing this

docs/HACKING.md
```diff
@@ -79,9 +80,10 @@ Additional flags that can be passed to `hack/run-ci-e2e-test.sh` are
   - `basic` creation, network and deletion tests are run
   - `upgrade` creation, upgrade, reconfiguration and deletion tests are run
 
-Example command to spin up 2 Windows nodes and retain them after test run:
+Example command to run the full test suite with 2 instances configured by the Windows Machine controller, and 1
+configured by the ConfigMap controller, skipping the deletion of all the nodes.
 ```shell script
-hack/run-ci-e2e-test.sh -s -n 2      
+hack/run-ci-e2e-test.sh -s -m 2 -c 1
```
aravindhp, line 86: Wonder what happens if we do this? 🤔

controllers/secret_controller.go
```diff
@@ -171,6 +171,12 @@ func (r *SecretReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 		escapedPubKeyAnnotation := strings.Replace(nodeconfig.PubKeyHashAnnotation, "/", "~1", -1)
 		patchData := fmt.Sprintf(`[{"op":"add","path":"/metadata/annotations/%s","value":""}]`, escapedPubKeyAnnotation)
 		for _, node := range nodes.Items {
+			// Only clear the annotation for Nodes associated with Machines, as the clearing of the annotation is used
+			// solely for to kick off the deletion and recreation of Machines.
```
aravindhp, line 175: `used soley to kick...`

Won't we need to do this for BYOH instances that need to be reconfigured?

test/e2e/create_test.go
```diff
@@ -44,50 +45,114 @@ func testWindowsNodeCreation(t *testing.T) {
...
+	// BYOH creation must occur after the Machine creation, as the MachineConfiguration tests change the private key
+	// multiple times, and BYOH doesnt have the functionality of rotating keys on the VMs. This would result in BYOH
+	// failing the pub key annotation validation as it compares the current private key secret with the annotation.
+	t.Run("ConfigMap controller", testCtx.testBYOHConfiguration)
+
+}
+
+// ensureWindowsInstanceConfigMapDeleted deletes the windows-instances configmap if it exists
+func (tc *testContext) ensureWindowsInstanceConfigMapDeleted() error {
+	err := tc.client.K8s.CoreV1().ConfigMaps(tc.namespace).Delete(context.TODO(), controllers.InstanceConfigMap, metav1.DeleteOptions{})
```
aravindhp, line 58: 120 col?

test/e2e/create_test.go
```diff
@@ -44,50 +45,114 @@ func testWindowsNodeCreation(t *testing.T) {
 	// Create a private key secret with the known private key.
 	require.NoError(t, testCtx.createPrivateKeySecret(true), "could not create known private key secret")
 
-	t.Run("Windows Machines without the Windows label are not configured", func(t *testing.T) {
```
aravindhp, line 47: So glad this test is gone 😄 This will reduce e2e test time.

test/e2e/create_test.go
```diff
@@ -44,50 +45,114 @@ func testWindowsNodeCreation(t *testing.T) {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: controllers.InstanceConfigMap,
+		},
+		Data: make(map[string]string),
+	}
+	for _, machine := range machines.Items {
+		for _, address := range machine.Status.Addresses {
+			if address.Type == v1.NodeInternalIP || address.Type == v1.NodeInternalDNS {
+				if tc.CloudProvider.GetType() == config.VSpherePlatformType && address.Type == v1.NodeInternalDNS {
+					// skip using DNS on vSphere as it is not working
```
aravindhp, line 85: If you know this for a fact, please bring this up with the SPLAT team.

test/e2e/create_test.go
```diff
@@ -44,50 +45,114 @@ func testWindowsNodeCreation(t *testing.T) {
...
+	require.NoError(t, err, "error waiting for Windows Machines to be provisioned")
+	if tc.CloudProvider.GetType() == config.AWSPlatformType {
+		// Replace the known private key with a randomly generated one.
+		err = tc.createPrivateKeySecret(false)
+		require.NoError(t, err, "error replacing private key secret")
+	}
+	err = tc.waitForWindowsNodes(gc.numberOfMachineNodes, false, false, false)
+	assert.NoError(t, err, "Windows node creation failed")
+
+	t.Run("Changing the private key ensures all Windows nodes use the new private key", func(t *testing.T) {
```
aravindhp, line 128: Now that you are moving the code, please rename the test to something succinct. `Change private key` or something to that effect.

test/e2e/create_test.go
```diff
@@ -108,10 +173,10 @@ func (tc *testContext) deleteMachineSet(ms *mapi.MachineSet) error {
 
 // waitForWindowsMachines waits for a certain amount of Windows Machines to reach a certain phase
 // if machineCount = 0, it implies we are only waiting for Machines to be deleted and the phase is
-// ignored in this case.
-func (tc *testContext) waitForWindowsMachines(machineCount int, phase string) error {
+// ignored in this case. Returns the set of Machines that matched the provided windowsLabel criteria.
+func (tc *testContext) waitForWindowsMachines(machineCount int, phase string, windowsLabel bool) (*mapi.MachineList, error) {
```
aravindhp, line 177: What about https://github.com/openshift/windows-machine-config-operator/pull/446#discussion_r633786904?

test/e2e/create_test.go
```diff
@@ -262,16 +332,37 @@ func (tc *testContext) waitForWindowsNodes(nodeCount int32, waitForAnnotations,
...
+			// applied at the end of configuration
+			if node.Annotations[controllers.BYOHAnnotation] == "true" {
+				if isBYOH {
+					windowsNodes = append(windowsNodes, node)
+				}
+			} else {
+				if !isBYOH {
+					windowsNodes = append(windowsNodes, node)
+				}
+			}
```
aravindhp, line 346: I still think this check is strange. Can't you first check `isBYOH` and only then check for the BYOH annotation?

test/e2e/delete_test.go
```diff
@@ -35,25 +35,35 @@ func testWindowsNodeDeletion(t *testing.T) {
 	}
 
 	require.NotNil(t, windowsMachineSetWithLabel, "could not find MachineSet with Windows label")
-	// Set the expected number of Windows nodes to 0
-	gc.numberOfNodes = 0
```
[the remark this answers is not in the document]
aravindhp: What if we add further tests where this number needs to be accurate? Figuring that it needed to be reset here is going to be hard.

aravindhp (changes requested):
Nice work, @sebsoto. We are getting closer.

test/e2e/create_test.go
saifshaikh48: Only other way I could see this refactored is making the clause a one-liner
```
if (isBYOH && node.Annotations[controllers.BYOHAnnotation] == "true") || 
		(!isBYOH && node.Annotations[controllers.BYOHAnnotation] != "true") {
	windowsNodes = append(windowsNodes, node)
}
```

controllers/configmap_controller.go
sebsoto: In this case the reason why its invalid is the most important, so I chose to do just an error.

test/e2e/create_test.go
sebsoto: I dont actually think this is the case, but instead the issue is that the Machine does not have an IP yet even though its in the provisioned state. Making that check in the waitForWindowsMachines function instead

test/e2e/create_test.go
sebsoto: Sorry, missed that.

My concern is that if waitForWindowsMachines takes in a list of machines, createWindowsMachineSet and waitForWindowsMachines are both waiting for a set of Windows machines. waitForWindowsMachines doing what  it does now, and createWindowsMachineSet waiting until Machines matching the label selector are created.

sebsoto: Would changing the name of `waitForWindowsMachines` fix the issue you have?

test/e2e/delete_test.go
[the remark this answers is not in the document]
sebsoto: I'm not fully understanding the case you are describing. My issue is that throughout all the other tests gc.numberOfNodes is set through test arguments. If this needs to be changed in the deletion tests I feel it should be passed in when initiating the tests. Otherwise, normal non-global variables should be used.

controllers/configmap_controller.go
sebsoto: Yes it is, that todo is created as part of the generated file. This PR takes the expected windows-instances and creates nodes out of them. The missing parts of this process were moved into other TODOs

controllers/secret_controller.go
```diff
@@ -137,7 +136,7 @@ func (r *SecretReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 		return reconcile.Result{}, errors.Wrapf(err, "unable to get secret %s", request.NamespacedName)
 	}
 	// Generate expected userData based on the existing private key
-	validUserData, err := secrets.GenerateUserData(privateKey)
+	validUserData, err := secrets.GenerateUserData(keySigner)
```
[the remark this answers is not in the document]
sebsoto: We might want to have a call to talk about this pattern when you are available

controllers/configmap_controller.go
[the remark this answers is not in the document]
aravindhp: >The logic to determine if the node needs to be reconfigured should be done in a function that uses this one.

Sounds good. Always better to separate concerns.

controllers/configmap_controller_test.go
aravindhp: Thanks for adding it. I was asking for a separate case but was not sure where to place the comment.

controllers/controllers.go
aravindhp: Platform is also used to decide whether to set the hostname or not and I was suggesting to encapsulate all of that in the `instanceInfo` object.

test/e2e/create_test.go
aravindhp: Lets discuss this during the meeting.

aravindhp (commented):
@sebsoto I paused on reviewing and figured it is easier to discuss all outstanding issues in a meeting.

README.md
```diff
@@ -1,40 +1,75 @@
...
 # Create secret containing the private key in the openshift-windows-machine-config-operator namespace
 oc create secret generic cloud-private-key --from-file=private-key.pem=/path/to/key -n openshift-windows-machine-config-operator
 ```
- We strongly recommend not using the same
+We strongly recommend not using the same
 [private key](https://docs.openshift.com/container-platform/4.6/installing/installing_azure/installing-azure-default.html#ssh-agent-using_installing-azure-default)
 used when installing the cluster
 
+### Configuring BYOH (Bring Your Own Host) Windows instances
+WARNING: This is not a fully developed feature. Nodes can be removed from the cluster by deleting the Node object,
```
EricPonvelle, line 41: Is there a point when this will be fully developed? Is that planned for this or a future release?

EricPonvelle (commented):
/lgtm

I had one question for my own knowledge/planning.

README.md
sebsoto: This epic tracks this feature set: https://issues.redhat.com/browse/WINC-502

It is currently planned for 3.0.0

controllers/controllers.go
aravindhp: Please remind me what we decided to do regarding this.

test/e2e/create_test.go
```diff
@@ -44,50 +45,111 @@ func testWindowsNodeCreation(t *testing.T) {
...
+	t.Run("Machine controller", testCtx.testMachineConfiguration)
+	// BYOH creation must occur after the Machine creation, as the MachineConfiguration tests change the private key
+	// multiple times, and BYOH doesnt have the functionality of rotating keys on the VMs. This would result in BYOH
+	// failing the pub key annotation validation as it compares the current private key secret with the annotation.
+	t.Run("ConfigMap controller", testCtx.testBYOHConfiguration)
+
+}
+
+// ensureWindowsInstanceConfigMapDeleted deletes the windows-instances configmap if it exists
+func (tc *testContext) ensureWindowsInstanceConfigMapDeleted() error {
```
aravindhp, line 58: Isn't this just `deleteWindowsInstanceConfigMap()` given you are always calling delete?

test/e2e/create_test.go
```diff
@@ -149,6 +218,24 @@ func (tc *testContext) waitForWindowsMachines(machineCount int, phase string) er
 			if machine.Status.Phase == nil || *machine.Status.Phase != phase {
 				return false, nil
 			}
+			// If waiting for a provisioned Machine, ensure the Machine is fully attached to the network and has an
+			// assigned IPv4 address.
+			if phase == "Provisioned" {
```
aravindhp, line 224: I suggest a negation check to reduce the indentation.
```suggestion
                      if phase != "Provisioned" {
                           return...
}
```

test/e2e/delete_test.go
```diff
@@ -35,25 +35,35 @@ func testWindowsNodeDeletion(t *testing.T) {
...
+
 	// Cleanup all the MachineSets created by us.
 	for _, machineSet := range e2eMachineSets.Items {
 		assert.NoError(t, testCtx.deleteMachineSet(&machineSet), "error deleting MachineSet")
 	}
 	// Phase is ignored during deletion, in this case we are just waiting for Machines to be deleted.
-	require.NoError(t, testCtx.waitForWindowsMachines(int(gc.numberOfNodes), ""), "Windows machine deletion failed")
+	_, err = testCtx.waitForWindowsMachines(int(expectedNodeCount), "", true)
+	_, err = testCtx.waitForWindowsMachines(int(expectedNodeCount), "", false)
+	require.NoError(t, err, "Windows machine deletion failed")
```
aravindhp, line 66: This will give a false positive if the first `waitForWindowsMachines` fails

test/e2e/metrics_test.go
```diff
@@ -88,12 +88,14 @@ func testPrometheus(t *testing.T) {
 		metrics.WindowsMetricsResource, metav1.GetOptions{})
 	require.NoError(t, err)
 
-	if gc.numberOfNodes == 0 {
+	if len(gc.allNodes()) == 0 {
 		// check if all entries in subset are deleted when there are no Windows Nodes
-		require.Equal(t, 0, len(windowsEndpoints.Subsets))
+		// TODO: BYOH cleanup isn't done yet so only check that Nodes from Machines are properly removed
```
aravindhp, line 93: Where are we deleting the BYOH Machine then?

test/e2e/network_test.go
```diff
@@ -27,8 +27,10 @@ func testNetwork(t *testing.T) {
 	// Populate the global test context
 	testCtx, err := NewTestContext()
 	require.NoError(t, err)
-	err = testCtx.waitForWindowsNodes(gc.numberOfNodes, true, false, false)
-	assert.NoError(t, err, "timed out waiting for Windows node")
+	err = testCtx.waitForWindowsNodes(gc.numberOfMachineNodes, false, false, false)
+	assert.NoError(t, err, "timed out waiting for Windows nodes configured from Machines")
```
aravindhp, line 31: `Windows Machine nodes` will suffice.

test/e2e/network_test.go
```diff
@@ -27,8 +27,10 @@ func testNetwork(t *testing.T) {
...
+	err = testCtx.waitForWindowsNodes(gc.numberOfBYOHNodes, false, false, true)
```
aravindhp, line 32: We should introduce wrappers for `waitForWindowsNodes` to clear demarcate whether we are waiting for Machines o BYOH nodes. The code is really hard to read now without committing the function signature to memory. But I won't insist.

test/e2e/upgrade_test.go
```diff
@@ -45,6 +45,12 @@ func upgradeTestSuite(t *testing.T) {
 	err = testCtx.configureUpgradeTest()
 	require.NoError(t, err, "error configuring upgrade")
 
+	// get current Windows node state
+	err = testCtx.waitForWindowsNodes(gc.numberOfMachineNodes, false, true, false)
+	require.NoError(t, err, "wrong number of Machine controller nodes found")
+	err = testCtx.waitForWindowsNodes(gc.numberOfBYOHNodes, false, true, true)
+	require.NoError(t, err, "wrong number of ConfigMap controller nodes found")
```
aravindhp, line 52: Why are we waiting for the BYOH nodes here given BYOH nodes are not upgraded?

test/e2e/create_test.go
```diff
@@ -340,6 +350,16 @@ func (tc *testContext) waitForWindowsNodes(nodeCount int32, expectError, checkVe
...
+			// verify the node CSR associated with the node is not waiting for approval.
+			csr, err := tc.findNodeCSR(node.GetName())
+			if err != nil {
+				log.Printf("error finding node CSR: %s", err)
+				return false, nil
+			}
+			if tc.isPending(csr) {
+				log.Printf("node csr is waiting for approval")
+				return false, nil
+			}
```
aravindhp, line 362: Please add comment with the Jira story link to clearly cal out that this section needs to be removed. This applies to all the sections that need to be removed. I am asking for this as I suspect that a clean revert won't work if there is more code churn in this file.

aravindhp (changes requested):

test/e2e/upgrade_test.go
sebsoto: This suite ends up calling `testPrometheus` which requires the BYOH nodes to loaded into the global node variable, which is done by waitForWindowsNodes as the tests currently stand.

I wanted to avoid this PR growing any more so I stuck with this pattern instead of introducing even more code and making things less consistent throughout the e2e's, I'm fine adding extra code to fix this if thats what you think is best, else it can be done within https://issues.redhat.com/browse/WINC-620

test/e2e/network_test.go
sebsoto: I'd like to do this when doing the larger refactor around waiting for nodes and Machines https://issues.redhat.com/browse/WINC-620

test/e2e/create_test.go
sebsoto: I was thinking this was a check that would be useful for the CSR approval ticket. I will add a TODO suggesting its removal

README.md
```diff
@@ -1,40 +1,75 @@
...
+```yaml
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: windows-instances
+  namespace: openshift-windows-machine-config-operator
+data:
+  10.1.42.1: |-
+    username=Administrator
+  instance.dns.com: |-
```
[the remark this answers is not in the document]
sebsoto: Yes

test/e2e/create_test.go
```diff
@@ -262,16 +339,49 @@ func (tc *testContext) waitForWindowsNodes(nodeCount int32, waitForAnnotations,
...
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to list nodes")
+	}
+	var windowsNodes []v1.Node
+	for _, node := range nodes.Items {
+		// filter out nodes that haven't been fully configured
+		if _, present := node.Labels[nodeconfig.VersionAnnotation]; !present {
+			continue
+		}
+		if (isBYOH && node.Annotations[controllers.BYOHAnnotation] == "true") ||
```
mansikulkarni96, line 404: a liitle confused as to why is this check required for appending, aren't we already filtering on the label?

test/e2e/delete_test.go
```diff
@@ -35,25 +35,36 @@ func testWindowsNodeDeletion(t *testing.T) {
...
 	require.NoError(t, err, "error updating Windows MachineSet")
 
 	// we are waiting 10 minutes for all windows machines to get deleted.
-	err = testCtx.waitForWindowsNodes(gc.numberOfNodes, true, true, false)
+	err = testCtx.waitForWindowsNodes(expectedNodeCount, true, false, false)
 	require.NoError(t, err, "Windows node deletion failed")
 
+	t.Run("BYOH node removal", func(t *testing.T) {
+		assert.NoError(t, testCtx.deleteWindowsInstanceConfigMap(),
+			"error removing windows-instances ConfigMap")
```
mansikulkarni96, line 52: nit: "error removing Windows instances from ConfigMap"

test/e2e/delete_test.go
```diff
@@ -35,25 +35,36 @@ func testWindowsNodeDeletion(t *testing.T) {
...
+	})
+
 	// Cleanup all the MachineSets created by us.
 	for _, machineSet := range e2eMachineSets.Items {
 		assert.NoError(t, testCtx.deleteMachineSet(&machineSet), "error deleting MachineSet")
 	}
 	// Phase is ignored during deletion, in this case we are just waiting for Machines to be deleted.
-	require.NoError(t, testCtx.waitForWindowsMachines(int(gc.numberOfNodes), ""), "Windows machine deletion failed")
+	_, err = testCtx.waitForWindowsMachines(int(expectedNodeCount), "", true)
+	require.NoError(t, err, "Windows machine deletion failed")
```
mansikulkarni96, line 65: can we add a specific error message to signify what type of Windows nodes we are deleting in each case.

pkg/secrets/secrets.go
```diff
@@ -9,25 +9,23 @@ import (
...
 	// PrivateKeySecret is the name of the private key secret provided by the user
 	PrivateKeySecret = "cloud-private-key"
 	// PrivateKeySecretKey is the key within the private key secret which holds the private key
 	PrivateKeySecretKey = "private-key.pem"
 )
 
 // GetPrivateKey fetches the specified secret and extracts the private key data
-func GetPrivateKey(secret kubeTypes.NamespacedName, c client.Client) ([]byte, error) {
-	// Fetch the private key secret
+func GetPrivateKey(namespace string, c client.Client) ([]byte, error) {
```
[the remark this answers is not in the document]
aravindhp: In future we should send out "decision notes" at the end of design meetings especially with large PRs. I definitely could be misremembering. But I am fine if you revert.

test/e2e/metrics_test.go
aravindhp: So won't this end up leaking BYOH Machines in CI?

test/e2e/network_test.go
aravindhp: I am fine with that. I updated the story,

test/e2e/upgrade_test.go
aravindhp: No I am fine with it. I keep forgetting `waitForWindowsNodes` has that side effect.

test/e2e/create_test.go
aravindhp: That's a good point. However that decision should be made by whoever is working on the CSR story.

test/e2e/delete_test.go
aravindhp: +1

test/e2e/metrics_test.go
aravindhp: Just to clarify, I want to ensure that we are deleting the `MachineSet` created without the Windows label. Just because BYOH cleanup is not implemented does not mean we can keep leaking Machines until then. Or am I missing the spot where this is getting deleted.

test/e2e/create_test.go
```diff
@@ -385,3 +408,91 @@ func (tc *testContext) listFullyConfiguredWindowsNodes(isBYOH bool) ([]v1.Node,
 	}
 	return windowsNodes, nil
 }
+
+// TODO: vSphere CSRs are not being approved due to a hostname mismatch. This is an issue only because the
```
aravindhp, line 412: Nit: this could have been added to a separate package. which we then could have deleted.

aravindhp (commented):
/almostapprove 😃 

My main concern is leaking of the BYOH MachineSet.

PROJECT
```diff
@@ -1,8 +1,14 @@
 domain: windowsmachineconfig.openshift.io
-layout: go.kubebuilder.io/v3
+layout:
+- go.kubebuilder.io/v3
```
alinaryan, line 3: Why does this need to move below `layout:`?

alinaryan, line 3: Why does `go.kubebuilder.io/v3` need to move below `layout:`?

README.md
```diff
@@ -1,40 +1,75 @@
 # Windows Machine Config Operator
 
 ## Introduction
-The Windows Machine Config Operator configures Windows Machines into nodes, enabling Windows container workloads to
-be ran within OKD/OCP clusters. The operator is configured to watch for [Machines](https://docs.openshift.com/container-platform/4.5/machine_management/creating_machinesets/creating-machineset-aws.html#machine-api-overview_creating-machineset-aws) with a `machine.openshift.io/os-id: Windows`
-label. The way a user will initiate the process is by creating a MachineSet which uses a Windows image with the Docker
-container runtime installed. The operator will do all the necessary steps to configure the underlying VM so that it can join
-the cluster as a worker node.
+The Windows Machine Config Operator configures Windows instances into nodes, enabling Windows container workloads to
```
alinaryan, line 4: Is instance an all encompassing word for instances and machines? Should we say `configures instances and machines into nodes`

controllers/configmap_controller.go
```diff
@@ -0,0 +1,242 @@
...
+	"github.com/openshift/windows-machine-config-operator/pkg/cluster"
+	"github.com/openshift/windows-machine-config-operator/pkg/instances"
+	"github.com/openshift/windows-machine-config-operator/pkg/metrics"
+	"github.com/openshift/windows-machine-config-operator/pkg/secrets"
+	"github.com/openshift/windows-machine-config-operator/pkg/signer"
+)
+
+//+kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups="",resources=configmaps/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups="",resources=configmaps/finalizers,verbs=update
```
alinaryan, line 44: What are these comments for?

alinaryan, line 44: What are these comments for?

jrvaldes: FYI:  https://github.com/openshift/windows-machine-config-operator/pull/406#discussion_r617502899

README.md
sebsoto: The terminology around this is pretty confusing, but its important to differentiate "Machines" and the actual VM that gets created as a result of the Machine.

A Machine is a kubernetes object that defines the desired state of virtual machines available to the cluster. Technically a Machine isnt something that can be configured, instead you would configure the VM associated with it. 

Used in this context, instance would be a cover-all term for VMs and other potential configuration targets, possibly including a bare metal host.

test/e2e/create_test.go
sebsoto: This check is figuring out if the node is a BYOH or Machine node. The previous checks were to ensure that 1) the node is Windows, 2) the node is fully configured. The BYOH/Machine check cannot be made until the node is fully configured, as the BYOH annotation is applied at the same time as the version annotation

test/e2e/delete_test.go
sebsoto: The "windows-instances" ConfigMap is a specific configmap :smiley:

test/e2e/metrics_test.go
sebsoto: Sorry, I misunderstood your first comment. By cleanup I mean cleanup that occurs when a BYOH Node is deleted, in this case the endpoints entries for the BYOH nodes arent being removed. The actual Machineset is being deleted along with the Machine controller MachineSet within the deletion suite.

sebsoto: Changing the comment to be more clear
`cleanup -> metrics cleanup`

test/e2e/create_test.go
```diff
@@ -262,16 +349,150 @@ func (tc *testContext) waitForWindowsNodes(nodeCount int32, waitForAnnotations,
...
+		}
+	}
+	return nil, errors.Errorf("could not find CSR named %s", expectedCSRName)
+}
+
+// TODO: vSphere CSRs are not being approved due to a hostname mismatch. This is an issue only because the
+//       ConfigMap controller does not have the ability to approve CSRs yet. This needs to be removed when that
+//       functionality is added in https://issues.redhat.com/browse/WINC-581
+// approveAllCSRs approves all CSRs that are waiting for an approval
+func (tc *testContext) approveAllCSRs() error {
```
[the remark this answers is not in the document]
sebsoto: Yep, just a stopgap to get through vSphere CI

PROJECT
sebsoto: This is just a generated file change, I'm not sure why it wasn't this way in the first place.

controllers/windowsmachine_controller.go
```diff
@@ -441,18 +433,18 @@ func (r *WindowsMachineReconciler) addWorkerNode(ipAddress, instanceID, machineN
...
+	err := r.client.Get(context.TODO(), kubeTypes.NamespacedName{Name: secrets.UserDataSecret,
+		Namespace: secrets.UserDataNamespace}, userDataSecret)
 	if err != nil {
-		return errors.Errorf("could not find Windows userData secret in required namespace: %v", err)
+		return errors.Wrap(err, "could not find Windows userData secret in required namespace")
 	}
 
 	secretData := string(userDataSecret.Data["userData"][:])
-	desiredUserDataSecret, err := secrets.GenerateUserData(privateKey)
+	desiredUserDataSecret, err := secrets.GenerateUserData(r.signer.PublicKey())
```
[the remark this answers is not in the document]
sebsoto: Added this check

controllers/configmap_controller.go
alinaryan: thanks @jrvaldes

README.md
alinaryan: Thanks for the context! Would it make sense to someday change our name to Windows Instance Config Operator (WICO)? :smiley:

sebsoto: That name would make sense, but it would be logistically hard to switch at this point 😅

README.md
```diff
@@ -1,40 +1,75 @@
...
+
+A ConfigMap named `windows-instances` must be created in the WMCO namespace, describing the instances that should be
+joined to a cluster. The required information to configure an instance is:
+* An address to SSH into the instance with. This can be a DNS name or an ipv4 address.
+* An administrator user with the [private key](#create-a-private-key-secret) set as an authorized SSH key. This must
+  be done within the Windows instance by the user.
+
+Each instance described in the ConfigMap must have the Docker container runtime installed.
+
+Each entry in the data section of the configmap should be formatted with the address as the key, and a value with the
```
aravindhp, line 52: `ConfigMap`

aravindhp (approved):
Nice work, @sebsoto. A giant leap towards BYOH 🚀 

/approve

mansikulkarni96:
/lgtm

saifshaikh48:
/lgtm

Great work put into this @sebsoto.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -248,6 +252,17 @@ func (nc *nodeConfig) addPubKeyHashAnnotation() {
 	nc.node.Annotations[PubKeyHashAnnotation] = nc.publicKeyHash
 }
 
+// addAdditionalLabels merges nc.additionalAnnotations into the annotations on nc.node. If the annotation
```
mansikulkarni96, line 255: @sebsoto missed this but this should be addAdditionalAnnotations*

PR #449, opened 2021-05-11
[hack] Stop hardcoding the AMI date in MachineSet generation

aravindhp:
Instead pick the latest 1809 image.

### DIFF
--- a/hack/machineset.sh
+++ b/hack/machineset.sh
@@ -76,12 +76,11 @@ get_aws_ms() {
   local provider=$4
 
   # get the AMI id for the Windows VM
-  ami_date="2021.01.13"
-  ami_id=$(aws ec2 describe-images --filters Name=name,Values=Windows_Server-2019-English-Full-ContainersLatest-${ami_date} --region ${region} --query 'Images[*].[ImageId]' --output text)
+  ami_id=$(aws ec2 describe-images --region ${region} --filters "Name=name,Values=Windows_Server-2019*English*Full*Containers*" "Name=is-public,Values=true" --query "reverse(sort_by(Images, &CreationDate))[*].{name: Name, id: ImageId}" --output json | jq -r '.[0].id')
   if [ -z "$ami_id" ]; then
-        error-exit "unable to find AMI ID for $ami_date"
+        error-exit "unable to find AMI ID for Windows Server 2019 1809"
   fi
-  
+
   cat <<EOF
 $(get_spec $infraID $az $provider)
       providerSpec:


hack/machineset.sh
```diff
@@ -76,12 +76,11 @@ get_aws_ms() {
   local provider=$4
 
   # get the AMI id for the Windows VM
-  ami_date="2021.01.13"
-  ami_id=$(aws ec2 describe-images --filters Name=name,Values=Windows_Server-2019-English-Full-ContainersLatest-${ami_date} --region ${region} --query 'Images[*].[ImageId]' --output text)
+  ami_id=$(aws ec2 describe-images --region=us-east-1  --filters "Name=name,Values=Windows_Server-2019*English*Full*Containers*" "Name=is-public,Values=true" --query "reverse(sort_by(Images, &CreationDate))[*].{name: Name, id: ImageId}"  | jq -r '.[0].id')
```
saifshaikh48, line 79: Any reason we have hardcoded `--region=us-east-1`?

hack/machineset.sh
```diff
@@ -76,12 +76,11 @@ get_aws_ms() {
   local provider=$4
 
   # get the AMI id for the Windows VM
-  ami_date="2021.01.13"
-  ami_id=$(aws ec2 describe-images --filters Name=name,Values=Windows_Server-2019-English-Full-ContainersLatest-${ami_date} --region ${region} --query 'Images[*].[ImageId]' --output text)
+  ami_id=$(aws ec2 describe-images --region ${region} --filters "Name=name,Values=Windows_Server-2019*English*Full*Containers*" "Name=is-public,Values=true" --query "reverse(sort_by(Images, &CreationDate))[*].{name: Name, id: ImageId}" | jq -r '.[0].id')
```
jrvaldes, line 79: Since using `jq` to pipe, I feel the `--output json` should be explicit. Otherwise a parsing error will appears while using `text` as default output (Invalid numeric literal)

i.e. 
`aws ec2 describe-images --output json --filters "Name=name,Values=Windows_Server-2019*English*Full*Containers*" "Name=is-public,Values=true" --query "reverse(sort_by(Images, &CreationDate))[*].{name: Name, id: ImageId}" | jq -r '.[0].id'`



[Out of scope] In line [#70](https://github.com/openshift/windows-machine-config-operator/pull/449/files#diff-1e8df6923adbd7e6daa04e1a753e5e5d61cd894f108dc29ef0f8714bf2647681R70) there maybe be a typo: <del>get_spec</del> get_aws_ms.

aravindhp: Fixed.

PR #453, opened 2021-05-20
[test] Ensure node resource usage is retrievable

saifshaikh48:
This commit adds a test to make sure CPU and memory usage statistics are
available for Windows nodes.

### DIFF
--- a/test/e2e/metrics_test.go
+++ b/test/e2e/metrics_test.go
@@ -30,6 +30,7 @@ func testMetrics(t *testing.T) {
 	t.Run("Windows Exporter configuration validation", testWindowsExporter)
 	t.Run("Prometheus configuration validation", testPrometheus)
 	t.Run("Windows Prometheus rules validation", testWindowsPrometheusRules)
+	t.Run("Windows Node resource usage info validation", testNodeResourceUsage)
 }
 
 // testWindowsExporter deploys Linux pod and tests that it can communicate with Windows node's metrics port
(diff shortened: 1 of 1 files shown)


test/e2e/metrics_test.go
```diff
@@ -287,3 +289,37 @@ func (tc *testContext) ensureMonitoringIsEnabled() error {
...
+			node, err := testCtx.client.K8s.CoreV1().Nodes().Get(context.TODO(), winNode.Name, metav1.GetOptions{})
+			require.NoErrorf(t, err, "could not retreive node object %s", winNode.Name)
+
+			cpuAvailable, isValidQuantity := node.Status.Allocatable.Cpu().AsInt64()
+			assert.Truef(t, isValidQuantity, "expected valid numeric quantity for node %s CPU usage", winNode.Name)
+			assert.NotZerof(t, cpuAvailable, "expected non-zero value for node %s CPU usage", winNode.Name)
+
+			memoryAvailable, isValidQuantity := node.Status.Allocatable.Memory().AsInt64()
+			assert.Truef(t, isValidQuantity, "expected valid numeric quantity for node %s memory usage", winNode.Name)
+			assert.NotZerof(t, memoryAvailable, "expected non-zero value for node %s memory usage", winNode.Name)
```
saifshaikh48, line 313: Following  up to https://github.com/openshift/windows-machine-config-operator/pull/279#discussion_r636279251, I think these tests are already covered by `testWindowsPrometheusRules` which ensures the below metrics are non-zero.
 
```
node_cpu_utilisation 
node_cpu_info
node_cpu
memory_MemAvailable
memory_MemTotal
filesystem_size_bytes
node_filesystem_size_avail
node_network_receive_bytes_total
node_network_transmit_bytes_total
node_filesystem_free_bytes
```

aravindhp (approved):

/approve

test/e2e/metrics_test.go
saifshaikh48: Just to document the reasoning behind keeping these tests (as no story associated with the work): 

> The CPU/memory stats tested here are served thru api-server by kubelet (which I assume collects via system calls as it is running on the hosts, same as windows_exporter). So if those tests fail then there are likely larger issues than just metrics errors - imo it would not hurt to add in the tests as it is testing a different path to the metrics (not queried through Prometheus like with windows_exporter)

aravindhp: @saifshaikh48 for future reference, a good place for such documentation would be the commit message.

PR #463, opened 2021-06-03
WINC-600: Add system-cluster-critical priority class

saifshaikh48:
This PR adds a "system-cluster-critical" priority tag to the operator so
that WMCO pods are not preempted by user workloads. This priority still allows
operator pods to be OOMKilled as WMCO should come up fine if it gets
rescheduled on another node.

Bundle change was generated using `make bundle`.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -24,6 +24,7 @@ spec:
         control-plane: controller-manager
     spec:
       hostNetwork: true
+      priorityClassName: system-cluster-critical
       serviceAccountName: windows-machine-config-operator
       containers:
       - command:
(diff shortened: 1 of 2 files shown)


config/manager/manager.yaml
```diff
@@ -24,7 +24,7 @@ spec:
         control-plane: controller-manager
     spec:
       hostNetwork: true
-      serviceAccountName: windows-machine-config-operator
```
aravindhp, line 27: Are you certain there are no other side effects? I suggest a separate PR for this.

saifshaikh48: There should be no issue as it is a duplicate field in the yaml (`serviceAccountName` defined at line 47 already).

aravindhp: This sort of fixe should go into a separate PR and not when introducing a new change. Consider if we need to revert this PR for some reason; that revert would including removing this useful change.

aravindhp (approved):

/approve

config/manager/manager.yaml
saifshaikh48: That makes sense, thanks. New PR opened [here](https://github.com/openshift/windows-machine-config-operator/pull/464)

mansikulkarni96:
/lgtm

Issue #349, opened 2021-03-18
Documentation typo


CCChou:
Hi 

I have tried the windows node on vsphere and find out there's a typo about the example [unattend.xml](https://github.com/openshift/windows-machine-config-operator/blob/master/docs/unattend.xml).

At line 5 the xmlns:xsi and processorArchitecture there's some extra space would make the vm broken after sysprep.

mansikulkarni96:
@CCChou, thanks for the report. I will open a PR to fix the issue.

PR #466, opened 2021-06-08
WINC-610: Add Windows 2004 template to vSphere

jrvaldes:
This commit configures the Windows Server 2004 virtual machine for the vSphere e2e test.


docs/vsphere_ci/README.md
```diff
@@ -37,9 +37,8 @@ The `scripts/autounattend.xml` file must be edited to change the value of `Windo
 
 The provided [autounattend.xml](scripts/autounattend.xml)
 - Installs VMWare tools
-- Runs powershell.ps1 script which configures SSH. This is different from [powershell script](../powershell.ps1) we have 
-earlier as it copies the file from floppy disk instead of copying from C drive
```
jrvaldes, line 41: Found the `powershell.ps1`  script with duplicated content, except for the [line that copies](https://github.com/jrvaldes/windows-machine-config-operator/blob/f1158e08355613f9a75c61c11ed6208c28f98a34/docs/vsphere_ci/scripts/powershell.ps1#L21-L21) the `authorized_keys` file. 

In order to consolidate them, the novel `install-openssh.ps1` script accept the path for the `authorized_keys` file as parameter.  See usage in the [autounattend.xml files](https://github.com/jrvaldes/windows-machine-config-operator/blob/6af27634325b58c29668a1d9bf29084bd7f3e953/docs/vsphere_ci/scripts/autounattend.xml#L106-L106)

docs/unattend.xml
```diff
@@ -2,7 +2,7 @@
 <!--A sample unattend.xml which helps in setting admin password and running scripts on first boot-->
 <unattend xmlns="urn:schemas-microsoft-com:unattend">
    <settings pass="specialize">
-      <component xmlns:wcm="http://schemas.microsoft.com/WMIConfig/2002/State" xmlns:xsi="http:// www.w3.org/2001/XMLSchema-instance" name="Microsoft-Windows-International-Core" processorArchitecture="am d64" publicKeyToken="31bf3856ad364e35" language="neutral" versionScope="nonSxS">
+      <component xmlns:wcm="http://schemas.microsoft.com/WMIConfig/2002/State" xmlns:xsi="http:// www.w3.org/2001/XMLSchema-instance" name="Microsoft-Windows-International-Core" processorArchitecture="amd64" publicKeyToken="31bf3856ad364e35" language="neutral" versionScope="nonSxS">
```
mansikulkarni96, line 5: there is also an extra space at ` xmlns:xsi="http:// www.w3.org/2001/XMLSchema-instance"` can you fix that as part of this commit.

jrvaldes: Good catch, will update. Thank you.

It's also present in the [documentation](https://github.com/openshift/openshift-docs/blob/master/modules/creating-the-vsphere-windows-vm-golden-image.adoc).

cc: @EricPonvelle

docs/vsphere-prerequisites.md
```diff
@@ -1,71 +1,34 @@
...
-on every node which helps in resolving the internal API server URL. The
-external API endpoint should have been
-created as part of the
-[cluster install](https://docs.openshift.com/container-platform/4.7/installing/installing_vsphere/installing-vsphere-installer-provisioned.html).
+## vSphere environment changes
 
-An example to automate the VM golden image creation can be found
-[here](vsphere_ci/README.md)
+WMCO downloads the ignition files from the internal API server endpoint. To enable this, add a new DNS entry for
+```html
```
sebsoto, line 20: Doesnt seem like this should be tagged as html

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,172 @@
+# Creating the vSphere Windows VM Golden Image
+
+This guide describes the thought process of creating a Windows virtual machine base image (the Golden Image) in vSphere.
+
+
+## 1. Select a compatible Windows Server version
+
+Currently the Windows Machine Config Operator (WMCO) stable version [2.0.0](REF?) only supports Windows Server version 1909 (build 18363.\*) Semi-Annual Channel, that includes the patch [KB4565351](https://support.microsoft.com/en-us/help/4565351/windows-10-update-kb4565351), required to get the operating system feature that allows setting the [hybrid OVN Kubernetes networking with a custom VXLAN port](setup-hybrid-OVNKubernetes-cluster.md#vSphere)
```
sebsoto, line 8: `(REF?)` doesnt point to anything

sebsoto: I would refrain from mentioning any version, it should be assumed that this is talking about the current version

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,172 @@
...
+
+### Setup VMware Tools
+
+Install VMware Tools version 11.0.6 or greater. See the [VMware Tools documentation](https://docs.vmware.com/en/VMware-Tools/11.2.0/com.vmware.vsphere.vmwaretools.doc/GUID-D8892B15-73A5-4FCE-AB7D-56C2C90BD951.html) for more information.
+
+Enable the VMware Tools configuration file [tools.conf](https://docs.vmware.com/en/VMware-Tools/11.2.0/com.vmware.vsphere.vmwaretools.doc/GUID-EA16729B-43C9-4DF9-B780-9B358E71B4AB.html) within the Windows virtual machine.
+
+For example in Windows Server version 1909 the configuration file is located at
+
+```powershell
```
sebsoto, line 29: This isnt really powershell so I would drop the powershell tag. In addition this can be done inline with back ticks, i.e `<path>`

docs/vsphere-prerequisites.md
jrvaldes: agree

docs/vsphere-golden-image.md
jrvaldes: Agree, left that as an open forum to review. Thanks

docs/vsphere-golden-image.md
jrvaldes: Agree, the powershell tag should be out of the equation.

mansikulkarni96:
Fixes #349

docs/vsphere-prerequisites.md
```diff
@@ -17,13 +17,16 @@ The follwing section describes the setup of the internal API server endpoint wit
...
 api-int.<cluster-name>.domain.com
 ```
+
 which points to the external API server URL
-```html
+
+```
 api.<cluster-name>.domain.com
 ```
+
```
mansikulkarni96, line 29: this commit has some changes unrelated to the commit subject, can that be fixed?

docs/vsphere_ci/README.md
```diff
@@ -16,169 +21,113 @@ In case of our vSphere environment, we need to install Packer 1.6.6 on the basti
...
 
-### Sample Packer build file
+## Packer build configuration file
+
+Packer needs a build file which specifies how the virtual machine template should be built. You can find a [reference 
+build file](build.json) in the repository.
 
-Packer needs a build file which specifies the how the VM template should be built. A sample build file is shown below.
-```
-{
```
jrvaldes, line 55: Removed this block to avoid duplicated definitions, since the build.json file is included.

docs/vsphere-prerequisites.md
jrvaldes: Good point, will adjust. Thanks @mansikulkarni96

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,179 @@
...
+## 1. Select a compatible Windows Server version
+
+Currently, the Windows Machine Config Operator (WMCO) stable version only supports Windows Server version 1909 (build 18363.\*)
+Semi-Annual Channel, that includes the patch [KB4565351](https://support.microsoft.com/en-us/help/4565351/windows-10-update-kb4565351). 
+Required to get the operating system feature that allows setting the [hybrid OVN Kubernetes networking with a custom 
+VXLAN port](setup-hybrid-OVNKubernetes-cluster.md#vsphere)
+
+> Please note that the above patch is not available for Windows Server version 2019.
+
+> Coming soon: Support for Windows Server versions 2004 (build 19041.\*) and 20H2 (build 19042.\*) Semi-Annual Channels.
```
saifshaikh48, line 18: This PR adds support for 2004 right? We should remove that from `Coming soon:` and mark `Windows Server versions 2004 (build 19041.\*)` as supported by WMCO.

jrvaldes: This PR "Adds Windows 2004 template to vSphere", could that be considered as declaring support for WMCO with Windows Server version 2004 (build 19041) ?

saifshaikh48: My thought was that since this doc is vSphere specific we call out WMCO support for 2004 (and not for other cloud providers). Though you are correct, my suggestion may be getting ahead of ourselves.

docs/vsphere_ci/scripts/autounattend.xml
```diff
@@ -104,9 +104,14 @@
                     <Order>4</Order>
                     <Description>Install and configure OpenSSH Server</Description>
                     <CommandLine>cmd.exe /c powershell.exe -File a:\install-openssh.ps1 "a:\authorized_keys"</CommandLine>
-                </SynchronousCommand>
+                </SynchronousCommand>
```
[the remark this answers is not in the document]
aravindhp: This has not been addressed.

docs/vsphere_ci/scripts/install-firewall-rules.ps1
```diff
@@ -0,0 +1,16 @@
...
+#    ./install-firewall-rules.ps1 
+
+# create firewall rule to allow Container Logs on port 10250
+New-NetFirewallRule -DisplayName "ContainerLogsPort" -LocalPort 10250 -Enabled True -Direction Inbound -Protocol TCP -Action Allow -EdgeTraversalPolicy Allow
+
+# check for OpenSSH-Server-In-TCP rule, it should be created automatically during setup
+if (-not($(Get-NetFirewallRule -DisplayName "OpenSSH Server (sshd)" | Get-NetFirewallPortFilter | Where-Object {($_.Protocol -eq "TCP") -and ($_.LocalPort -eq 22)}))) {
+    # If the firewall rule does not exist, create one
+    New-NetFirewallRule -DisplayName 'OpenSSH Server (sshd)' -LocalPort 22 -Enabled True -Direction Inbound -Protocol TCP -Action Allow 
+}
```
aravindhp, line 13: Why is this being done given we were not doing this before?

docs/unattend.xml
aravindhp: Please say `Fixes #349` in the commit or PR message

docs/unattend.xml
```diff
@@ -16,19 +16,22 @@
...
+            <Enabled>true</Enabled>
+            <Username>Administrator</Username>
             <Password>
-               <Value>MyPassword</Value>
+               <!-- Administrator password -->
+               <Value>_ADMINISTRATOR_PASSWORD_</Value>
                <PlainText>true</PlainText>
             </Password>
-            <Enabled>true</Enabled>
-            <Username>Administrator</Username>
```
aravindhp, line 31: You have made a lot of these sort of movements for aesthetic purposes from what I can see. Please do such changes in a separate PR in the future.

docs/unattend.xml
```diff
@@ -16,19 +16,22 @@
          <CEIPEnabled>0</CEIPEnabled>
       </component>
       <component xmlns:wcm="http://schemas.microsoft.com/WMIConfig/2002/State" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" name="Microsoft-Windows-Shell-Setup" processorArchitecture="amd64" publicKeyToken="31bf3856ad364e35" language="neutral" versionScope="nonSxS">
+         <!-- Windows host computer name -->
          <ComputerName>windows-host</ComputerName>
-         <ProductKey>My_Product_key</ProductKey>
+         <!-- Windows product key -->
+         <ProductKey>_WINDOWS_PRODUCT_KEY_</ProductKey>
```
aravindhp: Is there a reason for this change?

docs/unattend.xml
```diff
@@ -47,7 +50,8 @@
          <TimeZone>Eastern Standard Time</TimeZone>
          <UserAccounts>
             <AdministratorPassword>
-               <Value>MyPassword</Value>
+               <!-- Administrator password -->
+               <Value>_ADMINISTRATOR_PASSWORD_</Value>
```
aravindhp, line 52: Same question here.

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,172 @@
+# Creating the vSphere Windows VM Golden Image
+
+This guide describes the thought process of creating a Windows virtual machine base image (the Golden Image) in vSphere.
+
```
aravindhp, line 4: Nit: Extra line

docs/vsphere-golden-image.md
aravindhp: Please stick to 120 columns. Applies everywhere in this doc.

aravindhp: If this was left to an open forum, the conversation should not be marked as resolved.

Not sure what `REF` implies. I say we just call out the Windows version here.

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,172 @@
...
+
+## 1. Select a compatible Windows Server version
+
+Currently the Windows Machine Config Operator (WMCO) stable version [2.0.0](REF?) only supports Windows Server version 1909 (build 18363.\*) Semi-Annual Channel, that includes the patch [KB4565351](https://support.microsoft.com/en-us/help/4565351/windows-10-update-kb4565351), required to get the operating system feature that allows setting the [hybrid OVN Kubernetes networking with a custom VXLAN port](setup-hybrid-OVNKubernetes-cluster.md#vSphere)
+
+*Please note that the above patch is not available for Windows Server version 2019.*
+
+> Comming soon: Support for Windows Server versions 2004 (build 19041.\*) and 20H2 (build 19042.\*) Semi-Annual Channels.
```
aravindhp: This is the point of the PR. Please remove and moreover no need to mention 20H2.

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,172 @@
...
+
+
+## 2. Create the virtual machine
+
+To start with, create a new virtual machine in vSphere from the selected Windows Server distribution using the ISO image.
+See [vSphere documentation](https://docs.vmware.com/en/VMware-vSphere/index.html)
```
aravindhp, line 24: Please point to the exact spot in the docs.

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,172 @@
...
+
+
+### Setup VMware Tools
+
+Install VMware Tools version 11.0.6 or greater. See the [VMware Tools documentation](https://docs.vmware.com/en/VMware-Tools/11.2.0/com.vmware.vsphere.vmwaretools.doc/GUID-D8892B15-73A5-4FCE-AB7D-56C2C90BD951.html) for more information.
+
+Enable the VMware Tools configuration file [tools.conf](https://docs.vmware.com/en/VMware-Tools/11.2.0/com.vmware.vsphere.vmwaretools.doc/GUID-EA16729B-43C9-4DF9-B780-9B358E71B4AB.html) within the Windows virtual machine.
+
+For example in Windows Server version 1909 the configuration file is located at
```
aravindhp, line 33: `2004`

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,172 @@
...
+
+```powershell
+    C:\ProgramData\VMware\VMware Tools\tools.conf
+```
+
+In case the configuration file does not exist, VMware Tools installs an example configuration file in the same 
+directory as the location above or you can download it from the [open-vm-tools](https://raw.githubusercontent.com/vmware/open-vm-tools/master/open-vm-tools/tools.conf) repository.
+
+Ensure the virtual machine has a valid IP address in vCenter.
```
aravindhp, line 43: Please call out how you do that here.

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,172 @@
...
+
+Ensure the **VMTools** Windows service is running and will start on boot.
```
aravindhp, line 45: Please add the powershell commands here and what the user needs to look for.

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,172 @@
...
+
+
```
aravindhp, line 47: Extra line

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,172 @@
...
+Enable the following entry in the VMware Tools configuration file to ensure that the cloned vNIC generated on 
+the Windows VM by the hybrid-overlay won't be ignored and the virtual machine has a valid IP Address in vCenter
+
+```bash
+    exclude-nics=
+``` 
+
+In addition, you can run the following PowerShell script that downloads the example configuration file, excludes 
+the network interface and creates the `tools.conf` file.
```
aravindhp, line 58: Not in addition but alternatively

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,172 @@
...
+
+```powershell
+    Invoke-WebRequest -O tools.conf https://raw.githubusercontent.com/vmware/open-vm-tools/master/open-vm-tools/tools.conf
+    
+    (Get-Content -Path tools.conf) -replace '#exclude-nics=','exclude-nics=' | Set-Content -Force -Path tools.conf
+    
+    mv -Force tools.conf 'C:\ProgramData\VMware\VMware Tools\tools.conf'
+```
+
+> IPv6 is not supported and has to be disabled in the golden image.
```
aravindhp, line 68: Did you do this for the 2004 template VM you created?

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,172 @@
...
+
```
aravindhp, line 69: Extra line. I see you have been doing this consistently. But to be consistent with other docs please remove.

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,172 @@
...
+
+## 3. Setup SSH
+
+Install and configure the OpenSSH Server on the Windows virtual machine with automatic startup and key based-authentication.
+See [Microsoft documentation](https://docs.microsoft.com/en-us/windows-server/administration/openssh/openssh_install_firstuse) to install using PowerShell.
+
+Ensure the SSH service is running and that it will start on boot.
```
aravindhp, line 76: Please add the commands the user can check this.

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,172 @@
...
+
+
+### Add the Administrator Public Key
```
aravindhp, line 79: It is unclear from this paragraph how this key should be installed. You need to give the manual steps and the alternative is to use the script.

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,172 @@
...
+
+The `authorized_keys` file is used to configure the OpenSSH Server in the Windows VM.
+
+The provided PowerShell script [install-openssh.ps1](vsphere_ci/scripts/install-openssh.ps1) aims to programatically 
+install and configure the OpenSSH Server with the given `authorized_keys` file.
+
+
+## 4. Setup the container runtime
+
+> Currently `docker`, it will change to `containerd`
```
aravindhp, line 94: Please remove. We can update this when we move to containerd.

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,172 @@
...
+
+Install the `docker` container runtime following the [Microsoft's documentation](https://docs.microsoft.com/en-us/virtualization/windowscontainers/quick-start/set-up-environment?tabs=Windows-Server).
+
+In addition, you can run the provided PowerShell script [install-docker.ps1](vsphere_ci/scripts/install-docker.ps1) to programatically install `docker` in the Windows VM.
```
aravindhp, line 98: Alternatively

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,172 @@
...
+
+```powershell
+    Get-ComputerInfo | Select OsHardwareAbstractionLayer 
+```
+
+Example base images:
+- Windows Server 1909 SAC: `docker pull mcr.microsoft.com/windows/servercore:1909`
+- Windows Server 2004 SAC: `docker pull mcr.microsoft.com/windows/servercore:2004`
+
+For additional Windows Server Core available images see the [Docker Hub repository](https://hub.docker.com/_/microsoft-windows-servercore?tab=description).
```
aravindhp, line 117: Please remove this as we increased the image pull time out.

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,172 @@
...
+
+
+## 5. Setup incoming connection for container logs
+
+Create a new firewall rule in the Windows VM to allow incoming connections for Container Logs, usually
```
aravindhp, line 122: `container logs`

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,172 @@
...
+```
+
+In addition, you can run the provided PowerShell script [install-firewall-rules.ps1](vsphere_ci/scripts/install-firewall-rules.ps1) to programatically create the container logs rule along with the OpenSSH incoming connections rule.
+
+
+## 6. Generalize the virtual machine installation
+
+To deploy the Windows VM as a reusable image, you have to first generalize the VM removing computer-specific information 
+such as installed drivers. The `Sysprep` with an *unattend* answer file generalize the image and make it ready for 
+deployment, maintaining all the changes needed for the WMCO installation.
```
aravindhp, line 136: Do you mean `Running sysprep with an unattend answer file generalizes the image and makes...`?

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,172 @@
...
+
+To execute the `Sysprep` command use:
+
+```cmd
+    C:\Windows\System32\Sysprep\sysprep.exe /generalize /oobe /shutdown /unattend:<path/to/unattend.xml>
+```
+
+where `<path/to/unattend.xml>` is the path to the customized answer file.
+
+> There is [a limit](https://docs.microsoft.com/en-us/windows-hardware/manufacture/desktop/sysprep--generalize--a-windows-installation#limits-on-how-many-times-you-can-run-sysprep) on how many times you can run Sysprep
```
aravindhp, line 149: Please add or `Note:` or something to that affect rather than `>` which implies it was quoted.

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,172 @@
...
+## 6. Generalize the virtual machine installation
+
+To deploy the Windows VM as a reusable image, you have to first generalize the VM removing computer-specific information 
+such as installed drivers. The `Sysprep` with an *unattend* answer file generalize the image and make it ready for 
+deployment, maintaining all the changes needed for the WMCO installation. 
+
+You can find an example [unattend.xml](unattend.xml) file in this repository. It is just an example 
+and it **should not** be used directly. You must customize the example to fit your needs.
+
+To execute the `Sysprep` command use:
```
aravindhp, line 141: Is it `Sysprep` or `sysprep`?

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,172 @@
...
+where `<path/to/unattend.xml>` is the path to the customized answer file.
+
+> There is [a limit](https://docs.microsoft.com/en-us/windows-hardware/manufacture/desktop/sysprep--generalize--a-windows-installation#limits-on-how-many-times-you-can-run-sysprep) on how many times you can run Sysprep 
+
+
+## 7. Setup the virtual machine golden image
+
+Onces the `Sysprep` command completes the Windows virtual machine will power off. 
+
+You **can not** use such Windows virtual machine anymore. It is recommended to clone the newly created virtual machine
```
aravindhp, line 156: `You must not use the created virtual machine anymore`

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,172 @@
...
+
+
+## 7. Setup the virtual machine golden image
+
+Onces the `Sysprep` command completes the Windows virtual machine will power off. 
+
+You **can not** use such Windows virtual machine anymore. It is recommended to clone the newly created virtual machine 
+before executing the [Step #6](vsphere-golden-image.md#6-generalize-the-virtual-machine-installation) and run the `Sysprep` command in the clone. 
+
+Add the cloned Windows virtual machine as a template in the [machine set](../README.md#usage).
```
aravindhp, line 159: Please link to `../README.md#configuring-windows-instances-provisioned-through-machinesets`

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,172 @@
...
+    kind: MachineSet
+    spec:
+      template:
+        spec:
+          providerSpec:
+            template: <Cloned Windows Virtual Machine>
+```
+
+If you want to update the virtual machine golden image, make the changes in the original Windows virtual machine, 
+clone it and run `Sysprep` command again.
```
aravindhp, line 172: Does it need to be cloned?

docs/vsphere-prerequisites.md
```diff
@@ -1,71 +1,34 @@
...
-The vSphere cluster must be configured with [hybrid OVN Kubernetes networking](setup-hybrid-OVNKubernetes-cluster.md)
-with a custom VXLAN port to work around the pod-to-pod connectivity between
-hosts [issue](https://docs.microsoft.com/en-us/virtualization/windowscontainers/kubernetes/common-problems#pod-to-pod-connectivity-between-hosts-is-broken-on-my-kubernetes-cluster-running-on-vsphere).
+In order to successfully use the Windows Machine Config Operator (WMCO) on a vSphere Platform, the following pre-requisites are required:
 
-In order to use the Windows MachineConfig Operator to successfully create
-a Windows node, you will first need to create a Windows "golden image"
-VM. These steps are an overview of the required configuration of the
-golden image.
+* the vSphere cluster must be configured with a hybrid OVN-Kubernetes networking using a custom VXLAN port
```
aravindhp, line 5: `The`
Please link to hybrid networking docs.

docs/vsphere-prerequisites.md
```diff
@@ -1,71 +1,34 @@
...
+* a compatible Windows virtual machine base image must be provided
```
aravindhp, line 6: `A`

docs/vsphere-prerequisites.md
```diff
@@ -1,71 +1,34 @@
...
+* an internal API server endpoint exist in the vSphere environment
```
aravindhp, line 7: It is not the internal API server endpoint but rather the DNS entry for it.

docs/vsphere-prerequisites.md
```diff
@@ -1,71 +1,34 @@
...
-  start on boot.
-* Pull all the required Windows container base images
-you'll need for your applications. The images you'll pull
-is dependent on the Windows kernel being used. Please see
-[the offical matrix](https://docs.microsoft.com/en-us/virtualization/windowscontainers/manage-containers/container-base-images)
-to determine which images you need.
-* Sysprep the VM. Use an `unattend.xml` that will maintain all the
-changes needed for the WMCO. The most important one being that the
-Administrator's home directory stays intact with the SSH public key.
+Follow the provided documentation to configure the vSphere cluster with [hybrid OVN Kubernetes networking with a custom VXLAN port](setup-hybrid-OVNKubernetes-cluster.md#vSphere) to work around the pod-to-pod connectivity between hosts [issue](https://docs.microsoft.com/en-us/virtualization/windowscontainers/kubernetes/common-problems#pod-to-pod-connectivity-between-hosts-is-broken-on-my-kubernetes-cluster-running-on-vsphere).
```
aravindhp: hybrid OVN pre-req needs to be called out only once.

docs/vsphere-prerequisites.md
```diff
@@ -1,71 +1,34 @@
...
 
-  Here is an example command to `sysprep`
-  ```
-  C:\Windows\System32\Sysprep\sysprep.exe /generalize /oobe /shutdown /unattend:<path_to_unattend.xml>`
-  ```
-  You can find an example [unattend.xml](unattend.xml) in this repo. It
-  is provided as an example and it **should not** be used directly.
+To create a Windows worker node, you will first need to setup a Windows virtual machine base image (the Golden Image), find enclosed an overview of the process required for the [configuration of a Windows VM golden image in vSphere Platform](vsphere-golden-image.md).
```
aravindhp, line 12: Lets call it golden image no need to introduce a new terminology like base image here.

docs/vsphere-prerequisites.md
```diff
@@ -1,71 +1,34 @@
...
 
-## vSphere environment changes
-WMCO downloads the ignition files from the internal API server
-endpoint. To enable this, add a new DNS entry for
-`api-int.<cluster-name>.domain.com` which points to the external API
-server URL `api.<cluster-name>.domain.com`.
+In addition, a programmatic approach for the creation of the Windows VM golden image can be found [here](vsphere_ci/README.md)
 
-This can be a CNAME or an additional A record.
+The follwing section describes the setup of the internal API server endpoint within the environment.
```
aravindhp: We are not setting up the internal API server endpoint but just the DNS entry for it.

docs/vsphere-prerequisites.md
```diff
@@ -1,71 +1,34 @@
...
+```html
+api.<cluster-name>.domain.com
+```
+where `<cluster-name>` is the name of the Openshift cluster.
+
+> The DNS entry can be a CNAME or an additional A record.
+
+The above DNS entry ensures that Windows VM can download the ignition file from the internal API server 
+and the `kubelet` on the configured VM can communicate with the internal API server. In the case of Linux nodes,
+coreDNS is running on every node which helps in resolving the internal API server URL. The external API endpoint
```
aravindhp, line 37: `CoreDNS`

docs/vsphere-golden-image.md
```diff
@@ -5,7 +5,7 @@ This guide describes the thought process of creating a Windows virtual machine b
 
 ## 1. Select a compatible Windows Server version
 
-Currently the Windows Machine Config Operator (WMCO) stable version [2.0.0](REF?) only supports Windows Server version 1909 (build 18363.\*) Semi-Annual Channel, that includes the patch [KB4565351](https://support.microsoft.com/en-us/help/4565351/windows-10-update-kb4565351), required to get the operating system feature that allows setting the [hybrid OVN Kubernetes networking with a custom VXLAN port](setup-hybrid-OVNKubernetes-cluster.md#vSphere)
+Currently the Windows Machine Config Operator (WMCO) stable version only supports Windows Server version 1909 (build 18363.\*) Semi-Annual Channel, that includes the patch [KB4565351](https://support.microsoft.com/en-us/help/4565351/windows-10-update-kb4565351), required to get the operating system feature that allows setting the [hybrid OVN Kubernetes networking with a custom VXLAN port](setup-hybrid-OVNKubernetes-cluster.md#vSphere)
```
aravindhp: This change should not be in this commit. This applies to all the other changes being made here.

docs/unattend.xml
```diff
@@ -2,7 +2,7 @@
 <!--A sample unattend.xml which helps in setting admin password and running scripts on first boot-->
 <unattend xmlns="urn:schemas-microsoft-com:unattend">
    <settings pass="specialize">
-      <component xmlns:wcm="http://schemas.microsoft.com/WMIConfig/2002/State" xmlns:xsi="http:// www.w3.org/2001/XMLSchema-instance" name="Microsoft-Windows-International-Core" processorArchitecture="amd64" publicKeyToken="31bf3856ad364e35" language="neutral" versionScope="nonSxS">
+      <component xmlns:wcm="http://schemas.microsoft.com/WMIConfig/2002/State" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" name="Microsoft-Windows-International-Core" processorArchitecture="amd64" publicKeyToken="31bf3856ad364e35" language="neutral" versionScope="nonSxS">
```
aravindhp, line 5: I suggest collapsing this commit into the other commit where you are changing this line and calling it out in the commit message.

docs/HACKING.md
```diff
@@ -54,7 +54,7 @@ hack/olm.sh cleanup
 ```
 
 ### Running e2e tests on a cluster
```
aravindhp, line 56: s/Additional considerations/Additional changes in your commit message

docs/unattend.xml
```diff
@@ -18,8 +18,6 @@
       <component xmlns:wcm="http://schemas.microsoft.com/WMIConfig/2002/State" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" name="Microsoft-Windows-Shell-Setup" processorArchitecture="amd64" publicKeyToken="31bf3856ad364e35" language="neutral" versionScope="nonSxS">
          <!-- Windows host computer name -->
          <ComputerName>windows-host</ComputerName>
-         <!-- Windows product key -->
-         <ProductKey>_WINDOWS_PRODUCT_KEY_</ProductKey>
```
aravindhp: This is confusing given you add it in one commit and then remove it here. Do you need to add it in the first place?

docs/vsphere_ci/README.md
```diff
@@ -16,169 +21,113 @@ In case of our vSphere environment, we need to install Packer 1.6.6 on the basti
...
-- `<password>` - vSphere environment password
-- `<server_name>` - vSphere environment URL
-- `<user_name>` - vSphere username to login
-- `<Windows_os_password>` - Password for the Windows template
 
-## What actually happens during build
+In order to use the provided [reference build file](build.json) as a valid configuration with Packer, you **must** 
+adjust the following variables:
+
+- `<vmtools-iso-path>` Path where VMWare Tools ISO is available in vSphere datacenter (i.e. `[] /usr/lib/vmware/isoimages/windows.iso`)
```
aravindhp, line 69: Rather than `i.e.` do you mean `example:`? Why is the `[]` needed?

docs/vsphere_ci/scripts/autounattend.xml
```diff
@@ -5,7 +5,7 @@
             <SetupUILanguage>
                 <UILanguage>en-US</UILanguage>
             </SetupUILanguage>
-            <InputLocale>es-ES</InputLocale>
+            <InputLocale>en-US</InputLocale>
```
aravindhp, line 8: Please change the autounattend.xml file in one commit. You are forcing the reviewers to have to re-review the same file multiple times.

aravindhp (changes requested):
@jrvaldes thanks for working on this. Please address me comments.

In general it is fine to use `VM` rather than expand it out to `virtual machine`.

/cc @EricPonvelle @christianh814

docs/vsphere_ci/README.md
jrvaldes: > Rather than i.e. do you mean example:? 

No, [ that is ](https://thewordcounter.com/i-e-meaning/) a valid value, not an example.

> Why is the [] needed?

In vSphere, that's the default path to the VMWare Tools ISO image for Windows. `[]` means no vSAN Datastore is specified. Otherwise, `[WorkloadDatastore] path/to/folder` would be a valid path.

docs/vsphere_ci/scripts/install-firewall-rules.ps1
jrvaldes: Glad you ask, since I have mixed feeling too. 

This rule should be created automatically by the OpenSSH server installation, it did fails sometime running local tests, so I add it explicitly.

Happy to remove it, and just rely on the OpenSSH installation process.

docs/vsphere_ci/README.md
aravindhp: So is that a constant or can it change? `i.e.` implies it is constant for all vSphere clusters and I am not so sure if that is the case.

docs/vsphere_ci/scripts/install-firewall-rules.ps1
aravindhp: I would then add to the ssh script with a comment mentioning that it sometimes does not created by the setup.

docs/vsphere_ci/scripts/install-openssh.ps1
```diff
@@ -0,0 +1,58 @@
+# Powershell script to configure OpenSSH Server in Windows Server 1809 and later.
+#
+# USAGE
+#    ./install-openssh.ps1
```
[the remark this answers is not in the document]
jrvaldes: The `param()` function [defaults to `authorized_keys`](https://github.com/openshift/windows-machine-config-operator/blob/b6d41cdb3da46a77dd78f1650a94d4fb66285f1e/docs/vsphere_ci/scripts/install-openssh.ps1#L15) as the file name value [as documented](https://github.com/openshift/windows-machine-config-operator/blob/b6d41cdb3da46a77dd78f1650a94d4fb66285f1e/docs/vsphere_ci/scripts/install-openssh.ps1#L9) in the script.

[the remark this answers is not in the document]
aravindhp: That is not clear from the documentation as that seems to describe the `keyfile` option.

docs/unattend.xml
jrvaldes: Yes, to standardize the usage of parameters within the unattend.xml file.

aravindhp: Such changes should go into a separate PR for ease of review unless it is a required change. Moreover I [recall](https://github.com/openshift/windows-machine-config-operator/pull/466#discussion_r656400174) this being deleted in a later commit.

docs/vsphere_ci/scripts/install-openssh.ps1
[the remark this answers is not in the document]
christianh814: It can be made a bit more clear. Does it create an `authorized_keys` file in the current working directory? Or in the Admin's users home directory? Or in the `__PROGRAMDATA__/ssh/administrators_authorized_keys ` file? The script should attempt to do "the best practice" when no params is set.

docs/vsphere_ci/scripts/install-openssh.ps1
```diff
@@ -0,0 +1,56 @@
...
+$administratorsRule = New-Object system.security.accesscontrol.filesystemaccessrule("Administrators","FullControl","Allow")
+$acl.SetAccessRule($administratorsRule)
+# set full control for SYSTEM
+$systemRule = New-Object system.security.accesscontrol.filesystemaccessrule("SYSTEM","FullControl","Allow")
+$acl.SetAccessRule($systemRule)
+# apply file acl
+$acl | Set-Acl
+# restart service
+Restart-Service sshd
+# success
```
christianh814, line 55: RE: https://github.com/openshift/windows-machine-config-operator/pull/466/files#r656474446

Should you put a check here to see if the FW rule was created?

christianh814 (commented):
reviewed and commented.

docs/vsphere_ci/scripts/install-openssh.ps1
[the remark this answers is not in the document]
jrvaldes: Good point, will adjust.

docs/vsphere_ci/scripts/install-openssh.ps1
```diff
@@ -0,0 +1,58 @@
...
+# validate given keyfile
+if (-not(Test-Path -Path $keyfile -PathType Leaf)) {
+    # log error and stop
+    Write-Error -ErrorAction Stop -Message "Cannot find authorized key file: $keyfile" 
+}
+
+# install OpenSSH server (See: https://docs.microsoft.com/en-us/windows-server/administration/openssh/openssh_install_firstuse)
+Add-WindowsCapability -Online -Name OpenSSH.Server~~~~0.0.1.0
+# create firewall rule for Container Logs
+New-NetFirewallRule -DisplayName "ContainerLogsPort" -LocalPort "10250" -Direction Inbound -Action Allow -Protocol TCP -EdgeTraversalPolicy Allow
```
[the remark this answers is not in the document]
jrvaldes: Fair enough. However, I feel introducing a new script should be addressed in a more "atomic" approach using [a separate commit](https://github.com/openshift/windows-machine-config-operator/pull/466/commits/d0365c904e9abd3697eb28d79668e0ada9eb82b7).

docs/vsphere_ci/scripts/install-firewall-rules.ps1
jrvaldes: makes sense, thank you

docs/vsphere_ci/scripts/install-openssh.ps1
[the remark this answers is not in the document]
aravindhp: Point I am making is adding it here in this commit, removing it in another commit is just unnecessary code churn. Moreover given the [discussion](https://github.com/openshift/windows-machine-config-operator/pull/466#discussion_r656479631) as to why it is being added, it might be preferable for it to remain here.

docs/unattend.xml
jrvaldes: Agree, will rollback

docs/vsphere-golden-image.md
jrvaldes: That's definitely a good addition. However, I feel it should go in a separated/following PR.

aravindhp: Why? You are adding this information here without specifying how you did it.

docs/vsphere-golden-image.md
jrvaldes: No, I inferred that from [BZ1947052](https://bugzilla.redhat.com/show_bug.cgi?id=1947052#c0)

docs/vsphere_ci/scripts/install-openssh.ps1
jrvaldes: @christianh814  thanks for the feedback. As per [discussion](https://github.com/openshift/windows-machine-config-operator/pull/466#discussion_r656479631) and the fact that needs more testing/debugging I'm leaning towards to add the validation as a supporting code snippet in the documentation.

docs/vsphere-golden-image.md
jrvaldes: [Ticket created](https://issues.redhat.com/browse/WINC-630), PTAL and feel free to add more context.

docs/vsphere-golden-image.md
jrvaldes: Yes, since the `sysprep` command must run in the clone.

docs/vsphere_ci/README.md
jrvaldes: That's a good question. To my understanding it is not a constant, but rather is a default in vSphere. 

The iso images directory is [a symlink to /productLocker/vmtools](https://kb.vmware.com/s/article/1037405). 

Happy to follow your lead here, just wanted to present a valid default value.

docs/unattend.xml
jrvaldes: Mission accomplished!

docs/vsphere_ci/scripts/install-openssh.ps1
aravindhp: So it is possible that in the future we will prepare an image and this rule won't be present. Not sure if we want to do that especially given the Microsoft [docs](https://docs.microsoft.com/en-us/windows-server/administration/openssh/openssh_install_firstuse#start-and-configure-openssh-server) also seem to imply that this rule creation is flaky.

docs/HACKING.md
```diff
@@ -97,7 +97,7 @@ were made to the files in config/.
 If the operator version needs to be changed, the `WMCO_VERSION` variable at the top of the repo's Makefile should
 be set to the new version.
 
-If the bundle channel needs to be changed, the `CHANNELS` and/or `DEFAULT_CHANNELS` variable within the the Makefile
+If the bundle channel needs to be changed, the `CHANNELS` and/or `DEFAULT_CHANNELS` variable within the Makefile
```
aravindhp, line 100: Both these changes have nothing to do with this PR. Please revert. This is a general rule.

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,191 @@
+# Creating the vSphere Windows VM Golden Image
+
+This guide describes the thought process of creating a Windows virtual machine base image (the Golden Image) in vSphere.
+
+## 1. Select a compatible Windows Server version
+
+Currently, the Windows Machine Config Operator (WMCO) stable version only supports Windows Server version 1909 
+(build 18363.\*) Semi-Annual Channel, that includes the patch [KB4565351](https://support.microsoft.com/en-us/help/4565351/windows-10-update-kb4565351), 
+required to get the operating system feature that allows setting the [hybrid OVN Kubernetes networking with a 
+custom VXLAN port](setup-hybrid-OVNKubernetes-cluster.md#vSphere).
```
aravindhp, line 10: The point of this PR is adding support for 2004. So:
```
Windows Machine Config Operator (WMCO) only supports Windows Server Semi-Annual Channel (SAC): Windows Server 2004, that includes...
```
Lets be consistent with the verbiage used in the OpenShift docs and in https://github.com/openshift/windows-machine-config-operator/pull/476/

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,191 @@
...
+
+*Please note that the above patch is not available for Windows Server version 2019.*
```
aravindhp, line 12: Please say `Windows Server Long-Term Servicing Channel (LTSC): Windows Server 1809` and that it cannot be used.

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,191 @@
...
+`administrators_authorized_keys` in the *C:\ProgramData\ssh\\* directory to allow administration access via SSH, as 
+describe by the [Microsoft documentation.](https://docs.microsoft.com/en-us/windows-server/administration/openssh/openssh_keymanagement#administrative-user) 
+
+Alternatively, a PowerShell script [install-openssh.ps1](vsphere_ci/scripts/install-openssh.ps1) is provided to 
+automate the process of installing and configuring the OpenSSH Server with the public key file. For example, you can 
+the following command:
+
+```
+    ./install-openssh.ps1 <path/to/public_key_file>
+```
```
aravindhp, line 124: You have already mentioned this in step 3. So you can move the command up to step 3 and remove it from here and just mention this step is not needed if the script was used in step 3.

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,191 @@
...
+where `<path/to/unattend.xml>` is the path to the customized answer file.
+
+Note: There is [a limit](https://docs.microsoft.com/en-us/windows-hardware/manufacture/desktop/sysprep--generalize--a-windows-installation#limits-on-how-many-times-you-can-run-sysprep)
+on how many times you can run the `sysprep` command.
+
+## 7. Set up the virtual machine golden image
+
+Once the `sysprep` command completes the Windows virtual machine will power off. 
+
+You **must not** use such Windows virtual machine anymore. It is recommended to clone the newly created virtual machine
```
aravindhp, line 168: s/such/this

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,191 @@
...
+Note: There is [a limit](https://docs.microsoft.com/en-us/windows-hardware/manufacture/desktop/sysprep--generalize--a-windows-installation#limits-on-how-many-times-you-can-run-sysprep)
+on how many times you can run the `sysprep` command.
+
+## 7. Set up the virtual machine golden image
+
+Once the `sysprep` command completes the Windows virtual machine will power off. 
+
+You **must not** use such Windows virtual machine anymore. It is recommended to clone the newly created virtual machine 
+before executing the [Step #6](vsphere-golden-image.md#6-generalize-the-virtual-machine-installation) 
+and run the `sysprep` command in the clone.
```
aravindhp, line 170: I suggest you remove `It is recommended...` sentence. It actually makes me think I need to clone the VM again before I use in the `MachineSet`

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,191 @@
...
+
+## 7. Set up the virtual machine golden image
+
+Once the `sysprep` command completes the Windows virtual machine will power off. 
+
+You **must not** use such Windows virtual machine anymore. It is recommended to clone the newly created virtual machine 
+before executing the [Step #6](vsphere-golden-image.md#6-generalize-the-virtual-machine-installation) 
+and run the `sysprep` command in the clone. 
+
+Add the cloned Windows virtual machine as a template in the [machine set](../README.md#configuring-windows-instances-provisioned-through-machinesets).
```
aravindhp, line 172: `MachineSet`

docs/vsphere-prerequisites.md
```diff
@@ -1,71 +1,39 @@
...
-with a custom VXLAN port to work around the pod-to-pod connectivity between
-hosts [issue](https://docs.microsoft.com/en-us/virtualization/windowscontainers/kubernetes/common-problems#pod-to-pod-connectivity-between-hosts-is-broken-on-my-kubernetes-cluster-running-on-vsphere).
+In order to successfully use the Windows Machine Config Operator (WMCO) on a vSphere Platform, 
+the following pre-requisites are required:
 
-In order to use the Windows MachineConfig Operator to successfully create
-a Windows node, you will first need to create a Windows "golden image"
-VM. These steps are an overview of the required configuration of the
-golden image.
+* The vSphere cluster must be configured with a [hybrid OVN Kubernetes networking with a custom VXLAN port](setup-hybrid-OVNKubernetes-cluster.md#vSphere)
```
aravindhp, line 6: You can remove `a` before `hybrid`

docs/vsphere-prerequisites.md
```diff
@@ -1,71 +1,39 @@
...
-you'll need for your applications. The images you'll pull
-is dependent on the Windows kernel being used. Please see
-[the offical matrix](https://docs.microsoft.com/en-us/virtualization/windowscontainers/manage-containers/container-base-images)
-to determine which images you need.
-* Sysprep the VM. Use an `unattend.xml` that will maintain all the
-changes needed for the WMCO. The most important one being that the
-Administrator's home directory stays intact with the SSH public key.
+To create a Windows worker node, you will first need to set up a Windows virtual machine image (the golden image),
+find enclosed an overview of the process required for the [configuration of a Windows VM golden image in 
+vSphere Platform](vsphere-golden-image.md).
```
aravindhp, line 14: You can remove this and add a link to `vsphere-golden-image.md` in `set up a Windows virtual machine image (the golden image)`

docs/vsphere_ci/README.md
aravindhp: I would just say `... in the vSphere datacenter (default: '[] /usr/lib/vmware/isoimages/windows.iso\')`

aravindhp (changes requested):
@jrvaldes thanks for re-working the commits. It was much easier to review this time around.

docs/vsphere_ci/scripts/install-openssh.ps1
jrvaldes: Agree. 

Added both [a link to Microsoft docs](https://github.com/openshift/windows-machine-config-operator/blame/8f34eca7268bcba6241f52efbee08aae06ef5272/docs/vsphere-golden-image.md#L69) and the validation/creation [code snippets](https://github.com/openshift/windows-machine-config-operator/blame/8f34eca7268bcba6241f52efbee08aae06ef5272/docs/vsphere-golden-image.md#L94) to make it explicit in the docs.

docs/vsphere-golden-image.md
jrvaldes: Good catch, thank you!

docs/HACKING.md
jrvaldes: [PR opened](https://github.com/openshift/windows-machine-config-operator/pull/483)

docs/vsphere_ci/scripts/install-openssh.ps1
aravindhp: My point was we might want to add the check here given the docs is asking the user to check for the rule. But I will let this go. We just need to be aware that we might be creating flaky images in the future.

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,186 @@
+# Creating the vSphere Windows VM Golden Image
+
+This guide describes the thought process of creating a Windows virtual machine base image (the Golden Image) in vSphere.
+
+## 1. Select a compatible Windows Server version
+
+Currently, the Windows Machine Config Operator (WMCO) stable version only supports Windows Server Semi-Annual
+Channel (SAC): Windows Server 2004, that includes the patch [KB4565351](https://support.microsoft.com/en-us/help/4565351/windows-10-update-kb4565351), 
+required to get the operating system feature that allows setting the [hybrid OVN Kubernetes networking with a
```
aravindhp, line 9: `...allows using the [hybrid OVN Kubernetes networking with a custom VXLAN port](setup-hybrid-OVNKubernetes-cluster.md#vSphere) feature`

docs/vsphere-golden-image.md
```diff
@@ -0,0 +1,186 @@
...
+    mv -Force tools.conf 'C:\ProgramData\VMware\VMware Tools\tools.conf'
+```
+
+## 3. Set up SSH
+
+Install and configure the OpenSSH Server on the Windows virtual machine with automatic startup and key based-authentication.
+See [Microsoft documentation](https://docs.microsoft.com/en-us/windows-server/administration/openssh/openssh_install_firstuse)
+to install using PowerShell.
+
+The provided PowerShell script [install-openssh.ps1](vsphere_ci/scripts/install-openssh.ps1) aims to programmatically
```
aravindhp, line 73: `Alternatively, the provided...`

docs/vsphere-prerequisites.md
```diff
@@ -1,71 +1,34 @@
...
+In order to successfully use the Windows Machine Config Operator (WMCO) on a vSphere Platform, 
+the following pre-requisites are required:
 
-In order to use the Windows MachineConfig Operator to successfully create
-a Windows node, you will first need to create a Windows "golden image"
-VM. These steps are an overview of the required configuration of the
-golden image.
+* The vSphere cluster must be configured with [hybrid OVN Kubernetes networking with a custom VXLAN port](setup-hybrid-OVNKubernetes-cluster.md#vSphere)
+  to work around the pod-to-pod connectivity between hosts [issue](https://docs.microsoft.com/en-us/virtualization/windowscontainers/kubernetes/common-problems#pod-to-pod-connectivity-between-hosts-is-broken-on-my-kubernetes-cluster-running-on-vsphere)
+* Set up a compatible [Windows virtual machine golden image](vsphere-golden-image.md)
```
aravindhp, line 8: I would again say `Windows Server Semi-Annual Channel (SAC): Windows Server 2004 VM golden image` just to reinforce it at the higher level.

aravindhp (changes requested):

aravindhp:
@EricPonvelle please review the docs part of this PR. This too would need to be reflected in 4.7 and 4.8.

aravindhp (approved):
Nice work @jrvaldes 🎉 

/approve

EricPonvelle (approved):
/lgtm

PR #468, opened 2021-06-09
WINC-614: Remove test overlap in upgrade suite

saifshaikh48:
This PR aims to reduce the amount of time our end-to-end tests take to complete by dropping a redundant case from the upgrade tests and spreading platform-agnostic tests across cloud providers in CI.

### DIFF
--- a/test/e2e/upgrade_test.go
+++ b/test/e2e/upgrade_test.go
@@ -46,6 +46,8 @@ func upgradeTestSuite(t *testing.T) {
 	require.NoError(t, err, "error configuring upgrade")
 
 	// get current Windows node state
+	// TODO: waitForWindowsNodes currently loads nodes into global context, so we need this (even though BYOH
+	// 		 nodes are not being upgraded/tested here). Remove as part of https://issues.redhat.com/browse/WINC-620
 	err = testCtx.waitForWindowsNodes(gc.numberOfMachineNodes, false, true, false)
 	require.NoError(t, err, "wrong number of Machine controller nodes found")
 	err = testCtx.waitForWindowsNodes(gc.numberOfBYOHNodes, false, true, true)
@@ -67,11 +69,12 @@ func testUpgradeVersion(t *testing.T) {
 	testPrometheus(t)
 
 	// Test if there was any downtime for Windows workloads by checking the failure on the Job pods.
-	pods, err := testCtx.client.K8s.CoreV1().Pods(testCtx.workloadNamespace).List(context.TODO(), metav1.ListOptions{FieldSelector: "status.phase=Failed",
-		LabelSelector: "job-name=" + windowsWorkloadTesterJob + "-job"})
+	pods, err := testCtx.client.K8s.CoreV1().Pods(testCtx.workloadNamespace).List(context.TODO(),
+		metav1.ListOptions{FieldSelector: "status.phase=Failed",
+			LabelSelector: "job-name=" + windowsWorkloadTesterJob + "-job"})
 
 	require.NoError(t, err)
-	require.Equal(t, 0, len(pods.Items), "unable to access Windows workloads for significant amount of time during upgrade")
+	require.Equal(t, 0, len(pods.Items), "Windows workloads inaccessible for significant amount of time during upgrade")
 
 }
 
@@ -86,15 +89,16 @@ func (tc *testContext) configureUpgradeTest() error {
 		return err
 	}
 
-	// Override the Windows Node Version Annotation
-	nodes, err := tc.listFullyConfiguredWindowsNodes(false)
+	// tamper version annotation on nodes configured by the Machine controller
+	machineNodes, err := tc.listFullyConfiguredWindowsNodes(false)
 	if err != nil {
 		return errors.Wrap(err, "error getting list of fully configured nodes")
 	}
 
-	for _, node := range nodes {
+	for _, node := range machineNodes {
 		patchData := fmt.Sprintf(`{"metadata":{"annotations":{"%s":"%s"}}}`, nc.VersionAnnotation, "badVersion")
-		_, err := tc.client.K8s.CoreV1().Nodes().Patch(context.TODO(), node.Name, types.MergePatchType, []byte(patchData), metav1.PatchOptions{})
+		_, err := tc.client.K8s.CoreV1().Nodes().Patch(context.TODO(), node.Name, types.MergePatchType,
+			[]byte(patchData), metav1.PatchOptions{})
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


test/e2e/upgrade_test.go
```diff
@@ -100,35 +70,19 @@ func testTamperAnnotation(t *testing.T) {
...
 
-// configureUpgradeTest carries out steps required before running tests for upgrade scenario.
-// The steps include -
-// 1. Scale down the operator to 0.
-// 2. Change Windows node version annotation to an invalid value
-// 3. Scale up the operator to 1
+// configureUpgradeTest carries out steps required before running tests for upgrade scenario. The steps include -
+// 1. Ensuring the correct amount of nodes have a proper annotation to start
+// 2. Ensuring the operator is running with 1 replica
 func (tc *testContext) configureUpgradeTest() error {
```
saifshaikh48, line 76: Reviewers, as part of the individual test suites (network, upgrade, metrics, etc) within the e2e tests, which route should we take?
1. just assume that the cluster is in testable state (e.g. operator is up and, for this suite, correct # nodes initially have valid version annotation) -- this approach calls for the removal of this setup method
2. have guards to check, else fail fast/don't even run test case
3. try to coerce cluster to reach proper state

test/e2e/upgrade_test.go
```diff
@@ -100,35 +70,19 @@ func testTamperAnnotation(t *testing.T) {
...
-	// Scale down the WMCO deployment to 0
-	if err := tc.scaleWMCODeployment(0); err != nil {
-		return err
-	}
-
-	// Override the Windows Node Version Annotation
-	nodes, err := tc.client.K8s.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{LabelSelector: nc.WindowsOSLabel})
+	nodes, err := tc.client.K8s.CoreV1().Nodes().List(context.TODO(),
```
saifshaikh48, line 77: Example of option 2, guard + fail fast

test/e2e/upgrade_test.go
```diff
@@ -100,35 +70,19 @@ func testTamperAnnotation(t *testing.T) {
...
-	for _, node := range nodes.Items {
-		patchData := fmt.Sprintf(`{"metadata":{"annotations":{"%s":"%s"}}}`, nc.VersionAnnotation, "badVersion")
-		_, err := tc.client.K8s.CoreV1().Nodes().Patch(context.TODO(), node.Name, types.MergePatchType, []byte(patchData), metav1.PatchOptions{})
-		if err != nil {
-			return err
-		}
-		log.Printf("Node Annotation changed to %v", node.Annotations[nc.VersionAnnotation])
-	}
-
 	// Scale up the WMCO deployment to 1
```
saifshaikh48, line 86: Example of option 3, manual coercion

test/e2e/upgrade_test.go
saifshaikh48, line 76: Reviewers, as part of the individual test suites (network, upgrade, metrics, etc) within the e2e tests, which route should we take?
1. just assume that the cluster is in testable state (e.g. operator is up and, for this suite, correct # nodes initially have valid version annotation) -- this approach calls for the removal of this setup method
2. have guards to check, else fail fast/don't even run test case
3. try to coerce cluster to reach proper state

test/e2e/upgrade_test.go
```diff
@@ -100,35 +70,19 @@ func testTamperAnnotation(t *testing.T) {
...
-	// Override the Windows Node Version Annotation
-	nodes, err := tc.client.K8s.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{LabelSelector: nc.WindowsOSLabel})
+	nodes, err := tc.client.K8s.CoreV1().Nodes().List(context.TODO(),
+		metav1.ListOptions{LabelSelector: nc.WindowsOSLabel})
 	if err != nil {
 		return err
 	}
 	if len(nodes.Items) != int(gc.numberOfNodes) {
 		return errors.Wrapf(nil, "unexpected number of nodes %v", gc.numberOfNodes)
 	}
```
saifshaikh48, line 84: Example of option 2, guard + fail fast

test/e2e/upgrade_test.go
```diff
@@ -100,35 +70,19 @@ func testTamperAnnotation(t *testing.T) {
...
-		_, err := tc.client.K8s.CoreV1().Nodes().Patch(context.TODO(), node.Name, types.MergePatchType, []byte(patchData), metav1.PatchOptions{})
-		if err != nil {
-			return err
-		}
-		log.Printf("Node Annotation changed to %v", node.Annotations[nc.VersionAnnotation])
-	}
-
 	// Scale up the WMCO deployment to 1
 	if err := tc.scaleWMCODeployment(1); err != nil {
 		return err
```
saifshaikh48, line 88: Example of option 3, manual coercion

mansikulkarni96:
changes LGTM will /lgtm once approved.

test/e2e/upgrade_test.go
```diff
@@ -46,13 +46,14 @@ func upgradeTestSuite(t *testing.T) {
...
 	// get current Windows node state
+	// TODO: waitForWindowsNodes currently loads nodes into global context, so we need this (even though BYOH
+	// 		 nodes are not being upgraded/tested here). Remove as part of https://issues.redhat.com/browse/WINC-620
 	err = testCtx.waitForWindowsNodes(gc.numberOfMachineNodes, false, true, false)
 	require.NoError(t, err, "wrong number of Machine controller nodes found")
 	err = testCtx.waitForWindowsNodes(gc.numberOfBYOHNodes, false, true, true)
 	require.NoError(t, err, "wrong number of ConfigMap controller nodes found")
 
 	t.Run("Operator version upgrade", testUpgradeVersion)
-	t.Run("Version annotation tampering", testTamperAnnotation)
```
sebsoto, line 55: In this commit, this line + the deletion of the function are really the only changes that are accomplishing the goals you outlined in the commit message.
Everything you've done is beneficial for the codebase as a whole, but it would be better to separate these related small quality of life refactors into a separate commit, for the sake of reviewers. It makes it easier for us to see what should have a functional impact, and what shouldn't :)

sebsoto (changes requested):
The work done here LGTM but I'd prefer to see the commit separation done, not a blocker from me though.

test/e2e/upgrade_test.go
saifshaikh48: ACK, broke it out into separate commits.

saifshaikh48:
> changes LGTM will /lgtm once approved.

@mansikulkarni96 this has been updated & approved, PTAL

PR #472, opened 2021-06-11
[submodule][kubelet] Hack for projected volumes

aravindhp:
https://github.com/kubernetes/kubernetes/commit/3db7275b549559696c42c0b5f51c9a2397e9571d introduced os.Chown() which is not supported in Windows. Workaround that by only calling it for non-Windows OSes.

Co-authored-by: Ryan Phillips <rphillips@redhat.com>


sebsoto:
/retest
registry issue should be resolved

PR #474, opened 2021-06-17
WINC-582: [test] Ensure all Win services are running

sebsoto:
Adds test code to inspect the state of each Windows node's services, and test that they are in the expected state.


test/e2e/validation_test.go
```diff
@@ -143,3 +150,261 @@ func testNodeTaint(t *testing.T) {
...
+		return "", errors.Wrap(err, "error getting debug image")
+	}
+
+	// username will be Administrator on all cloud providers, except Azure where it is "capi"
+	var username string
+	if tc.CloudProvider.GetType() == config.AzurePlatformType {
+		username = "capi"
+	} else {
+		username = "Administrator"
+	}
```
saifshaikh48, line 301: This is duplicated elsewhere in the e2e tests. Do you think it is worth pulling this out into a helper method `GetUsername`?

test/e2e/validation_test.go
```diff
@@ -143,3 +150,261 @@ func testNodeTaint(t *testing.T) {
...
+
+	// Copy the current WMCO private key to the test namespace, so that the SSH pod which will run in the test
+	// namespace will have access to it
+	if err := tc.copyKeySecret(); err != nil {
+		return "", errors.Wrap(err, "error copying private-key secret to test namespace")
+	}
+
+	// The SSH pod will use the openshift/tools image, the same image used by `oc debug`. This image is available on
+	// all OpenshiftClusters, and has SSH pre-installed.
```
saifshaikh48, line 289: nit: `OpenShift clusters`

test/e2e/validation_test.go
```diff
@@ -143,3 +150,261 @@ func testNodeTaint(t *testing.T) {
...
+			return nil, errors.New("expected comma separated values, found: " + line)
+		}
+		services[strings.Trim(fields[0], "\"")] = strings.Trim(fields[1], "\"")
+	}
+	return services, nil
+}
+
+// testExpectedServicesRunning tests that for each node all the expected services are running
+func testExpectedServicesRunning(t *testing.T) {
+	expectedServices := []string{"kubelet", "kube-proxy", "hybrid-overlay-node"}
```
saifshaikh48, line 393: `windows-exporter` also runs as a Windows service.

test/e2e/validation_test.go
```diff
@@ -143,3 +150,261 @@ func testNodeTaint(t *testing.T) {
...
+	}
+	sa := core.ServiceAccount{ObjectMeta: meta.ObjectMeta{Name: tc.workloadNamespace}}
+	_, err = saClient.Create(context.TODO(), &sa, meta.CreateOptions{})
+	if err != nil {
+		return errors.Wrap(err, "unable to create SA")
+	}
+	return nil
+}
+
+// ensureTestRunnerRoleBinding ensures the proper Role exists, a requirement for SSHing into a Windows node
```
aravindhp, line 169: Incorrect doc comment

test/e2e/validation_test.go
```diff
@@ -143,3 +150,261 @@ func testNodeTaint(t *testing.T) {
...
+// copyKeySecret copies the current private-key secret from the WMCO namespace into the testContext's test namespace
+func (tc *testContext) copyKeySecret() error {
+	currentSecret, err := tc.client.K8s.CoreV1().Secrets(tc.namespace).Get(context.TODO(), secrets.PrivateKeySecret, meta.GetOptions{})
+	if err != nil {
+		return errors.Wrap(err, "unable to get current private key secret from WMCO namespace")
+	}
+
+	// Remove secret from test namespace if it already exists, so the new secret can be created
+	secretClient := tc.client.K8s.CoreV1().Secrets(tc.workloadNamespace)
+	err = secretClient.Delete(context.TODO(), secrets.PrivateKeySecret, meta.DeleteOptions{})
```
aravindhp, line 247: I again don't follow why you have to delete it?

test/e2e/validation_test.go
```diff
@@ -143,3 +150,261 @@ func testNodeTaint(t *testing.T) {
...
+	if err := tc.copyKeySecret(); err != nil {
+		return "", errors.Wrap(err, "error copying private-key secret to test namespace")
+	}
+
+	// The SSH pod will use the openshift/tools image, the same image used by `oc debug`. This image is available on
+	// all OpenshiftClusters, and has SSH pre-installed.
+	debugImage, err := tc.getOpenShiftToolsImage()
+	if err != nil {
+		return "", errors.Wrap(err, "error getting debug image")
+	}
```
aravindhp, line 293: These should again be one time calls.

test/e2e/validation_test.go
aravindhp, line 301: You should do this once and pass in the user into this function.

test/e2e/validation_test.go
aravindhp, line 393: `windows-exporter`?

aravindhp (changes requested):
Thanks for working on this @sebsoto. Please address my comments.

saifshaikh48 (changes requested):
Nice work @sebsoto, almost LGTM.

Question not directly related to review: when is it necessary or recommended to run `go mod tidy`? Any time new dependencies are introduced?

test/e2e/validation_test.go
```diff
@@ -143,3 +150,261 @@ func testNodeTaint(t *testing.T) {
 		assert.Equal(t, hasTaint, true, "expected Windows Taint to be present on the Node: %s", node.GetName())
 	}
 }
+
+// ensureTestRunnerSA ensures the proper ServiceAccount exists, a requirement for SSHing into a Windows node
```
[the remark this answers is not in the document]
sebsoto: My thought process behind doing this was that it would make things easier when running the test locally, especially in situations where the resource specifications are being iterated on. Thinking about it more thats not the way that this should be done as it just complicates things, when like you implied, running in CI should be the main focus. Instead I have changed it so that if the resource already exists, these functions return nil. That maintains the ability to re-run these tests locally without too much headache, while removing the delete calls.

test/e2e/validation_test.go
```diff
@@ -143,3 +150,261 @@ func testNodeTaint(t *testing.T) {
...
+	}
+	return image, nil
+}
+
+// runSSHJob creates and waits for a Kubernetes job to run. The command provided will be executed on the host specified
+// by the provided IP.
+func (tc *testContext) runSSHJob(name, command, ip string) (string, error) {
+	if err := tc.sshSetup(); err != nil {
+		return "", errors.Wrap(err, "unable to fufill SSH requirements")
+	}
```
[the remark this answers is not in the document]
sebsoto: These resources are all scoped to the wmco-test namespace so they will be cleaned up when that namespace is cleaned up.

test/e2e/validation_test.go
sebsoto: Removing the duplicate code doesn't seem possible without making the username result part of the testContext struct. I've made that change for now, but I'm not sure if thats what you wanted. I can also make this a helper function like @saifshaikh48 suggested.

The username needs to be known in the creation suite when creating the windows-instance configmap and when SSHing in to validate creation, as well as in the deletion suite in the future when SSHing in to validate proper deconfiguration.

test/e2e/validation_test.go
saifshaikh48: +1 
Should this noop if the key already exists in the test namespace, similar to `sshSetup`?

aravindhp:
/retitle WINC-582: [test] Ensure all Win services are running

test/e2e/validation_test.go
aravindhp: This has not been addressed.

test/e2e/validation_test.go
```diff
@@ -143,3 +148,218 @@ func testNodeTaint(t *testing.T) {
...
+}
+
+// runSSHJob creates and waits for a Kubernetes job to run. The command provided will be executed on the host specified
+// by the provided IP.
+func (tc *testContext) runSSHJob(name, command, ip string) (string, error) {
+	// Copy the current WMCO private key to the test namespace, so that the SSH pod which will run in the test
+	// namespace will have access to it
+	if err := tc.copyKeySecret(); err != nil {
+		return "", errors.Wrap(err, "error copying private-key secret to test namespace")
+	}
```
aravindhp, line 256: Shouldn't this also be a one time operation similar to `sshSetup()`?

sebsoto: The private key secret changes multiple times throughout the creation suite so when SSHing, the current key has to be copied over and used. It should be possible to pull this out of this function, but for simplicity sake I thought it made sense to keep it in here, since its so subject to change.

Thats also the reasoning behind deleting the existing key and copying over the current one https://github.com/openshift/windows-machine-config-operator/pull/474#discussion_r657406633

test/e2e/validation_test.go
sebsoto: See https://github.com/openshift/windows-machine-config-operator/pull/474#discussion_r657425994

test/e2e/validation_test.go
aravindhp: Then this should be done where the key is being changed and not all the time. Is that really hard to do? We are going to be making a lot of SSH calls and this sort of churn is not required when it can be easily avoided.

saifshaikh48 (approved):
/lgtm

test/e2e/validation_test.go
sebsoto: Moved into the secret creation/deletion function and updated the commit message

aravindhp (approved):
Nice work @sebsoto 
/approve

PR #485, opened 2021-06-24
[test] Fix flakey configmap unit test

sebsoto:
Random ordering of elements caused this unit test to fail occasionally.
ElementsMatch should be used instead so that the order does not matter.

Co-authored-by: Aravindh Puthiyaparambil <aravindh@redhat.com>


aravindhp (approved):
Thanks for fixing this @sebsoto 

/approve

PR #498, opened 2021-06-28
WINC-582: [wmco] BYOH node removal

sebsoto:
This commit enables the removal of BYOH nodes from a cluster. A new
username annotation has been added to BYOH nodes in order to help
accomplish this.

Instances that are removed from the windows-instances ConfigMap will be
cordoned, drained, and deconfigured, returning them to the state they
were in before WMCO configured them.

### DIFF
--- a/controllers/configmap_controller.go
+++ b/controllers/configmap_controller.go
@@ -46,6 +46,8 @@ import (
 const (
 	// BYOHAnnotation is an an anotation that should be applied to all Windows nodes not associated with a Machine.
 	BYOHAnnotation = "windowsmachineconfig.openshift.io/byoh"
+	// UsernameAnnotation is a node annotation that contains the username used to log into the Windows instance
+	UsernameAnnotation = "windowsmachineconfig.openshift.io/username"
 	// InstanceConfigMap is the name of the ConfigMap where VMs to be configured should be described.
 	// TODO: Possibly make this a singleton that WMCO creates https://issues.redhat.com/browse/WINC-612
 	InstanceConfigMap = "windows-instances"
@@ -181,8 +183,11 @@ func (r *ConfigMapReconciler) reconcileNodes(ctx context.Context, instances *cor
 			return errors.Wrapf(err, "error configuring host with address %s", host.Address)
 		}
 	}
-	// TODO: Remove BYOH nodes that are not associated with the parsed host list
-	//       https://issues.redhat.com/browse/WINC-582
+
+	// Ensure that only instances currently specified by the ConfigMap are joined to the cluster as nodes
+	if err = r.deconfigureInstances(hosts, nodes); err != nil {
+		return errors.Wrap(err, "error removing undesired nodes from cluster")
+	}
 
 	// Once all the proper Nodes are in the cluster, configure the prometheus endpoints.
 	if err := r.prometheusNodeConfig.Configure(); err != nil {
@@ -200,10 +205,31 @@ func (r *ConfigMapReconciler) ensureNodeIsConfigured(instance *instances.Instanc
 		return nil
 	}
 
-	if err := r.configureInstance(instance, map[string]string{BYOHAnnotation: "true"}); err != nil {
+	if err := r.configureInstance(instance, map[string]string{BYOHAnnotation: "true",
+		UsernameAnnotation: instance.Username}); err != nil {
 		return errors.Wrap(err, "error configuring node")
 	}
 
+	return nil
+}
+
+// deconfigureInstances removes all BYOH nodes that are not specified in the given instances slice, and
+// deconfigures the instances associated with them.
+func (r *ConfigMapReconciler) deconfigureInstances(instances []*instances.InstanceInfo, nodes *core.NodeList) error {
+	for _, node := range nodes.Items {
+		// Only looking at BYOH nodes
+		if _, present := node.Annotations[BYOHAnnotation]; !present {
+			continue
+		}
+		// Check for instances associated with this node
+		if hasEntry := hasAssociatedInstance(&node, instances); hasEntry {
+			continue
+		}
+		// no instance found in the provided list, remove the node from the cluster
+		if err := r.deconfigureInstance(&node); err != nil {
+			return errors.Wrapf(err, "unable to deconfigure instance with node %s", node.GetName())
+		}
+	}
 	return nil
 }
 
@@ -218,6 +244,18 @@ func findNode(address string, nodes *core.NodeList) (*core.Node, bool) {
 		}
 	}
 	return nil, false
+}
+
+// hasAssociatedInstance returns true if the given node is associated with an instance in the given slice
+func hasAssociatedInstance(node *core.Node, instances []*instances.InstanceInfo) bool {
+	for _, instance := range instances {
+		for _, nodeAddress := range node.Status.Addresses {
+			if instance.Address == nodeAddress.Address {
+				return true
+			}
+		}
+	}
+	return false
 }
 
 // SetupWithManager sets up the controller with the Manager.
--- a/controllers/controllers.go
+++ b/controllers/controllers.go
@@ -1,9 +1,12 @@
 package controllers
 
 import (
+	"net"
+
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 	"golang.org/x/crypto/ssh"
+	core "k8s.io/api/core/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/record"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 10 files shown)


pkg/windows/windows.go
```diff
@@ -580,6 +621,34 @@ func (vm *windows) stopService(svc *service) error {
 	return nil
 }
 
+// deleteService deletes the specified Windows service
+func (vm *windows) deleteService(svc *service) error {
+	if svc == nil {
+		return errors.New("service object should not be nil")
```
jrvaldes, line 627: s/service object should not be nil/service object cannot be nil

pkg/windows/windows.go
```diff
@@ -75,8 +75,15 @@ const (
...
-// filesToTransfer is a map of what files should be copied to the Windows VM and where they should be copied to
-var filesToTransfer map[*payload.FileInfo]string
+var (
+	// filesToTransfer is a map of what files should be copied to the Windows VM and where they should be copied to
+	filesToTransfer map[*payload.FileInfo]string
+	// InstalledSVCs is a list of Windows services installed by WMCO
+	InstalledSVCs = []string{windowsExporterServiceName, kubeProxyServiceName, hybridOverlayServiceName,
+		kubeletServiceName}
+	// RequiredDirectories is a list of directories to be created by WMCO
+	RequiredDirectories = []string{k8sDir, remoteDir, cniDir, cniConfDir, logDir, kubeProxyLogDir, hybridOverlayLogDir}
```
jrvaldes, line 85: Consider breaking them in multiple lines [as before](https://github.com/openshift/windows-machine-config-operator/pull/498/files#diff-4d8fe3bb66a872f0e4a277f820d60d2efa722f09aa89516be74b98c06b57f3bdL434), for better visibility. 

Same with [InstalledSVCs](https://github.com/openshift/windows-machine-config-operator/pull/498/files#diff-4d8fe3bb66a872f0e4a277f820d60d2efa722f09aa89516be74b98c06b57f3bdR81) .

test/e2e/delete_test.go
```diff
@@ -2,21 +2,90 @@ package e2e
...
+	}
+	command += "exit 0\""
+	out, err := tc.runSSHJob("check-win-dirs", command, address)
+	if err != nil {
+		return false, errors.Wrapf(err, "error confirming directories do not exist %s", out)
+	}
+	if strings.Contains(out, "exists") {
+		return false, nil
+	}
+	return true, nil
```
saifshaikh48, line 86: Can be simplified to `return !strings.Contains(out, "exists"), nil`

Overall though, the function is a bit strange to me as the positive case reads as a double negative: `false, the required dirs do not exist`
Maybe rename it to `requiredDirsExist`, and invert the return values and usage?

controllers/configmap_controller.go
```diff
@@ -181,8 +183,11 @@ func (r *ConfigMapReconciler) reconcileNodes(ctx context.Context, instances *cor
 			return errors.Wrapf(err, "error configuring host with address %s", host.Address)
 		}
 	}
-	// TODO: Remove BYOH nodes that are not associated with the parsed host list
-	//       https://issues.redhat.com/browse/WINC-582
+
+	// Ensure that only instances specified by the ConfigMap are joined to the cluster as nodes
```
mansikulkarni96, line 187: removeUndesiredNodes seems confusing to me, having a specific comment like instances *currrently specified by ConfigMap might  be more clear.
Would it make sense to name this function to something like syncNodesToConfigMap? 
On a similar note, might it be worth considering a flow where we sync the configmap instances with current list of nodes -> decide to remove or add nodes.

controllers/controllers.go
```diff
@@ -48,3 +51,49 @@ func (r *instanceReconciler) configureInstance(instance *instances.InstanceInfo,
...
+				continue
+			}
+			return addr.Address, nil
+		}
+	}
+	return "", errors.New("no usable address")
+}
+
+// removeNode removes the specified node from the cluster
+func (r *instanceReconciler) removeNode(node *core.Node) error {
```
mansikulkarni96, line 84: should this be deconfigureInstance() like we have configureInstance(), we would have to find the instance info from the node previous to this step and pass it in here in that case.

mansikulkarni96 (commented):
Thanks for working on this @sebsoto.
PTAL at my comments.

test/e2e/delete_test.go
```diff
@@ -2,21 +2,90 @@ package e2e
...
+			dirsCleanedUp, err := tc.requiredDirsDoNotExist(addr)
+			require.NoError(t, err, "error determining if created directories exist")
+			assert.True(t, dirsCleanedUp, "directories not removed")
+		})
+	}
+}
+
+// requiredDirsDoNotExist checks if the required directories created by WMCO exist on the instance with the provided address
+func (tc *testContext) requiredDirsDoNotExist(address string) (bool, error) {
+	command := "powershell.exe -NonInteractive -ExecutionPolicy Bypass -Command \""
```
aravindhp, line 74: Is there a reason to use PowerShell here when we have not been using it in other places?

test/e2e/delete_test.go
aravindhp: I suggest calling it `checkDirsDoNotExist()`

pkg/windows/windows.go
```diff
@@ -75,8 +75,15 @@ const (
 	serviceNotFound = "status 1060"
 )
 
-// filesToTransfer is a map of what files should be copied to the Windows VM and where they should be copied to
-var filesToTransfer map[*payload.FileInfo]string
+var (
+	// filesToTransfer is a map of what files should be copied to the Windows VM and where they should be copied to
+	filesToTransfer map[*payload.FileInfo]string
+	// InstalledSVCs is a list of Windows services installed by WMCO
```
aravindhp, line 81: I suggest calling them `requiredServices`

pkg/windows/windows.go
```diff
@@ -258,6 +265,39 @@ func (vm *windows) EnsureRequiredServicesStopped() error {
 	return nil
 }
 
+// removeInstalledServices removes all services installed by WMCO
```
aravindhp, line 268: I suggest calling this `removeServices`

test/e2e/delete_test.go
```diff
@@ -63,12 +125,20 @@ func testWindowsNodeDeletion(t *testing.T) {
 	// Phase is ignored during deletion, in this case we are just waiting for Machines to be deleted.
 	_, err = testCtx.waitForWindowsMachines(int(expectedNodeCount), "", true)
 	require.NoError(t, err, "Machine controller Windows machine deletion failed")
-	_, err = testCtx.waitForWindowsMachines(int(expectedNodeCount), "", false)
-	require.NoError(t, err, "ConfigMap controller Windows machine deletion failed")
+
+	// TODO: Currently on vSphere it is impossible to delete a Machine after its node has been deleted.
```
aravindhp, line 129: I am obviously missing something. Where is the call that deletes the BYOH Machine?

test/e2e/delete_test.go
```diff
@@ -2,21 +2,90 @@ package e2e
...
+	config "github.com/openshift/api/config/v1"
 	mapi "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
+	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	"github.com/openshift/windows-machine-config-operator/controllers"
 	"github.com/openshift/windows-machine-config-operator/pkg/secrets"
+	win "github.com/openshift/windows-machine-config-operator/pkg/windows"
```
aravindhp, line 18: Is there a reason for using `win`. I suggest we only introduce aliases when really needed. Applies to all places.

controllers/controllers.go
```diff
@@ -48,3 +51,49 @@ func (r *instanceReconciler) configureInstance(instance *instances.InstanceInfo,
...
+
+// instanceFromNode returns an instance object for the given node. Requires a username that can be used to SSH into the
+// instance to be annotated on the node.
+func (r *instanceReconciler) instanceFromNode(node *core.Node) (*instances.InstanceInfo, error) {
+	if node.Annotations[UsernameAnnotation] == "" {
+		return nil, errors.New("node is missing valid username annotation")
+	}
+	addr, err := getAddress(node.Status.Addresses)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to get node address")
```
aravindhp, line 63: Will I know which node based on this return?

aravindhp (changes requested):

pkg/windows/windows.go
```diff
@@ -258,6 +275,41 @@ func (vm *windows) EnsureRequiredServicesStopped() error {
 	return nil
 }
 
+// removeServices removes all services installed by WMCO
```
aravindhp, line 278: `ensures all services installed by WMCO are removed`

Might want to rename this `ensureServicesAreRemoved()`

pkg/windows/windows.go
```diff
@@ -75,8 +75,25 @@ const (
...
-var filesToTransfer map[*payload.FileInfo]string
+var (
+	// filesToTransfer is a map of what files should be copied to the Windows VM and where they should be copied to
+	filesToTransfer map[*payload.FileInfo]string
+	// RequiredServices is a list of Windows services installed by WMCO
+	RequiredServices = []string{
+		windowsExporterServiceName,
+		kubeProxyServiceName,
+		hybridOverlayServiceName,
+		kubeletServiceName}
```
aravindhp, line 86: It seems ordering is important here so please add the comment `// This slice order matters due to service dependencies`

pkg/windows/windows.go
```diff
@@ -247,9 +266,7 @@ func (vm *windows) Reinitialize() error {
 
 func (vm *windows) EnsureRequiredServicesStopped() error {
 	// This slice order matters due to service dependencies
```
aravindhp, line 268: This can be removed from here.

aravindhp (approved):
/approve

Caveat: minor change requests are addressed before lgtm

pkg/nodeconfig/nodeconfig.go
```diff
@@ -346,6 +346,35 @@ func (nc *nodeConfig) configureCNI() error {
 	return nil
 }
 
+// Deconfigure removes the node from the cluster, reverting changes made by the Configure function
+func (nc *nodeConfig) Deconfigure() error {
+	drainHelper := &drain.Helper{Ctx: context.TODO(), Client: nc.k8sclientset}
+	// Set nc.node to the existing node
+	if err := nc.setNode(true); err != nil {
```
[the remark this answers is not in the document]
sebsoto: I see, the source of the issue is that the `IP()` method of the windows interface should really be `Address()`, or `NetworkAddress()` and the struct should have `address` instead of `ipAddress`.

I can open a followup PR for these changes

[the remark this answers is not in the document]
sebsoto: To be clear, I dont see a functionality issue, just a naming issue here.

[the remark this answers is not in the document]
aravindhp: I see a functionality issue here, won't
```
if nc.IP() == nodeAddress.Address 
```
fail if only DNS is set?

test/e2e/delete_test.go
aravindhp: FYI, my comment was a question and I would have been satisfied with your answer given this was a situation where the alias was really needed.

pkg/windows/windows.go
```diff
@@ -703,3 +785,8 @@ func (vm *windows) newFileInfo(path string) (*payload.FileInfo, error) {
 func mkdirCmd(dirName string) string {
 	return "if not exist " + dirName + " mkdir " + dirName
 }
+
+// rmDirCmd returns the Windows command to recursively remove a directory if it exists
+func rmDirCmd(dirName string) string {
+	return "if exist " + dirName + " rmdir " + dirName + " /s /q"
```
saifshaikh48, line 791: Are we pushing to standardize Powershell usage over interacting with command prompt?

saifshaikh48:
/lgtm

Left one question, not a major thing though.

mansikulkarni96:
/hold
 incase you want to address https://github.com/openshift/windows-machine-config-operator/pull/498#discussion_r661700128

pkg/windows/windows.go
sebsoto: There I'm going with the pattern set by the existing mkDirCmd function. I've just created https://issues.redhat.com/browse/WINC-636

PR #499, opened 2021-06-30
 WINC-581: Approve CSRs of BYOH instances

mansikulkarni96:
This PR adds the ability to authorize CSRs from BYOH instances by verifying their contents.
If the CSR's are authorized as a valid kubelet client or server CSR, they are approved by WMCO


aravindhp:
/retitle [WIP] WINC-581: Authorize & approve CSRs of BYOH instances

`WINC-581:` Needed for linking to Jira

pkg/csr/csr.go
```diff
@@ -0,0 +1,219 @@
+package csr
+
+import (
+	"context"
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+	"github.com/pkg/errors"
```
sebsoto, line 8: Should be moved down a group

pkg/csr/csr.go
```diff
@@ -0,0 +1,219 @@
...
+	certificatesv1.UsageServerAuth,
+}
+
+const (
+	nodeGroup    = "system:nodes"
+	nodePrefix   = "system:node:"
+	systemPrefix = "system:authenticated"
+)
+
+var log = ctrl.Log.WithName("Config")
```
sebsoto, line 43: Is this used for anything?

pkg/csr/csr.go
```diff
@@ -0,0 +1,219 @@
...
+
+// Config holds the information required to authorize and approve a CSR
+type Config struct {
```
sebsoto, line 46: I dont understand the naming here, this doesnt really seem to be configuring anything

pkg/csr/csr.go
```diff
@@ -0,0 +1,219 @@
...
+	k8sClientSet      *kubernetes.Clientset
+	instanceAddresses []string
+}
+
+// NewConfig returns a pointer to the Config
+func NewConfig(k8sclienSet *kubernetes.Clientset, instanceAddresses []string) *Config {
+	if len(instanceAddresses) == 0 || k8sclienSet == nil {
+	}
```
sebsoto, line 54: Empty statement

pkg/csr/csr.go
```diff
@@ -0,0 +1,219 @@
...
+var log = ctrl.Log.WithName("Config")
+
+// Config holds the information required to authorize and approve a CSR
+type Config struct {
+	k8sClientSet      *kubernetes.Clientset
+	instanceAddresses []string
+}
+
+// NewConfig returns a pointer to the Config
+func NewConfig(k8sclienSet *kubernetes.Clientset, instanceAddresses []string) *Config {
```
sebsoto, line 52: k8sclienSet -> k8sclientSet

controllers/certificatesigningrequests_controller.go
```diff
@@ -46,18 +62,109 @@ type CertificateSigningRequestsReconciler struct {
...
+	for _, host := range hosts {
+		instanceAddresses = append(instanceAddresses, host.Address)
+	}
+
+	certificateSigningRequest := csr.NewConfig(r.k8sclientset, instanceAddresses)
+
+	if authorize, err := certificateSigningRequest.Authorize(request); !authorize {
+		return errors.Wrapf(err, "could not authorize CSR %s", request.Name)
+	}
+	if err := certificateSigningRequest.Approve(request); err != nil {
```
sebsoto, line 156: Whats the reasoning behind having Authorize and Approve as separate operations?
Is there a case where authorizing is required and approving is not? If not, it would make more sense for exported function in the CSR package to be Approve().

pkg/csr/csr.go
```diff
@@ -0,0 +1,219 @@
...
+	if isNodeClientCert(csr, parsedCSR) {
+		// this is done before the instance becomes a node
+		return c.authorizeNodeClientCSR(dnsAddr)
+	} else {
+		return c.authorizeServerCSR(csr, parsedCSR)
+	}
+}
+
+// authorizeServerCSR authorizes a kubelet server CSR
+func (c *Config) authorizeServerCSR(req *certificatesv1.CertificateSigningRequest, csr *x509.CertificateRequest) (bool, error) {
```
sebsoto, line 91: Server as in a master node?
We should never approve a master node CSR.

If this is not about master nodes then a different name and more descriptive comment should be used

pkg/csr/csr.go
```diff
@@ -0,0 +1,219 @@
...
+
+// NewConfig returns a pointer to the Config
+func NewConfig(k8sclienSet *kubernetes.Clientset, instanceAddresses []string) *Config {
+	if len(instanceAddresses) == 0 || k8sclienSet == nil {
+	}
+	return &Config{k8sclienSet, instanceAddresses}
+}
+
+// Authorize returns true if the CSR request contents can be validated
+func (c *Config) Authorize(csr *certificatesv1.CertificateSigningRequest) (bool, error) {
```
sebsoto, line 59: Authorize can be defined as :`give official permission for or approval to`
To reduce confusion I recommend renaming this to `validate()`.
(In addition to making this a private function called as part of `Approve()`)

pkg/csr/csr.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+	return true, nil // Approve node server cert
+}
+
+// authorizeNodeClientCSR authorizes a kubelet client CSR
+func (c *Config) authorizeNodeClientCSR(nodeName string) (bool, error) {
+	if nodeName == "" {
+		return false, errors.New("node name should not be empty")
+	}
+	_, err := c.k8sClientSet.CoreV1().Nodes().Get(context.Background(), nodeName, metav1.GetOptions{})
```
sebsoto, line 128: Use `context.TODO()` for now, in the future we will most likely want to be using a context passed down from main, but I'm not sure if thats correct practice at the moment.

test/e2e/create_test.go
```diff
@@ -293,15 +301,6 @@ func (tc *testContext) waitForWindowsNodes(nodeCount int32, expectError, checkVe
...
 	err = wait.Poll(nodeRetryInterval, time.Duration(math.Max(float64(nodeCount), 1))*creationTime, func() (done bool, err error) {
-		// TODO: vSphere CSRs are not being approved due to a hostname mismatch. This is an issue only because the
-		//       ConfigMap controller does not have the ability to approve CSRs yet. This needs to be removed when that
-		//       functionality is added in https://issues.redhat.com/browse/WINC-581
-		if tc.CloudProvider.GetType() == config.VSpherePlatformType {
-			if err := tc.approveAllCSRs(); err != nil {
-				log.Printf("error approving CSRs: %s", err)
-				return false, nil
-			}
-		}
```
sebsoto, line 304: Removing these changes in a separate commit would make this easier to read, will give the tests a better look after that

test/e2e/validation_test.go
```diff
@@ -334,3 +337,64 @@ func testExpectedServicesRunning(t *testing.T) {
...
+
+// parseCSR extracts the CSR from the API object and decodes it.
+func parseCSR(csr *certificates.CertificateSigningRequest) (*x509.CertificateRequest, error) {
+	// extract PEM from request object
+	block, _ := pem.Decode(csr.Spec.Request)
+	if block == nil || block.Type != "CERTIFICATE REQUEST" {
+		return nil, fmt.Errorf("PEM block type must be CERTIFICATE REQUEST")
+	}
+	return x509.ParseCertificateRequest(block.Bytes)
+}
```
sebsoto, line 400: Since this is repeated code from the CSR package you could export it there + write a unit test and re-use it here

test/e2e/validation_test.go
```diff
@@ -334,3 +337,64 @@ func testExpectedServicesRunning(t *testing.T) {
...
+					if c.Type != certificates.CertificateApproved || c.Reason != "WMCOApproved" {
+						return false
+					}
+				}
+				return true
+			}()
+			assert.Equal(t, isWMCOApproved, true, "expected BYOH node CSR to be approved by WMCO: %s", node.GetName())
+		}
+	}
+
```
sebsoto, line 365: extra newline

sebsoto (changes requested):
Thanks for doing the research and work here @mansikulkarni96, general idea looks great. PTAL at my requests

controllers/certificatesigningrequests_controller.go
```diff
@@ -0,0 +1,170 @@
...
+//+kubebuilder:rbac:groups="certificates.k8s.io",resources=certificatesigningrequests/approval,verbs=update
+//+kubebuilder:rbac:groups="certificates.k8s.io",resources=certificatesigningrequests,verbs=get;list;watch
+//+kubebuilder:rbac:groups="certificates.k8s.io",resources=signers,verbs=approve,resourceNames=kubernetes.io/kube-apiserver-client-kubelet;kubernetes.io/kubelet-serving
+//+kubebuilder:rbac:groups=certificates.k8s.io,resources=certificatesigningrequests/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=certificates.k8s.io,resources=certificatesigningrequests/finalizers,verbs=update
+//+kubebuilder:rbac:groups="",resources=configmaps,verbs=get
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+// the CertificateSigningRequests object against the actual cluster state, and then
```
alinaryan, line 59: "the CertificateSigningRequests object against the actual cluster state" This sentence is confusing to me for some reason. Could we use "compare" instead of "object"?

controllers/certificatesigningrequests_controller.go
```diff
@@ -0,0 +1,170 @@
...
+			return pendingCertFilter(e.Object)
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool { return false },
+	}
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&certificatesv1.CertificateSigningRequest{}, builder.WithPredicates(certificateSigningRequestsPredicate)).
+		Complete(r)
+}
+
+// pendingCertFilter returns true if the CSR is pending approval
```
alinaryan, line 129: The comment of this function sounds really similiar to the isPending() function. Could we call out the difference?

pkg/csr/csr.go
```diff
@@ -0,0 +1,219 @@
...
+// Approve approves a CSR by updating it's status conditions
+func (c *Config) Approve(csr *certificatesv1.CertificateSigningRequest) error {
+	csr.Status.Conditions = append(csr.Status.Conditions, certificatesv1.CertificateSigningRequestCondition{
+		Type:           certificatesv1.CertificateApproved,
+		Status:         "True",
+		Message:        "This CSR was approved by the WMCO CSR Approver",
+		LastUpdateTime: metav1.Now(),
+		Reason:         "WMCOApproved",
+	})
+	if _, err := c.k8sClientSet.CertificatesV1().CertificateSigningRequests().UpdateApproval(context.Background(), csr.Name, csr, metav1.UpdateOptions{}); err != nil {
```
alinaryan, line 146: over 120 column

controllers/certificatesigningrequests_controller.go
```diff
@@ -0,0 +1,170 @@
...
+// reconcileCSR handles the CSR authorization and approval
+func (r *certificateSigningRequestsReconciler) reconcileCSR(request *certificatesv1.CertificateSigningRequest, instances *core.ConfigMap) error {
+	if request == nil || instances == nil {
+		return errors.New("CSR or instance ConfigMap cannot be nil")
+	}
+
+	// Get the list of instances that are expected to be Nodes
+	hosts, err := parseHosts(instances.Data)
+	if err != nil {
+		return errors.Wrapf(err, "unable to parse hosts from configmap")
```
jrvaldes, line 144: No need for **Wrapf** since no params in message.

controllers/certificatesigningrequests_controller.go
```diff
@@ -0,0 +1,170 @@
...
+	return ok && isPending(cert)
+}
+
+// reconcileCSR handles the CSR authorization and approval
+func (r *certificateSigningRequestsReconciler) reconcileCSR(request *certificatesv1.CertificateSigningRequest, instances *core.ConfigMap) error {
+	if request == nil || instances == nil {
+		return errors.New("CSR or instance ConfigMap cannot be nil")
+	}
+
+	// Get the list of instances that are expected to be Nodes
```
jrvaldes, line 141: s/instances/hosts

pkg/csr/csr.go
```diff
@@ -0,0 +1,219 @@
...
+var kubeletServerUsages = []certificatesv1.KeyUsage{
+	certificatesv1.UsageKeyEncipherment,
+	certificatesv1.UsageDigitalSignature,
+	certificatesv1.UsageServerAuth,
+}
+
+const (
+	nodeGroup    = "system:nodes"
+	nodePrefix   = "system:node:"
+	systemPrefix = "system:authenticated"
```
jrvaldes, line 40: Looks like there is an extra colon (:) in `"system:node:"`
 
 Since `nodePrefix` and `systemPrefix` are both `Prefix`, they should have the same schema.

pkg/csr/csr.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+	if !hasMatch {
+		return false, errors.Errorf("CSR is not from an authorized source %s", csr.Name)
+	}
+
+	if isNodeClientCert(csr, parsedCSR) {
+		// this is done before the instance becomes a node
+		return c.authorizeNodeClientCSR(dnsAddr)
+	} else {
+		return c.authorizeServerCSR(csr, parsedCSR)
```
jrvaldes, line 86: No need for a `else` statement, since this is the default return.

test/e2e/validation_test.go
```diff
@@ -334,3 +337,64 @@ func testExpectedServicesRunning(t *testing.T) {
...
+		for _, csr := range csrs {
+			isWMCOApproved := func() bool {
+				for _, c := range csr.Status.Conditions {
+					if c.Type != certificates.CertificateApproved || c.Reason != "WMCOApproved" {
+						return false
+					}
+				}
+				return true
+			}()
```
saifshaikh48, line 361: Can CSRs be approved by multiple entities and still be considered valid for our use case? If so, then current logic does not hold.

Assume a CSR has 2 Conditions, `c1` and `c2`, in its status:
`c1.Type = certificates.CertificateApproved` and `c1.Reason != "SomeApproverThatIsNotWMCO"`
`c2.Type = certificates.CertificateApproved` and `c2.Reason != "WMCOApproved"`

Loop would return false (even if `c2` gets processed first) even though the CSR is WMCO approved.

```suggestion
			isWMCOApproved := func() bool {
				for _, c := range csr.Status.Conditions {
					if c.Type == certificates.CertificateApproved && c.Reason == "WMCOApproved" {
						return true
					}
				}
				return false
			}()
```

pkg/csr/csr.go
aravindhp: It seems this function is also checking if this is a CSR generated for a Windows instance by checking the DNS of the node.

pkg/csr/csr.go
```diff
@@ -0,0 +1,219 @@
...
+		Reason:         "WMCOApproved",
+	})
+	if _, err := c.k8sClientSet.CertificatesV1().CertificateSigningRequests().UpdateApproval(context.Background(),
+		csr.Name, csr, metav1.UpdateOptions{}); err != nil {
+		return err
+	}
+	return nil
+}
+
+// isNodeClientCert returns true if the CSR is a kubelet client CSR
```
aravindhp, line 152: Please link to documentation where this is defined i.e. how did you find out what makes a valid kubelet client CSR?

pkg/csr/csr.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+	if !hasOrg {
+		return false, errors.Errorf("CSR does not contain required organization: %s", req.Name)
+	}
+	return true, nil // Approve node server cert
+}
+
+// authorizeNodeClientCSR validates a kubelet client CSR by verifying if it already
+// exists as a node because client CSR's are issued by an instance before becoming a node.
+func (c *Config) authorizeNodeClientCSR(nodeName string) (bool, error) {
```
aravindhp, line 123: I don't see this function doing any authorization. It seems to be checking if a node object exists.

pkg/csr/csr.go
```diff
@@ -0,0 +1,219 @@
...
+
+	if isNodeClientCert(csr, parsedCSR) {
+		// this is done before the instance becomes a node
+		return c.authorizeNodeClientCSR(dnsAddr)
+	}
+
+	return c.authorizeKubeletServingCSR(csr, parsedCSR)
+}
+
+// authorizeKubeletServingCSR authorizes a kubelet serving CSR
```
aravindhp, line 88: I don't see this function doing any authorization

pkg/csr/csr.go
```diff
@@ -0,0 +1,219 @@
...
+// isNodeClientCert returns true if the CSR is a kubelet client CSR
+func isNodeClientCert(csr *certificatesv1.CertificateSigningRequest, x509cr *x509.CertificateRequest) bool {
+	if !reflect.DeepEqual([]string{nodeGroup}, x509cr.Subject.Organization) {
+		return false
+	}
+	if (len(x509cr.DNSNames) > 0) || (len(x509cr.EmailAddresses) > 0) || (len(x509cr.IPAddresses) > 0) {
+		return false
+	}
+	// Check usages include: digital signature, key encipherment and client auth
+	if !hasExactUsages(csr, kubeletClientUsages) {
```
aravindhp, line 161: Why does this need to be checked here and in `authorizeKubeletServingCSR`.

test/e2e/create_test.go
aravindhp: Why? This commit is introducing changes that requires a change in the e2e tests. Question to ask is, can I revert this individual commit?

test/e2e/create_test.go
```diff
@@ -146,7 +146,15 @@ func (tc *testContext) testBYOHConfiguration(t *testing.T) {
 	if gc.numberOfBYOHNodes == 0 {
 		t.Skip("BYOH testing disabled")
 	}
-	_, err := tc.createWindowsMachineSet(gc.numberOfBYOHNodes, false)
+
+	// Scale the Cluster Machine Approver Deployment pods to 0
+	// This is required for testing BYOH CSR approval feature so that BYOH instances
+	// CSR's are not approved by Cluster Machine Approver
+	expectedPodCount := int32(0)
+	err := tc.scaleMachineApproverDeployment(&expectedPodCount)
```
aravindhp, line 154: How is this working? Won't CVO stomp on this change? I also don't understand why @sebsoto did not have to do this in https://github.com/openshift/windows-machine-config-operator/pull/446

aravindhp (changes requested):
Thanks for working on this, @mansikulkarni96. PTAL at my comments. I am also puzzled by why there is no mention of the bootstrap CSR like we did in the [playbook](https://github.com/openshift/windows-machine-config-bootstrapper/blob/release-4.5/tools/ansible/tasks/wsu/main.yaml#L289).

pkg/csr/csr.go
```diff
@@ -0,0 +1,219 @@
...
+	// check DNS of node requesting CSR matches instance IP/DNS address
+	var hasMatch bool
+	for _, instanceAdd := range c.instanceAddresses {
+		if areEqualAddresses(instanceAdd, dnsAddr) {
+			hasMatch = true
+			break
+		}
+	}
+	if !hasMatch {
+		return false, errors.Errorf("CSR is not from an authorized source %s", csr.Name)
```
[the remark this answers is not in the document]
mansikulkarni96: handled this by not returning this error instead logging where required.

test/e2e/validation_test.go
aravindhp: Couldn't the CSR be approved by the `cluster-machine-approver` in a real world cluster? I am a bit worried that given we are scaling that deployment down to 0, we are never testing that scenario.

mansikulkarni96: I have accepted @saifshaikh48 suggestion on this. Additionally I don't think BYOH CSR's will be approved by cluster-machine-approver because the node to machine mapping will not be found which is used by cluster-machine-approver for validation.

pkg/csr/csr_test.go
```diff
@@ -0,0 +1,84 @@
...
+		expectedErr bool
+	}{
+		{
+			name:        "valid CSR",
+			input:       []byte(goodCSR),
+			expectedErr: false,
+		},
+		{
+			name:        "invalid CSR",
+			input:       []byte(emptyCSR),
```
aravindhp, line 68: In addition shouldn't there be a test for a  CSR with an invalid cert?

pkg/csr/csr_test.go
```diff
@@ -0,0 +1,84 @@
...
+		input       []byte
+		expectedErr bool
+	}{
+		{
+			name:        "valid CSR",
+			input:       []byte(goodCSR),
+			expectedErr: false,
+		},
+		{
+			name:        "invalid CSR",
```
aravindhp, line 67: `empty CSR`

pkg/csr/csr.go
```diff
@@ -0,0 +1,236 @@
...
+		if areAddressesEqual(instanceAdd, dnsAddr) {
+			hasMatch = true
+			break
+		}
+	}
+	if !hasMatch {
+		// CSR is not from a BYOH Windows instance, don't deny it or return error
+		// since it might be from a linux node.
+		return false, nil
+	}
```
aravindhp, line 85: Can't this validation be done in the predicate?

pkg/csr/csr.go
```diff
@@ -0,0 +1,236 @@
...
+	// Both types are validated based on their contents
+	var validated bool
+	if isNodeClientCert(csr, parsedCSR) {
+		// Node client bootstrapper CSR is received before the instance becomes a node
+		// hence we should not proceed if a corresponding node already exists
+		_, err := c.k8sClientSet.CoreV1().Nodes().Get(context.TODO(), dnsAddr, metav1.GetOptions{})
+		if err != nil && !apierrors.IsNotFound(err) {
+			return false, errors.Wrapf(err, "unable to get node %s", dnsAddr)
+		} else if err == nil {
+			return false, errors.Wrapf(err, "%s node already exists, cannot validate CSR", dnsAddr)
```
aravindhp, line 98: Is `dnsAddr` equivalent to the `node.Name`? Because that is what this error is hinting at. 

Is this how the `cluster-machine-approver` is handling this scenario?

pkg/csr/csr.go
```diff
@@ -0,0 +1,236 @@
...
+
+// validateKubeletServingCSR validates a kubelet serving CSR for its contents
+func (c *CSRApprover) validateKubeletServingCSR(req *certificatesv1.CertificateSigningRequest, csr *x509.CertificateRequest) (bool, error) {
+	if req == nil || csr == nil {
+		return false, errors.New("CSR or request should not be nil")
+	}
+
+	// Check groups, we need at least: system:nodes, system:authenticated
+	if len(req.Spec.Groups) < 2 {
+		return false, errors.Errorf("CSR contains too few groups: %s", req.Name)
```
aravindhp, line 116: `CSR %s contains invalid number of groups: %d`

I suggest reflecting this pattern in other errors also.

pkg/csr/csr.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+	groupSet := sets.NewString(req.Spec.Groups...)
```
aravindhp, line 118: I suggest naming this `groups`. It is not good practice to reflect the data structure being used in the variable name.

pkg/csr/csr.go
aravindhp: Thank you for the explanation. My bad that I did not notice that the inputs are different at the call sites.

pkg/csr/csr.go
```diff
@@ -0,0 +1,236 @@
...
+	if !hasExactUsages(csr, kubeletClientUsages) {
+		return false
+	}
+	if !strings.HasPrefix(x509cr.Subject.CommonName, NodeUserPrefix) {
+		return false
+	}
+	return true
+}
+
+// hasExactUsages verifies if the required usages exist in the CSR spec
```
aravindhp, line 184: I suggest naming it `hasUsages` given the input is a list of usages that you are checking against.

pkg/csr/csr.go
```diff
@@ -0,0 +1,236 @@
...
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/kubernetes"
+)
+
+//+kubebuilder:rbac:groups="certificates.k8s.io",resources=certificatesigningrequests/approval,verbs=update
+//+kubebuilder:rbac:groups="certificates.k8s.io",resources=certificatesigningrequests,verbs=get;list;watch
+//+kubebuilder:rbac:groups="certificates.k8s.io",resources=signers,verbs=Approve,resourceNames=kubernetes.io/kube-apiserver-client-kubelet;kubernetes.io/kubelet-serving
+
+// kubeletClientUsages contains the permitted key usages from a kube-apiserver-client-kubelet signer
+var kubeletClientUsages = []certificatesv1.KeyUsage{
```
aravindhp, line 26: Looking at how this is being used, wouldn't it be easier if this was a set?

pkg/csr/csr.go
```diff
@@ -0,0 +1,236 @@
...
+			return false
+		}
+	}
+
+	return true
+}
+
+// getDnsAddresses returns list of dns addresses corresponding to the given IPv4 address
+func getDnsAddresses(ipAddr string) []string {
+	addr, _ := net.LookupAddr(ipAddr)
```
aravindhp, line 206: I should have caught this in my previous review. You are doing a reverse look  up here. What if you are in an environment where this does not resolve? Moreover every IP needn't have a DNS name associated with it. We should discuss this. I would like to understand why this is needed.

test/e2e/create_test.go
aravindhp: Thanks for the explanation. We should add the override in this spot.

aravindhp (changes requested):

pkg/csr/csr.go
mansikulkarni96: I think the naming caused some confusion here, it should be named nodeName. The cluster-machine-appover handles this by matching the nodeName with the machine.status.address && type internalDNSAddress.

pkg/csr/csr.go
mansikulkarni96: this was taken from the way cluster-machine-approver uses it, I have added the link to approver code in the commit msg and the CSR package header now.
https://github.com/openshift/cluster-machine-approver/blob/master/pkg/controller/node_client.go#L32

pkg/csr/csr.go
mansikulkarni96: modified this to checking the platform type and deciding to get the actual hostname vs lookup the hostname depending on the usecase.

sebsoto (changes requested):

controllers/certificatesigningrequests_controller.go
```diff
@@ -19,45 +19,195 @@ package controllers
...
+	// Get the list of hosts that are expected to be Nodes
+	hosts, err := parseHosts(configMap.Data)
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to parse hosts from ConfigMap %s", configMap.Name)
+	}
+	var requireVMHostName bool
+	// platform type set None or vSphere indicates hostName required is the actual hostName of the VM
+	if r.platform == oconfig.NonePlatformType || r.platform == oconfig.VSpherePlatformType {
+		requireVMHostName = true
+	}
```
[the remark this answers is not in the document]
mansikulkarni96: If the platform type is set to vSphere or None, the future node name that gets constructed is using the actual hostname of the VM, otherwise it is constructed using the instance host address(mapped from instance configmap). I ll try to add a more clear comment, if that's the issue here.

controllers/certificatesigningrequests_controller.go
```diff
@@ -19,45 +19,195 @@ package controllers
...
+
+	var hostNames []string
+	for _, host := range hosts {
+		hostName := host.Address
+		if requireVMHostName {
+			nc, err := nodeconfig.NewNodeConfig(r.k8sclientset, r.clusterServiceCIDR, r.vxlanPort, host, r.signer,
```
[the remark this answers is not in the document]
mansikulkarni96: yes that's what I am using nodeconfig for but I used it here to avoid getting into populating `nodeConfigCache.workerIgnitionEndPoint ` when calling https://github.com/openshift/windows-machine-config-operator/blob/4a90df34920a4c295c9cfefaa896cfa799576748/pkg/nodeconfig/nodeconfig.go#L114

pkg/csr/csr.go
```diff
@@ -0,0 +1,248 @@
...
+		return false, errors.Errorf("CSR %s does not contain required subject organization", c.Name)
+	}
+	return true, nil
+}
+
+// Approve approves a CSR by updating it's status conditions to true if it is a valid CSR
+func (c *Approver) Approve() error {
+	if validated, err := c.validateCSRContents(); !validated {
+		c.log.Info("CSR contents are invalid for approval by WMCO CSR Approver", "CSR", c.Name)
+		return errors.Wrapf(err, "could not validate CSR %s contents for approval", c.Name)
```
aravindhp, line 156: Will we constantly keep trying to reconcile this invalid CSR?

pkg/csr/csr.go
```diff
@@ -0,0 +1,248 @@
...
+		if _, ok := usageMap[u]; !ok {
+			return false
+		}
+	}
+
+	return true
+}
+
+// getDnsAddresses returns list of dns addresses corresponding to the given IPv4 address
+func getDnsAddresses(ipAddr string) []string {
```
aravindhp, line 215: If this no longer being used we should remove this function.

controllers/certificatesigningrequests_controller.go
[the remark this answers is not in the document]
sebsoto: Elaborated more on my thoughts offline, quick recap:
* Nodeconfig has a purpose of Configuring and Deconfiguring nodes, it having a `Windows` interface present in it is a means to achieving that.
* In this case what is needed is not configuring instances but being able to run a single command against an instance, for that, instantiating a Windows object is more proper.
* We shouldnt need `workerIgnitionEndPoint` to run a command against an instance, but that is a current restriction put in place by the implementation of `windows.New`
* If removing that restriction is too much work, I am fine with the change not being made, but I think a TODO and an associated ticket should be created for this.

controllers/certificatesigningrequests_controller.go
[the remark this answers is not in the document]
aravindhp: So a reverse look up is not needed? I thought from the Slack thread, the direction was to try doing the reverse lookup and if that fails, use the hostname. Or did I misunderstand?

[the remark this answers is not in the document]
sebsoto: Also talked about this, I think it would be great to have the high level CSR approval process documented. I am having a hard time wrapping my head around all the changes made here, and having all the considerations laid out either at the top of the csr.go file, or as a separate doc in the `docs` folder would be very helpful

pkg/csr/csr.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+	return x509.ParseCertificateRequest(block.Bytes)
+}
+
+// areAddressesEqual returns true if the dns address in CSR subject matches with the instance address
+// If the address passed is an IP address, we do a reverse lookup to find the host address
+func (c *Approver) areAddressesEqual(instanceAdd string, csrAddr string) bool {
+	// reverse lookup the host if the address is an IP address
+	if parseAddr := net.ParseIP(instanceAdd); parseAddr != nil {
+		dnsAddresses, _ := net.LookupAddr(instanceAdd)
```
aravindhp, line 238: Why are you ignoring the error? I am also struggling to understand the flow. Won't this fail if the reverse look up fails?

aravindhp (changes requested):
Thanks for working on this, @mansikulkarni96. I suggest calling a meeting to resolve the issues brought up here.

pkg/csr/csr.go
mansikulkarni96: no we wouldn't, this will return error only if the err in `return errors.Wrapf(err, "could not validate CSR %s contents for approval", c.Name)` is not nil, which could be returned due to parsing errors. If CSR is not a BYOH CSR we are only logging here.

pkg/csr/csr.go
mansikulkarni96: I ll add a error check here, I am thinking if it makes more sense to do the platform check here for the fall back option if this fails.

pkg/csr/csr.go
```diff
@@ -0,0 +1,260 @@
...
+}
+
+// areAddressesEqual returns true if the DNS address in CSR subject matches with the instance address
+// If the address passed is an IP address, we do a reverse lookup to find the host address
+func (c *Approver) areAddressesEqual(instanceAdd string, nodeName string) bool {
+	// reverse lookup the host if the address is an IP address
+	if parseAddr := net.ParseIP(instanceAdd); parseAddr != nil {
+		dnsAddresses, err := net.LookupAddr(instanceAdd)
+		if err != nil {
+			c.log.Info("failed to lookup hosts using", "IP", instanceAdd)
```
jrvaldes, line 246: Not quite sure about logging a "failed" with `Info` level.

Following the calls stack, [c.validateCSRContents()](https://github.com/openshift/windows-machine-config-operator/pull/499/files#diff-666ed6f6b4dbed560da68743f3814f61952cd552103ea310e478cde0c5cbed2eR76) throws an error if validated==false.

In that sense, returning an error here may be helpful, while reduce the burden of maintaining a boolean flag to control the execution flow.

pkg/csr/csr.go
```diff
@@ -0,0 +1,260 @@
...
+		}
+	}
+
+	return true
+}
+
+// ParseCSR extracts the CSR from the API object and decodes it.
+func ParseCSR(csr []byte) (*x509.CertificateRequest, error) {
+	if len(csr) == 0 {
+		return nil, fmt.Errorf("CSR request spec should not be empty")
```
jrvaldes, line 229: no need for formatting `Errorf`

pkg/csr/csr.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+	// extract PEM from request object
+	block, _ := pem.Decode(csr)
+	if block == nil || block.Type != "CERTIFICATE REQUEST" {
+		return nil, fmt.Errorf("PEM block type must be CERTIFICATE REQUEST")
```
jrvaldes, line 234: ditto

pkg/csr/csr.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+	return x509.ParseCertificateRequest(block.Bytes)
+}
+
+// areAddressesEqual returns true if the DNS address in CSR subject matches with the instance address
+// If the address passed is an IP address, we do a reverse lookup to find the host address
+func (c *Approver) areAddressesEqual(instanceAdd string, nodeName string) bool {
```
jrvaldes, line 241: Would be nice to unit test this logic.

pkg/csr/csr.go
```diff
@@ -0,0 +1,260 @@
...
+
+// kubeletClientUsages contains the permitted key usages from a kube-apiserver-client-kubelet signer
+var kubeletClientUsages = []certificates.KeyUsage{
+	certificates.UsageKeyEncipherment,
+	certificates.UsageDigitalSignature,
+	certificates.UsageClientAuth,
+}
+
+// kubeletServerUsages contains the permitted key usages from a kubelet-serving signer
+var kubeletServerUsages = []certificates.KeyUsage{
```
jrvaldes, line 40: This is a slice that can be mutated. 

Does `kubeletServerUsages` meant to be a constant? If so, consider something like:
```
func getKubeletServerUsages() []certificates.KeyUsage {
        // where you return a copy of a fixed array
	return [...]certificates.KeyUsage{    }
}
```

pkg/csr/csr.go
mansikulkarni96: We do not want to return an error in cases where validation fails due to the instance address mismatch since the CSR could be from a linux instance.
 If validated == false we are only logging the error, the return errors.wrap [here](https://github.com/openshift/windows-machine-config-operator/pull/499/files#diff-666ed6f6b4dbed560da68743f3814f61952cd552103ea310e478cde0c5cbed2eR78)  would return nil if the err = nil.
I am trying to only return errors in CSR validation failure except the address mismatch.

controllers/certificatesigningrequests_controller.go
```diff
@@ -19,45 +19,192 @@ package controllers
...
+	}
+
+	certificateSigningRequest, err := csr.NewApprover(request, r.k8sclientset, instanceAddresses, hostNames, r.log)
+	if err != nil {
+		return errors.Wrapf(err, "could not create WMCO CSR Approver")
+	}
+
+	if err := certificateSigningRequest.Approve(); err != nil {
+		return errors.Wrapf(err, "WMCO CSR Approver could not approve %s CSR", request.Name)
+	}
```
sebsoto, line 153: In the case we handle a CSR we do not want to approve, this will return error and the CSR will requeue indefinitely.

sebsoto (changes requested):

controllers/certificatesigningrequests_controller.go
mansikulkarni96: no it will not return error if we do not want to approve CSR, it will only log in that case if validated = false(this is assuming you are talking about CSR's from non-BYOH instances).
It will only return error if there is error validating CSR due to parsing /the CSR contents are invalid(once it has been validated to be from an instance present in configmap)/ error with updating approval conditions.
[ValidateCSRContents() ](https://github.com/openshift/windows-machine-config-operator/pull/499/files#diff-666ed6f6b4dbed560da68743f3814f61952cd552103ea310e478cde0c5cbed2eR76) is something to look at to understand when the error is being returned.

I would think we need to requeue in the cases mentioned above as the node would not become available if we don't.

pkg/csr/csr.go
aravindhp: Do we really need to do this given this is a private member that is not exposed to the outside world?

pkg/csr/csr.go
aravindhp: The error should be `failed to lookup DNS for IP`. 

What happens if we fail to reverse lookup an `instanceAddress` and it is not of a Linux node but that of a valid Windows instance whose CSR we should have approved? How are we going to convey that to the user?

pkg/csr/csr.go
```diff
@@ -0,0 +1,260 @@
...
+	if block == nil || block.Type != "CERTIFICATE REQUEST" {
+		return nil, fmt.Errorf("PEM block type must be CERTIFICATE REQUEST")
+	}
+	return x509.ParseCertificateRequest(block.Bytes)
+}
+
+// areAddressesEqual returns true if the DNS address in CSR subject matches with the instance address
+// If the address passed is an IP address, we do a reverse lookup to find the host address
+func (c *Approver) areAddressesEqual(instanceAdd string, nodeName string) bool {
+	// reverse lookup the host if the address is an IP address
```
aravindhp, line 242: s/host/instance

controllers/configmap_controller.go
```diff
@@ -123,7 +123,7 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 }
 
 // parseHosts gets the lists of hosts specified in the configmap's data
-func (r *ConfigMapReconciler) parseHosts(configMapData map[string]string) ([]*instances.InstanceInfo, error) {
+func parseHosts(configMapData map[string]string) ([]*instances.InstanceInfo, error) {
```
aravindhp, line 126: Given you are changing this, please rename the function to `parseInstances` and mention it in the commit.

pkg/csr/csr.go
aravindhp: `areAddressesEqual` does not convey the intent of this function. You are trying to see if the `nodeName` maps to `instanceAdd`

pkg/csr/csr.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+	// check if the node name matches host names of the instances
+	for _, hostName := range c.hostNames {
+		if strings.Contains(hostName, nodeName) {
+			return true
+		}
+	}
+	// check if the node name matches the lookup of address in configMap
+	for _, address := range c.instanceAddresses {
+		return c.areAddressesEqual(address, nodeName)
```
aravindhp, line 152: By returning here won't you be checking with just the first entry in `c.instanceAddresses`?

pkg/csr/csr.go
```diff
@@ -0,0 +1,260 @@
...
+}
+
+// validateNodeName returns true if the node name passed here matches
+// either the actual host name of the VM'S or the reverse lookup of
+// the instance addresses present in the configMap
+func (c *Approver) validateNodeName(nodeName string) bool {
+	if len(c.hostNames) == 0 {
+		c.log.Info("no host names found to validate the CSR")
+		return false
+	}
```
aravindhp, line 143: This validation is a hint that you should be collecting the `hostNames` here.

aravindhp (changes requested):

pkg/csr/csr.go
```diff
@@ -0,0 +1,357 @@
...
+var kubeletServerUsages = []certificates.KeyUsage{
+	certificates.UsageKeyEncipherment,
+	certificates.UsageDigitalSignature,
+	certificates.UsageServerAuth,
+}
+
+const (
+	nodeGroup          = "system:nodes"
+	nodeUserName       = "system:node"
+	NodeUserNamePrefix = nodeUserName + ":"
```
saifshaikh48, line 61: I don't see where else `nodeUserName` is used. Should we consolidate into one var `NodeUserNamePrefix = "system:node:"`?

pkg/csr/csr.go
```diff
@@ -0,0 +1,357 @@
...
+	}
+	return &Approver{client,
+		clientSet,
+		csr,
+		log,
+		recorder,
+		watchNamespace}, nil
+}
+
+// Approve approves a CSR by updating it's status conditions to true if it is a valid CSR
```
saifshaikh48, line 93: nit 
`...updating its status...`

pkg/csr/csr.go
```diff
@@ -0,0 +1,357 @@
...
+}
+
+// validateKubeletServingCSR validates a kubelet serving CSR for its contents
+func (a *Approver) validateKubeletServingCSR(parsedCsr *x509.CertificateRequest) (bool, error) {
+	if a.csr == nil || parsedCsr == nil {
+		return false, errors.New("CSR or request should not be nil")
+	}
+
+	// Check groups, we need at least: system:nodes, system:authenticated
+	if len(a.csr.Spec.Groups) < 2 {
```
saifshaikh48, line 227: Is this check enough or should this iterate over the groups to the right values exist? Could be a helper similar to `hasUsages`

saifshaikh48 (commented):
Thanks for taking this on @mansikulkarni96. Just a few comments

pkg/csr/csr.go
```diff
@@ -0,0 +1,357 @@
...
+	if _, err := a.k8sclientset.CertificatesV1().CertificateSigningRequests().UpdateApproval(context.Background(),
+		a.csr.Name, a.csr, meta.UpdateOptions{}); err != nil {
+		return errors.Wrapf(err, "could not update conditions for CSR approval %s", a.csr.Name)
+	}
+	a.log.Info("CSR approved", "CSR", a.csr.Name)
+	return nil
+}
+
+// validateCSRContents returns true if the CSR request contents are valid
+func (a *Approver) validateCSRContents() (bool, error) {
```
sebsoto, line 121: Whats the point in the bool return type if it only ever returns
false + error
or
true + no error?
error always implies false and no error implies true

pkg/csr/csr.go
```diff
@@ -0,0 +1,357 @@
...
+// Approve approves a CSR by updating it's status conditions to true if it is a valid CSR
+func (a *Approver) Approve() error {
+	if a.k8sclientset == nil {
+		return errors.New("Kubernetes clientSet should not be nil")
+	}
+
+	if validated, err := a.validateCSRContents(); !validated {
+		a.log.Info("CSR contents are invalid for approval by WMCO CSR Approver", "CSR", a.csr.Name)
+		return errors.Wrapf(err, "could not validate CSR %s contents for approval", a.csr.Name)
+	}
```
sebsoto, line 102: This doesnt seem to be filtering things out correctly. If err == nil and !validated it should return nil here? If not its going to cause a requeue on something we dont want to approve

pkg/csr/csr.go
```diff
@@ -0,0 +1,357 @@
...
+	if nodeName == "" {
+		return false, errors.Errorf("CSR subject name does not contain the required node user prefix %s",
+			a.csr.Name)
+	}
+
+	// lookup the node name against the instance configMap/host names
+	if valid, err := a.validateNodeName(nodeName); !valid {
+		// CSR is not from a BYOH Windows instance, don't deny it or return error
+		// since it might be from a linux node.
+		return false, errors.Wrapf(err, "error validating node name %s for CSR %s", nodeName, a.csr.Name)
```
sebsoto, line 137: Comment and functionality dont match, you are returning error here.

pkg/csr/csr.go
```diff
@@ -0,0 +1,357 @@
...
+	hasEntry, err = hasAssociatedHost(nodeName, windowsInstances, instanceSigner)
+	if err != nil {
+		return false, errors.Wrapf(err, "unable to map node name to the host names of instances in configMap %s",
+			configMap.Name)
+	}
+	if !hasEntry {
+		return false, nil
+	}
+	// validate node name for DNS RFC1123 naming conventions
+	// ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names
```
sebsoto, line 210: Are you sure nodes names need to follow this standard or the more strict DNS label standards? https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names

pkg/csr/csr.go
```diff
@@ -0,0 +1,357 @@
...
+		a.recorder.Eventf(a.csr, core.EventTypeWarning, "nodeNameValidationFailed",
+			"node name %s does not comply with naming rules defined in RFC1123: "+
+				"Requirements for internet hosts", nodeName)
+		return false, nil
+	}
+	return true, nil
+}
+
+// validateKubeletServingCSR validates a kubelet serving CSR for its contents
+func (a *Approver) validateKubeletServingCSR(parsedCsr *x509.CertificateRequest) (bool, error) {
```
sebsoto, line 221: this function also only returns false + error or true + nil.

sebsoto (changes requested):
Good work mansi, I have a few more requests

pkg/csr/csr.go
mansikulkarni96: It was returning false + nil from the validateNodeName() call if the match wasn't found
Realized that could be confusing to read, hence modified the function to return false, nil if node name validation fails with no error.

pkg/csr/csr.go
mansikulkarni96: updated to only return error

main.go
```diff
@@ -191,11 +191,12 @@ func main() {
 		os.Exit(1)
 	}
 
-	if err = (&controllers.CertificateSigningRequestsReconciler{
-		Client: mgr.GetClient(),
-		Log:    ctrl.Log.WithName("controllers").WithName("CertificateSigningRequests"),
-		Scheme: mgr.GetScheme(),
-	}).SetupWithManager(mgr); err != nil {
+	certificateSigningRequestsReconciler, err := controllers.NewCertificateSigningRequestsReconciler(mgr, clusterConfig, watchNamespace)
```
aravindhp, line 194: Going past 120 col

sebsoto (changes requested):
Thanks for the changes and clarifications @mansikulkarni96
Just a few more requests

pkg/csr/csr.go
aravindhp: Does this need to be a global variable though?

pkg/csr/csr.go
```diff
@@ -0,0 +1,350 @@
...
+	"github.com/openshift/windows-machine-config-operator/pkg/windows"
+	"github.com/openshift/windows-machine-config-operator/pkg/wiparser"
+)
+
+//+kubebuilder:rbac:groups="certificates.k8s.io",resources=certificatesigningrequests,verbs=update/approval
+//+kubebuilder:rbac:groups="certificates.k8s.io",resources=certificatesigningrequests,verbs=get;list;watch
+//+kubebuilder:rbac:groups="certificates.k8s.io",resources=signers,verbs=Approve,resourceNames=kubernetes.io/kube-apiserver-client-kubelet;kubernetes.io/kubelet-serving
+
+// kubeletClientUsages contains the permitted key usages from a kube-apiserver-client-kubelet signer
+var kubeletClientUsages = []certificates.KeyUsage{
```
aravindhp, line 44: Does this need to be a global variable? I see it used only in one spot.

pkg/csr/csr.go
```diff
@@ -0,0 +1,350 @@
...
+func (a *Approver) validateCSRContents() (bool, error) {
+	parsedCSR, err := ParseCSR(a.csr.Spec.Request)
+	if err != nil {
+		return false, errors.Wrapf(err, "error parsing CSR %s", a.csr.Name)
+	}
+
+	nodeName := strings.TrimPrefix(parsedCSR.Subject.CommonName, NodeUserNamePrefix)
+	if nodeName == "" {
+		return false, errors.Errorf("CSR subject name does not contain the required node user prefix %s",
+			a.csr.Name)
```
aravindhp, line 131: This implies the node user prefix is `a.csr.Name`. Is that correct or should `%s` be moved after `CSR`?

pkg/csr/csr.go
aravindhp: The call site of this function looks strange as we are not checking the error there. I wish we could be more explicit about it. But I will let it go for now.

pkg/wiparser/wiparser.go
```diff
@@ -0,0 +1,101 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/openshift/windows-machine-config-operator/pkg/instances"
+)
+
+// InstanceConfigMap is the name of the ConfigMap where VMs to be configured should be described.
+const InstanceConfigMap = "windows-instances"
+
+// FindInstances returns a list of Windows instances by parsing the Windows instance configMap.
+func FindInstances(client client.Client, namespace string) ([]*instances.InstanceInfo, error) {
```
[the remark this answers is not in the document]
aravindhp: I suggest calling it `getInstances()` or if you name the package `windowsinstances`, this would be `Get()`.

pkg/csr/csr.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+	// validate node name for DNS RFC1123 naming conventions
+	// ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names
+	if errs := validation.IsDNS1123Subdomain(nodeName); len(errs) > 0 {
+		a.recorder.Eventf(a.csr, core.EventTypeWarning, "nodeNameValidationFailed",
+			"node name %s does not comply with naming rules defined in RFC1123: "+
+				"Requirements for internet hosts", nodeName)
+		return false, errors.Errorf("node name %s should comply with naming rules defined in RFC1123: "+
+			"Requirements for internet hosts", nodeName)
+	}
```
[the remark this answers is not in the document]
aravindhp: +1

pkg/csr/csr.go
```diff
@@ -0,0 +1,350 @@
...
+}
+
+// findHostName returns the actual host name of the instance by running the 'hostname' command
+func findHostName(instance *instances.InstanceInfo, instanceSigner ssh.Signer) (string, error) {
+	win, err := windows.New("", "", instance, instanceSigner)
+	if err != nil {
+		return "", errors.Wrap(err, "error instantiating Windows instance")
+	}
+	// get the instance host name  by running hostname command on remote VM
+	hostName, err := win.Run("hostname", true)
```
aravindhp, line 308: Is `hostname` a powershell command?

aravindhp (changes requested):

pkg/csr/csr.go
mansikulkarni96: tried to answer this earlier in one of the [discussions](https://github.com/openshift/windows-machine-config-operator/pull/499#discussion_r673292797)

pkg/csr/csr.go
```diff
@@ -0,0 +1,355 @@
...
+		return false, errors.Wrap(err, "unable to map node name to the host names of Windows instances")
+	}
+	if !hasEntry {
+		// CSR is not from a BYOH instance
+		return false, nil
+	}
+	// validate node name for DNS RFC1123 naming conventions
+	// ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names
+	if errs := validation.IsDNS1123Subdomain(nodeName); len(errs) > 0 {
+		a.recorder.Eventf(a.csr, core.EventTypeWarning, "nodeNameValidationFailed",
```
aravindhp, line 196: Sorry should have caught this earlier. `NodeNameValidationFailed` to be consistent with the existing event naming pattern.

aravindhp (approved):
Nice work, @mansikulkarni96 

/approve

pkg/csr/csr.go
mansikulkarni96: done

saifshaikh48 (commented):
/lgtm

aravindhp:
/hold

For lgtm from @sebsoto as he had requested some changes

PR #503, opened 2021-07-01
[wmco] Rename IP() to Address()

sebsoto:
This commit renames the IP() method of the Windows interface to
Address(). This should be done as either an ip or dns address can be
used, and calling it IP is a misnomer.


jrvaldes (approved):
/lgtm

aravindhp (approved):

/approve

PR #508, opened 2021-07-08
WINC-632: Secure BYOH username annotation

saifshaikh48:
This PR introduces symmetric encryption and decryption into the usage of the instance username annotation to secure potentially sensitive user info. When a user changes the private key used to SSH into the instances, each BYOH node's annotation is updated to be encrypted with the new key.


saifshaikh48:
/hold

Holding until CI (vsphere) is fixed.

saifshaikh48:
This PR is blocked by https://github.com/openshift/windows-machine-config-operator/pull/518, which fixes an error when applying node annotations.

aravindhp:
Why is this PR blocked by https://github.com/openshift/windows-machine-config-operator/pull/518? Doesn't the Node object get updated on the second try with `Update()`? At least that is what I have been observing.

saifshaikh48:
> Why is this PR blocked by #518? Doesn't the Node object get updated on the second try with `Update()`? At least that is what I have been observing.

You're right, badly worded on my end -- #518 should ideally merge first.

controllers/secret_controller.go
```diff
@@ -164,33 +165,59 @@ func (r *SecretReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
-			// Only clear the annotation for Nodes associated with Machines, as the clearing of the annotation is used
-			// solely to kick off the deletion and recreation of Machines.
 			if _, present := node.Annotations[BYOHAnnotation]; present {
-				continue
-			}
+				// For BYOH nodes, patch the username annotation, encrpyting with the new private key
+				username, err := r.usernameFromNode(ctx, &node)
+				if err != nil {
+					return reconcile.Result{}, err
+				}
```
saifshaikh48, line 180: Reviewing the logic here -- could I get another pair of eyes:

If `usernameFromNode` finds no associated instance then that means its entry has been removed from the ConfigMap, and the instance should be deconfigured (or is currently being deconfigured, since the configmap controller can be running in parallel). Therefore, this isn't an error scenario and should `return reconcile.Result{}, nil`.

alinaryan: I think we still want an err check so maybe you could instead add in the empty instance check that returns `reconcile.Result{}, nil`

test/e2e/create_test.go
```diff
@@ -376,6 +400,40 @@ func (tc *testContext) waitForWindowsNodes(nodeCount int32, expectError, checkVe
...
+		nodeAddressStr, err := controllers.GetAddress(node.Status.Addresses)
+		if err != nil {
+			return "", errors.Errorf("unable to find address for node %s", node.Name)
+		}
+		if instance.Address == nodeAddressStr {
+			return instance.Username, nil
+		}
+	}
+	return "", errors.Errorf("unable to find Windows instance associated with node %s", node.Name)
+}
```
sebsoto, line 435: I dont think this function is necessary at all. the test suites knows what the username is as it creates it.
We supply the expected value so we know what it should be, the username should be read directly from the configmap, instead of calling `ParseInstances`. You already have the node so you know the ConfigMap entry key

test/e2e/secrets_test.go
```diff
@@ -128,6 +129,27 @@ func testUserDataTamper(t *testing.T) {
...
+		err = tc.createPrivateKeySecret(false)
+		require.NoError(t, err, "error replacing known private key secret with a random key")
+		// Ensure public key hash and encrypted username annotations are updated correctly
+		err = tc.waitForWindowsNodes(gc.numberOfBYOHNodes, false, false, true)
+		assert.NoError(t, err, "error waiting for Windows nodes updated with newly created private key")
+	})
+
+	// Re-create the known private key so SSH connection can be re-established
+	require.NoError(t, tc.createPrivateKeySecret(true), "error confirming known private key secret exists")
+}
```
sebsoto, line 151: Shouldn't this be done at the same time the privatekey change is being validated for Machines, using the same new private key?

This will require the reordering of some tests

pkg/instances/instances.go
```diff
@@ -57,3 +61,64 @@ func (i *InstanceInfo) UpgradeRequired() bool {
 	// fully deconfigured before being configured by the current version.
 	return true
 }
+
+// ParseInstances gets the list of instances specified in the ConfigMap's data. This function should be passed a list
+// of Nodes in the cluster, as each instance returned will contain a reference to its associated Node, if it has one
+// in the given NodeList. If an instance does not have an associated node from the NodeList, the node reference will
+// be nil.
+func ParseInstances(configMapData map[string]string, nodes *core.NodeList) ([]*InstanceInfo, error) {
```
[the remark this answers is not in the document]
sebsoto: What about moving it to controllers.go?

sebsoto (changes requested):

pkg/secrets/secrets.go
```diff
@@ -80,3 +85,60 @@ func GenerateUserData(publicKey ssh.PublicKey) (*core.Secret, error) {
...
 }
+
+// Encrypt attempts to encrpyt a plaintext message using the given key as an encoded passphrase
+func Encrypt(plainText string, key []byte) (string, error) {
+	if key == nil {
+		return "", errors.New("encryption passphrase cannot be nil")
+	}
+	// prepare PGP block to hold encrpyted data
+	msgBuffer := bytes.NewBuffer(nil)
+	encoder, _ := armor.Encode(msgBuffer, "ENCRYPTED DATA", nil)
```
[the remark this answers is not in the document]
saifshaikh48: It's really just a tag that the wraps encrypted data contains
```
-----BEGIN ENCRYPTED DATA-----
base64 encoded data
-----END ENCRYPTED DATA----
```
Since it's not necessary, I'll just remove it and pass in the empty string.

test/e2e/create_test.go
saifshaikh48: You're right, thanks for pointing this out.

pkg/secrets/secrets.go
```diff
@@ -80,3 +85,60 @@ func GenerateUserData(publicKey ssh.PublicKey) (*core.Secret, error) {
...
+
+	writer, _ := openpgp.SymmetricallyEncrypt(encoder, key, nil, nil)
+	_, err := writer.Write([]byte(plainText))
+	if err != nil {
+		return "", err
+	}
+	writer.Close()
+	encoder.Close()
```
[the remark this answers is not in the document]
saifshaikh48: The closing has to be done before reading the encrypted bytes from the buffer

pkg/instances/instances.go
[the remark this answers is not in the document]
saifshaikh48: The more I look at it, the less convinced I am that it needs to move at all. I'll keep it in configmap_controller.go as a helper.

test/e2e/secrets_test.go
saifshaikh48: I am not clear on the advantage of doing so (bit quicker/less API call I suppose). The reasoning behind the current separation is that, when priv key changes, BYOH nodes are just patched whereas Machine nodes are deleted and recreated - different test suites in my mind, reconfiguration vs creation.

sebsoto: The expected result is different yes, but the advantage is that the setup is the same, so overall testing time is saved, as well as less duplication occuring in the codebase so the tests are easier to understand. You can structure the test like so:
```
func testPrivateKeyChange(t *testing.T) {
  changeKey()
  t.Run("testing expected BYOH results", testFunction)
  t.Run("testing expected Machine results", testFunction)
}
```

pkg/secrets/secrets.go
```diff
@@ -80,3 +85,69 @@ func GenerateUserData(publicKey ssh.PublicKey) (*core.Secret, error) {
 
 	return userDataSecret, nil
 }
+
+// Encrypt creates an ecnrypted block of text from a plaintext message using the given key
+func Encrypt(plainText string, key []byte) (string, error) {
+	if key == nil {
```
mansikulkarni96, line 91: is nil an adequate check for empty bytes?

pkg/secrets/secrets.go
```diff
@@ -80,3 +85,69 @@ func GenerateUserData(publicKey ssh.PublicKey) (*core.Secret, error) {
...
+		return "", errors.New("encryption passphrase cannot be nil")
+	}
+
+	// prepare PGP block with a wrapper tag around the encrypted data
+	tag := "ENCRYPTED DATA"
+	msgBuffer := bytes.NewBuffer(nil)
+	encoder, _ := armor.Encode(msgBuffer, tag, nil)
+
+	writer, _ := openpgp.SymmetricallyEncrypt(encoder, key, nil, nil)
```
mansikulkarni96, line 100: can there be an error returned here?

pkg/secrets/secrets.go
```diff
@@ -80,3 +85,69 @@ func GenerateUserData(publicKey ssh.PublicKey) (*core.Secret, error) {
...
+	_, err := writer.Write([]byte(plainText))
+	if err != nil {
+		return "", err
+	}
+	writer.Close()
+	encoder.Close()
+
+	cipherText := string(msgBuffer.Bytes())
```
mansikulkarni96, line 108: you can just return this, if the variable is not doing much here.

pkg/secrets/secrets.go
```diff
@@ -80,3 +85,69 @@ func GenerateUserData(publicKey ssh.PublicKey) (*core.Secret, error) {
...
+	// Unwrap encoded block holding the message content
+	msgBuffer := bytes.NewBuffer([]byte(cipherText))
+	armorBlock, err := armor.Decode(msgBuffer)
+	if err != nil {
+		return "", errors.Wrapf(err, "unable to process given cipher %s", cipherText)
+	}
+
+	msgBody, err := readMessage(armorBlock.Body, key)
+	if err != nil {
+		return "", err
```
mansikulkarni96, line 127: can add a wrapper here too.

pkg/secrets/secrets.go
```diff
@@ -80,3 +85,69 @@ func GenerateUserData(publicKey ssh.PublicKey) (*core.Secret, error) {
...
+			return nil, errors.New("invalid passphrase supplied")
+		}
+		attempted = true
+		return key, nil
+	}
+	message, err := openpgp.ReadMessage(reader, nil, promptFunction, nil)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to decrypt message using given key")
+	}
+	return message.UnverifiedBody, nil
```
mansikulkarni96, line 152: can we be sure message.UnverifiedBody will always have a value?

mansikulkarni96 (commented):
PTAL at my comments.

pkg/secrets/secrets.go
saifshaikh48: Empty bytes should be fine, covered in [unit tests](https://github.com/openshift/windows-machine-config-operator/pull/508/commits/5004d8efe302d755d5f0812a8472b6d20771af8a#diff-0be37f7149d650a2c98fc7c079b136182f607b5e2f98587dce5f3f00657d38b7R86-R92)

pkg/secrets/secrets.go
saifshaikh48: Yes, as long as there is no error from the `ReadMessage` call. Meaning encrypted message is in proper format and the provided key is the same one used to encrypt.

pkg/secrets/secrets.go
```diff
@@ -80,3 +85,69 @@ func GenerateUserData(publicKey ssh.PublicKey) (*core.Secret, error) {
 
 	return userDataSecret, nil
 }
+
+// Encrypt creates an ecnrypted block of text from a plaintext message using the given key
+func Encrypt(plainText string, key []byte) (string, error) {
```
aravindhp, line 90: Why is `Encrypt()`part of the `secrets` package. AFAICT this has nothing to do with the private key secret. Just because you are using the same private key to encrypt and decrypt doesn't mean it belongs to this package.

pkg/secrets/secrets.go
aravindhp: Yes an error can be returned and it needs to be handled

pkg/secrets/secrets.go
```diff
@@ -80,3 +85,69 @@ func GenerateUserData(publicKey ssh.PublicKey) (*core.Secret, error) {
...
+	if key == nil {
+		return "", errors.New("encryption passphrase cannot be nil")
+	}
+
+	// prepare PGP block with a wrapper tag around the encrypted data
+	tag := "ENCRYPTED DATA"
+	msgBuffer := bytes.NewBuffer(nil)
+	encoder, _ := armor.Encode(msgBuffer, tag, nil)
```
aravindhp, line 98: Please handle the error that is returned.

pkg/secrets/secrets.go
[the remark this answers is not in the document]
aravindhp: - Please add a comment saying this
- If an error is returned by `writer.Write()`, you will be leaking `encoder` and that needs to be handled

pkg/secrets/secrets.go
aravindhp: Only if you are going to add more context and information to the error.

controllers/configmap_controller.go
```diff
@@ -239,19 +252,6 @@ func (r *ConfigMapReconciler) deconfigureInstances(instances []*instances.Instan
...
-	for _, node := range nodes.Items {
-		for _, nodeAddress := range node.Status.Addresses {
-			if address == nodeAddress.Address {
-				return &node, true
-			}
-		}
-	}
-	return nil, false
-}
-
```
aravindhp, line 241: Why are you moving this function?

controllers/configmap_controller_test.go
```diff
@@ -146,67 +209,3 @@ func TestParseInstances(t *testing.T) {
 		})
 	}
 }
-
-func TestIsValidConfigMap(t *testing.T) {
```
aravindhp, line 150: Again, why are you moving this function around? Doing is forcing the reviewer to re-review this function. If you want to mechanically move functions around, open separate PRs.

controllers/secret_controller.go
```diff
@@ -162,36 +163,58 @@ func (r *SecretReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+			annotationsToApply := make(map[string]string)
 			if _, present := node.Annotations[BYOHAnnotation]; present {
-				continue
-			}
+				// For BYOH nodes, update the username annotation and public key hash annotation using new private key
+				instanceUsername, err := r.usernameFromNode(ctx, &node, nodes)
+				if err != nil {
+					return reconcile.Result{}, err
+				}
+				usernameCipher, err := secrets.Encrypt(instanceUsername, privateKeyBytes)
```
aravindhp, line 188: `usernameCipher` is misleading. This is the encrypted username and not a cipher function given the definition of a cipher. Comment applies to all places you are using the cipher in a variable name.

controllers/secret_controller.go
```diff
@@ -162,36 +163,58 @@ func (r *SecretReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+				if err != nil {
+					return reconcile.Result{}, err
+				}
+				// Obfuscate newline characters to make encrypted data compatible as JSON Patch request body
+				patchCombatibleCipher := strings.Replace(usernameCipher, "\n", "<newline>", -1)
```
aravindhp, line 193: You have repeated code here and converting these lines into a helper method will fix that like I called out before.

controllers/secret_controller.go
```diff
@@ -162,36 +163,58 @@ func (r *SecretReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
-				return reconcile.Result{}, errors.Wrapf(err, "error clearing public key annotation on node %s",
-					node.GetName())
+				annotationsToApply = map[string]string{
+					UsernameAnnotation:              patchCombatibleCipher,
+					nodeconfig.PubKeyHashAnnotation: expectedPubKeyAnno,
+				}
+				if err := r.applyAnnotations(ctx, node, annotationsToApply); err != nil {
+					return reconcile.Result{},
+						errors.Wrapf(err, "error updating username & public key annotations on node %s", node.GetName())
+				}
```
aravindhp, line 202: Can't this be moved out of the `if else` given it will always be called now.

controllers/secret_controller.go
```diff
@@ -162,36 +163,58 @@ func (r *SecretReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+			} else {
+				// For Nodes associated with Machines, clear the public key annotation, as the clearing of the
+				// annotation is used solely to kick off the deletion and recreation of Machines.
+				existingPubKeyAnno := node.Annotations[nodeconfig.PubKeyHashAnnotation]
+				if existingPubKeyAnno == expectedPubKeyAnno {
```
aravindhp, line 207: Can't you do a similar check for the encrypted username?

controllers/secret_controller.go
```diff
@@ -206,6 +229,49 @@ func (r *SecretReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 	}
 }
 
+// usernameFromNode identifies the username associated with a given node
+func (r *SecretReconciler) usernameFromNode(ctx context.Context, node *core.Node, nodes *core.NodeList) (string, error) {
```
aravindhp, line 233: I find this function construction to be a bit strange. You are passing a `node` and a list of nodes to get the username from the node. This is hinting that `ParseInstances` needs to be refactored better. Moreover why do you need to cross-reference the list of nodes with the `ConfigMap` here?

test/e2e/secrets_test.go
```diff
@@ -25,18 +25,18 @@ import (
 	"github.com/openshift/windows-machine-config-operator/test/e2e/clusterinfo"
 )
 
-// getExpectedPublicKey returns the public key associated with the private key within the cloud-private-key secret
-func (tc *testContext) getExpectedPublicKey() (ssh.PublicKey, error) {
+// getExpectedKeys returns the private key within the cloud-private-key secret and the associated public key
```
aravindhp, line 29: I suggest calling this `getExpectedKeyPair()`

test/e2e/secrets_test.go
aravindhp: +1

aravindhp (changes requested):

controllers/configmap_controller.go
```diff
@@ -179,14 +179,27 @@ func (r *ConfigMapReconciler) reconcileNodes(ctx context.Context, windowsInstanc
...
 	// configuration effort for configurable hosts will not be blocked by a specific host that has issues with
 	// configuration.
 	for _, instance := range instances {
-		err := r.ensureInstanceIsUpToDate(instance, map[string]string{BYOHAnnotation: "true",
-			UsernameAnnotation: instance.Username})
+		usernameCipher, err := secrets.Encrypt(instance.Username, privateKeyBytes)
+		if err != nil {
+			return errors.Wrapf(err, "unable to encrypt username for instance %s", instance.Address)
+		}
+		// Obfuscate newline characters to make encrypted data compatible as JSON Patch request body
```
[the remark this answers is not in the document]
saifshaikh48: Removing the `\n`s would make it very difficult to decrypt, as we need the original encrypted message. We wouldn't know where to put back the line breaks.

controllers/secret_controller.go
saifshaikh48: I see, sure I'll make the changes where applicable.

controllers/configmap_controller.go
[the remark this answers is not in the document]
aravindhp: Please explain this better in your comment. You are not really obfuscating newlines but adding markers for them to help in decryption.

pkg/secrets/secrets.go
saifshaikh48: I do see your point. Addressed this by moving the encrypt and decrypt functions to their own `crypto` package

test/e2e/secrets_test.go
saifshaikh48: I tried my hand at this, please let me know if my implementation captures the ideas discussed here.

controllers/secret_controller.go
saifshaikh48: The node list being passed in 1. to then pass along to `parseInstances` and 2. just to reduce the amount of API calls (since we already have the node list at the site of this function call). I agree this is a bit strange for now. Improvement can be made based on the `windows-instances parser pkg that will be introduced in #499, but I am not clear how I could better this in this PR currently.

aravindhp: >Moreover why do you need to cross-reference the list of nodes with the ConfigMap here?

You still have not answered this. Moreover having `usernameFromNode` as part of the `SecretReconciler` is also strange. I suggest pairing with @sebsoto to find a better solution for this.

saifshaikh48: @sebsoto and I refactored the logic here. We moved the username retrieval to a helper `getNodeUsername` no longer associated with the secret controller.

controllers/secret_controller.go
saifshaikh48: Revisiting this, I don't think that this check on the username is necessary -- checking the public key annotation is sufficient for both. If the public key anno changed, that means the private key secret has changed, so both `PubKeyHashAnnotation` and `UsernameAnnotation` must be updated.

aravindhp: What if the user changed the private key secret but didn't actually update the key? Like added a label or annotation.

saifshaikh48: Both annotations' values are made using the bytes present in private key secret data. So no change necessary if a label/annotation is added

controllers/controllers.go
```diff
@@ -75,14 +79,45 @@ func (r *instanceReconciler) ensureInstanceIsUpToDate(instance *instances.Instan
...
+	if usernameAnnotation == "" {
 		return nil, errors.New("node is missing valid username annotation")
 	}
 	addr, err := GetAddress(node.Status.Addresses)
 	if err != nil {
 		return nil, err
 	}
-	return instances.NewInstanceInfo(addr, node.Annotations[UsernameAnnotation], "", node), nil
+
+	// Decypt username annotation to plain text using private key
```
alinaryan, line 91: nit: spelling err in decrypt

pkg/crypto/crypto.go
```diff
@@ -0,0 +1,82 @@
...
+	"bytes"
+	"io"
+	"io/ioutil"
+
+	"github.com/pkg/errors"
+	"golang.org/x/crypto/openpgp"
+	"golang.org/x/crypto/openpgp/armor"
+)
+
+// Encrypt creates an ecnrypted block of text from a plaintext message using the given key
```
alinaryan, line 13: nit: spelling err encrypted

controllers/secret_controller.go
```diff
@@ -206,6 +217,39 @@ func (r *SecretReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+		return "", err
+	}
+	encryptedUsername, err := encryptUsername(instanceUsername, key)
+	if err != nil {
+		return "", errors.Wrapf(err, "error encrypting node %s username", node.GetName())
+	}
+	return encryptedUsername, nil
+}
+
+// applyAnnotations applies all the given annotations to the given Node resource
```
aravindhp, line 239: Same comment here.

controllers/controllers.go
```diff
@@ -75,14 +79,45 @@ func (r *instanceReconciler) ensureInstanceIsUpToDate(instance *instances.Instan
...
+	}
+	username, err := decryptUsername(usernameAnnotation, privateKeyBytes)
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to decrypt username annotation for node %s", node.Name)
+	}
+
+	return instances.NewInstanceInfo(addr, username, "", node), nil
+}
+
+// encryptUsername returns the encrypted value to be used as a username annotation
```
aravindhp, line 105: Should this be moved to the crypto package?

controllers/controllers.go
```diff
@@ -75,14 +79,45 @@ func (r *instanceReconciler) ensureInstanceIsUpToDate(instance *instances.Instan
...
+	if err != nil {
+		return "", err
+	}
+	// Make encrypted data compatible as JSON Patch request body.
+	// Newline characters are marked with text placeholder in order to be easily reversed during decryption
+	return strings.Replace(encryptedUsername, "\n", "<newline>", -1), nil
+}
+
+// decryptUsername returns the plaintext value of an encrypted username annotation
+func decryptUsername(encrypted string, key []byte) (string, error) {
```
aravindhp, line 117: Should this be moved to the crypto package?

aravindhp (changes requested):
Looks much better now, @saifshaikh48

controllers/secret_controller.go
```diff
@@ -163,35 +163,46 @@ func (r *SecretReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+			Name: secrets.PrivateKeySecret}, r.client)
 		if err != nil {
-			return reconcile.Result{}, errors.Wrapf(err, "error creating public key annotation add request")
+			return reconcile.Result{}, err
 		}
 		for _, node := range nodes.Items {
-			// Only clear the annotation for Nodes associated with Machines, as the clearing of the annotation is used
-			// solely to kick off the deletion and recreation of Machines.
-			if _, present := node.Annotations[BYOHAnnotation]; present {
+			if node.Annotations[nodeconfig.PubKeyHashAnnotation] == expectedPubKeyAnno {
```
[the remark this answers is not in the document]
saifshaikh48: Yep, these 2 annotations are [applied at the same time](https://github.com/openshift/windows-machine-config-operator/blob/2ceed3d84b140aac958071c34e812a39a2efbde1/pkg/nodeconfig/nodeconfig.go#L177-L178) and updated together if needed. Since both are dependent on the private key secret, I concluded the username check was therefore redundant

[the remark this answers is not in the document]
aravindhp: I suggest adding a comment stating that here.

controllers/secret_controller.go
```diff
@@ -163,35 +163,46 @@ func (r *SecretReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
 		// userdata secret data does not match what is expected
-		// Mark nodes configured with the previous private key for deletion
 		nodes := &core.NodeList{}
 		err = r.client.List(ctx, nodes, client.MatchingLabels{core.LabelOSStable: "windows"})
 		if err != nil {
 			return reconcile.Result{}, errors.Wrapf(err, "error getting node list")
 		}
+
+		// Modify annotations on nodes configured with the previous private key, if it has changed
 		expectedPubKeyAnno := nodeconfig.CreatePubKeyHashAnnotation(keySigner.PublicKey())
```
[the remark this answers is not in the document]
saifshaikh48: Noted. I will refrain from renaming this in this PR, though.

controllers/secret_controller.go
saifshaikh48: I can see why this could be moved to the `metadata` package. I'll make this change.

controllers/secret_controller.go
```diff
@@ -206,6 +217,39 @@ func (r *SecretReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 	}
 }
 
+// getUsernameAnno retrieves the username associated with a given node and ecrypts it using the given key
+func (r *SecretReconciler) getUsernameAnno(ctx context.Context, node core.Node, key []byte) (string, error) {
```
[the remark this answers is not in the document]
saifshaikh48: As far as the function name, I'll change it to `getEncryptedUsername` for clarity. However, I think it should stay as part of the reconciler since it needs information about the configmap's namespace, and access to the configmap parsing which is in configmap_controller.

test/e2e/secrets_test.go
```diff
@@ -12,6 +12,7 @@ import (
 	"testing"
 	"time"
 
+	config "github.com/openshift/api/config/v1"
```
jrvaldes, line 15: s/config/oconfig

test/e2e/secrets_test.go
```diff
@@ -128,6 +129,31 @@ func testUserDataTamper(t *testing.T) {
...
 
+// testPrivateKeyChange alters the private key used to SSH into instances and ensures nodes are updated properly
+func testPrivateKeyChange(t *testing.T) {
+	tc, err := NewTestContext()
+	require.NoError(t, err)
+
+	// This test cannot be run on vSphere because this random key is not part of the vSphere template image.
+	// Moreover this test is platform agnostic so is not needed to be run for every supported platform.
+	if tc.CloudProvider.GetType() == config.VSpherePlatformType {
+		t.Skipf("Skipping for %s", tc.CloudProvider.GetType())
```
jrvaldes, line 140: No need for `tc.CloudProvider.GetType()`, since it will always be `config.VSpherePlatformType`

test/e2e/secrets_test.go
```diff
@@ -128,6 +129,31 @@ func testUserDataTamper(t *testing.T) {
 	}
 }
 
+// testPrivateKeyChange alters the private key used to SSH into instances and ensures nodes are updated properly
+func testPrivateKeyChange(t *testing.T) {
```
jrvaldes, line 133: s/testPrivateKeyChange/privateKeyChangeTest

test/e2e/secrets_test.go
```diff
@@ -68,7 +68,7 @@ func testUserData(t *testing.T) {
 	tc, err := NewTestContext()
 	require.NoError(t, err)
 
-	pubKey, err := tc.getExpectedPublicKey()
+	_, pubKey, err := tc.getExpectedKeyPair()
 	require.NoError(t, err, "error determining expected public key")
```
jrvaldes, line 73: the private key processing can fail too, leading to a  "error determining expected private key"
Consider:
"[error getting the expected public/private key pair](https://github.com/openshift/windows-machine-config-operator/pull/508/commits/6f17a39559bf0cfc1bcd5c6eb2d98d5960103d99#diff-f08d14d1ac64442e0171050a9b9b70b50cfc7e1f0bc0eab1badc1e4a827252e3R273)"

test/e2e/validation_test.go
```diff
@@ -35,7 +35,7 @@ func testNodeMetadata(t *testing.T) {
 	clusterKubeletVersion, err := tc.getClusterKubeVersion()
 	require.NoError(t, err, "could not get cluster kube version")
 
-	pubKey, err := tc.getExpectedPublicKey()
+	_, pubKey, err := tc.getExpectedKeyPair()
 	require.NoError(t, err, "error getting the expected public key")
```
jrvaldes, line 39: ditto

controllers/configmap_controller.go
```diff
@@ -168,6 +168,29 @@ func validateAddress(address string) error {
...
+		}
+	}
+	return "", errors.Errorf("unable to find instance associated with node %s", node.GetName())
+}
+
+// extractUsernameFromEntry returns the username string from data in the form username=<username>
+func extractUsernameFromEntry(value string) (string, error) {
+	splitData := strings.SplitN(value, "=", 2)
+	if len(splitData) == 0 || splitData[0] != "username" {
+		return "", errors.New("data has an incorrect format")
```
jrvaldes, line 189: consider logging the incorrectly formatted data (value)

controllers/configmap_controller.go
```diff
@@ -168,6 +168,29 @@ func validateAddress(address string) error {
...
+	for _, address := range node.Status.Addresses {
+		if data, found := configMapData[address.Address]; found {
+			return extractUsernameFromEntry(data)
+		}
+	}
+	return "", errors.Errorf("unable to find instance associated with node %s", node.GetName())
+}
+
+// extractUsernameFromEntry returns the username string from data in the form username=<username>
+func extractUsernameFromEntry(value string) (string, error) {
```
jrvaldes, line 186: s/value/data

controllers/configmap_controller.go
```diff
@@ -168,6 +168,29 @@ func validateAddress(address string) error {
...
+	// Find entry in ConfigMap that is associated to node via address
+	for _, address := range node.Status.Addresses {
+		if data, found := configMapData[address.Address]; found {
+			return extractUsernameFromEntry(data)
+		}
+	}
+	return "", errors.Errorf("unable to find instance associated with node %s", node.GetName())
+}
+
+// extractUsernameFromEntry returns the username string from data in the form username=<username>
```
jrvaldes, line 185: What's an `Entry`? I feel `extractUsername` as function name is good.

controllers/configmap_controller.go
jrvaldes: Would be nice to unit test `func extractUsernameFromEntry(value string) (string, error) {}`.

jrvaldes (changes requested):
Thanks for working on this @saifshaikh48 , PTAL to my comments

controllers/configmap_controller.go
saifshaikh48: `value` refers to the second part of the key/value pair that makes up each ConfigMap entry. From our README: 
> Each entry in the data section of the ConfigMap should be formatted with the address as the key, and a value with the format of username=<username>

IMO `extractUsernameFromEntry` does not need separate tests as there are already unit tests for `getNodeUsername`. It is just a helper for `getNodeUsername` and moreover is not used in unit tests (or elsewhere).

I will leave this is as.

test/e2e/secrets_test.go
saifshaikh48: `testFooBar` seems to be standard naming within our e2e suite

controllers/configmap_controller.go
```diff
@@ -174,19 +124,30 @@ func (r *ConfigMapReconciler) reconcileNodes(ctx context.Context, windowsInstanc
 	}
 
 	// Get the list of instances that are expected to be Nodes
-	instances, err := r.parseInstances(windowsInstances.Data, nodes)
+	instances, err := instances.ParseInstances(windowsInstances.Data, nodes)
 	if err != nil {
 		return errors.Wrap(err, "unable to parse hosts from ConfigMap")
```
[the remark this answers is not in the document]
saifshaikh48: IMO, this does not seem within scope of this PR.

controllers/configmap_controller.go
alinaryan: +1

test/e2e/secrets_test.go
jrvaldes: I'm following [reconfigurationTest](https://github.com/openshift/windows-machine-config-operator/blob/1ed33ba363613f96840f327419a4739e89abd4a0/test/e2e/reconfigure_test.go#L16)

controllers/configmap_controller.go
jrvaldes: That's a valid concern; But, whoever has access to the logs, is authenticated already

controllers/controllers.go
```diff
@@ -75,14 +78,27 @@ func (r *instanceReconciler) ensureInstanceIsUpToDate(instance *instances.Instan
 // instanceFromNode returns an instance object for the given node. Requires a username that can be used to SSH into the
 // instance to be annotated on the node.
 func (r *instanceReconciler) instanceFromNode(node *core.Node) (*instances.InstanceInfo, error) {
-	if node.Annotations[UsernameAnnotation] == "" {
+	usernameAnnotation := node.Annotations[UsernameAnnotation]
```
alinaryan, line 81: should we avoid "api call stutters" like Aravindh mentioned yesterday? Maybe you could keep this line how it was it was previously

saifshaikh48: Don't think that is applicable here. Also `usernameAnnotation` is used in multiple places so I think it warrants a variable.

controllers/configmap_controller.go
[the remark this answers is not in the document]
aravindhp: Given you are in the area, it is fine to make minor changes like this.

test/e2e/secrets_test.go
aravindhp: Why? This should be done only if we have a conflict with using `config`. Places where we use `oconfig` is because we have a conflict with `config`. It is not a hard and fast rule.

test/e2e/secrets_test.go
aravindhp: Actually `reconfigurationTest` is not following the Go standard naming.

controllers/configmap_controller.go
aravindhp: Please don't log the username to be on the safe side

controllers/configmap_controller.go
```diff
@@ -178,7 +201,7 @@ func (r *ConfigMapReconciler) reconcileNodes(ctx context.Context, windowsInstanc
 	// Get the list of instances that are expected to be Nodes
 	instances, err := r.parseInstances(windowsInstances.Data, nodes)
 	if err != nil {
-		return errors.Wrap(err, "unable to parse hosts from ConfigMap")
+		return errors.Wrap(err, "unable to parse instances from ConfigMap")
```
aravindhp, line 204: This change should go into `[wmco] Refactor BYOH instance retrieval helper` because changing it there is warranted but not in this commit.

pkg/crypto/crypto.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+// EncryptUsername returns the encrypted value to be used as a username annotation
+func EncryptUsername(username string, key []byte) (string, error) {
+	encryptedUsername, err := encrypt(username, key)
+	if err != nil {
+		return "", err
+	}
+	// Make encrypted data compatible as JSON Patch request body.
+	// Newline characters are marked with text placeholder in order to be easily reversed during decryption
```
aravindhp, line 21: Does encryption add newline characters? Sorry I am still puzzled by what introduces the newline character. Are you getting it from the ConfigMap data?

pkg/crypto/crypto.go
```diff
@@ -0,0 +1,104 @@
...
+	"io/ioutil"
+	"strings"
+
+	"github.com/pkg/errors"
+	"golang.org/x/crypto/openpgp"
+	"golang.org/x/crypto/openpgp/armor"
+)
+
+// EncryptUsername returns the encrypted value to be used as a username annotation
+func EncryptUsername(username string, key []byte) (string, error) {
```
aravindhp, line 15: So re-reading and seeing it here, this looks like a generic function that can encrypt any string. Not sure if this should be called `EncryptUsername`. So maybe `Encrypt()` at which point the only thing this function does in addition to `encrypt()` is the funny business with the newline.

test/e2e/secrets_test.go
```diff
@@ -12,6 +12,7 @@ import (
 	"testing"
 	"time"
 
+	oconfig "github.com/openshift/api/config/v1"
```
aravindhp, line 15: Like I mentioned in the [comment](https://github.com/openshift/windows-machine-config-operator/pull/508#discussion_r680095803), `oconfig` is not needed.

test/e2e/secrets_test.go
aravindhp: Actually it is a test suite and hence named `reconfigurationTest` which @saifshaikh48 has now correctly named `reconfigurationTestSuite`.

aravindhp (changes requested):

pkg/crypto/crypto.go
saifshaikh48: The encryption itself introduces the newlines. The encrypted data is in the form of an [armor block](https://pkg.go.dev/golang.org/x/crypto/openpgp/armor#Block) with many `\n` characters

aravindhp: Please call that out in your comment as it is not immediately obvious. Is `<newline>` is visible in the annotation? Should it be made less obvious? `<wmcomarker>`?

pkg/crypto/crypto.go
aravindhp: This comment still holds. Basically `Encrypt()` function returns an encrypted value that can be used as a JSON string.

pkg/crypto/crypto.go
```diff
@@ -0,0 +1,104 @@
...
+	encryptedUsername, err := encrypt(username, key)
+	if err != nil {
+		return "", err
+	}
+	// Make encrypted data compatible as JSON Patch request body.
+	// Newline characters are marked with text placeholder in order to be easily reversed during decryption
+	return strings.Replace(encryptedUsername, "\n", "<newline>", -1), nil
```
aravindhp, line 22: Now that I understand what's going on, we should make `<newline>` a constant then.

pkg/crypto/crypto.go
saifshaikh48: This is true. Should I go ahead with a renaming of `EncryptUsername` & `DecryptUsername` to `EncryptAsJSONString`  & `DecrpytJSONString`?

aravindhp: That would be `EncryptToJSONString` and `DecryptToJSONString`

saifshaikh48: I will assume you mean `DecryptFromJSONString` :slightly_smiling_face:

controllers/configmap_controller.go
aravindhp: We are logging the username when initializing the connection to the instance. Not sure if we should be removing that too now.

saifshaikh48: I think it is a good idea to remove, given the work we are doing to protect the plaintext value. I'll add a separate commit removing from that log.

aravindhp: I can take care of it in my logging PR. Please remind me during the PR review.

aravindhp: On second thought that belongs in this PR.

test/e2e/wmco_test.go
```diff
@@ -49,7 +49,11 @@ func TestWMCO(t *testing.T) {
 	t.Run("create", creationTestSuite)
 	t.Run("network", testNetwork)
 	t.Run("upgrade", upgradeTestSuite)
-	t.Run("reconfigure", reconfigurationTest)
+	// The reconfigurationTestSuite must be run directly before the deletionTestSuite. This is because we do not
+	// currently wait for nodes to fully reconcile after changing the private key back to the valid key. Any tests
+	// added/moved in between these two suites may fail.
+	// This limitation will be removed with https://issues.redhat.com/browse/WINC-655
```
aravindhp, line 55: This also needs to be mentioned in `./hack/run-ci-e2e-test.sh` which is what actually kicks  off the test.

saifshaikh48: Sure, I will update.

aravindhp (approved):
Nice work @saifshaikh48 

/approve
/hold

Adding a hold that @sebsoto can remove given I did not review the e2e tests very closely.

test/e2e/secrets_test.go
```diff
@@ -128,6 +129,31 @@ func testUserDataTamper(t *testing.T) {
...
+	// Ensure Machines nodes are re-created and configured using the new private key
+	err = tc.waitForWindowsNodes(gc.numberOfMachineNodes, false, false, false)
+	assert.NoError(t, err, "error waiting for Windows nodes configured with newly created private key")
+	// Ensure public key hash and encrypted username annotations are updated correctly on BYOH nodes
+	err = tc.waitForWindowsNodes(gc.numberOfBYOHNodes, false, false, true)
+	assert.NoError(t, err, "error waiting for Windows nodes updated with newly created private key")
+
+	// Re-create the known private key so SSH connection can be re-established
+	// TODO: Remove dependency on this secret by rotating keys as part of https://issues.redhat.com/browse/WINC-655
+	require.NoError(t, tc.createPrivateKeySecret(true), "error confirming known private key secret exists")
```
sebsoto, line 154: I'm still on the fence about this but it seems to be working...
This operation is going to bleed into the deletion suite as the WMCO will be handling this first and could taint the results that we expect when doing certain deletion actions. But what we are testing for in the BYOH reconfiguration, specifically the removal of binaries and services, shouldnt be triggered by this.

I'm fine with this going in but we need to handle the TODO story ASAP so that we dont open ourselves up to missing certain edge cases

PR #513, opened 2021-07-12
WINC-580: Upgrade BYOH Nodes configured by previous WMCO versions

sebsoto:
Deconfigures and then configures any instances associated with a node
that has an mismatched WMCO version annotation.

### DIFF
--- a/controllers/configmap_controller_test.go
+++ b/controllers/configmap_controller_test.go
@@ -12,61 +12,131 @@ import (
 	"github.com/openshift/windows-machine-config-operator/pkg/instances"
 )
 
-func TestParseHosts(t *testing.T) {
+func TestParseInstances(t *testing.T) {
 	r := ConfigMapReconciler{}
 
 	testCases := []struct {
 		name        string
 		input       map[string]string
+		nodeList    *core.NodeList
 		expectedOut []*instances.InstanceInfo
 		expectedErr bool
 	}{
 		{
 			name:        "invalid username",
 			input:       map[string]string{"localhost": "notusername=core"},
+			nodeList:    &core.NodeList{},
 			expectedOut: nil,
 			expectedErr: true,
 		},
 		{
 			name:        "invalid DNS address",
 			input:       map[string]string{"notlocalhost": "username=core"},
+			nodeList:    &core.NodeList{},
 			expectedOut: nil,
 			expectedErr: true,
 		},
 		{
 			name:        "invalid username and DNS",
 			input:       map[string]string{"invalid": "invalid"},
+			nodeList:    &core.NodeList{},
 			expectedOut: nil,
 			expectedErr: true,
 		},
 		{
 			name:        "valid ipv6 address",
 			input:       map[string]string{"::1": "username=core"},
+			nodeList:    &core.NodeList{},
 			expectedOut: nil,
 			expectedErr: true,
 		},
 		{
 			name:        "valid dns address",
 			input:       map[string]string{"localhost": "username=core"},
+			nodeList:    &core.NodeList{},
 			expectedOut: []*instances.InstanceInfo{{Address: "localhost", Username: "core"}},
 			expectedErr: false,
 		},
 		{
 			name:        "valid ip address",
 			input:       map[string]string{"127.0.0.1": "username=core"},
+			nodeList:    &core.NodeList{},
 			expectedOut: []*instances.InstanceInfo{{Address: "127.0.0.1", Username: "core"}},
 			expectedErr: false,
 		},
 		{
-			name:        "valid dns and ip addresses",
+			name:        "valid dns and ip addresses with no nodes",
 			input:       map[string]string{"localhost": "username=core", "127.0.0.1": "username=Admin"},
+			nodeList:    &core.NodeList{},
 			expectedOut: []*instances.InstanceInfo{{Address: "localhost", Username: "core"}, {Address: "127.0.0.1", Username: "Admin"}},
+			expectedErr: false,
+		},
+		{
+			name:  "valid dns and ip addresses with unassociated nodes",
+			input: map[string]string{"localhost": "username=core", "127.0.0.1": "username=Admin"},
+			nodeList: &core.NodeList{
+				Items: []core.Node{
+					{
+						ObjectMeta: meta.ObjectMeta{
+							Name: "wrong-node",
+						},
+						Status: core.NodeStatus{
+							Addresses: []core.NodeAddress{
+								{Address: "111.1.1.1", Type: core.NodeInternalIP},
+							},
+						},
+					},
+				},
+			},
+			expectedOut: []*instances.InstanceInfo{
+				{Address: "127.0.0.1", Username: "Admin", Node: nil},
+				{Address: "localhost", Username: "core", Node: nil},
+			},
+			expectedErr: false,
+		},
+		{
+			name:  "valid dns and ip addresses with associated nodes",
+			input: map[string]string{"localhost": "username=core", "127.0.0.1": "username=Admin"},
+			nodeList: &core.NodeList{
+				Items: []core.Node{
+					{
+						ObjectMeta: meta.ObjectMeta{
+							Name: "ip-node",
+						},
+						Status: core.NodeStatus{
+							Addresses: []core.NodeAddress{
+								{Address: "127.0.0.1", Type: core.NodeInternalIP},
+							},
+						},
+					},
+					{
+						ObjectMeta: meta.ObjectMeta{
+							Name: "dns-node",
+						},
+						Status: core.NodeStatus{
+							Addresses: []core.NodeAddress{
+								{Address: "localhost", Type: core.NodeInternalDNS},
+							},
+						},
+					},
+				},
+			},
+			expectedOut: []*instances.InstanceInfo{
+				{Address: "127.0.0.1", Username: "Admin", Node: &core.Node{ObjectMeta: meta.ObjectMeta{Name: "ip-node"},
+					Status: core.NodeStatus{Addresses: []core.NodeAddress{{Address: "127.0.0.1",
+						Type: core.NodeInternalIP}},
+					}}},
+				{Address: "localhost", Username: "core", Node: &core.Node{ObjectMeta: meta.ObjectMeta{Name: "dns-node"},
+					Status: core.NodeStatus{Addresses: []core.NodeAddress{{Address: "localhost",
+						Type: core.NodeInternalDNS}},
+					}}},
+			},
 			expectedErr: false,
 		},
 	}
 	for _, test := range testCases {
 		t.Run(test.name, func(t *testing.T) {
-			out, err := r.parseHosts(test.input)
+			out, err := r.parseInstances(test.input, test.nodeList)
 			if test.expectedErr {
 				assert.Error(t, err)
 				return
--- a/controllers/controllers.go
+++ b/controllers/controllers.go
@@ -66,7 +66,7 @@ func (r *instanceReconciler) instanceFromNode(node *core.Node) (*instances.Insta
 	if err != nil {
 		return nil, err
 	}
-	return instances.NewInstanceInfo(addr, node.Annotations[UsernameAnnotation], ""), nil
+	return instances.NewInstanceInfo(addr, node.Annotations[UsernameAnnotation], "", node), nil
 }
 
 // GetAddress returns a non-ipv6 address that can be used to reach a Windows node. This can be either an ipv4
--- a/controllers/controllers.go
+++ b/controllers/controllers.go
@@ -15,6 +15,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	"github.com/openshift/windows-machine-config-operator/pkg/instances"
+	"github.com/openshift/windows-machine-config-operator/pkg/metadata"
 	"github.com/openshift/windows-machine-config-operator/pkg/metrics"
 	"github.com/openshift/windows-machine-config-operator/pkg/nodeconfig"
 	"github.com/openshift/windows-machine-config-operator/version"
@@ -118,7 +119,7 @@ func windowsNodePredicate(byoh bool) predicate.Funcs {
 				(!byoh && e.Object.GetAnnotations()[BYOHAnnotation] == "true") {
 				return false
 			}
-			if e.Object.GetAnnotations()[nodeconfig.VersionAnnotation] != version.Get() {
+			if e.Object.GetAnnotations()[metadata.VersionAnnotation] != version.Get() {
 				return true
 			}
 			return false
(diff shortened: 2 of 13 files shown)


jrvaldes:
/test vsphere-e2e-operator

since golden image has been re-created

controllers/controllers.go
```diff
@@ -49,10 +49,24 @@ func (r *instanceReconciler) configureInstance(instance *instances.InstanceInfo,
...
-		return errors.Wrap(err, "failed to configure Windows instance")
+
+	// if instance does not have an associated node, configure it
+	if instance.Node == nil {
+		return nc.Configure()
 	}
-	return nil
+
+	// Version annotation equal to the current version means that the node has been fully configured by this version
+	// of WMCO. No further action is required.
```
aravindhp, line 59: This again seems to be a change in current behavior. Previously we would have reconfigured the `Machine` instance irrespective of the version annotation. Why are you doing this?

controllers/controllers.go
```diff
@@ -49,10 +49,24 @@ func (r *instanceReconciler) configureInstance(instance *instances.InstanceInfo,
 	if err != nil {
 		return errors.Wrap(err, "failed to create new nodeconfig")
 	}
-	if err := nc.Configure(); err != nil {
-		return errors.Wrap(err, "failed to configure Windows instance")
+
+	// if instance does not have an associated node, configure it
+	if instance.Node == nil {
+		return nc.Configure()
```
aravindhp, line 55: This seems to be a change in current behavior. Previously we would have configured the Machine instance irrespective if the node object was present. Are you sure this does not lead to corner cases where we end up with a node object and instance that has not been fully configured?

aravindhp (changes requested):
Thanks for working on this, @sebsoto. Please address my comments

- Please associate the PR with the Jira issue
- Rather than `wrong version annotation` how about using `older version`.

Regarding:
>This currently only
applies to BYOH nodes, however this logic was placed in the
InstanceReconciler methods, and not the ConfigMapReconciler so that the
code can be shared between Machine and BYOH nodes in the future.

This is fine as long as this code sharing is not changing existing behavior with Machine instances.

controllers/controllers.go
```diff
@@ -49,10 +49,24 @@ func (r *instanceReconciler) configureInstance(instance *instances.InstanceInfo,
...
 	}
-	return nil
+
+	// Version annotation equal to the current version means that the node has been fully configured by this version
+	// of WMCO. No further action is required.
+	annotation, present := instance.Node.Annotations[nodeconfig.VersionAnnotation]
+	if annotation == version.Get() {
+		return nil
+	} else if present && annotation != version.Get() {
```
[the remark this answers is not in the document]
sebsoto: We still need to check if the annotation was present on the node object, if it is then we need to deconfigure it.

controllers/controllers.go
sebsoto: This code is confusing, working on streamlining it.
Right now the function is structured like this:

1) If the instance does not have an associated node, return early and configure it.
2) The instance is confirmed to have a node, if the version annotation has an older value, deconfigure the instance. If it has the current value do nothing and return.
3) All instances that have a node are configured.

Changing it to:
1) If the instance has a node and the version annotation is wrong, deconfigure the instance. If it is correct do nothing and return.
2) Configure all instances

Same result but easier to read

controllers/controllers.go
sebsoto: In practice are never doing that. In the case that the annotation has the incorrect value [we immediately attempt to delete the machine](https://github.com/openshift/windows-machine-config-operator/blob/master/controllers/windowsmachine_controller.go#L253-L271).

This function is only called by the WindowsMachine controller if the version annotation is not present

saifshaikh48 (commented):
It looks good to me overall, haven't put LGTM due to https://github.com/openshift/windows-machine-config-operator/pull/513/files#r668246198. Also I'm not 100% sure how the second commit falls in line with the PR's purpose, but fine with including the changes in here.

controllers/configmap_controller_test.go
```diff
@@ -15,55 +17,102 @@ func TestParseHosts(t *testing.T) {
...
 			expectedErr: false,
 		},
 		{
 			name:        "valid dns and ip addresses",
 			input:       map[string]string{"localhost": "username=core", "127.0.0.1": "username=Admin"},
+			nodeList:    &core.NodeList{},
 			expectedOut: []*instances.InstanceInfo{{Address: "localhost", Username: "core"}, {Address: "127.0.0.1", Username: "Admin"}},
 			expectedErr: false,
 		},
+		{
```
mansikulkarni96, line 73: can we add a check here for an invalid case?

pkg/instances/instances.go
```diff
@@ -1,14 +1,21 @@
...
 type InstanceInfo struct {
-	Address     string
-	Username    string
+	// Address is the network address of the instance
+	Address string
+	// Username is the name of a user that can be ssh'd into.
+	Username string
+	// NewHostname being set means that the instance's hostname should be changed. An empty value is a no-op.
 	NewHostname string
+	//nodeRef is an optional pointer to the Node object associated with the instance, if it has one.
```
mansikulkarni96, line 13: incorrect doc name

controllers/controllers.go
```diff
@@ -50,10 +50,26 @@ func (r *instanceReconciler) configureInstance(instance *instances.InstanceInfo,
 	if err != nil {
 		return errors.Wrap(err, "failed to create new nodeconfig")
 	}
-	if err := nc.Configure(); err != nil {
-		return errors.Wrap(err, "failed to configure Windows instance")
+
+	// If the given instance has a Node already, check which version of WMCO configured it
+	if instance.Node != nil {
```
mansikulkarni96, line 55: why not check instance.Node == nil and configure first to reduce indentation.

mansikulkarni96 (commented):
Thanks for working on this @sebsoto PTAL at my comments.

controllers/controllers.go
sebsoto: See https://github.com/openshift/windows-machine-config-operator/pull/513#discussion_r668811315 as to why I changed it from that

controllers/configmap_controller_test.go
```diff
@@ -15,55 +17,125 @@ func TestParseHosts(t *testing.T) {
...
+							},
+						},
+					},
+				},
+			},
+			expectedOut: []*instances.InstanceInfo{
+				{Address: "127.0.0.1", Username: "Admin", Node: nil},
+				{Address: "localhost", Username: "core", Node: nil},
+			},
+			expectedErr: false,
```
mansikulkarni96, line 94: I am a little confused shouldn't we expect error here and in above case?

controllers/controllers.go
aravindhp: My [comment](https://github.com/openshift/windows-machine-config-operator/pull/513#discussion_r668292078) still stands even with this change. Or am I missing something?

aravindhp (changes requested):

saifshaikh48 (commented):
/lgtm

alinaryan:
Thanks for working on this, the only suggestion I have is to maybe call out why you moved from using `host` -> `instance`

controllers/controllers.go
```diff
@@ -41,19 +41,61 @@ type instanceReconciler struct {
...
-		annotations)
+		annotationsToApply)
 	if err != nil {
 		return errors.Wrap(err, "failed to create new nodeconfig")
 	}
-	if err := nc.Configure(); err != nil {
-		return errors.Wrap(err, "failed to configure Windows instance")
+
+	// Check if the instance was configured by a previous version of WMCO and must be deconfigured before being
+	// configured again.
```
aravindhp, line 64: I ran into an issue where an instance was added to the ConfigMap which resulted in it getting configured. I then removed it from the ConfigMap causing it to be deconfigured. Now when I tried adding it back to the ConfigMap, it failed to get configured because hybrid-overlay failed to come update the new Node object with the annotation because it ignored creating the already existing HNS network. Have you run into this issue?

controllers/controllers.go
```diff
@@ -41,19 +41,61 @@ type instanceReconciler struct {
...
+		}
 	}
-	return nil
+
+	return nc.Configure()
+}
+
+// instanceUpToDate returns true if the instance was configured by the current WMCO version
+func instanceUpToDate(nodeAnnotations map[string]string) bool {
+	versionAnnotation, _ := nodeAnnotations[nodeconfig.VersionAnnotation]
```
aravindhp, line 76: I assume this handles the case of the annotation not being present.

controllers/controllers.go
```diff
@@ -41,19 +41,61 @@ type instanceReconciler struct {
...
+	annotation, present := nodeAnnotations[nodeconfig.VersionAnnotation]
+
+	// Version annotation not being present means that the node has been created but not fully configured.
+	// The upgrade process is not required, the node should just be configured normally.
+	if !present {
+		return false
+	}
+
+	// Version annotation equal to the current version means that the node has been fully configured by this version
+	// of WMCO. No further action is required.
```
aravindhp, line 91: Is `No further action is required` accurate? Returning `false` here just implies that the instance will not be `deconfigured` but the instance will go through  configuration.

controllers/controllers.go
```diff
@@ -41,19 +41,61 @@ type instanceReconciler struct {
...
+	if annotation == version.Get() {
```
aravindhp, line 92: Furthermore I also don't understand how this case will ever fire, given this is the equivalent of `if instance.Node != nil && instanceUpToDate(instance.Node.Annotations)`. Or am I missing something?

controllers/controllers_test.go
```diff
@@ -62,3 +65,62 @@ func TestGetAddress(t *testing.T) {
...
+		input       map[string]string
+		expectedOut bool
+	}{
+		{
+			name:        "Version annotation missing",
+			input:       map[string]string{"wrong-annotation": version.Get()},
+			expectedOut: false,
+		},
+		{
+			name:        "Version annotation incorrect",
```
aravindhp, line 81: Instead of `incorrect`, `mismatch` seems more apt.

controllers/controllers_test.go
```diff
@@ -62,3 +65,62 @@ func TestGetAddress(t *testing.T) {
...
```
aravindhp, line 110: Same here

aravindhp (changes requested):
This looks much cleaner, @sebsoto

pkg/nodeconfig/nodeconfig.go
```diff
@@ -380,10 +380,11 @@ func (nc *nodeConfig) Deconfigure() error {
 		return errors.Wrap(err, "error deconfiguring instance")
 	}
 
-	// Delete the Node object
-	err := nc.k8sclientset.CoreV1().Nodes().Delete(context.TODO(), nc.node.GetName(), meta.DeleteOptions{})
+	// Clear the version annotation from the node object to indicate the node is not configured
```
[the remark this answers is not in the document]
sebsoto: Sorry, meant to cover this in the meeting for my PR.
I cant think of any at the moment, besides trying to make a `Not ready` node `Ready`. But that should be handled by MachineHealthChecks for Machines, and as for BYOH I am not sure.

controllers/configmap_controller_test.go
sebsoto: There shouldnt be an error as not finding an associated node is not an error.

controllers/controllers.go
sebsoto: Yes as if the annotation is not present it will return the default value of "".  Going to add the present check just to be more clear

controllers/controllers.go
sebsoto: Ill rephrase

controllers/controllers.go
sebsoto: Made some changes to the function to reuse the `instanceUpToDate` function.
I wanted to cover all the possible cases in these functions, in case the way they are used in `ensureInstanceIsUpToDate()` is changed, the functions stand on their own.
For these functions even though the logic is very little, I thought it worked well to contain it, in order to simplify the function they are called.

controllers/controllers.go
sebsoto: I have not noticed that. Let me double check.

controllers/controllers.go
```diff
@@ -41,19 +41,59 @@ type instanceReconciler struct {
...
+// applied to it.
+func (r *instanceReconciler) ensureInstanceIsUpToDate(instance *instances.InstanceInfo, annotationsToApply map[string]string) error {
+	if instance == nil {
+		return errors.New("instance cannot be nil")
+	}
+
+	// Instance is up to date, do nothing
+	if instance.Node != nil && instanceUpToDate(instance.Node.Annotations) {
+		return nil
+	}
```
aravindhp, line 55: Re-reading this, I think the function `instanceUpToDate()` should take `InstanceInfo` as the input and you should do all the checks in the function.

controllers/controllers.go
```diff
@@ -41,19 +41,59 @@ type instanceReconciler struct {
...
 		return errors.Wrap(err, "failed to create new nodeconfig")
 	}
-	if err := nc.Configure(); err != nil {
-		return errors.Wrap(err, "failed to configure Windows instance")
+
+	// Check if the instance was configured by a previous version of WMCO and must be deconfigured before being
+	// configured again.
+	if instance.Node != nil && upgradeRequired(instance.Node.Annotations) {
+		if err := nc.Deconfigure(); err != nil {
+			return err
```
aravindhp, line 67: Same here. Also for consistency maybe call the function `instanceUpgradeRequired()`. Or remove the `instance` prefix from both functions.

aravindhp (changes requested):
Almost approve 😉

controllers/controllers.go
```diff
@@ -41,19 +41,88 @@ type instanceReconciler struct {
...
+	if instance.Node == nil {
+		return false, nil
+	}
+
+	versionAnnotation, present := instance.Node.GetAnnotations()[nodeconfig.VersionAnnotation]
+	return present && versionAnnotation == version.Get(), nil
+}
+
+// instanceUpgradeRequired returns true if the instance needs to go through the upgrade process.
+func instanceUpgradeRequired(instance *instances.InstanceInfo) (bool, error) {
```
aravindhp, line 97: Sorry @sebsoto , I was not thinking things through with my previous [comment](https://github.com/openshift/windows-machine-config-operator/pull/513#discussion_r673283175). These functions should actually be part of the `instance` struct. But given it was an oversight on my part, I can open a follow up PR if you just want to get this merged.

aravindhp (approved):

/approve

controllers/controllers.go
aravindhp: If you decide to make the change please ensure the calls don't stutter.

sebsoto: In order to do that the VersionAnnotation const would need to pulled out of the nodeconfig package, as currently trying to use it from the `instances` package would result in a circular dependency. I can give it a shot.
I don't want to do the alternative which would be passing in the annotation key to the NewInstance function.

config/manager/kustomization.yaml
```diff
@@ -14,4 +14,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: REPLACE_IMAGE
+  newName: quay.io/ssoto/test
+  newTag: wmco
```
aravindhp, line 18: I am guessing this was an oversight

aravindhp (changes requested):

config/manager/kustomization.yaml
sebsoto: Fixed

aravindhp (approved):
/approve

pkg/metadata/metadata.go
```diff
@@ -0,0 +1,4 @@
+package metadata
```
saifshaikh48, line 1: I feel like there should be more in this package e.g. `PubKeyHashAnnotation`. If you agree, we can create a story to follow-up on this.

PR #515, opened 2021-07-13
[docs] Update scripts and docs for vSphere golden image

jrvaldes:
This PR fixes formatting issues and typos, aligns the
Packer builder script (build.json) with the documentation,
and adds a missing step in the PowerShell script (install-docker.ps1)
that installs docker as part of the golden images generalization process.

Highlights:
 - build.json file formatted and re-arranged grouping
   relevant keys together
 - `variables` key moved to the top of build.json script
   for easy access during customization
 - `vsphere-network` parameter removed from build.json script
 - `vsphere-folder` parameter renamed to `vm-template-folder`
 - `winadmin-password` parameter renamed to `vm-elevated-password`
 - Increased the provisioner pause to 3 mins
 - Added default values to `os-iso-path` and `vmtools-iso-path`
   parameters to follow the naming convention adopted
   by https://issues.redhat.com/browse/DPP-7909


docs/vsphere_ci/build.json
```diff
@@ -1,82 +1,82 @@
...
+      "max_retries": "10"
+    }
+  ],
+  "builders": [
+    {
+      "CPUs": "2",
+      "RAM": "4096",
+      "RAM_reserve_all": true,
+      "communicator": "ssh",
+      "convert_to_template": "true",
```
aravindhp, line 37: You had mentioned that the resulting entity is not a template but a VM. So why is this set?

aravindhp (changes requested):
@jrvaldes I could not get this to work with these changes.

```
$ PACKER_LOG=1 ./packer build build.json
2021/07/16 22:55:51 [INFO] Packer version: 1.6.6 [go1.15.6 linux amd64]
2021/07/16 22:55:51 Checking 'PACKER_CONFIG' for a config file path
2021/07/16 22:55:51 'PACKER_CONFIG' not set; checking the default config file path
2021/07/16 22:55:51 Attempting to open config file: /home/aravindh/.packerconfig
2021/07/16 22:55:51 [WARN] Config file doesn't exist: /home/aravindh/.packerconfig
2021/07/16 22:55:51 Setting cache directory: /home/aravindh/packer/packer_cache
2021/07/16 22:55:51 Creating plugin client for path: /home/aravindh/packer/packer
2021/07/16 22:55:51 Starting plugin: /home/aravindh/packer/packer []string{"/home/aravindh/packer/packer", "plugin", "packer-builder-vsphere-iso"}
2021/07/16 22:55:51 Waiting for RPC address for: /home/aravindh/packer/packer
2021/07/16 22:55:51 packer-builder-vsphere-iso plugin: [INFO] Packer version: 1.6.6 [go1.15.6 linux amd64]
2021/07/16 22:55:51 Received unix RPC address for /home/aravindh/packer/packer: addr is /tmp/packer-plugin275536640
2021/07/16 22:55:51 packer-builder-vsphere-iso plugin: Checking 'PACKER_CONFIG' for a config file path
2021/07/16 22:55:51 packer-builder-vsphere-iso plugin: 'PACKER_CONFIG' not set; checking the default config file path
2021/07/16 22:55:51 packer-builder-vsphere-iso plugin: Attempting to open config file: /home/aravindh/.packerconfig
2021/07/16 22:55:51 packer-builder-vsphere-iso plugin: [WARN] Config file doesn't exist: /home/aravindh/.packerconfig
2021/07/16 22:55:51 packer-builder-vsphere-iso plugin: Setting cache directory: /home/aravindh/packer/packer_cache
2021/07/16 22:55:51 packer-builder-vsphere-iso plugin: args: []string{"packer-builder-vsphere-iso"}
2021/07/16 22:55:51 packer-builder-vsphere-iso plugin: Plugin address: unix /tmp/packer-plugin275536640
2021/07/16 22:55:51 packer-builder-vsphere-iso plugin: Waiting for connection...
2021/07/16 22:55:51 packer-builder-vsphere-iso plugin: Serving a plugin connection...
2021/07/16 22:55:51 Creating plugin client for path: /home/aravindh/packer/packer
2021/07/16 22:55:51 Starting plugin: /home/aravindh/packer/packer []string{"/home/aravindh/packer/packer", "plugin", "packer-provisioner-powershell"}
2021/07/16 22:55:51 Waiting for RPC address for: /home/aravindh/packer/packer
2021/07/16 22:55:51 packer-provisioner-powershell plugin: [INFO] Packer version: 1.6.6 [go1.15.6 linux amd64]
2021/07/16 22:55:51 packer-provisioner-powershell plugin: Checking 'PACKER_CONFIG' for a config file path
2021/07/16 22:55:51 packer-provisioner-powershell plugin: 'PACKER_CONFIG' not set; checking the default config file path
2021/07/16 22:55:51 packer-provisioner-powershell plugin: Attempting to open config file: /home/aravindh/.packerconfig
2021/07/16 22:55:51 packer-provisioner-powershell plugin: [WARN] Config file doesn't exist: /home/aravindh/.packerconfig
2021/07/16 22:55:51 packer-provisioner-powershell plugin: Setting cache directory: /home/aravindh/packer/packer_cache
2021/07/16 22:55:51 packer-provisioner-powershell plugin: args: []string{"packer-provisioner-powershell"}
2021/07/16 22:55:51 Received unix RPC address for /home/aravindh/packer/packer: addr is /tmp/packer-plugin725954232
2021/07/16 22:55:51 packer-provisioner-powershell plugin: Plugin address: unix /tmp/packer-plugin725954232
2021/07/16 22:55:51 packer-provisioner-powershell plugin: Waiting for connection...
2021/07/16 22:55:51 Preparing build: winsrv-2004-builder
2021/07/16 22:55:51 packer-provisioner-powershell plugin: Serving a plugin connection...
winsrv-2004-builder: output will be in this color.

2021/07/16 22:55:51 Build debug mode: false
2021/07/16 22:55:51 Force build: false
2021/07/16 22:55:51 On error: 
2021/07/16 22:55:51 Waiting on builds to complete...
2021/07/16 22:55:51 Starting build run: winsrv-2004-builder
2021/07/16 22:55:51 Running builder: vsphere-iso
2021/07/16 22:55:51 [INFO] (telemetry) Starting builder vsphere-iso
2021/07/16 22:55:51 packer-builder-vsphere-iso plugin: No URLs were provided to Step Download. Continuing...
2021/07/16 22:55:51 packer-builder-vsphere-iso plugin: No CD files specified. CD disk will not be made.
==> winsrv-2004-builder: Creating VM...
==> winsrv-2004-builder: Customizing hardware...
==> winsrv-2004-builder: Mounting ISO images...
2021/07/16 22:55:54 packer-builder-vsphere-iso plugin: Check if ISO path is a Content Library path
2021/07/16 22:55:55 packer-builder-vsphere-iso plugin: ISO path not identified as a Content Library path
2021/07/16 22:55:55 packer-builder-vsphere-iso plugin: Using [WorkloadDatastore] windows-iso-images/winsrv2004_sep_2020_x64_dvd.iso as the datastore path
2021/07/16 22:55:55 packer-builder-vsphere-iso plugin: Creating CD-ROM on controller '&{{{} 200 0xc000b12f60 <nil> <nil> <nil> 0 <nil>} 0 []}' with iso '[WorkloadDatastore] windows-iso-images/winsrv2004_sep_2020_x64_dvd.iso'
2021/07/16 22:55:55 packer-builder-vsphere-iso plugin: Check if ISO path is a Content Library path
2021/07/16 22:55:55 packer-builder-vsphere-iso plugin: ISO path not identified as a Content Library path
2021/07/16 22:55:55 packer-builder-vsphere-iso plugin: Using [WorkloadDatastore] windows-iso-images/vmtools-v11333-windows.iso as the datastore path
2021/07/16 22:55:55 packer-builder-vsphere-iso plugin: Creating CD-ROM on controller '&{{{} 200 0xc000c87e00 <nil> <nil> <nil> 0 <nil>} 0 [3000]}' with iso '[WorkloadDatastore] windows-iso-images/vmtools-v11333-windows.iso'
==> winsrv-2004-builder: Adding configuration parameters...
==> winsrv-2004-builder: Creating floppy disk...
    winsrv-2004-builder: Copying files flatly from floppy_files
    winsrv-2004-builder: Copying file: scripts/autounattend.xml
2021/07/16 22:55:56 packer-builder-vsphere-iso plugin: Floppy path: /tmp/packer225898079
2021/07/16 22:55:56 packer-builder-vsphere-iso plugin: Initializing block device backed by temporary file
2021/07/16 22:55:56 packer-builder-vsphere-iso plugin: Formatting the block device with a FAT filesystem...
2021/07/16 22:55:56 packer-builder-vsphere-iso plugin: Initializing FAT filesystem on block device
2021/07/16 22:55:56 packer-builder-vsphere-iso plugin: Reading the root directory from the filesystem
    winsrv-2004-builder: Copying file: scripts/install-vm-tools.cmd
    winsrv-2004-builder: Copying file: scripts/configure-vm-tools.ps1
    winsrv-2004-builder: Copying file: scripts/install-openssh.ps1
    winsrv-2004-builder: Copying file: scripts/install-firewall-rules.ps1
    winsrv-2004-builder: Copying file: scripts/authorized_keys
    winsrv-2004-builder: Copying file: scripts/install-updates.ps1
    winsrv-2004-builder: Copying file: scripts/install-docker.ps1
    winsrv-2004-builder: Done copying files from floppy_files
    winsrv-2004-builder: Collecting paths from floppy_dirs
    winsrv-2004-builder: Resulting paths from floppy_dirs : []
    winsrv-2004-builder: Done copying paths from floppy_dirs
==> winsrv-2004-builder: Uploading created floppy image
==> winsrv-2004-builder: Adding generated Floppy...
==> winsrv-2004-builder: Set boot order temporary...
==> winsrv-2004-builder: Power on VM...
2021/07/16 22:55:58 packer-builder-vsphere-iso plugin: [INFO] Waiting for IP, up to total timeout: 30m0s, settle timeout: 5s
==> winsrv-2004-builder: Waiting for IP...
==> winsrv-2004-builder: Timeout waiting for IP.
==> winsrv-2004-builder: Clear boot order...
==> winsrv-2004-builder: Power off VM...
==> winsrv-2004-builder: Deleting Floppy image ...
2021/07/16 23:25:59 packer-builder-vsphere-iso plugin: Deleting floppy disk: /tmp/packer225898079
==> winsrv-2004-builder: Destroying VM...
2021/07/16 23:25:59 [INFO] (telemetry) ending vsphere-iso
==> Wait completed after 30 minutes 8 seconds
2021/07/16 23:25:59 machine readable: error-count []string{"1"}
==> Some builds didn't complete successfully and had errors:
2021/07/16 23:25:59 machine readable: winsrv-2004-builder,error []string{"Timeout waiting for IP."}
==> Builds finished but no artifacts were created.
Build 'winsrv-2004-builder' errored after 30 minutes 8 seconds: Timeout waiting for IP.

==> Wait completed after 30 minutes 8 seconds

==> Some builds didn't complete successfully and had errors:
--> winsrv-2004-builder: Timeout waiting for IP.

==> Builds finished but no artifacts were created.
2021/07/16 23:25:59 Cancelling builder after context cancellation context canceled
2021/07/16 23:25:59 [INFO] (telemetry) Finalizing.
2021/07/16 23:26:00 waiting for all plugin processes to complete...
2021/07/16 23:26:00 /home/aravindh/packer/packer: plugin process exited
2021/07/16 23:26:00 /home/aravindh/packer/packer: plugin process exited
```

docs/vsphere_ci/build.json
jrvaldes: Didn't went well last time I tuned this one, still will take a closer look.

docs/vsphere_ci/build.json
```diff
@@ -1,82 +1,82 @@
...
+      "datastore": "{{user `vsphere-datastore`}}",
+      "disk_controller_type": "lsilogic-sas",
+      "firmware": "bios",
+      "floppy_files": [
+        "scripts/autounattend.xml",
+        "scripts/install-vm-tools.cmd",
+        "scripts/install-openssh.ps1",
+        "scripts/install-firewall-rules.ps1",
+        "scripts/authorized_keys",
+        "scripts/win-updates.ps1",
```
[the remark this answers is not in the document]
jrvaldes: Good catch, it's not in use should be removed?

docs/vsphere_ci/build.json
```diff
@@ -1,82 +1,82 @@
...
+      ],
+      "guest_os_type": "windows9Server64Guest",
+      "insecure_connection": "true",
+      "iso_paths": [
+        "{{user `os-iso-path`}}",
+        "{{user `vmtools-iso-path`}}"
+      ],
+      "network_adapters": [
+        {
+          "network": "dev-segment",
```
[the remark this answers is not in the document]
jrvaldes: Correct, while CI the ci-segment network is used.

docs/vsphere_ci/scripts/install-docker.ps1
```diff
@@ -1,3 +1,4 @@
+Install-PackageProvider -Name NuGet -MinimumVersion 2.8.5.201 -Force
```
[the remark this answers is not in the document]
jrvaldes: Before, the install-openssh.ps1 script was installing it, and it ran first.

jrvaldes (commented):
@aravindhp  Thanks for the feedback.

docs/vsphere_ci/build.json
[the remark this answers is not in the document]
aravindhp: That depends on if we want to control the exact version we are running. I am fine installing updates as long as we track what patch version the current golden image is at.

docs/vsphere_ci/scripts/install-docker.ps1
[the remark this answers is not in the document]
aravindhp: Yes but why are you adding it back? `NuGet` was needed before to install `sshutils` which is no longer required.

docs/vsphere_ci/build.json
aravindhp: Sorry I don't follow. If this did not get well before, why did you create the current golden image with this option?

docs/vsphere_ci/scripts/install-docker.ps1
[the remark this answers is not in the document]
jrvaldes: `NuGet` provider is a requirement to install `docker`, as per [documentation](https://docs.microsoft.com/en-us/virtualization/windowscontainers/quick-start/set-up-environment?tabs=Windows-Server); 

> If you're prompted to install the NuGet provider, type Y to install it as well.

Even using the  `-Force` flag with `Install-Module -Name DockerMsftProvider -Repository PSGallery`  a manual confirmation is needed. So to avoid that, `NuGet` provider  is getting installed first.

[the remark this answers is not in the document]
aravindhp: Please call that out by linking to the doc as it is not immediately obvious why it is needed.

docs/vsphere_ci/build.json
jrvaldes: Let me add more color to that statement.

As per current [documentation](https://github.com/openshift/windows-machine-config-operator/blob/master/docs/vsphere_ci/README.md#post-packer-build), the process creates a template (given this flag is set to `true`) and need to be manually converted to a VM in the vCenter.

Had mixed feelings about that too, so explored to following two scenarios:
1- disabling this flag (`convert_to_template: false`) and directly producing a VM, to be used as the golden image.
2- leaving this flag enabled, and using the template, as is in the [template field](https://github.com/openshift/windows-machine-config-operator/blob/820507d6ad0bcfb5d8cd6bad7686a8bbfa087612/test/e2e/providers/vsphere/vsphere.go#L57) of the e2e test, where the [VSphereMachineProviderSpec](https://github.com/openshift/windows-machine-config-operator/blob/595f2828d9aea5c6aefb1a168d4264ed78c83d30/vendor/github.com/openshift/machine-api-operator/pkg/apis/vsphereprovider/v1beta1/vsphereproviderconfig_types.go#L27) is configured.

Back then, none of the above seemed to generate a valid golden image. Since, there were a few inconsistencies in the process will revisit them.

docs/vsphere_ci/README.md
```diff
@@ -71,7 +71,6 @@ adjust the following variables:
 - `<vsphere-cluster>` Name of the vSphere cluster
```
aravindhp, line 74: Commit message comment:

> name property added to build.json script for best practices,
   otherwise it default to type and limit the number of builder
   definitions due name collisions

Not sure what this implies and might need rewording.

docs/vsphere_ci/build.json
```diff
@@ -1,82 +1,82 @@
...
-      "vsphere-server":"vcenter_url",
-      "vsphere-user":"Your_username",
-      "winadmin-password":"Some random password"
-   }
+  "variables": {
+    "os-iso-path": "[WorkloadDatastore] windows-iso-images/winsrv2004_sep_2020_x64_dvd.iso",
+    "vmtools-iso-path": "[WorkloadDatastore] windows-iso-images/vmtools-v11333-windows.iso",
+    "vm-template-folder": "windows-golden-images",
+    "vm-template-name": "vm-winsrv-2004-golden-image-template",
+    "vm-elevated-password": "Some random password",
```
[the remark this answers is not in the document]
aravindhp: Looks like you fixed this in another commit. Please stop introducing something new in one commit and then fixing it in a follow up commit. You rather need to amend this commit.

docs/vsphere_ci/build.json
```diff
@@ -1,82 +1,82 @@
...
+    "vsphere-cluster": "Cluster-1",
+    "vsphere-datacenter": "SDDC-Datacenter",
+    "vsphere-datastore": "WorkloadDatastore",
+    "vsphere-server": "vcenter_url",
```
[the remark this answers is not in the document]
aravindhp: Same here.

aravindhp (changes requested):

docs/vsphere_ci/README.md
jrvaldes: Good catch, actually no need to introduce the `name` property, so comment no longer relevant.
For further references see [packer documentation](https://www.packer.io/docs/templates/legacy_json_templates/builders#named-builds)

saifshaikh48:
/lgtm

@jrvaldes Thanks for working on this.

docs/vsphere-golden-image.md
```diff
@@ -158,13 +160,24 @@ where `<path/to/unattend.xml>` is the path to the customized answer file.
...
-## 7. Set up the virtual machine golden image
+## 7. Set up the virtual machine template
 
 Once the `sysprep` command completes the Windows virtual machine will power off. 
 
-You **must not** use this Windows virtual machine anymore. 
+You **must not** use or power on this virtual machine anymore.
+
+Note: In case, you need to update or change the configuration of the golden image, make the changes in the 
+original Windows virtual machine (before running the `sysprep` command), clone it and then execute `sysprep` command.
```
aravindhp, line 170: I suggest leaving this out. It is a bit confusing. I know this is in the existing docs and we are trying to cover all cases but lets leave it out for simplicity.

aravindhp (changes requested):

aravindhp:
/hold

Until:
- These steps are reproducible
- We figure out why the the VM created from the template has an open console sitting at the Administrator prompt

docs/vsphere-golden-image.md
```diff
@@ -147,6 +147,8 @@ makes it ready for future deployments, maintaining all the changes needed for th
 You can find an example [unattend.xml](unattend.xml) file in this repository. It is just an example and, it **should 
 not** be used directly. You must customize the example to fit your needs.
 
+Before executing the `sysprep` command, clone the VM and then run `sysprep` command in the cloned VM.
```
aravindhp, line 150: We need to call out the hostname requirements somewhere in these steps.

docs/vsphere_ci/scripts/autounattend.xml
```diff
@@ -101,16 +101,21 @@
                 </SynchronousCommand>
                 <SynchronousCommand wcm:action="add">
                     <Order>4</Order>
+                    <Description>Configure VMware Tools</Description>
```
aravindhp, line 104: Don't you need to add `<ComputerName>` somewhere in this file? Following the instructions from this PR is resulting in VMs that have hostname set in CAPS which causes the kubelet configuration to fail.

jrvaldes: Good catch, thank you. Looks like missed it in the rebase process.

docs/vsphere_ci/build.json
```diff
@@ -1,82 +1,80 @@
...
-      "vsphere-datacenter":"SDDC-Datacenter",
-      "vsphere-datastore":"WorkloadDatastore",
-      "vsphere-folder":"windows-eval-iso",
-      "vsphere-network":"dev-segment",
-      "vsphere-password":"Your_password",
-      "vsphere-server":"vcenter_url",
-      "vsphere-user":"Your_username",
-      "winadmin-password":"Some random password"
-   }
+  "variables": {
```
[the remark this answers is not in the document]
jrvaldes: Agree

aravindhp (changes requested):
When I tried the packer steps, I could not get past:
```
2021-07-22T23:57:15.502Z        DEBUG   VM 172.31.251.240       initializing SSH connection     {"user": "Administrator"}
2021-07-22T23:57:45.676Z        INFO    VM 172.31.251.240       configuring
```
This is a regression for me as previously it was failing at kubelet stage due to the bad hostname.

docs/vsphere-golden-image.md
jrvaldes: You can find it [here](https://github.com/openshift/windows-machine-config-operator/pull/515/commits/3f983dffb1af3d1d8c2e35e70d8b1b09b1f20cb2#diff-eff7340e15df41ff3a96507abf744eb5fea3f670fbb381edc624b52cda6b1b79R150), it was added as part of this [commit](https://github.com/openshift/windows-machine-config-operator/pull/515/commits/68fac906e2a151b4c049efd41b0db29c4e441695#diff-eff7340e15df41ff3a96507abf744eb5fea3f670fbb381edc624b52cda6b1b79R150) where t[he computer name is introduced](https://github.com/openshift/windows-machine-config-operator/pull/515/commits/68fac906e2a151b4c049efd41b0db29c4e441695#diff-f95d5d0ee98dedf392fb54c0fb93f7e3df3c9699b01d08ca8fd00d30ccba8244R71)

aravindhp: None of these links work anymore which is expected when you force push a commit.

docs/vsphere-golden-image.md
```diff
@@ -147,6 +147,9 @@ makes it ready for future deployments, maintaining all the changes needed for th
 You can find an example [unattend.xml](unattend.xml) file in this repository. It is just an example and, it **should 
 not** be used directly. You must customize the example to fit your needs.
 
+For example, the value of the `ComputerName` property in the [unattend.xml](unattend.xml) file must follows the
```
aravindhp, line 150: Rather than `For example,`, you should say `Please note that the`. This is not an optional change but a must have. You should also call out that if the customer is apply Guest OS customization when launching from the resulting template, the name there again needs to match the k8s specs.

docs/vsphere-golden-image.md
aravindhp: You have added docs for this in the CI steps but not in [vsphere-golden-image.md](https://github.com/openshift/windows-machine-config-operator/blob/master/docs/vsphere-golden-image.md)

docs/vsphere-golden-image.md
```diff
@@ -176,12 +186,9 @@ Add the cloned Windows virtual machine as a template in the [machineset](../READ
       template:
         spec:
           providerSpec:
-            template: <Cloned Windows Virtual Machine>
+            template: <Windows Virtual Machine Template Name>
```
aravindhp, line 189: We also need to call out that the network used here needs to match the network that other Linux workers are on. You need to mention this in [vsphere-golden-image.md](https://github.com/openshift/windows-machine-config-operator/blob/master/docs/vsphere-golden-image.md) also.

docs/vsphere-golden-image.md
jrvaldes: My bad, links updated. PTAL

aravindhp (changes requested):
I think I might have figured out why the image failed during configuration. I was blindly following the steps and it does not call out anything regarding the network that needs to be used. I am verifying that now.

docs/vsphere-golden-image.md
jrvaldes: Not quite sure about such level of details, do you anticipate calling out the infrastructure ID and node labels too? Those are presented in the [documentation](https://docs.openshift.com/container-platform/4.7/machine_management/creating_machinesets/creating-machineset-vsphere.html#machineset-yaml-vsphere_creating-machineset-vsphere) as well.

aravindhp: Without this "level of detail", a customer could launch a VM connected to a different network and open a bug saying it is not getting configured. So this is a missing piece in our docs.

aravindhp: Maybe a better spot for this would be [configuring-windows-instances-provisioned-through-machinesets](https://github.com/openshift/windows-machine-config-operator#configuring-windows-instances-provisioned-through-machinesets).

jrvaldes: Agree, leaving [vsphere-golden-image.md ](https://github.com/openshift/windows-machine-config-operator/blob/master/docs/vsphere-golden-image.md) with direct emphasis on golden image creation and usage.

jrvaldes: > Maybe a better spot for this would be [configuring-windows-instances-provisioned-through-machinesets](https://github.com/openshift/windows-machine-config-operator#configuring-windows-instances-provisioned-through-machinesets).

I feel that should come in a separated PR, thoughts?

aravindhp: I am fine with that.

aravindhp:
/retitle [docs] Update scripts and docs for vSphere golden image

docs/vsphere-golden-image.md
```diff
@@ -151,6 +151,8 @@ Please note that, the value of the `ComputerName` property in the [unattend.xml]
 [Kubernetes' names specification](https://kubernetes.io/docs/concepts/overview/working-with-objects/names). The latter,
 also applies to Guest OS customization performed on the resulting template while creating new VMs. 
 
+Before executing the `sysprep` command, clone the VM and then run `sysprep` command in the cloned VM.
```
aravindhp, line 154: @EricPonvelle please review this and reflect it in the OpenShift docs.

aravindhp (approved):
Thanks for working on this, @jrvaldes 

/approve

docs/vsphere-golden-image.md
jrvaldes: [PR](https://github.com/openshift/windows-machine-config-operator/pull/545) opened

jrvaldes:
/hold cancel

Looking forward to addressing @EricPonvelle review

aravindhp:
@jrvaldes please fix the 50/80 commit formats in this PR.

docs/vsphere_ci/README.md
```diff
@@ -63,18 +63,20 @@ In order to use the provided [reference build file](build.json) as a valid confi
 adjust the following variables:
 
 - `<vmtools-iso-path>` Path where VMWare Tools ISO is available in vSphere datacenter
-  (default: `[] /usr/lib/vmware/isoimages/windows.iso`)  
+  (default: `[] /usr/lib/vmware/isoimages/windows.iso`)
```
aravindhp, line 68: Commit message should read `[docs] [vsphere_ci] Update Packer build

docs/vsphere-golden-image.md
```diff
@@ -144,8 +144,13 @@ To deploy the Windows VM as a reusable image, you have to first generalize the V
...
 
-You can find an example [unattend.xml](unattend.xml) file in this repository. It is just an example and, it **should 
-not** be used directly. You must customize the example to fit your needs.
+A reference [unattend.xml](unattend.xml) file is provided, it **cannot** be used directly. You must customize it
+replacing the `WindowsPassword` parameter with the desired password for the Administrator user. The Password must 
+meet complexity requirements as per [Microsoft documentation](https://docs.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/password-must-meet-complexity-requirements).
+
+Please note that, the value of the `ComputerName` property in the [unattend.xml](unattend.xml) file must follows the
+[Kubernetes' names specification](https://kubernetes.io/docs/concepts/overview/working-with-objects/names). The latter,
+also applies to Guest OS customization performed on the resulting template while creating new VMs.
```
aravindhp, line 153: Please move this to a commit where you are making non-CI changes.

docs/vsphere_ci/README.md
```diff
@@ -107,23 +107,11 @@ Packer mounts the Windows iso and starts the VM.
...
-    kind: MachineSet
-    spec:
-      template:
-        spec:
-          providerSpec:
-            template: <Windows Virtual Machine Name>
-```
+Once the Packer build completes successfully, a new VM template with name `<vm-template-name>` will be created in
+the folder `<vm-template-folder>` following the [Variables](#variables). The later VM template is ready to use as a
+golden image, as described in [the documentation](../vsphere-golden-image.md#8-using-the-virtual-machine-template).
```
aravindhp, line 114: This commit should not have anything CI specific.

aravindhp (changes requested):

docs/vsphere-golden-image.md
```diff
@@ -144,8 +144,15 @@ To deploy the Windows VM as a reusable image, you have to first generalize the V
 such as installed drivers. Running the `sysprep` command with a *unattend* answer file generalizes the image and 
 makes it ready for future deployments, maintaining all the changes needed for the WMCO installation. 
 
-You can find an example [unattend.xml](unattend.xml) file in this repository. It is just an example and, it **should 
-not** be used directly. You must customize the example to fit your needs.
+A reference [unattend.xml](unattend.xml) file is provided, it **cannot** be used directly. You must customize it
+replacing the `WindowsPassword` parameter with the desired password for the Administrator user. The Password must
+meet complexity requirements as per [Microsoft documentation](https://docs.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/password-must-meet-complexity-requirements).
```
aravindhp, line 149: `password` in lower case please. This is a best practice and not a `must`. So something to the effect of `We suggest following the Microsoft best practices for picking a password...`

docs/vsphere-golden-image.md
```diff
@@ -144,8 +144,15 @@ To deploy the Windows VM as a reusable image, you have to first generalize the V
...
+
+Please note that, the value of the `ComputerName` property in the [unattend.xml](unattend.xml) file must follows the
+[Kubernetes' names specification](https://kubernetes.io/docs/concepts/overview/working-with-objects/names). The latter,
+also applies to Guest OS customization performed on the resulting template while creating new VMs.
+
+Before executing the `sysprep` command, clone the VM and then run `sysprep` command in the cloned VM.
```
aravindhp, line 155: Please move this to the same line as `To execute the `sysprep` command use:` following the 120 column constraint.

docs/vsphere-golden-image.md
```diff
@@ -158,13 +165,21 @@ where `<path/to/unattend.xml>` is the path to the customized answer file.
 Note: There is [a limit](https://docs.microsoft.com/en-us/windows-hardware/manufacture/desktop/sysprep--generalize--a-windows-installation#limits-on-how-many-times-you-can-run-sysprep)
 on how many times you can run the `sysprep` command.
 
-## 7. Set up the virtual machine golden image
+## 7. Set up the virtual machine template
```
aravindhp, line 168: Has your PR merged that uses the template?

aravindhp (changes requested):

docs/vsphere-golden-image.md
jrvaldes: No, just missing a LGTM

See https://github.com/openshift/windows-machine-config-operator/pull/544

docs/vsphere-golden-image.md
```diff
@@ -144,10 +144,16 @@ To deploy the Windows VM as a reusable image, you have to first generalize the V
 such as installed drivers. Running the `sysprep` command with a *unattend* answer file generalizes the image and 
 makes it ready for future deployments, maintaining all the changes needed for the WMCO installation. 
 
-You can find an example [unattend.xml](unattend.xml) file in this repository. It is just an example and, it **should 
-not** be used directly. You must customize the example to fit your needs.
+A reference [unattend.xml](unattend.xml) file is provided, it **cannot** be used directly. You must customize it
+replacing the `WindowsPassword` parameter with the desired password for the Administrator user. It's recommended to
```
aravindhp, line 148: `WindowsPassword` is not a parameter but the value of the `Password` and `AdministratorPassword` parameters / properties. The intent here should be to indicate to replace the `Password` and `AdministratorPassword` parameter / property values. And if they need to match, it needs to be called out here.

docs/vsphere-golden-image.md
```diff
@@ -144,10 +144,16 @@ To deploy the Windows VM as a reusable image, you have to first generalize the V
...
+follow the [Microsoft best practices for choosing the password.](https://docs.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/password-must-meet-complexity-requirements)
 
-To execute the `sysprep` command use:
+Please note that, the value of the `ComputerName` property in the [unattend.xml](unattend.xml) file must follows the
```
aravindhp, line 151: Is `property` the correct terminology for XML?

docs/unattend.xml
```diff
@@ -46,7 +46,7 @@
          <TimeZone>Eastern Standard Time</TimeZone>
          <UserAccounts>
             <AdministratorPassword>
-               <Value>MyPassword</Value>
+               <Value>WindowsPassword</Value>
```
aravindhp, line 49: Looking at the docs PR this seems like an unnecessary change.

docs/unattend.xml
```diff
@@ -16,14 +16,14 @@
          <CEIPEnabled>0</CEIPEnabled>
       </component>
       <component xmlns:wcm="http://schemas.microsoft.com/WMIConfig/2002/State" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" name="Microsoft-Windows-Shell-Setup" processorArchitecture="amd64" publicKeyToken="31bf3856ad364e35" language="neutral" versionScope="nonSxS">
-         <ComputerName>windows-host</ComputerName>
+         <ComputerName>winhost</ComputerName>
       </component>
    </settings>
    <settings pass="oobeSystem">
       <component xmlns:wcm="http://schemas.microsoft.com/WMIConfig/2002/State" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" name="Microsoft-Windows-Shell-Setup" processorArchitecture="amd64" publicKeyToken="31bf3856ad364e35" language="neutral" versionScope="nonSxS">
          <AutoLogon>
```
aravindhp, line 24: Is `AutoLogon` really needed? Please see https://github.com/openshift/windows-machine-config-operator/issues/579#issuecomment-896962856

aravindhp (changes requested):

docs/vsphere-golden-image.md
jrvaldes: Good catch; `element` it is. Still, simplicity will re-frame it as:

"Please note that, the `ComputerName` value in the [unattend.xml](unattend.xml) file must follows the..."

docs/vsphere-golden-image.md
jrvaldes: Correct, `WindowsPassword` is not a parameter, is a placeholder. Thanks for calling this out, will adjust.

docs/unattend.xml
jrvaldes: fixed in https://github.com/openshift/windows-machine-config-operator/pull/515/commits/90d680d529aa57c2585b9a641d1d80e5d0324399

jrvaldes:
/hold

for @openshift/openshift-team-windows-containers review

docs/unattend.xml
```diff
@@ -45,18 +41,12 @@
...
          <UserAccounts>
+            <!-- Prevents the built-in Administrator account from having a blank password by default. -->
             <AdministratorPassword>
-               <Value>WindowsPassword</Value>
+                <!-- Replace the `MyPassword` placeholder with the password for the Administrator account -->
+               <Value>MyPassword</Value>
                <PlainText>true</PlainText>
             </AdministratorPassword>
-            <LocalAccounts>
-               <LocalAccount wcm:action="add">
```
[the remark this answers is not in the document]
aravindhp: This implies that @alinaryan will have to retest with this changes.

saifshaikh48 (commented):
LGTM, though we may want to leave official /lgtm up to Docs folks.

docs/vsphere-golden-image.md
jrvaldes: #544 merged

docs/unattend.xml
```diff
@@ -16,18 +16,14 @@
...
       <component xmlns:wcm="http://schemas.microsoft.com/WMIConfig/2002/State" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" name="Microsoft-Windows-Shell-Setup" processorArchitecture="amd64" publicKeyToken="31bf3856ad364e35" language="neutral" versionScope="nonSxS">
          <AutoLogon>
-            <Password>
-               <Value>MyPassword</Value>
-               <PlainText>true</PlainText>
-            </Password>
-            <Enabled>true</Enabled>
-            <Username>Administrator</Username>
+            <!-- Disable automatic logon. This is the default value. -->
+            <Enabled>false</Enabled>
```
aravindhp, line 26: If this is the default, then why do we need this?

aravindhp (approved):
/approve

docs/unattend.xml
jrvaldes: It's safer to be explicit in this particular case.

aravindhp: Sure but why?

docs/unattend.xml
```diff
@@ -16,18 +16,14 @@
          <CEIPEnabled>0</CEIPEnabled>
       </component>
       <component xmlns:wcm="http://schemas.microsoft.com/WMIConfig/2002/State" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" name="Microsoft-Windows-Shell-Setup" processorArchitecture="amd64" publicKeyToken="31bf3856ad364e35" language="neutral" versionScope="nonSxS">
-         <ComputerName>windows-host</ComputerName>
```
aravindhp, line 19: @codyhoag please review this commit. We need this reflected in [Creating the vSphere Windows VM golden image](https://docs.openshift.com/container-platform/4.8/windows_containers/creating_windows_machinesets/creating-windows-machineset-vsphere.html#creating-the-vsphere-windows-vm-golden-image_creating-windows-machineset-vsphere)

docs/unattend.xml
jrvaldes: Given the number of issues/escalations received regarding the golden image creating process and, in particular, the security issue of having an open terminal with Administrator privileges at boot; led to make this default visible. 

It leaves an explicit record that the decision is being made, otherwise there will be no record the in the VCS, rather than the commit message, which is not exposed to the users consuming this file from the Openshift docs.

In addition, an explanatory comment (`Disable automatic logon` ) adds context and helps users to identify the proposed configuration; the default one.

docs/vsphere-golden-image.md
```diff
@@ -158,13 +165,21 @@ where `<path/to/unattend.xml>` is the path to the customized answer file.
...
+Next, you need to convert this virtual machine, in Power-Off status, to a Template as shown in Figure 1.
+
+|![Convert VM to Template](images/vcenter-vm-to-template.png)|
+|:---|
+|Figure 1. Steps to convert a VM to Template in vCenter|
 
-Add the cloned Windows virtual machine as a template in the [machineset](../README.md#configuring-windows-instances-provisioned-through-machinesets).
+## 8. Using the virtual machine template
+
+In order to use the recently created template, enter the template name in the [machineset](../../README.md#configuring-windows-instances-provisioned-through-machinesets).
```
codyhoag, line 182: The machineset link doesn't work. Perhaps it should point back to `../README.md#configuring-windows-instances-provisioned-through-machinesets`?

docs/unattend.xml
aravindhp: I am afraid `<!-- Disable automatic logon. This is the default value. -->` is not sufficient and you need to say more like you have in your above response. You need to call out the security issue as a reason to explicitly add this default value.

codyhoag (commented):
Just a few minor wording suggestions; looks good otherwise!

docs/unattend.xml
codyhoag: Done

docs/unattend.xml
jrvaldes: Makes senses, will adjust.

alinaryan:
Thanks for working on this, Jose! Only question I have is how do you recommend getting the helper scripts (i.e. the VM tools script and SSH install script) over to the VM given that copy/paste is disabled and scp can't be used until after ssh is configured in step 3?

jrvaldes:
> how do you recommend getting the helper scripts (i.e. the VM tools script and SSH install script) over to the VM given that copy/paste is disabled and scp can't be used until after ssh is configured in step 3?

That's is a good question; There are few walkarounds a) the [clipboard could be enabled](https://kb.vmware.com/s/article/57122) in the vCenter, b) the files could be downloaded into the Windows VM (using the `Invoke-WebRequest` powershell command, for example) or c) move the OpenSSH server installation to the top, allowing/suggesting the connectivity over SSH with the Windows VM from the very beginning.

Thoughts?

aravindhp:
Do we really need to give guidance around this? Plus you are assuming this is always vSphere so option A is out. Moreover this (getting files over to the VM) has not been the point of contention from customers but rather the steps themselves being wrong. We should focus on that.

alinaryan:
> Do we really need to give guidance around this? Plus you are assuming this is always vSphere so option A is out. Moreover this (getting files over to the VM) has not been the point of contention from customers but rather the steps themselves being wrong. We should focus on that.

Ok, I see your point on why we don't need to give guidance around that here, but I do like the idea of adding OpenSSH install step to the top. Could you make that change? @jrvaldes

docs/vsphere-golden-image.md
```diff
@@ -144,10 +144,17 @@ To deploy the Windows VM as a reusable image, you have to first generalize the V
...
 
-You can find an example [unattend.xml](unattend.xml) file in this repository. It is just an example and, it **should 
-not** be used directly. You must customize the example to fit your needs.
+A reference [unattend.xml](unattend.xml) file is provided, it **cannot** be used directly. You must customize it
+replacing the `MyPassword` placeholder with the desired password for the Administrator account, it prevents the built-in
+Administrator account from having a blank password by default. It's recommended to follow the [Microsoft best 
+practices for choosing the password.](https://docs.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/password-must-meet-complexity-requirements)
 
-To execute the `sysprep` command use:
+Please note that, the `ComputerName` value in the [unattend.xml](unattend.xml) file must follows the
```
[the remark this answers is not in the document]
jrvaldes: Fixed in https://github.com/openshift/windows-machine-config-operator/pull/515/commits/2a33a4a1762187c6005b5ee9176a5fb371d8b9c3

docs/vsphere-golden-image.md
jrvaldes: Fixed in https://github.com/openshift/windows-machine-config-operator/pull/515/commits/2a33a4a1762187c6005b5ee9176a5fb371d8b9c3

docs/vsphere-golden-image.md
```diff
@@ -71,14 +71,15 @@ See [Microsoft documentation](https://docs.microsoft.com/en-us/windows-server/ad
 to install using PowerShell.
 
 Alternatively, the provided PowerShell script [install-openssh.ps1](vsphere_ci/scripts/install-openssh.ps1) aims to 
-programmatically install and configure the OpenSSH Server with the public key file from section [Add the Administrator 
-Public Key](#add-the-administrator-public-key). For example, you can run the following command:
+programmatically install and configure the OpenSSH Server using the public key corresponding to the [private key](../README.md#create-a-private-key-secret)
+provided as part of WMCO setup. For example, you can run the following command:
```
aravindhp, line 75: I suggest removing, `For example` given you are changing this. This is not an example but rather the only command used. This key is also not provided as part of WMCO setup.

`programmatically install and configure the OpenSSH Server using the public key corresponding to the private key used in the [secret](../README.md#create-a-private-key-secret)`

This is also pointing at the fact that this file needs to be present on the VM, which leads to more [confusion](https://github.com/openshift/windows-machine-config-operator/pull/515#issuecomment-900546406) given SSH is not setup. So maybe we list the commands that they need to run here instead. But yes, they still need to figure out how to get the key over.

docs/vsphere-golden-image.md
```diff
@@ -71,14 +71,15 @@ See [Microsoft documentation](https://docs.microsoft.com/en-us/windows-server/ad
...
 
 ```
     ./install-openssh.ps1 <path/to/public_key_file>
 ```
 
-where, `<path/to/public_key_file>` is the path to the public key file corresponding to the [private key.](../README.md#create-a-private-key-secret)
+where, `<path/to/public_key_file>` is the path to the public key file used to enable key-based authentication in the
+OpenSSH Server.
```
aravindhp, line 82: Using the public key file does not enable key-based auth.

docs/vsphere-golden-image.md
```diff
@@ -101,26 +102,20 @@ In case no firewall rule exist, you must create it by running the following Powe
     New-NetFirewallRule -DisplayName 'OpenSSH Server (sshd)' -LocalPort 22 -Enabled True -Direction Inbound -Protocol TCP -Action Allow 
 ```
 
-### Add the Administrator Public Key
+### Set up key-based authentication
```
aravindhp, line 105: I suggest removing this section. It is not really adding any useful information here.

docs/vsphere-golden-image.md
```diff
@@ -101,26 +102,20 @@ In case no firewall rule exist, you must create it by running the following Powe
...
 
-This section provides an overview of how to use key-based authentication with OpenSSH server, similar to Linux 
-environments commonly setup of public-key/private-key pairs to drive authentication without passwords. Key pairs refer 
-to the public and private key files that are used by certain authentication protocols.
+To enable key-based authentication in the OpenSSH Server, you need a public/private key pair.
 
-**Important**: You need to have OpenSSH Server installed first. Please see [Set up SSH](#3-set-up-ssh).
+#### Using the private key secret
```
aravindhp, line 109: Same with this section given we have linked to the private key secret before

docs/vsphere-golden-image.md
```diff
@@ -101,26 +102,20 @@ In case no firewall rule exist, you must create it by running the following Powe
...
-#### Private key generation
-
-To use key-based authentication, you first need to generate public/private key pairs. 
-
-As the private key file let's reuse the exiting [the secret](../README.md#create-a-private-key-secret) created while 
-installing WMCO, along with the corresponding public key file.
+The creation of the private key is described as part of the WMCO setup, see [Create a private key secret](../README.md#create-a-private-key-secret)
+section.
 
 #### Deploying the public key
```
aravindhp, line 114: This section can be removed too.

aravindhp (changes requested):
@jrvaldes I suggest just having a section `Set up SSH` where everything is described. It is having these separate subsections that is causing confusion.

docs/vsphere-golden-image.md
```diff
@@ -68,17 +68,18 @@ the network interface and creates the `tools.conf` file.
 
 Install and configure the OpenSSH Server on the Windows virtual machine with automatic startup and key based-authentication.
```
jrvaldes, line 69: @alinaryan,  as per [comment](https://github.com/openshift/windows-machine-config-operator/pull/515#issuecomment-901179106), the **Set up SSH** section is the following step, after creating the Windows VM and configuring VMWare Tools. Both are platform' specific steps for vSphere.

codyhoag (approved):
LGTM!

docs/vsphere-golden-image.md
```diff
@@ -68,17 +68,18 @@ the network interface and creates the `tools.conf` file.
...
 See [Microsoft documentation](https://docs.microsoft.com/en-us/windows-server/administration/openssh/openssh_install_firstuse)
-to install using PowerShell.
+to install using PowerShell. Given we are setting up SSH access for an administrator, please follow the
+[Microsoft's documentation](https://docs.microsoft.com/en-us/windows-server/administration/openssh/openssh_keymanagement#administrative-user)
```
aravindhp, line 72: `Microsoft`. Yes this was an error I introduced but you should proof read 😉

aravindhp (approved):
/approve

with a minor change request.

docs/vsphere-golden-image.md
jrvaldes: Fixed

/cc @alinaryan

PR #517, opened 2021-07-15
[deploy] Fix e2e log collection by switching back to more descriptive WMCO label

sebsoto:
Changes the label 'control-plane: controller-manager', to
'name: windows-machine-config-operator'. This change happened as part of
the switch to the kubebuilder style repo, and has broken the WMCO pod
log collection in the e2e tests.
The monitor.yaml and auth_proxy_service.yaml are not included in the
bundle that we ship, but changing them in order to be consistent.


aravindhp (approved):

/approve

PR #518, opened 2021-07-15
[wmco] Patch node when adding annotations

saifshaikh48:
This PR opts to use a JSON Patch instead of Update when adding/removing 
annotations to node objects. Update calls can result in failures as the node 
object  can be a stale reference (if some other entity changes the Node 
while WMCO is configuring for example). A patch is best used when 
modifying parts of a resource, which fits the node annotation use case;
this is why a patch was favored over a Get -> Update loop.

This PR also introduces new packages to construct and format Patch data for arbitrary k8s resources.

### DIFF
--- a/controllers/secret_controller.go
+++ b/controllers/secret_controller.go
@@ -2,8 +2,6 @@ package controllers
 
 import (
 	"context"
-	"fmt"
-	"strings"
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
@@ -24,6 +22,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	"github.com/openshift/windows-machine-config-operator/pkg/annotations"
 	"github.com/openshift/windows-machine-config-operator/pkg/nodeconfig"
 	"github.com/openshift/windows-machine-config-operator/pkg/secrets"
 	"github.com/openshift/windows-machine-config-operator/pkg/signer"
@@ -171,8 +170,10 @@ func (r *SecretReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 			return reconcile.Result{}, errors.Wrapf(err, "error getting node list")
 		}
 		expectedPubKeyAnno := nodeconfig.CreatePubKeyHashAnnotation(keySigner.PublicKey())
-		escapedPubKeyAnnotation := strings.Replace(nodeconfig.PubKeyHashAnnotation, "/", "~1", -1)
-		patchData := fmt.Sprintf(`[{"op":"add","path":"/metadata/annotations/%s","value":""}]`, escapedPubKeyAnnotation)
+		patchData, err := annotations.GenerateAddPatch(map[string]string{nodeconfig.PubKeyHashAnnotation: ""})
+		if err != nil {
+			return reconcile.Result{}, errors.Wrapf(err, "error creating public key annotation add request")
+		}
 		for _, node := range nodes.Items {
 			// Only clear the annotation for Nodes associated with Machines, as the clearing of the annotation is used
 			// solely to kick off the deletion and recreation of Machines.
@@ -185,7 +186,7 @@ func (r *SecretReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 				continue
 			}
 			node.Annotations[nodeconfig.PubKeyHashAnnotation] = ""
-			err = r.client.Patch(ctx, &node, client.RawPatch(kubeTypes.JSONPatchType, []byte(patchData)))
+			err = r.client.Patch(ctx, &node, client.RawPatch(kubeTypes.JSONPatchType, patchData))
 			if err != nil {
 				return reconcile.Result{}, errors.Wrapf(err, "error clearing public key annotation on node %s",
 					node.GetName())
@@ -219,12 +220,14 @@ func (r *SecretReconciler) RemoveInvalidAnnotationsFromLinuxNodes(config *rest.C
 	}
 	// The public key hash was accidentally added to Linux nodes in WMCO 2.0 and must be removed.
 	// The `/` in the annotation key needs to be escaped in order to not be considered a "directory" in the path.
-	escapedPubKeyAnnotation := strings.Replace(nodeconfig.PubKeyHashAnnotation, "/", "~1", -1)
-	patchData := fmt.Sprintf(`[{"op":"remove","path":"/metadata/annotations/%s"}]`, escapedPubKeyAnnotation)
+	patchData, err := annotations.GenerateRemovePatch([]string{nodeconfig.PubKeyHashAnnotation})
+	if err != nil {
+		return errors.Wrapf(err, "error creating public key annotation add request")
+	}
 	for _, node := range nodes.Items {
 		if _, present := node.Annotations[nodeconfig.PubKeyHashAnnotation]; present == true {
 			_, err = kc.CoreV1().Nodes().Patch(context.TODO(), node.GetName(), kubeTypes.JSONPatchType,
-				[]byte(patchData), meta.PatchOptions{})
+				patchData, meta.PatchOptions{})
 			if err != nil {
 				return errors.Wrapf(err, "error removing public key annotation from node %s", node.GetName())
 			}
--- a/pkg/annotations/annotations.go
+++ b/pkg/annotations/annotations.go
@@ -1,0 +1,45 @@
+package annotations
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/openshift/windows-machine-config-operator/pkg/patch"
+)
+
+// generateAnnotationPatch creates a patch applying the given operation onto each given annotation key and value
+func generateAnnotationPatch(op string, annotations map[string]string) ([]byte, error) {
+	if len(annotations) == 0 {
+		return nil, errors.New("annotations to format cannot be empty or nil")
+	}
+	var patches []*patch.JSONPatch
+	for key, value := range annotations {
+		patches = append(patches, patch.NewJSONPatch(op, formatPathValue(key), value))
+	}
+	return json.Marshal(patches)
+}
+
+// GenerateAddPatch creates a comma-separated list of operations to add all given annotations from an object
+// An "add" patch overwrites existing value if an annotation already exists
+func GenerateAddPatch(annotations map[string]string) ([]byte, error) {
+	return generateAnnotationPatch("add", annotations)
+}
+
+// GenerateRemovePatch creates a comma-separated list of operations to remove all given annotations from an object
+// A "remove" patch fails transactionally if any of the annotations do not exist
+func GenerateRemovePatch(annotations []string) ([]byte, error) {
+	annotationMap := make(map[string]string)
+	for _, annotation := range annotations {
+		annotationMap[annotation] = ""
+	}
+	return generateAnnotationPatch("remove", annotationMap)
+}
+
+// formatPathValue formats the path value specifying which attribute a JSON Patch operation should be applied to
+func formatPathValue(key string) string {
+	// The `/` in the annotation key needs to be escaped in order to not be considered a "directory" in the path
+	escapedKey := strings.Replace(key, "/", "~1", -1)
+	return fmt.Sprintf("/metadata/annotations/%s", escapedKey)
+}
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -20,6 +20,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
 	"github.com/openshift/windows-machine-config-operator/pkg/nodeconfig"
+	"github.com/openshift/windows-machine-config-operator/pkg/patch"
 )
 
 //+kubebuilder:rbac:groups="",resources=services;services/finalizers,verbs=create;get;delete
@@ -66,16 +67,6 @@ type Config struct {
 	namespace string
 	// recorder to generate events
 	recorder record.EventRecorder
-}
-
-// patchEndpoint contains information regarding patching metrics Endpoint
-type patchEndpoint struct {
-	// op defines patch operation to be performed on the Endpoints object
-	Op string `json:"op"`
-	// path defines the location of the patch
-	Path string `json:"path"`
-	// value defines the data to be patched
-	Value []v1.EndpointSubset `json:"value"`
 }
 
 // NewPrometheuopsNodeConfig creates a new instance for prometheusNodeConfig  to be used by the caller.
@@ -125,11 +116,7 @@ func (pc *PrometheusNodeConfig) syncMetricsEndpoint(nodeEndpointAdressess []v1.E
 		}}
 	}
 
-	patchData := []patchEndpoint{{
-		Op:    "replace",
-		Path:  "/subsets",
-		Value: subsets,
-	}}
+	patchData := []*patch.JSONPatch{patch.NewJSONPatch("replace", "/subsets", subsets)}
 	// convert patch data to bytes
 	patchDataBytes, err := json.Marshal(patchData)
 	if err != nil {
--- a/pkg/nodeconfig/nodeconfig.go
+++ b/pkg/nodeconfig/nodeconfig.go
@@ -21,6 +21,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	crclientcfg "sigs.k8s.io/controller-runtime/pkg/client/config"
 
+	"github.com/openshift/windows-machine-config-operator/pkg/annotations"
 	"github.com/openshift/windows-machine-config-operator/pkg/cluster"
 	"github.com/openshift/windows-machine-config-operator/pkg/instances"
 	"github.com/openshift/windows-machine-config-operator/pkg/metadata"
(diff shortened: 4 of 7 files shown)


pkg/nodeconfig/nodeconfig.go
```diff
@@ -172,14 +173,12 @@ func (nc *nodeConfig) Configure() error {
...
-		nc.addPubKeyHashAnnotation()
-		node, err := nc.k8sclientset.CoreV1().Nodes().Update(context.TODO(), nc.node, meta.UpdateOptions{})
-		if err != nil {
-			return errors.Wrapf(err, "error updating public key hash and additional annotations on node %s",
-				nc.node.GetName())
+		if err := nc.addAdditionalAnnotations(); err != nil {
+			return errors.Wrapf(err, "error updating additional annotations on node %s", nc.node.GetName())
+		}
+		if err := nc.addAnnotation(PubKeyHashAnnotation, nc.publicKeyHash); err != nil {
+			return errors.Wrapf(err, "error updating public key hash on node %s", nc.node.GetName())
```
sebsoto, line 180: Can this be done as one operation? I'd think we'd like to avoid increasing the amount of API calls as much as possible

pkg/nodeconfig/nodeconfig.go
```diff
@@ -255,25 +251,30 @@ func (nc *nodeConfig) configureNetwork() error {
...
-func (nc *nodeConfig) addAdditionalAnnotations() {
-	if nc.additionalAnnotations == nil {
-		return
-	}
+func (nc *nodeConfig) addAdditionalAnnotations() error {
 	for key, value := range nc.additionalAnnotations {
-		nc.node.Annotations[key] = value
+		if err := nc.addAnnotation(key, value); err != nil {
+			return err
+		}
```
sebsoto, line 260: Same idea as previous comment

sebsoto (changes requested):

pkg/nodeconfig/nodeconfig.go
```diff
@@ -172,14 +173,10 @@ func (nc *nodeConfig) Configure() error {
...
 		// Ensure we are annotating the node as soon as the Node object is created, so that we can identify which
 		// controller should be watching it
-		nc.addAdditionalAnnotations()
-		nc.addPubKeyHashAnnotation()
-		node, err := nc.k8sclientset.CoreV1().Nodes().Update(context.TODO(), nc.node, meta.UpdateOptions{})
-		if err != nil {
-			return errors.Wrapf(err, "error updating public key hash and additional annotations on node %s",
-				nc.node.GetName())
+		nc.additionalAnnotations[PubKeyHashAnnotation] = nc.publicKeyHash
+		if err := nc.addAnnotations(nc.additionalAnnotations); err != nil {
```
mansikulkarni96, line 177: why do we need to pass in nc.additionalAnnotations here? can't we just retrieve it at the call site from the struct, as it is already assigned a value.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -255,25 +249,27 @@ func (nc *nodeConfig) configureNetwork() error {
...
+		patchData += fmt.Sprintf(`{"op":"add","path":"/metadata/annotations/%s","value":"%s"},`,
+			escapedAnnotation, value)
 		nc.node.Annotations[key] = value
 	}
+	// Replace last character (extra comma) with a closing brace
+	patchData = string(patchData[:len(patchData)-1]) + "]"
+	node, err := nc.k8sclientset.CoreV1().Nodes().Patch(context.TODO(), nc.node.GetName(), kubeTypes.JSONPatchType,
+		[]byte(patchData), meta.PatchOptions{})
+	if err != nil {
+		return errors.Wrapf(err, "error updating annotations on node %s", nc.node.GetName())
```
mansikulkarni96, line 269: maybe add the patch data to the error here or add a log somewhere that helps us to debug this format?

mansikulkarni96 (changes requested):
PTAL at my comments.

pkg/nodeconfig/nodeconfig.go
saifshaikh48: `addAnnotations` is more of a helper that just handles what is passed in instead of retrieving annotations from `nc` as the function is used multiple times for different annotations.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -255,25 +249,40 @@ func (nc *nodeConfig) configureNetwork() error {
...
-func (nc *nodeConfig) addVersionAnnotation() {
-	nc.node.Annotations[VersionAnnotation] = version.Get()
-}
-
-// addPubKeyHashAnnotation adds the public key annotation to nc.node
-func (nc *nodeConfig) addPubKeyHashAnnotation() {
-	nc.node.Annotations[PubKeyHashAnnotation] = nc.publicKeyHash
+// addIdentifyingAnnotations adds the public key hash annotation as well as any additional annotations
+// that could help identify what type of node is being configured
+func (nc *nodeConfig) addIdentifyingAnnotations() error {
```
aravindhp, line 254: This obfuscates what we are doing and a bit confusing. `could help` implies it may or may not help. I would rather see `nc.addAnnotations()` being called at the call site with a map of annotations and this function removed.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -255,25 +249,40 @@ func (nc *nodeConfig) configureNetwork() error {
...
+	}
+	return nc.addAnnotations(annotationsToAdd)
 }
 
-// addAdditionalAnnotations merges nc.additionalAnnotations into the annotations on nc.node. If the annotation
-// already existed, its value will be overwritten.
-func (nc *nodeConfig) addAdditionalAnnotations() {
-	if nc.additionalAnnotations == nil {
-		return
+// addAnnotations adds all the given annotations to node object and updates the nodeconfig
```
aravindhp, line 262: Please name this `applyAnnotation` to be explicit as to what the method does and it syncs with k8s terminology. 

`updates the Node object in NodeConfig`

aravindhp (changes requested):
Thanks for the quick turnaround with this, @saifshaikh48. Please address my comments.

pkg/nodeconfig/nodeconfig.go
saifshaikh48: Agreed. I have addressed this.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -255,25 +253,32 @@ func (nc *nodeConfig) configureNetwork() error {
...
-// addAdditionalAnnotations merges nc.additionalAnnotations into the annotations on nc.node. If the annotation
-// already existed, its value will be overwritten.
-func (nc *nodeConfig) addAdditionalAnnotations() {
-	if nc.additionalAnnotations == nil {
-		return
+// applyAnnotations applies all the given annotations and updates the Node object in NodeConfig
+// If an annotation already exists, its value will be overwritten.
+func (nc *nodeConfig) applyAnnotations(annotations map[string]string) error {
+	if annotations == nil {
+		return errors.New("expected at least one annotation to apply but received none")
```
jrvaldes, line 260:
```suggestion
		return errors.New("annotations cannot be nil")
```

pkg/nodeconfig/nodeconfig.go
```diff
@@ -255,25 +253,32 @@ func (nc *nodeConfig) configureNetwork() error {
...
+	patchData := "["
+	for key, value := range annotations {
+		// The `/` in the annotation key needs to be escaped in order to not be considered a "directory" in the path.
+		escapedAnnotation := strings.Replace(key, "/", "~1", -1)
+		patchData += fmt.Sprintf(`{"op":"add","path":"/metadata/annotations/%s","value":"%s"},`,
+			escapedAnnotation, value)
 		nc.node.Annotations[key] = value
 	}
+	// Replace last character (extra comma) with a closing brace
+	patchData = string(patchData[:len(patchData)-1]) + "]"
```
jrvaldes, line 274: `string` concatenations is usually slow, and may lead to syntactical error, like maintaining the comma (`,`) at the end of each annotation json object, or keeping track of open brackets (`patchData := "["`).

Consider, collecting the annotation items in a list, and then join the items with the respective separator (`,`), prefix (`[`) and suffix (`]`)

pkg/nodeconfig/nodeconfig.go
```diff
@@ -255,25 +253,32 @@ func (nc *nodeConfig) configureNetwork() error {
...
+	node, err := nc.k8sclientset.CoreV1().Nodes().Patch(context.TODO(), nc.node.GetName(), kubeTypes.JSONPatchType,
```
jrvaldes, line 275: I feel `nc.k8sclientset.CoreV1().Nodes().Patch(...)` can be placed in a function by it's own (e.g. `patchNodes(...)`), 
a) can be reused
b) leaves room for `applyAnnotations()` becoming `getAnnotations()` and returning a `patchData` that can be unit tested, without the need of mock/stub the `nc.k8sclientset.CoreV1().Nodes().Patch()` call.

jrvaldes (changes requested):
@saifshaikh48  thanks for working on this.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -255,25 +253,32 @@ func (nc *nodeConfig) configureNetwork() error {
...
-	for key, value := range nc.additionalAnnotations {
+	// Accumulate the string representation of a JSON-formatted list of annotation requests
+	// For example: [{"op":"add","path":"/metadata/annotations/foo","value":"bar"},
+	//               {"op":"add","path":"/metadata/annotations/biz","value": "baz"}]
+	patchData := "["
+	for key, value := range annotations {
+		// The `/` in the annotation key needs to be escaped in order to not be considered a "directory" in the path.
+		escapedAnnotation := strings.Replace(key, "/", "~1", -1)
+		patchData += fmt.Sprintf(`{"op":"add","path":"/metadata/annotations/%s","value":"%s"},`,
+			escapedAnnotation, value)
```
sebsoto, line 270: This logic is being done in multiple places in  our codebase, It might be best to break it off into a separate function in a new package.

sebsoto (changes requested):

pkg/nodeconfig/nodeconfig.go
aravindhp: >I feel nc.k8sclientset.CoreV1().Nodes().Patch(...) can be placed in a function by it's own (e.g. patchNodes(...)),

This is just creating a wrapper that will need to be provided the same set of inputs needed for `Patch()` which now needs to be input validated. The call site is going to still going to need to check for errors. I see very little advantage in doing this.

pkg/nodeconfig/nodeconfig.go
saifshaikh48: Thanks for the suggestion. I ended up going going this route.

pkg/nodeconfig/nodeconfig.go
saifshaikh48: Annotations pkg created.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -172,14 +173,14 @@ func (nc *nodeConfig) Configure() error {
...
 		// controller should be watching it
-		nc.addAdditionalAnnotations()
-		nc.addPubKeyHashAnnotation()
-		node, err := nc.k8sclientset.CoreV1().Nodes().Update(context.TODO(), nc.node, meta.UpdateOptions{})
-		if err != nil {
+		annotationsToApply := map[string]string{PubKeyHashAnnotation: nc.publicKeyHash}
+		for key, value := range nc.additionalAnnotations {
+			annotationsToApply[key] = value
+		}
+		if err := nc.applyAnnotations(annotationsToApply); err != nil {
```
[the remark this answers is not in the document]
saifshaikh48: If you mean making two calls `applyAnnotations(map[string]string{PubKeyHashAnnotation: nc.publicKeyHash})` and `applyAnnotations(nc.additionalAnnotations)` -- this is avoided as `applyAnnotations` makes an API call. 

I don't see a way around this (I would rather process the annotation map 2x than make 2 API calls).

pkg/annotations/annotations.go
```diff
@@ -0,0 +1,55 @@
...
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
+)
+
+// Represents a single operation in a JSON Patch request body
+// Examples: {"op":"add","path":"/metadata/annotations/foo","value":"bar"}
+//           {"op":"remove","path":"/metadata/annotations/baz"}
+type patchOperation map[string]string
```
sebsoto, line 13: I feel like what you are trying to do here is better achieved by using a struct. I dont see any benefit in defining the `patchOperation` in this way.

sebsoto (changes requested):

pkg/annotations/annotations.go
saifshaikh48: The reasoning behind keeping it more fluid was that `remove` operations do not have the `value` field.

Do you think creating 2 separate structs, one for `add` operations and one for `remove` would be more clear?

alinaryan: I like the 2 struct idea, unless @sebsoto has another suggestion

mansikulkarni96: maybe you could reuse the patching we do here using [struct](https://github.com/openshift/windows-machine-config-operator/blob/892d1850ad645f57097c670a2560c397f3f9a244/pkg/metrics/metrics.go#L72) in the metrics pkg.

test/e2e/reconfigure_test.go
```diff
@@ -28,14 +27,13 @@ func reconfigurationTest(t *testing.T) {
 	require.NoError(t, err)
 
 	// Remove the version annotation of one of each type of node
-	// Forward slash within a path is escaped as '~1'
-	escapedVersionAnnotation := strings.Replace(nc.VersionAnnotation, "/", "~1", -1)
-	patchData := fmt.Sprintf("[{\"op\": \"remove\", \"path\": \"/metadata/annotations/%s\"}]", escapedVersionAnnotation)
+	patchData, err := annotations.GenerateRemovePatch([]string{nc.VersionAnnotation})
```
aravindhp, line 30: If you want to use this in test code, please add a unit test for the function

controllers/secret_controller.go
```diff
@@ -171,8 +170,10 @@ func (r *SecretReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 			return reconcile.Result{}, errors.Wrapf(err, "error getting node list")
 		}
 		expectedPubKeyAnno := nodeconfig.CreatePubKeyHashAnnotation(keySigner.PublicKey())
-		escapedPubKeyAnnotation := strings.Replace(nodeconfig.PubKeyHashAnnotation, "/", "~1", -1)
-		patchData := fmt.Sprintf(`[{"op":"add","path":"/metadata/annotations/%s","value":""}]`, escapedPubKeyAnnotation)
```
aravindhp, line 175: Please make the introduction of `GenerateAddPatch` a separate commit.

Replace the `Update()` with a `Patch()` should be a follow on commit.

pkg/annotations/annotations.go
aravindhp: +1 to @mansikulkarni96 comment. We should have one method for of patching in our code. 

Moreover, standalone functions like these need unit tests.

aravindhp (changes requested):
Thanks for doing this, @saifshaikh48. Please address my comments.

pkg/annotations/annotations.go
saifshaikh48: I am not clear on we could consolidate things to one Patch method. The patching done the metrics pkg is a different operation (`replace`) on a different k8s object (`Endpoints` vs. `Nodes`), so the patch body will be different. At minimum there would have to be separate method to process the each type of object.

test/e2e/reconfigure_test.go
saifshaikh48: Added (in other commit).

aravindhp (approved):

/approve

alinaryan:
Nice work! 
/lgtm

controllers/secret_controller.go
```diff
@@ -218,13 +219,15 @@ func (r *SecretReconciler) RemoveInvalidAnnotationsFromLinuxNodes(config *rest.C
 		return errors.Wrap(err, "error getting Linux node list")
 	}
 	// The public key hash was accidentally added to Linux nodes in WMCO 2.0 and must be removed.
-	// The `/` in the annotation key needs to be escaped in order to not be considered a "directory" in the path.
-	escapedPubKeyAnnotation := strings.Replace(nodeconfig.PubKeyHashAnnotation, "/", "~1", -1)
-	patchData := fmt.Sprintf(`[{"op":"remove","path":"/metadata/annotations/%s"}]`, escapedPubKeyAnnotation)
+	patchData, err := annotations.GenerateRemovePatch([]string{nodeconfig.PubKeyHashAnnotation})
+	if err != nil {
+		return errors.Wrapf(err, "error creating public key annotation add request")
+
```
sebsoto, line 225: extra newline

pkg/metrics/metrics.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/openshift/windows-machine-config-operator/pkg/patch"
```
sebsoto, line 7: This should be in the third import group

pkg/annotations/annotations.go
```diff
@@ -0,0 +1,46 @@
...
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/openshift/windows-machine-config-operator/pkg/patch"
+)
+
+// generateAnnotationPatch creates a comma-separated list of operations to add all given annotations from an object
+// An "add" patch overwrites existing value if an annotation already exists
```
sebsoto, line 13: Doc needs to be changed, it was copied from `GenerateAddPatch`

sebsoto (changes requested):
@saifshaikh48 just a few formatting things, otherwise lgtm

sebsoto:
/hold
can be removed when formatting is fixed

PR #519, opened 2021-07-16
WIP: WINC-512: Ensure APIServerInternal IP address at configuration

jrvaldes:
This commit adds the ensureAPIServerInternal() function to the windows
package, in order to ensure the IP address of the APIServerInternal
while configuring the VM.

If the APIServerInternal is available with an IP address, it's a no-op.

Otherwise, while no valid IP address resolution for APIServerInternal, the
APIServerInternal IP address is discovered from the cluster infrastructure,
specifically from the platform status, and included as a DNS alias entry
in the Windows hosts file.

Supported platform types: vsphere

### DIFF
--- a/pkg/windows/windows.go
+++ b/pkg/windows/windows.go
@@ -1,16 +1,22 @@
 package windows
 
 import (
+	"context"
 	"fmt"
+	"net/url"
 	"path/filepath"
 	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
+	config "github.com/openshift/api/config/v1"
+	clientset "github.com/openshift/client-go/config/clientset/versioned"
 	"github.com/pkg/errors"
 	"golang.org/x/crypto/ssh"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	ctrl "sigs.k8s.io/controller-runtime"
+	crclientcfg "sigs.k8s.io/controller-runtime/pkg/client/config"
 
 	"github.com/openshift/windows-machine-config-operator/pkg/instances"
 	"github.com/openshift/windows-machine-config-operator/pkg/nodeconfig/payload"
@@ -316,6 +322,9 @@ func (vm *windows) Configure() error {
 	if err := vm.EnsureRequiredServicesStopped(); err != nil {
 		return errors.Wrap(err, "unable to stop required services")
 	}
+	if err := vm.ensureAPIServerInternal(); err != nil {
+		return errors.Wrap(err, "error resolving APIServerInternal IP address")
+	}
 	// Set the hostName of the Windows VM if needed
 	if vm.hostName != "" {
 		if err := vm.ensureHostName(); err != nil {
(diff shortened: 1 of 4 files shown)


pkg/windows/windows.go
```diff
@@ -1,16 +1,22 @@
...
 import (
+	"context"
 	"fmt"
+	"net/url"
 	"path/filepath"
 	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
+	config "github.com/openshift/api/config/v1"
```
jrvaldes, line 12: found several ways to name this import reference in the code base, for example [oconfig](https://github.com/openshift/windows-machine-config-operator/blob/c2ec776ed0474386311f400b20c2cf07d926c990/controllers/windowsmachine_controller.go#L9) and [config](https://github.com/openshift/windows-machine-config-operator/blob/26fa97a5f3822f0f3eb1787707e8996b02a984ce/test/e2e/create_test.go#L11). 

What would be the consensus here?

pkg/windows/windows.go
```diff
@@ -335,6 +344,203 @@ func (vm *windows) Configure() error {
...
+	// flush DNS cache
+	if err := vm.clearDnsClientCache(); err != nil {
+		return err
+	}
+	vm.log.Info("Finding API server internal IP address for hostname",
+		"hostname", apiServerInternalHostname,
+		"platformType", status.PlatformStatus.Type)
+	// check existing DNS resolution for apiServerInternalHostname including
+	// existing hosts entries that may be pre-configured in the golden image
+	ipAddresses, err := vm.resolveIpAddresses(apiServerInternalHostname, false)
```
jrvaldes, line 502: Does it makes sense to ignore any existing hosts file entries, that may be pre-configured in the golden image?

pkg/windows/windows.go
```diff
@@ -335,6 +344,203 @@ func (vm *windows) Configure() error {
...
+		return "", errors.Errorf("cannot to parse URL: %s", urlString)
+	}
+	// return hostname
+	return u.Hostname(), nil
+}
+
+// resolveIpAddresses returns a list of strings representing the IP addresses
+// associated with the given hostname. If noHostsFile==true the entries in the
+// Windows hosts file are skipped when resolving the DNS query
+func (vm *windows) resolveIpAddresses(hostname string, noHostsFile bool) ([]string, error) {
```
jrvaldes, line 397: the `noHostsFile` param can be removed, depending on [this](https://github.com/openshift/windows-machine-config-operator/pull/519/files#diff-4d8fe3bb66a872f0e4a277f820d60d2efa722f09aa89516be74b98c06b57f3bdR502)

pkg/windows/windows.go
```diff
@@ -335,6 +344,203 @@ func (vm *windows) Configure() error {
...
+		return platformStatus.VSphere.APIServerInternalIP, nil
+		// TODO: uncomment as platforms become supported
+		// case config.BareMetalPlatformType:
+		// 	return platformStatus.BareMetal.APIServerInternalIP, nil
+		// case config.OpenStackPlatformType:
+		// 	return platformStatus.OpenStack.APIServerInternalIP, nil
+		// case config.OvirtPlatformType:
+		// 	return platformStatus.Ovirt.APIServerInternalIP, nil
+		// case config.KubevirtPlatformType:
+		// 	return platformStatus.Kubevirt.APIServerInternalIP, nil
```
jrvaldes, line 470: [YAGNI](https://martinfowler.com/bliki/Yagni.html) Alert!

the `APIServerInternalIP` is available for all of the above `platformStatus.Type`s, currently only `VSphere` is supported in WMCO. 

Happy to remove the comments. Just wanted to run this by the team.

jrvaldes (commented):
General considerations:
- no fixed/hardcoded APIServerInternal endpoint URL
- the APIServerInternal URL endpoint is loaded from the platform status
- the APIServerInternal IP address is loaded from the platform status (if available)
- the DNS client flushing could be remove, in case we trust the golden image

Follow up:
- the discovery of the APIServerInternal IP address currently occurs on every VM configuration, [need to be generalized](https://coreos.slack.com/archives/GLKK7Q40J/p1625694162007800?thread_ts=1625687987.006400&cid=GLKK7Q40J).

PR #520, opened 2021-07-16
Bug 1982991: Remove Service account file from bundle manifests

mansikulkarni96:
This PR removes the serviceaccount.yaml file from bundle manifests since it contains the same information as
held by the CSV spec for SA generation.
This causes an issue with the operator during upgrade with the CSV requirements not being met and the message: "Service account is owned by another ClusterServiceVersion"


jrvaldes:
Consider including `Fixes BZ#1982991`  in the commit message

aravindhp (approved):

/approve

PR #527, opened 2021-07-19
WINC-612: [wmco] Handle deletion of windows-instances

aravindhp:
If the admin deletes the `windows-instances` ConfigMap, the operation is viewed as a call to deconfigure all BYOH Windows instances. This PR handles that scenario.

### DIFF
--- a/controllers/configmap_controller.go
+++ b/controllers/configmap_controller.go
@@ -54,7 +54,6 @@ const (
 	// UsernameAnnotation is a node annotation that contains the username used to log into the Windows instance
 	UsernameAnnotation = "windowsmachineconfig.openshift.io/username"
 	// InstanceConfigMap is the name of the ConfigMap where VMs to be configured should be described.
-	// TODO: Possibly make this a singleton that WMCO creates https://issues.redhat.com/browse/WINC-612
 	InstanceConfigMap = "windows-instances"
 )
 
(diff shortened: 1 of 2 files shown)


controllers/configmap_controller.go
```diff
@@ -282,6 +279,13 @@ func (r *ConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		UpdateFunc: func(e event.UpdateEvent) bool {
 			return r.isValidConfigMap(e.ObjectNew)
 		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			// If DeleteStateUnknown is true it implies that the Delete event was missed  and we can ignore it
+			if e.DeleteStateUnknown {
+				return false
+			}
```
sebsoto, line 286: In this case why should we skip the reconciliation? We should still ensure that everything is as expected should we not?

aravindhp: I went with the documented [example](https://sdk.operatorframework.io/docs/building-operators/golang/references/event-filtering/#using-predicates). Not sure if we should be reacting to objects that have not been fully deleted.

aravindhp:
/retitle WINC-612: [wmco] Handle deletion of windows-instances

controllers/configmap_controller.go
sebsoto: If we are definitely going to get a second event when the object is fully deleted that makes sense to me

aravindhp: So should I remove this check or leave it in?

sebsoto: I was thinking that it would be fine to leave the check, but having looked at the [controller-runtime code](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/event/event.go#L44-L45) I am not so sure.

It being true does not mean that it hasnt been fully deleted, but that the event was missed. 
```
	// DeleteStateUnknown is true if the Delete event was missed but we identified the object
	// as having been deleted.
```

Digging more into how the delete event is actually generated:
* How we create a controller: https://github.com/kubernetes-sigs/controller-runtime/blob/865d56a824cc295f4895aca40af48737e188bd0c/alias.go#L101
* When we call `.Watches(...)` on a controller we declare that we want to watch for events for a specific object: https://github.com/kubernetes-sigs/controller-runtime/blob/865d56a824cc295f4895aca40af48737e188bd0c/pkg/source/source.go#L88
* The internal.EventHandler is always used: https://github.com/kubernetes-sigs/controller-runtime/blob/865d56a824cc295f4895aca40af48737e188bd0c/pkg/source/source.go#L133 
* Delete events are handled and DeleteStateUnknown set: https://github.com/kubernetes-sigs/controller-runtime/blob/865d56a824cc295f4895aca40af48737e188bd0c/pkg/source/internal/eventsource.go#L101-L105
```
	// Deal with tombstone events by pulling the object out.  Tombstone events wrap the object in a
	// DeleteFinalStateUnknown struct, so the object needs to be pulled out.
	// Copied from sample-controller
	// This should never happen if we aren't missing events, which we have concluded that we are not
	// and made decisions off of this belief.  Maybe this shouldn't be here?
```
* And finally the DeleteFinalStateUnknown definition: https://github.com/kubernetes/client-go/blob/e337077829155da74648d511bfa560a148fd965d/tools/cache/delta_fifo.go#L740-L743
```
// DeletedFinalStateUnknown is placed into a DeltaFIFO in the case where an object
// was deleted but the watch deletion event was missed while disconnected from
// apiserver. In this case we don't know the final "resting" state of the object, so
// there's a chance the included `Obj` is stale.
```

What I'm getting from this is that the issue with handling these events is the given object might be stale. That shouldn't be an issue, as we retrieve the current state of the object when handling the event. It does not look like we will get another delete event after this one. I think its safer to not filter these out. 
The docs seem to be misleading. Are you getting the same conclusion as me from the code?

aravindhp: @sebsoto I have removed the check. PTAL.

PR #537, opened 2021-07-23
[release-4.7] Bug 1985349: Remove SA file from bundle manifests

mansikulkarni96:
This PR removes the serviceaccount.yaml file from bundle manifests since it contains the same information as
held by the CSV spec for SA generation.
This causes an issue with the operator during upgrade with the CSV requirements not being met and the message
"Service account is owned by another ClusterServiceVersion"


saifshaikh48 (commented):
Should we include `Fixes bz#1985349` in the commit message?

PR #540, opened 2021-07-26
[wmco] Further modularize BYOH instance configuration

sebsoto:
This commit refactors the reconcileNodes() function present in the
ConfigMap controller, by breaking out instance configuration logic into
a separate ensureInstancesAreUpToDate() function.

### DIFF
--- a/controllers/configmap_controller.go
+++ b/controllers/configmap_controller.go
@@ -180,19 +180,10 @@ func (r *ConfigMapReconciler) reconcileNodes(ctx context.Context, windowsInstanc
 		return errors.Wrap(err, "unable to parse hosts from ConfigMap")
 	}
 
-	// For each host, ensure that it is configured into a node. On error of any host joining, return error and requeue.
-	// It is better to return early like this, instead of trying to configure as many nodes as possible in a single
-	// reconcile call, as it simplifies error collection. The order the map is read from is psuedo-random, so the
-	// configuration effort for configurable hosts will not be blocked by a specific host that has issues with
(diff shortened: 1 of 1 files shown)


controllers/configmap_controller.go
```diff
@@ -206,6 +197,22 @@ func (r *ConfigMapReconciler) reconcileNodes(ctx context.Context, windowsInstanc
 	return nil
 }
 
+// configureInstances configures all instances that require configuration
+func (r *ConfigMapReconciler) configureInstances(instances []*instances.InstanceInfo) error {
+	for _, instance := range instances {
+		err := r.ensureInstanceIsUpToDate(instance, map[string]string{BYOHAnnotation: "true",
```
mansikulkarni96, line 203: If configureInstance is calling ensureInstanceIsUpToDate, wouldn't it end up calling deconfigure if upgrade is required. Correct me if I am wrong, my thinking was we had introduced ensureInstanceIsUpToDate to avoid this issue in the previous implementation.

aravindhp (approved):

/approve

controllers/configmap_controller.go
mansikulkarni96: @sebsoto I can lgtm once the function is renamed to what we discussed during standup.

sebsoto: Thanks @mansikulkarni96 I have made the change

mansikulkarni96:
Thanks for the modularization @sebsoto 
/lgtm

PR #545, opened 2021-07-28
[docs] Update vSphere machineSet requirement

jrvaldes:
Add network name requirements in the machineSet example
for vSphere, since must match the network name where other
Linux workers are in the cluster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ oc get -o jsonpath='{.status.infrastructureName}{"\n"}' infrastructure cluster
 ```
 The following template variables need to be replaced as follows with values from your vSphere environment:
 * *\<Windows_VM_template\>*: template name
-* *\<VM Network Name\>*: network name
+* *\<VM Network Name\>*: network name, must match the network name where other Linux workers are in the cluster
 * *\<vCenter DataCenter Name\>*: datacenter name
 * *\<Path to VM Folder in vCenter\>*: path where your OpenShift cluster is running
 * *\<vCenter Datastore Name\>*: datastore name


README.md
```diff
@@ -101,14 +101,14 @@ oc get -o jsonpath='{.status.infrastructureName}{"\n"}' infrastructure cluster
...
-* *\<VM Network Name\>*: network name
+* *\<VM Network Name\>*: network name, must match the network name where other Linux workers are in the cluster
 * *\<vCenter DataCenter Name\>*: datacenter name
 * *\<Path to VM Folder in vCenter\>*: path where your OpenShift cluster is running
 * *\<vCenter Datastore Name\>*: datastore name
 * *\<vCenter Server FQDN/IP\>*: IP address or FQDN of the vCenter server
 
 *IMPORTANT*:
-- The template used in the MachineSet must be a Windows Server 1909
+- The template used in the MachineSet must be a Windows Server Semi-Annual Channel (SAC): Windows Server 2004
```
aravindhp, line 111: Please mention you are making this change in the commit message and reframe it.

aravindhp (changes requested):
Thanks for doing this, @jrvaldes. Please follow the [format of the commit messages.](https://github.com/openshift/windows-machine-config-operator/blob/master/CONTRIBUTION.md#format-of-the-commit-message)

aravindhp (changes requested):
None of comments have been addressed. `[docs] Describe the network name requirements in the machineset` implies that is the only change you are making. It needs to read something like `[docs] Update vSphere MachineSet requirement`. And you did not fix the 50/80 issue in the comment.

aravindhp (approved):
/approve

mansikulkarni96:
Thanks for fixing this @jrvaldes 
/lgtm

EricPonvelle:
/lgtm

[I created this PR](https://github.com/openshift/openshift-docs/pull/35036) to reflect these changes in the documentation; however, with the WMCO content currently hidden, I think the preview may not render correctly.

README.md
```diff
@@ -101,14 +101,14 @@ oc get -o jsonpath='{.status.infrastructureName}{"\n"}' infrastructure cluster
 ```
 The following template variables need to be replaced as follows with values from your vSphere environment:
 * *\<Windows_VM_template\>*: template name
-* *\<VM Network Name\>*: network name
+* *\<VM Network Name\>*: network name, must match the network name where other Linux workers are in the cluster
```
aravindhp, line 104: I just realized we need to call this out for all MachineSets. And yes, in this PR and not a follow up one. Just reframe the PR and make it a generic update.

aravindhp (changes requested):

README.md
jrvaldes: makes sense

aravindhp: Looking more closely, we seem to have explained this correctly for other `MachineSets`. However for Azure for `image`, we should just say `<image> should be a WindowsServer image offering that defines the 2019-Datacenter-with-Containers SKU`. All verbiage regarding the issue we ran into before needs to be removed. The Azure change however has been reflected in the OpenShift docs so @EricPonvelle does not need to update anything.

docs/machineset-azure.md
```diff
@@ -9,20 +17,21 @@ oc get -o jsonpath='{.status.infrastructureName}{"\n"}' infrastructure cluster
...
+`<image>` <image> should be a WindowsServer image offering that defines the 2019-Datacenter-with-Containers SKU
+
+Run the following command to list compatible Azure images:
 ```shell script
   az vm image list --all --location <location> \
       --publisher MicrosoftWindowsServer \
       --offer WindowsServer \
-      --sku 2019-Datacenter-with-Containers \
-      --query "[?contains(version, '17763.1457.2009030514')]"
+      --sku 2019-Datacenter-with-Containers
```
aravindhp, line 27: Did you run this command and verify that it does what it says? Moreover we only added it because of the version, so I suggest removing it.

docs/machineset-azure.md
```diff
@@ -9,20 +17,21 @@ oc get -o jsonpath='{.status.infrastructureName}{"\n"}' infrastructure cluster
...
 ```
-This is to work around Windows containers behind a Kubernetes load balancer
-becoming unreachable [issue](https://github.com/microsoft/Windows-Containers/issues/78).
 
 Please note that on Azure, Windows Machine names cannot be more than 15 characters long.
 The MachineSet name can therefore not be more than 9 characters long, due to the way Machine names are generated from it.
+
+## Sample YAML for a Windows MachineSet object on Azure
+
```
aravindhp, line 34: Same comment as before.

aravindhp (changes requested):

docs/machineset-azure.md
jrvaldes: > Did you run this command and verify that it does what it says?

Yes

>Moreover we only added it because of the version, so I suggest removing it.

Doesn't hurt, it helps to find the correct `<image>` value

aravindhp: Then please add it for AWS also.

README.md
jrvaldes: /cc @EricPonvelle

docs/machineset-azure.md
```diff
@@ -1,6 +1,14 @@
...
-`<infrastructureID>` should be replaced with the output of:
+Find enclosed a sample YAML, that defines a Windows MachineSet object running on Microsoft Azure that
+the Windows Machine Config Operator (WMCO) can react upon.
+
+## Parameters
+
+`<infrastructureID>` should be replaced with the infrastructure ID that is based on the cluster ID that
+you set when you provisioned the cluster.
+
+To obtain the infrastructure ID run the following command:
```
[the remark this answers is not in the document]
aravindhp: This needs to be done in the main README also.

aravindhp (changes requested):

docs/machineset-aws.md
```diff
@@ -1,19 +1,36 @@
...
- oc get -o jsonpath='{.status.infrastructureName}{"\n"}' infrastructure cluster
+oc get -o jsonpath='{.status.infrastructureName}{"\n"}' infrastructure cluster
 ```
-_\<region\>_ should be replaced with a valid AWS region like `us-east-1`.
-_\<zone\>_ should be replaced with a valid AWS availability zone like `us-east-1a`.
 
+`<region>` should be replaced with a valid AWS region like `us-east-1`.
+
+`<zone>` should be replaced with a valid AWS availability zone like `us-east-1a`.
+
```
aravindhp, line 19: Please remove these extra lines you are introducing in between. This is again introducing inconsistencies between the docs.

aravindhp (changes requested):

docs/machineset-aws.md
jrvaldes: @aravindhp   as per [the master branch](https://github.com/openshift/windows-machine-config-operator/blob/master/docs/machineset-aws.md) the parameters are mixed in the same line, which makes them difficult to read. 

The proposed "extra lines" help to easily identify them in the document. In addition, the [machineset-azure.md](https://github.com/openshift/windows-machine-config-operator/blob/master/docs/machineset-azure.md) document  entertains an itemized list of parameters, that was used as the target reference for this change.

aravindhp: Why given we have not done so anywhere else? This seems like some GitHub markdown foo is needed.

jrvaldes: Reverted to the main objective, the "network requirement in vSphere". Will follow-up with the formatting issues in a low low priority PR.

Thanks @aravindhp  for pulling back.

jrvaldes:
@EricPonvelle PTAL

EricPonvelle (approved):
/lgtm

aravindhp (approved):
/approve

Issue #579, opened 2021-08-11
Feature request: Support airgapped setup on vSphere, make pause image location configurable/prepull pause image


jomeier:
Hi,

as of https://github.com/openshift/windows-machine-config-operator/issues/141#issuecomment-887574152

the only reason, why airgapped OCP installations with Windows Containers are not supported seems to be that the pause image must be pulled from the internet.

Because we must provide our customers support for Windows Containers in production we are definitely interested to get support for this from RH.

An idea to mitigate this problem (as we did it in our own OCP Windows container test setup):
* we prepared a golden Windows VM as described in the documentation (with modifications we had to reverse engineer because the docs are not complete) in an environment where we had internet access
* we prepulled the pause image from the internet to the golden Windows VM
* we sysprepped the golden Windows VM
* we created the MachineSet for vSphere using the previously configured golden Windows VM in our airgapped cluster

In our setup everything works great with Windows containers, no problems in our airgapped environment.

The second option would be to make the hard coded pause image configurable (environment variable of WMCO operator?) so OCP users can store the pause image in an airgapped registry on premises. This should also be rather simple.

In our opinion airgapped clusters are very convenient with OCP intalled on vSphere. So this feature definitely should be support or at least the workaround with prepulled pause images on the golden Windows VM should be supported.

Could you have a look on that, please? It's very important for us.

Thanks and greetings,

Josef

saifshaikh48:
There does not seem to be any problems to me in the workaround you have tried and outlined. Glad to hear Windows containers ran without issue.

Air-gapped environment support is still something that is not officially supported, until we fully test and add documentation around it ourselves. I have created a few stories ([WINC-662](https://issues.redhat.com/browse/WINC-662) & [WINC-663](https://issues.redhat.com/browse/WINC-663)) to put this work on our radar; the team will work to prioritize these soon.

jomeier:
@saifshaikh48 
@aravindhp 
Thank you !

@philipp1992
Could you provide the unattend.xml we used to create the initial setup of the workers and the steps we provided to create the golden image, please?

philipp1992:
> @philipp1992
> Could you provide the unattend.xml we used to create the initial setup of the workers and the steps we provided to create the golden image, please?

we used the unattend.xml provided in the docs minus the product-key section

[...]

PR #601, opened 2021-08-16
WIP: [docs] VSphere golden image for airgapped network

jrvaldes:
Include instructions to pre-pull the Pause Image into the golden
image during the creation process, to support disconnected network
environments.

Follow-up for #579

### DIFF
--- a/docs/vsphere-golden-image.md
+++ b/docs/vsphere-golden-image.md
@@ -129,6 +129,17 @@ Install the `docker` container runtime following the [Microsoft's documentation]
 Alternatively, you can run the provided PowerShell script [install-docker.ps1](vsphere_ci/scripts/install-docker.ps1) 
 to programmatically install `docker` in the Windows VM.
 
+### Set up for disconnected network environment
+
+If you plan to use the golden image in an air-gapped or disconnected network environment, you must pre-pull a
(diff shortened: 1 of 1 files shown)


docs/vsphere-golden-image.md
```diff
@@ -129,6 +129,17 @@ Install the `docker` container runtime following the [Microsoft's documentation]
 Alternatively, you can run the provided PowerShell script [install-docker.ps1](vsphere_ci/scripts/install-docker.ps1) 
 to programmatically install `docker` in the Windows VM.
 
+### Set up for disconnected network environment
```
saifshaikh48, line 132: Could someone using another cloud provider also be interested in this step? Just wondering if this should be in a more generic place (like the top-level README)

jrvaldes: Good point, given the creation of a golden image is only a pre-req for vSphere; here is definitely a good fit for this.
As per OpenShift [docs](https://docs.openshift.com/container-platform/4.8/installing/installing-preparing.html#supported-installation-methods-for-different-platforms), _restricted network_ is also supported in AWS with IPI, not the case for Azure.

aravindhp:
/hold

until the actual work can be prioritized and tested

PR #603, opened 2021-08-16
[wmco] Fix issues with removing HNS networks

mansikulkarni96:
This PR fixes the HNSCommand suffix in the Run() method to avoid printing out the error logging message for PS timeouts when deleting HNS networks.
This PR also fixes the flake on Azure platform with the failure to remove all HNS networks.

### DIFF
--- a/pkg/windows/windows.go
+++ b/pkg/windows/windows.go
@@ -797,15 +797,15 @@ func (vm *windows) newFileInfo(path string) (*payload.FileInfo, error) {
 func (vm *windows) removeHNSNetworks() error {
 	vm.log.Info("removing HNS networks")
 	// VIP HNS endpoint created by the operator is also deleted when the HNS networks are deleted.
-	cmd := "\"$net0 = Get-HnsNetwork | where { $_.Name -eq '" + BaseOVNKubeOverlayNetwork + "'};" +
-		"$net1 = Get-HnsNetwork | where { $_.Name -eq '" + OVNKubeOverlayNetwork + "'};" +
-		"$net0 | Remove-HnsNetwork;" +
-		"$net1 | Remove-HnsNetwork;\""
-	_, err := vm.Run(cmd, true)
-	// PowerShell returns error waiting without exit status or signal error when the HNS networks are removed.
-	if err != nil && !strings.Contains(err.Error(), cmdExitNoStatus) {
-		return errors.Wrapf(err, "error removing %s and %s required HNS networks",
-			BaseOVNKubeOverlayNetwork, OVNKubeOverlayNetwork)
+	for _, network := range []string{BaseOVNKubeOverlayNetwork, OVNKubeOverlayNetwork} {
+		cmd := "\"Get-HnsNetwork | where { $_.Name -eq '" + network + "'}| Remove-HnsNetwork;\""
+		// PowerShell returns error waiting without exit status or signal error when the OVNKubeOverlayNetwork is removed.
+		if _, err := vm.Run(cmd, true); err != nil && !(network == OVNKubeOverlayNetwork && strings.Contains(err.Error(), cmdExitNoStatus)) {
+			return errors.Wrapf(err, "error removing %s HNS network", network)
+		}
+		if err := vm.Reinitialize(); err != nil {
+			return errors.Wrapf(err, "error reinitializing VM after removing %s HNS network", network)
+		}
 	}
 	return nil
 }
--- a/pkg/windows/windows.go
+++ b/pkg/windows/windows.go
@@ -253,7 +253,7 @@ func (vm *windows) Run(cmd string, psCmd bool) (string, error) {
 		// Hack to not print the error log for "sc.exe qc" returning 1060 for non existent services
 		// and not print error when the command takes too long to return after removing HNS networks.
 		if !(strings.HasPrefix(cmd, serviceQueryCmd) && strings.HasSuffix(err.Error(), serviceNotFound)) &&
-			!(strings.Contains(err.Error(), cmdExitNoStatus) && strings.HasSuffix(cmd, removeHNSCommand+";")) {
+			!(strings.Contains(err.Error(), cmdExitNoStatus) && strings.HasSuffix(cmd, removeHNSCommand+";\"")) {
 			vm.log.Error(err, "error running", "cmd", cmd, "out", out)
 		}
 		return out, errors.Wrapf(err, "error running %s", cmd)


aravindhp (approved):

mansikulkarni96:
azure test [passing](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/603/pull-ci-openshift-windows-machine-config-operator-master-azure-e2e-operator/1427741571461156864) with the fix.
/test azure-e2e-operator

pkg/windows/windows.go
```diff
@@ -800,6 +800,7 @@ func (vm *windows) removeHNSNetworks() error {
 	cmd := "\"$net0 = Get-HnsNetwork | where { $_.Name -eq '" + BaseOVNKubeOverlayNetwork + "'};" +
 		"$net1 = Get-HnsNetwork | where { $_.Name -eq '" + OVNKubeOverlayNetwork + "'};" +
 		"$net0 | Remove-HnsNetwork;" +
+		"Start-Sleep -s 10;" +
```
aravindhp, line 803: This warrants a comment in the code too. I however don't understand the fix. How can the command return before executing fully? By the same logic isn't it possible for the command to return without executing the sleep? 

I am also wondering if the order of deletion is important. Do you know deleting which network causes a blip in the connectivity?

I suspect you are running into issues because the network is getting reconfigured. You might have to split the call into separate commands and re-initialize the SSH connection in between.

aravindhp (changes requested):

aravindhp:
I see the same [failure](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/603/pull-ci-openshift-windows-machine-config-operator-master-azure-e2e-operator/1427791968527716352) even with the fix.

pkg/windows/windows.go
mansikulkarni96: Agree with your statement. My earlier thought process was the sleep might give time for the first network to cleanup before executing the next command.
Removing the second network  OVNKubeOverlayNetwork causes loss in connectivity. That is why it is the last one being removed.
Separating into two different calls seems to be the best solution. I made the changes for that. Testing out locally to see if the issue persists.

mansikulkarni96:
/test azure-e2e-operator
failed with cluster bring up.

aravindhp:
The tests actually passed but the must-gather step failed.

mansikulkarni96:
> The tests actually passed but the must-gather step failed
That's right I must have overlooked, tests passed but [failed must-gather](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/603/pull-ci-openshift-windows-machine-config-operator-master-azure-e2e-operator/1428190545959718912) again.

saifshaikh48 (commented):
/lgtm

pkg/windows/windows.go
```diff
@@ -797,15 +799,19 @@ func (vm *windows) newFileInfo(path string) (*payload.FileInfo, error) {
...
+	baseOvnCmd := "\"Get-HnsNetwork | where { $_.Name -eq '" + BaseOVNKubeOverlayNetwork + "'}| Remove-HnsNetwork;\""
+	_, err := vm.Run(baseOvnCmd, true)
+	if err != nil {
+		return errors.Wrapf(err, "error removing %s required HNS networks", BaseOVNKubeOverlayNetwork)
+	}
+	if err = vm.Reinitialize(); err != nil {
+		return errors.Wrapf(err, "error reinitializing VM after removing %s HNS network", BaseOVNKubeOverlayNetwork)
+	}
+	ovnCmd := "\"Get-HnsNetwork | where { $_.Name -eq '" + OVNKubeOverlayNetwork + "'}| Remove-HnsNetwork;\""
+	_, err = vm.Run(ovnCmd, true)
```
jrvaldes, line 811: Following this [comment](https://github.com/openshift/windows-machine-config-operator/pull/603#discussion_r691295828), the OVNKubeOverlayNetwork removal is the one  that drops the connectivity; Consider moving `vm.Reinitialize()` after, so no need to [reinitialize SSH client should if executing a step after removing networks](https://github.com/openshift/windows-machine-config-operator/pull/603/files#diff-4d8fe3bb66a872f0e4a277f820d60d2efa722f09aa89516be74b98c06b57f3bdR317).

pkg/windows/windows.go
```diff
@@ -797,15 +799,19 @@ func (vm *windows) newFileInfo(path string) (*payload.FileInfo, error) {
...
-		"$net1 | Remove-HnsNetwork;\""
-	_, err := vm.Run(cmd, true)
-	// PowerShell returns error waiting without exit status or signal error when the HNS networks are removed.
+	baseOvnCmd := "\"Get-HnsNetwork | where { $_.Name -eq '" + BaseOVNKubeOverlayNetwork + "'}| Remove-HnsNetwork;\""
+	_, err := vm.Run(baseOvnCmd, true)
+	if err != nil {
+		return errors.Wrapf(err, "error removing %s required HNS networks", BaseOVNKubeOverlayNetwork)
+	}
+	if err = vm.Reinitialize(); err != nil {
+		return errors.Wrapf(err, "error reinitializing VM after removing %s HNS network", BaseOVNKubeOverlayNetwork)
```
jrvaldes, line 808: s/reinitializing VM/reinitializing ssh client

`reinitializing VM` may lead to misunderstanding, for example: re-configuring/rebooting the VM

jrvaldes (commented):
Thanks for the fix @mansikulkarni96, comments added PTAL, otherwise LGTM

pkg/windows/windows.go
mansikulkarni96: We are not really  rebooting or reconfiguring the VM, we only [re-initalize the ssh client](https://github.com/openshift/windows-machine-config-operator/blob/374053964b55c49aec644f4398789100c70667ea/pkg/windows/windows.go#L151), the wording is consistent with a call [here](https://github.com/openshift/windows-machine-config-operator/blob/374053964b55c49aec644f4398789100c70667ea/pkg/windows/windows.go#L395) and the error message is expanded in the [vm.Reinitialize()](https://github.com/openshift/windows-machine-config-operator/blob/374053964b55c49aec644f4398789100c70667ea/pkg/windows/windows.go#L267) call.

pkg/windows/windows.go
mansikulkarni96: @jrvaldes re-initializing the ssh client in between is necessary to fix the flake. Removal of the BaseOvnKubeOverlayNetwork network causes reconfigurations and results into the flake of skipping the removal of OvnKubeOverlayNetwork on some instances.
Although I could add another vm.Reinitialize() at the end to remove the comment if that's the consensus. I did not add it because this is the last step in deconfiguration.

pkg/windows/windows.go
```diff
@@ -313,6 +313,8 @@ func (vm *windows) Deconfigure() error {
 	if err := vm.removeDirectories(); err != nil {
 		return errors.Wrap(err, "unable to remove created directories")
 	}
+	// Removing HNS networks is the last step in deconfigure process, it results in the ssh connection
+	// to be closed. Ssh client should be reinitialized if executing a step after removing networks.
```
sebsoto, line 317: `SSH`

pkg/windows/windows.go
```diff
@@ -797,15 +799,19 @@ func (vm *windows) newFileInfo(path string) (*payload.FileInfo, error) {
...
-		"$net1 = Get-HnsNetwork | where { $_.Name -eq '" + OVNKubeOverlayNetwork + "'};" +
-		"$net0 | Remove-HnsNetwork;" +
-		"$net1 | Remove-HnsNetwork;\""
-	_, err := vm.Run(cmd, true)
-	// PowerShell returns error waiting without exit status or signal error when the HNS networks are removed.
+	baseOvnCmd := "\"Get-HnsNetwork | where { $_.Name -eq '" + BaseOVNKubeOverlayNetwork + "'}| Remove-HnsNetwork;\""
+	_, err := vm.Run(baseOvnCmd, true)
+	if err != nil {
+		return errors.Wrapf(err, "error removing %s required HNS networks", BaseOVNKubeOverlayNetwork)
+	}
```
sebsoto, line 806: Can be condensed to 
```

	if _, err := vm.Run(baseOvnCmd, true); err != nil {
		return errors.Wrapf(err, "error removing %s required HNS networks", BaseOVNKubeOverlayNetwork)
	}
```

pkg/windows/windows.go
sebsoto: > Removal of the BaseOvnKubeOverlayNetwork network causes reconfigurations

Reconfigurations by what entiity?

additionally:
could this be done in a loop?

something like:
```
for _, network := range []string{OVNKubeOverlayNetwork,BaseOVNKubeOverlayNetwork} {
  //removal logic
}
vm.reinitialize()
```

sebsoto (changes requested):

pkg/windows/windows.go
mansikulkarni96: can be done in a loop with the vm.Reinitialize call placed in the loop. I am not a 100% sure what re-configurations happen but I am here referring to HNSEndpoint and other networks that get deleted as part of the BaseOVNKubeOverlayNetwork deletion.

pkg/windows/windows.go
```diff
@@ -797,15 +797,15 @@ func (vm *windows) newFileInfo(path string) (*payload.FileInfo, error) {
...
-	_, err := vm.Run(cmd, true)
-	// PowerShell returns error waiting without exit status or signal error when the HNS networks are removed.
-	if err != nil && !strings.Contains(err.Error(), cmdExitNoStatus) {
-		return errors.Wrapf(err, "error removing %s and %s required HNS networks",
-			BaseOVNKubeOverlayNetwork, OVNKubeOverlayNetwork)
+	for _, network := range []string{BaseOVNKubeOverlayNetwork, OVNKubeOverlayNetwork} {
+		cmd := "\"Get-HnsNetwork | where { $_.Name -eq '" + network + "'}| Remove-HnsNetwork;\""
+		// PowerShell returns error waiting without exit status or signal error when the OVNKubeOverlayNetwork is removed.
+		if _, err := vm.Run(cmd, true); err != nil && !strings.Contains(err.Error(), cmdExitNoStatus) {
+			return errors.Wrapf(err, "error removing %s required HNS networks", network)
```
aravindhp, line 804: `error removing %s HNS network`

pkg/windows/windows.go
```diff
@@ -797,15 +797,15 @@ func (vm *windows) newFileInfo(path string) (*payload.FileInfo, error) {
...
-		"$net0 | Remove-HnsNetwork;" +
-		"$net1 | Remove-HnsNetwork;\""
-	_, err := vm.Run(cmd, true)
-	// PowerShell returns error waiting without exit status or signal error when the HNS networks are removed.
-	if err != nil && !strings.Contains(err.Error(), cmdExitNoStatus) {
-		return errors.Wrapf(err, "error removing %s and %s required HNS networks",
-			BaseOVNKubeOverlayNetwork, OVNKubeOverlayNetwork)
+	for _, network := range []string{BaseOVNKubeOverlayNetwork, OVNKubeOverlayNetwork} {
+		cmd := "\"Get-HnsNetwork | where { $_.Name -eq '" + network + "'}| Remove-HnsNetwork;\""
+		// PowerShell returns error waiting without exit status or signal error when the OVNKubeOverlayNetwork is removed.
```
aravindhp, line 802: Given this, shouldn't you be doing `!strings.Contains(err.Error(), cmdExitNoStatus)` only for `OVNKubeOverlayNetwork`?

aravindhp (changes requested):
Please add `[wmco]` or `[windows]` prefix to your 2nd commit message

pkg/windows/windows.go
```diff
@@ -797,15 +797,15 @@ func (vm *windows) newFileInfo(path string) (*payload.FileInfo, error) {
...
+		if _, err := vm.Run(cmd, true); err != nil && !(network == OVNKubeOverlayNetwork && strings.Contains(err.Error(), cmdExitNoStatus)) {
+			return errors.Wrapf(err, "error removing %s required HNS network", network)
+		}
+		if err := vm.Reinitialize(); err != nil {
```
aravindhp, line 806: I assume you are calling `vm.Reinitialize()` on purpose after every network removal.

aravindhp (approved):
/approve

pkg/windows/windows.go
mansikulkarni96: that is correct, going with the consensus removed the comment added earlier to mention vm.Reinitialize() needs to be done in future.

pkg/windows/windows.go
```diff
@@ -797,15 +797,15 @@ func (vm *windows) newFileInfo(path string) (*payload.FileInfo, error) {
...
-	_, err := vm.Run(cmd, true)
-	// PowerShell returns error waiting without exit status or signal error when the HNS networks are removed.
-	if err != nil && !strings.Contains(err.Error(), cmdExitNoStatus) {
-		return errors.Wrapf(err, "error removing %s and %s required HNS networks",
-			BaseOVNKubeOverlayNetwork, OVNKubeOverlayNetwork)
+	for _, network := range []string{BaseOVNKubeOverlayNetwork, OVNKubeOverlayNetwork} {
+		cmd := "\"Get-HnsNetwork | where { $_.Name -eq '" + network + "'}| Remove-HnsNetwork;\""
+		// PowerShell returns error waiting without exit status or signal error when the OVNKubeOverlayNetwork is removed.
+		if _, err := vm.Run(cmd, true); err != nil && !(network == OVNKubeOverlayNetwork && strings.Contains(err.Error(), cmdExitNoStatus)) {
+			return errors.Wrapf(err, "error removing %s required HNS network", network)
```
jrvaldes, line 804: s/error removing %s required HNS network/error removing %s HNS network

Removing something that is required, doesn't sounds.

jrvaldes (changes requested):
Minor change request, as per https://github.com/openshift/windows-machine-config-operator/pull/603#discussion_r692431678

jrvaldes (approved):
/lgtm

mansikulkarni96:
/test azure-e2e-operator

making sure flake is fixed.

PR #618, opened 2021-08-19
WINC-571: [wmco] Prevent OLM upgrades when operator is busy

saifshaikh48:
This PR introduces the reporting of operator conditions back to OLM.
Specifically, the `Upgradeable` condition is set to `False` when WMCO is
processing any child components (Windows nodes, private key secret, CSRs, 
etc), blocking OLM from replacing the operator CSV during this time.
This ensures smooth transitions from one EUS version of OpenShift to another.

The `Upgradeable` condition is set back to `True` when no processing is
currently ongoing, meaning that controllers:
- reconciled without error
- did not request a requeue

A mutex lock is maintained to secure access to the busy controllers set
and the patching of the OperatorCondition resource.


controllers/certificatesigningrequests_controller.go
```diff
@@ -69,21 +71,40 @@ func NewCertificateSigningRequestsReconciler(mgr manager.Manager, clusterConfig
 // Reconcile is part of the main kubernetes reconciliation loop which reads that state of the cluster for a
 // CertificateSigningRequests object and aims to move the current state of the cluster closer to the desired state.
 func (r *certificateSigningRequestsReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	_ = r.log.WithValues("certificatesigningrequests", req.NamespacedName)
+	log := r.log.WithValues("certificatesigningrequests", req.NamespacedName)
+
+	// Update operator condition to prevent OLM from updating WMCO while CSRs are being processed
+	if err := conditions.PatchUpgradeable(r.client, r.watchNamespace, meta.ConditionFalse,
+		conditions.UpgradeableFalseMessage, conditions.UpgradeableFalseReason); err != nil {
+		log.Info("unable to set Upgradeable condition, be cautious of automatic operator upgrades", "error", err)
```
aravindhp, line 79: I suggest sending an event in addition to the log here. This applies to all places where an error is happening.

While I think not returning an error is fine, please check with the OLM team also.

controllers/certificatesigningrequests_controller.go
```diff
@@ -69,21 +71,40 @@ func NewCertificateSigningRequestsReconciler(mgr manager.Manager, clusterConfig
...
 	if err := r.client.Get(ctx, req.NamespacedName, certificateSigningRequest); err != nil {
 		if k8sapierrors.IsNotFound(err) {
 			// Request object not found, could have been deleted after reconcile request.
 			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
 			// Return and don't requeue
+			// Communicate current operator condition Upgradeable=True to allow OLM to update WMCO if needed
+			if err := conditions.PatchUpgradeable(r.client, r.watchNamespace, meta.ConditionTrue,
+				conditions.UpgradeableTrueMessage, conditions.UpgradeableTrueReason); err != nil {
+				log.Info("unable to set Upgradeable condition, manual operator upgrades may be needed", "error", err)
+			}
```
aravindhp, line 92: This can be a helper. Something like `condition.DisableOLMUpgrade()` and you wouldn't expose `PatchUpgradeable` and all the other constants.

controllers/certificatesigningrequests_controller.go
```diff
@@ -69,21 +71,40 @@ func NewCertificateSigningRequestsReconciler(mgr manager.Manager, clusterConfig
...
-	return ctrl.Result{}, r.reconcileCSR(certificateSigningRequest)
+	if err := r.reconcileCSR(certificateSigningRequest); err != nil {
+		return ctrl.Result{}, err
+	}
+	// Communicate current operator condition Upgradeable=True to allow OLM to update WMCO if needed
+	if err := conditions.PatchUpgradeable(r.client, r.watchNamespace, meta.ConditionTrue,
+		conditions.UpgradeableTrueMessage, conditions.UpgradeableTrueReason); err != nil {
+		log.Info("unable to set Upgradeable condition, manual operator upgrades may be needed", "error", err)
+	}
+	return ctrl.Result{}, nil
```
aravindhp, line 107: And this would be `condition.EnableOLMUpgrade()`

pkg/conditions/conditions.go
```diff
@@ -0,0 +1,56 @@
+package conditions
```
aravindhp, line 1: Shouldn't this be `condition`?

main.go
```diff
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	mapi "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
+	operatorframework "github.com/operator-framework/api/pkg/operators/v2"
```
aravindhp, line 11: `operators` please in case we have to add more from `operator-framework`

aravindhp (changes requested):

mansikulkarni96:
Design LGTM, with the suggestion of adding helpers for updating to safe and unsafe upgrade.

pkg/conditions/conditions.go
saifshaikh48: Sure, as this package is centered around the OperatorCondition CR.

controllers/certificatesigningrequests_controller.go
saifshaikh48: The OLM team mentioned that operators would usually retry on transient errors (e.g. connection issues). Because of the high number of places/code paths that set `Upgradeable` (multiple places in each controller), I think we mitigate that. So sending and event and logging should be fine.

aravindhp: I suggest retrying given the feedback from the OLM team. Having multiple places where we set `Upgradable` is not a solution to the problem.

pkg/condition/condition.go
```diff
@@ -0,0 +1,85 @@
...
+	UpgradeableTrueMessage  = "upgradeIsSafe"
+	UpgradeableTrueReason   = "The operator is safe for upgrade."
+
+	// OperatorConditionName is an environment variable set by OLM identifying the operator's OperatorCondition CR
+	OperatorConditionName = "OPERATOR_CONDITION_NAME"
+)
+
+// ToggleOLMUpgrade sets the current operator Upgradeable condition to either prevent or allow operator version upgrades
+// Sends warning event and logs error in case of failure. No-op if operator is not managed by OLM
+func ToggleOLMUpgrade(c client.Client, recorder record.EventRecorder, log logr.Logger, ocNamespace string, allow bool) {
```
aravindhp, line 32: I am not a fan of exposing such a method. I prefer having explicit helpers like enable and disable. This makes reading the code much easier at the call site.

saifshaikh48:
/retitle [WINC-571] Prevent upgrades when operator is busy

controllers/certificatesigningrequests_controller.go
saifshaikh48: After discussion with OLM and Operator SDK folks:
- will retry on error
- will wait for `Upgradeable` to be updated in `OperatorCondition` status, not just spec

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
```diff
@@ -309,6 +309,16 @@ spec:
           - securitycontextconstraints
```
aravindhp, line 309: Spelling error in `reconcile` in the commit message.

pkg/condition/condition.go
```diff
@@ -0,0 +1,109 @@
...
+	return set(c, watchNamespace, operators.Upgradeable, meta.ConditionFalse, upgradeableFalseReason,
+		upgradeableFalseMessage)
+}
+
+// set sets the given condition with the given values within the associated OperatorCondition.
+// No-op if operator is not managed by OLM
+func set(c client.Client, ocNamespace, conditionType string, status meta.ConditionStatus,
+	reason, message string) error {
+	// Check if operator is OLM-managed
+	if ocName, present := os.LookupEnv(OperatorConditionName); present {
```
aravindhp, line 45: I take it that the env var not being present is not an error. Correct?

I also suggest checking for `!present` and returning nil. That way majority of the code will be less indented and easier to read.

pkg/condition/condition.go
```diff
@@ -0,0 +1,109 @@
...
+			return err
+		}
+		// Treat condition setting as a lock mechanism -- ensure spec change is actually picked up
+		return waitForOperatorCondition(c, ocNamespace, ocName, conditionType, status)
+	}
+	return nil
+}
+
+// waitForOperatorCondition repeatedly checks if the OperatorCondition resource's Status has been updated
+func waitForOperatorCondition(c client.Client, ocNamespace, ocName, conditionType string,
```
aravindhp, line 60: Are you really waiting for the OperatorCondition or for it to be updated? I suggest renaming the method to `waitForOperatorConditioUpdate` or even to `wait`.

I also don't understand the logic to wait and keep retrying the get(). Shouldn't you be retrying the patch()` which is what can fail?

pkg/condition/condition.go
```diff
@@ -0,0 +1,109 @@
...
+	expectedStatus meta.ConditionStatus) error {
+	err := wait.Poll(10*time.Second, time.Minute, func() (bool, error) {
```
aravindhp, line 62: Please add the time value as a constant to the `retry` package.

pkg/condition/condition.go
```diff
@@ -0,0 +1,109 @@
...
+	// Check if operator is OLM-managed
+	if ocName, present := os.LookupEnv(OperatorConditionName); present {
+		oc, err := get(c, ocNamespace, ocName)
+		if err != nil {
+			return err
+		}
+		if err = patch(c, oc, conditionType, status, reason, message); err != nil {
+			return err
+		}
+		// Treat condition setting as a lock mechanism -- ensure spec change is actually picked up
```
aravindhp, line 53: Not sure if I would call this a lock given a lock should work even if we are multi-threaded.

test/e2e/reconfigure_test.go
```diff
@@ -49,6 +49,10 @@ func reconfigurationTest(t *testing.T) {
 	_, err = testCtx.client.K8s.CoreV1().Nodes().Patch(context.TODO(), machineNodes[0].Name, types.JSONPatchType,
 		patchData, metav1.PatchOptions{})
 	require.NoError(t, err)
+	// Ensure operator communicates to OLM that upgrade is not safe when processing Machine nodes
+	err = testCtx.validateUpgradeableCondition(metav1.ConditionFalse)
```
aravindhp, line 53: These test seem timing dependent. Are you sure they are not flaky?

aravindhp (changes requested):

pkg/condition/condition.go
saifshaikh48: I'll remove that part as I am not sure if it would work as intended with multithreading.

test/e2e/reconfigure_test.go
saifshaikh48: I have not seen any failures locally or in the e2e tests. Waiting for the condition to be reflected in status should prevent any timing issues I thought.

pkg/condition/condition.go
saifshaikh48: The purpose was that to check that after a successful patch, does the status actually get updated? If we don't wait then a controller start processing things while OLM still thinks upgrades are safe.

However, I see your point, we should retry the patch itself as well.

pkg/condition/condition.go
```diff
@@ -0,0 +1,149 @@
...
+	"encoding/json"
+	"fmt"
+	"os"
+	"strings"
+
+	operators "github.com/operator-framework/api/pkg/operators/v2"
+	"github.com/pkg/errors"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kubeTypes "k8s.io/apimachinery/pkg/types"
+	kubeWait "k8s.io/apimachinery/pkg/util/wait"
```
aravindhp, line 14: We only need to introduce prefixes if there is a collision. I suggest leaving `kube` out.

test/e2e/validation_test.go
```diff
@@ -393,3 +400,29 @@ func (tc *testContext) findNodeCSRs(nodeName string) ([]certificates.Certificate
...
+
+// validateUpgradeableCondition ensures that the operator's Upgradeable condition is correctly communicated to OLM
+func (tc *testContext) validateUpgradeableCondition(expected meta.ConditionStatus) error {
+	ocName, present := os.LookupEnv(condition.OperatorConditionName)
+	if !present {
+		// Implies operator is not OLM-managed
+		return nil
+	}
+
+	err := wait.Poll(10*time.Second, time.Minute, func() (bool, error) {
```
aravindhp, line 412: Is it possible to use some of the existing constants?

test/e2e/secrets_test.go
```diff
@@ -109,7 +109,7 @@ func testUserDataTamper(t *testing.T) {
 			}
 
 			// wait for userData secret creation / update to take effect.
-			err := wait.Poll(5*time.Second, 20*time.Second, func() (done bool, err error) {
+			err := wait.Poll(retry.Interval, retry.ResourceChangeTimeout, func() (done bool, err error) {
```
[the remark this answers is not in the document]
saifshaikh48: I was seeing a flake here as setting the Upgradeable condition takes ~20secs itself so the previous timeout was too low. Also for consistency. Updated the commit message to make it more clear.

test/e2e/validation_test.go
saifshaikh48: Refactored to reduce code duplication.

pkg/condition/condition.go
```diff
@@ -0,0 +1,149 @@
+package condition
```
aravindhp, line 1: Please collect the exported methods and internal helpers in separate sections. Makes reading and managing the code easier.

pkg/condition/condition.go
```diff
@@ -0,0 +1,149 @@
...
+	if OcName == "" {
+		return nil
+	}
+
+	oc, err := get(c, ocNamespace)
+	if err != nil {
+		return err
+	}
+	// Do not change the Upgradeable condition if a different controller is busy. Prevents potential timing errors
+	// if controller A, which started after controller B, finishes reconciling before controller B finishes
```
aravindhp, line 64: Do we need a lock?

test/e2e/validation_test.go
```diff
@@ -393,3 +395,12 @@ func (tc *testContext) findNodeCSRs(nodeName string) ([]certificates.Certificate
...
 	return nodeCSRs, nil
 }
+
+// validateUpgradeableCondition ensures that the operator's Upgradeable condition is correctly communicated to OLM
+func (tc *testContext) validateUpgradeableCondition(expected meta.ConditionStatus) error {
+	if condition.OcName == "" {
+		// Implies operator is not OLM-managed
+		return nil
+	}
+	return condition.Wait(tc.client.Cache, tc.namespace, operators.Upgradeable, expected)
```
aravindhp, line 405: Not a huge fan of using product code in the e2e test when the method is not being unit tested.

pkg/condition/condition.go
saifshaikh48: I don't think a lock is needed. With a lock, if 2 controllers are reconciling at the same time, a go routine from one controller would be stuck waiting for the lock in order to proceed despite the Upgradeable condition already reflecting the proper operator state.

aravindhp (changes requested):

pkg/retry/retry.go
```diff
@@ -9,4 +9,6 @@ const (
 	Interval = 15 * time.Second
 	// Timeout is the total time we will wait for an event to occur.
 	Timeout = time.Minute * 10
+	// ResourceChangeTimeout is the total time waited for a change (create/update/delete) to take place
+	ResourceChangeTimeout = time.Minute * 2
```
aravindhp, line 13: I would prefer this change be done in a separate PR.

saifshaikh48: Change is directly related to this PR as discussed here
https://github.com/openshift/windows-machine-config-operator/pull/618/commits/2998aa57b35da5f8380458269c635f245f920c38#r698607183

saifshaikh48: Not sure why the link doesn't work so here's the conversation:

 aravindhp on Aug 30
> You are introducing a behavior change here. Why are you doing that?

 saifshaikh48 on Aug 31
> I was seeing a flake here as setting the Upgradeable condition takes ~20secs itself so the previous timeout was too low. Also for consistency. Updated the commit message to make it more clear.

controllers/certificatesigningrequests_controller.go
```diff
@@ -34,6 +34,11 @@ import (
 	"github.com/openshift/windows-machine-config-operator/pkg/csr"
 )
 
+const (
```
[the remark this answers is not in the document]
saifshaikh48: These constants are used in the next commit in [isAnotherControllerBusy()](https://github.com/openshift/windows-machine-config-operator/pull/618/commits/25e14b17a4c39c654d5ab98d1bfab8b677adb5d8#diff-4e95dc26dacdca908d7eb173740544f27a74edf6474f9fbbeaa5461f57089fc7R113), passed in from each respective controller.

pkg/condition/condition.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	oc, err := get(c, ocNamespace)
+	if err != nil {
+		return err
+	}
+
+	if conditionType == operators.Upgradeable {
+		// Do not change the Upgradeable condition if a different controller is busy. Prevents potential timing issue
+		// if controller A, which started after controller B, finishes reconciling before controller B finishes
```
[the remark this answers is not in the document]
saifshaikh48: Valid concern. You're right, locking during the resource patching would handle this scenario.

pkg/retry/retry.go
aravindhp: OK, please call out in the commit message that the reason for increasing it was Upgradeable condition taking longer to be updated.

controllers/certificatesigningrequests_controller.go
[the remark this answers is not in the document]
aravindhp: OK, fair enough. Though I suspect this will no longer be needed if you introduce a lock.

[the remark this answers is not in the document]
saifshaikh48: I don't see how a lock would remove the need for this without also removing the parallel processing of controllers. This may warrant a meeting.

pkg/condition/condition.go
```diff
@@ -0,0 +1,171 @@
...
+		Message: message,
+	}
+	if err = patch(c, oc, newCond); err != nil {
+		return err
+	}
+	// Wait to ensure change is actually picked up
+	return wait(c, ocNamespace, newCond.Type, newCond.Status)
+}
+
+// get retrieves the OperatorCondition resource of the given name from the given namespace
```
sebsoto, line 124: Only the namespace is given, not the name

pkg/condition/condition.go
```diff
@@ -0,0 +1,171 @@
...
+	// More than one controller is busy, meaning Upgradeable condition is already False. No need to update value.
+	if len(busyControllers) > 1 {
+		return nil
+	}
+	return set(c, watchNamespace, operators.Upgradeable, meta.ConditionFalse, upgradeableFalseReason,
+		fmt.Sprintf("%s: %s", upgradeableFalseMessageBase, controllerName))
+}
+
+// Validate checks that the given condition is present and proper within the given list
+func Validate(conditions []meta.Condition, conditionType string, expectedStatus meta.ConditionStatus) bool {
```
sebsoto, line 87: Don't see this used outside of the package in this commit, maybe its used in a later commit

sebsoto (changes requested):

pkg/condition/condition.go
sebsoto: yep was used in test commit

pkg/condition/condition.go
```diff
@@ -0,0 +1,177 @@
...
+	busyControllers[controllerName] = struct{}{}
+
+	// If Upgradeable condition is already False, no-op to avoid redundant API calls
+	if Validate(oc.Spec.Conditions, operators.Upgradeable, meta.ConditionFalse) {
+		return nil
+	}
+	return set(c, oc, operators.Upgradeable, meta.ConditionFalse, upgradeableFalseReason, upgradeableFalseMessage)
+}
+
+// Validate checks that the given condition is present and proper within the given list
```
aravindhp, line 101: Please expand proper. Proper can mean anything without context.

pkg/condition/condition.go
```diff
@@ -0,0 +1,177 @@
...
+	// OperatorConditionName is an environment variable set by OLM identifying the operator's OperatorCondition CR
+	OperatorConditionName = "OPERATOR_CONDITION_NAME"
+)
+
+var (
+	// ocName is the name of the OLM-managed OperatorCondition resource. Empty if operator is not OLM-managed
+	ocName string
+	// busyControllers is a representation of a set that holds the names of controllers that are currently reconciling
+	busyControllers map[string]struct{}
+	// resourceLock is a mutex lock to ensure thread-safe access of shared entities
```
aravindhp, line 34: Please call out the entities this lock is for. Also given that this is protecting the map and the OperatorCondition it needs to be named appropriately.

pkg/condition/condition.go
```diff
@@ -0,0 +1,177 @@
...
+	resourceLock.Lock()
+	defer resourceLock.Unlock()
+
+	oc, err := get(c, watchNamespace)
+	if err != nil {
+		return err
+	}
+
+	// Add given controller to busy controllers set. struct{} used as value since it takes up 0 bytes
+	busyControllers[controllerName] = struct{}{}
```
aravindhp, line 92: Shouldn't this happen as soon as you grab the lock?

pkg/condition/condition.go
```diff
@@ -0,0 +1,177 @@
...
+	resourceLock.Lock()
+	defer resourceLock.Unlock()
+
+	oc, err := get(c, watchNamespace)
+	if err != nil {
+		return err
+	}
+
+	// Remove given controller from busy controllers set
+	delete(busyControllers, controllerName)
```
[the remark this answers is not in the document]
aravindhp: Ah yes. I forgot about that. But I hope you see the confusion this is causing by reading the code.

controllers/certificatesigningrequests_controller.go
```diff
@@ -73,9 +74,18 @@ func NewCertificateSigningRequestsReconciler(mgr manager.Manager, clusterConfig
...
+func (r *certificateSigningRequestsReconciler) Reconcile(ctx context.Context,
+	req ctrl.Request) (res ctrl.Result, reconcileErr error) {
 	_ = r.log.WithValues(CSRController, req.NamespacedName)
 
+	// Prevent WMCO upgrades while CSRs are being processed
+	if err := condition.MarkAsBusy(r.client, r.watchNamespace, CSRController); err != nil {
+		return ctrl.Result{}, err
+	}
+	defer func() {
+		reconcileErr = tryToMarkAsFree(r.client, r.watchNamespace, CSRController, res.Requeue, reconcileErr)
```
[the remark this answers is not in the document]
saifshaikh48: Won't the `defer func` always be called after `reconcileErr` is populated? We only want to call `MarkAsFree` depending on what `Reconcile` returns. This is the cleanest way I found to do that -- other options were: 

-  having to [call `MarkAsFree` everywhere we return with no error + no requeue](https://github.com/openshift/windows-machine-config-operator/commit/ea4e6b83360b441ae59a4ab01a57a451770495e5#diff-c5861f88c2edbbdc81d4ec5d1910af145b82e6acefe40f4c019b25ad112338dfL278-R298) in each controller
- Making `Reconcile()` a wrapper, something like
```suggestion
func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
	result, err := actualReconciler(ctx, request)
	if result.Requeue || err != nil {
		return result, err
	}
	return result, condition.MarkAsFree()
}

func actualReconciler(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
	// all the reconcile logic
}
```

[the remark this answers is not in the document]
saifshaikh48: If we opt for a ["best effort" policy](https://github.com/openshift/windows-machine-config-operator/pull/618#discussion_r723729906) when setting the condition, I suspect this logic would be a bit cleaner btw.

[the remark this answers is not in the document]
aravindhp: Not a option given what you [discovered](https://github.com/openshift/windows-machine-config-operator/pull/618#discussion_r694921000) from the SDK folks

pkg/condition/condition.go
saifshaikh48: Having trouble keeping the name descriptive yet short. Going with `ocAndControllerSetLock`, LMK if you object/have other suggestions.

pkg/condition/condition.go
```diff
@@ -0,0 +1,177 @@
...
+	"context"
+	"encoding/json"
+	"os"
+	"sync"
+
+	operators "github.com/operator-framework/api/pkg/operators/v2"
+	"github.com/pkg/errors"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	kubeWait "k8s.io/apimachinery/pkg/util/wait"
```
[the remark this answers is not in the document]
sebsoto: I thought there should be no issues with a helper function have the same name as an imported package?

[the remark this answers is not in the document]
saifshaikh48: Does not seem so, if I drop the `kubeWait` alias from the import I get a `DuplicateDecl` compiler error

pkg/condition/condition.go
```diff
@@ -0,0 +1,178 @@
...
+	opCond, err := get(c, watchNamespace)
+	if err != nil {
+		return err
+	}
+
+	// If at least one other controller is still busy, no-op as Upgradeable should not be set to True yet
+	// Also, if Upgradeable condition is already True, no-op to avoid redundant API calls
+	if len(busyControllers) > 0 || validate(opCond.Spec.Conditions, operators.Upgradeable, meta.ConditionTrue) {
+		return nil
+	}
```
sebsoto: This seems weird to combine these. You can check if len(busyControllers) >0 before you do the get.
I'm not sure if this is an effort to solve the issue described [here](https://github.com/openshift/windows-machine-config-operator/pull/618#discussion_r723420711)?

Is this issue actually solved? Don't you always have to validate the opCond?

saifshaikh48: > This seems weird to combine these. You can check if len(busyControllers) >0 before you do the get

Fair point

> Don't you always have to validate the opCond?

In the [`MarkAsBusy` method](https://github.com/openshift/windows-machine-config-operator/pull/618/commits/3d0c2ff167758d578c7601d85b250502d44828a8#diff-4e95dc26dacdca908d7eb173740544f27a74edf6474f9fbbeaa5461f57089fc7R94), yes we always have to validate the operator condition. 
But in the `MarkAsFree` case, we want to be a bit stricter before we can say upgrades are safe. If either of the conditions fails at this point, we cannot set `Upgradeable` to True. Here's all the scenarios:

- if at least one controller is busy (meaning upgradeable _should_ be false) and Upgradeable is currently False
  - no-op
- **important case:** if at least one controller is busy (meaning upgradeable _should_ be false) and Upgradeable is currently True
  - no-op
  - another controller had added itself to `busyControllers` set but hit an error when setting status in OperatorCondition
- if 0 controllers are busy (meaning upgradeable _should_ be true) and Upgradeable is currently True
  - no-op
- if 0 controllers are busy (meaning upgradeable _should_ be true) and Upgradeable is currently False
  - happy path, go ahead and set status in OperatorCondition to True

controllers/controllers.go
```diff
@@ -189,3 +190,13 @@ func isValidWindowsNode(o client.Object, byoh bool) bool {
 	}
 	return true
 }
+
+// tryToMarkAsFree is called after a controller's Reconcile function returns. If the given controller finished
+// reconciling without error or requesting a requeue event, the controller is marked as free.
+// When all controllers are free, WMCO upgrades are unblocked.
```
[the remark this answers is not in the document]
selansen: LIke @aravindhp  mentioned , if we caught in endless loop, should we have a timer or something so that we will allow upgrade instead of getting into deadloack ?

controllers/certificatesigningrequests_controller.go
```diff
@@ -75,15 +75,16 @@ func NewCertificateSigningRequestsReconciler(mgr manager.Manager, clusterConfig
 // Reconcile is part of the main kubernetes reconciliation loop which reads that state of the cluster for a
 // CertificateSigningRequests object and aims to move the current state of the cluster closer to the desired state.
 func (r *certificateSigningRequestsReconciler) Reconcile(ctx context.Context,
-	req ctrl.Request) (res ctrl.Result, reconcileErr error) {
+	req ctrl.Request) (result ctrl.Result, reconcileErr error) {
```
aravindhp, line 78: Why are you making this changes?

In general `[condition] Send an event on OperatorCond failure ` belongs to the commit where you are introducing the `[wmco] Opt in to Upgradeable operator condition. Look at this you are logging errors, so it should be done in the commit where you are introducing the feature  and it will also reduce code churn.

pkg/condition/condition.go
aravindhp: Nice explanation @saifshaikh48. How about expanding the comment to include your explanation in a more succinct manner.

pkg/condition/condition.go
aravindhp: Not sure why this and all its associated changes leaked into this commit. This belongs to `[wmco] Opt in to Upgradeable operator condition`. @sebsoto was your comment asking for this to be moved to this commit?

saifshaikh48:
/cherry-pick release-4.7

Backporting back to 4.7 as that's when the OperatorCondition CR was first introduced by OLM.

PR #628, opened 2021-08-23
WINC-645: [controllers] Handle generic events

aravindhp:
- Simplify the check for a Windows node
- Update all controllers to handle generic events

### DIFF
--- a/controllers/controllers.go
+++ b/controllers/controllers.go
@@ -149,13 +149,8 @@ func (r *instanceReconciler) deconfigureInstance(node *core.Node) error {
 func windowsNodePredicate(byoh bool) predicate.Funcs {
 	return predicate.Funcs{
 		CreateFunc: func(e event.CreateEvent) bool {
-			if !isValidWindowsNode(e.Object, byoh) {
-				return false
-			}
-			if e.Object.GetAnnotations()[metadata.VersionAnnotation] != version.Get() {
-				return true
-			}
-			return false
+			return isValidWindowsNode(e.Object, byoh) &&
+				e.Object.GetAnnotations()[metadata.VersionAnnotation] != version.Get()
 		},
 		UpdateFunc: func(e event.UpdateEvent) bool {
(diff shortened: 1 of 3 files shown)


controllers/controllers.go
```diff
@@ -187,3 +184,15 @@ func isValidWindowsNode(o client.Object, byoh bool) bool {
...
+// isWindowsNodeWithVersionMismatch returns true if the node object is a valid Windows node and if there is a
+// version mismatch which will require the node to be updated.
+func isWindowsNodeWithVersionMismatch(o client.Object, byoh bool) bool {
+	if !isValidWindowsNode(o, byoh) {
+		return false
+	}
+	if o.GetAnnotations()[metadata.VersionAnnotation] != version.Get() {
+		return true
+	}
+	return false
```
saifshaikh48, line 197: Can this function be reduced to `return isValidWindowsNode(o, byoh) && o.GetAnnotations()[metadata.VersionAnnotation] != version.Get()`? And if so, is there a need to keep this in a helper? I am fine either way, just floating the idea.

saifshaikh48 (commented):
Thanks for taking this on @aravindhp, PTAL at my comment.

controllers/controllers.go
aravindhp: I am fine removing the helper.

mansikulkarni96: +1 not sure we need another helper, rest LGTM.

aravindhp: Addressed. PTAL.

mansikulkarni96: The PR description needs an update removing first bullet point.

aravindhp: Addressed. PTAL.

jrvaldes (approved):
/lgtm

controllers/controllers.go
saifshaikh48: /lgtm

mansikulkarni96: LGTM

PR #638, opened 2021-08-24
[docs] Document projected volume issues

aravindhp:


docs/projected-volume.md
```diff
@@ -0,0 +1,27 @@
...
+
+A [projected volume](https://kubernetes.io/docs/concepts/storage/volumes/#projected) allows mapping of several existing
+volume sources into the same directory in the container. With the [proposal for file permission handling in projected
+service account volume](https://github.com/kubernetes/enhancements/pull/1598) enhancement, the projected files have
+the correct permissions set including container user ownership for Linux pods when `RunAsUser` is set in the Pod 
+`SecurityContext`. However, for Windows pods this is not the case when `RunAsUsername` is set, due to differences in 
+the way user accounts are managed in Windows. Windows stores and manages local user and group accounts in a database
+file called Security Account Manager (SAM). This database is not shared between the Windows container host and the
+running containers. This prevents the kubelet from setting correct ownership on the files in the projected volume.
+This problem is tracked by [Windows Pod with RunAsUserName and a Projected Volume does not honor file
```
deads2k, line 11: What permissions will the files have and what owner will the files have?

docs/projected-volume.md
```diff
@@ -0,0 +1,27 @@
...
+permissions in the volume](https://github.com/kubernetes/kubernetes/issues/102849) issue.
+
+##Bound Service Account Token Volume
+The [Bound Service Account Token Volume](https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#bound-service-account-token-volume)
+feature results in the ServiceAccount admission controller adding a projected volume instead of a Secret-based volume
+for the non-expiring service account token created by the Token Controller. This implies that on all Kubernetes 1.22+ 
+clusters, a projected volume with service account token will be present on all Pods. Now for Windows Pods`RunAsUserName`
```
deads2k, line 18: space between pods and runasusername

docs/projected-volume.md
```diff
@@ -0,0 +1,27 @@
...
+is set in the Pod `SecurityContext`, the ownership will not be set correctly. This problem can get exacerbated when
```
deads2k, line 19: who will the owner be?

docs/projected-volume.md
```diff
@@ -0,0 +1,27 @@
...
+feature results in the ServiceAccount admission controller adding a projected volume instead of a Secret-based volume
+for the non-expiring service account token created by the Token Controller. This implies that on all Kubernetes 1.22+ 
+clusters, a projected volume with service account token will be present on all Pods. Now for Windows Pods`RunAsUserName`
+is set in the Pod `SecurityContext`, the ownership will not be set correctly. This problem can get exacerbated when 
+used in conjunction with [HostPath](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath) where best practices
+are not followed. For example, giving a Pod access to ` c:\var\lib\kubelet\pods\` will result in that Pod being able to
+access service account tokens from other Pods.
+
+##RunAsUser and RunAsUsername
+Creating a Windows Pod with RunAsUser and RunAsUsername in it's `SecurityContext` will result in the Pod being stuck at
```
deads2k, line 25: This is the state of windows pods on openshift in 4.8?  Is there a benefit to allowing such pods through admission or should we ship a patch on 4.8 or 1.21 to directly fail when a windows pod has this combination of thing?

docs/projected-volume.md
aravindhp: I have added what the default permissions will look like. 

>This implies all administrator users like `ContainerAdministrator` will have read, write and execute access while,
non-administrator users will have read and execute access.

docs/projected-volume.md
aravindhp: In OpenShift 4.8 we introduced a workaround https://github.com/openshift/kubernetes/pull/804 for the combination.

aravindhp (commented):
@deads2k PTAL

docs/projected-volume.md
aravindhp: >Will the container still be able to read the credentials from the bound service account token?

Yes the container will have read access to the credentials.

docs/projected-volume.md
```diff
@@ -0,0 +1,42 @@
...
+Owner  : BUILTIN\Administrators
+Group  : NT AUTHORITY\SYSTEM
+Access : NT AUTHORITY\SYSTEM Allow  FullControl
+         BUILTIN\Administrators Allow  FullControl
+         BUILTIN\Users Allow  ReadAndExecute, Synchronize
+Audit  :
+Sddl   : O:BAG:SYD:AI(A;ID;FA;;;SY)(A;ID;FA;;;BA)(A;ID;0x1200a9;;;BU)
+```
+This implies all administrator users like `ContainerAdministrator` will have read, write and execute access while,
+non-administrator users will have read and execute access.
```
deads2k, line 26: does this mean that one container could (if it were able to mount it), read the content of another serviceaccount's secret?

docs/projected-volume.md
deads2k: > In OpenShift 4.8 we introduced a workaround [openshift/kubernetes#804](https://github.com/openshift/kubernetes/pull/804) for the combination.

This should be explained here, along with the impact.  If I've understood it all correctly, the workaround created in 804 means that RunAsUser and RunAsUsername are not honored for projected volumes on Windows.  IOW, it fails open.  I think that we chose to fail open rather than to perma-fail pods in the manner you describe on these lines.

If so, can we link the PR, describe the behavior and explain the alternative we avoided?

docs/projected-volume.md
aravindhp: Correct. I.e. if `HostPath` is used to give the Pod access to `c:\var\lib\kubelet\pods\`.

docs/projected-volume.md
aravindhp: >If I've understood it all correctly, the workaround created in 804 means that RunAsUser and RunAsUsername are not honored for projected volumes on Windows.

To be clear even if just `RunAsUsername` is specified, the situation is the same i.e. the ownership is not set to username specified. However because the default permission for regular users is `BUILTIN\Users Allow  ReadAndExecute, Synchronize`, the `RunAsUsername` will have read access to files.

And yes, I will call this out and link to the PR.

aravindhp: I have now updated the PR with this info,

docs/projected-volume.md
```diff
@@ -0,0 +1,46 @@
...
+
+##Bound Service Account Token Volume
```
saifshaikh48, line 28: Seems you need a space after the `##` to render as proper markdown.

docs/projected-volume.md
```diff
@@ -0,0 +1,46 @@
...
+feature results in the ServiceAccount admission controller adding a projected volume instead of a Secret-based volume
+for the non-expiring service account token created by the Token Controller. This implies that on all Kubernetes 1.22+ 
+clusters, a projected volume will be present on all Pods. Now for Windows Pods `RunAsUserName` is set in the Pod 
+`SecurityContext`, the ownership will not be set correctly and will have the default permissions as shown above. This 
+problem can get exacerbated when used in conjunction with
+[HostPath](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath) where best practices are not followed.
+For example, giving a Pod access to ` c:\var\lib\kubelet\pods\` will result in that Pod being able to access service
+account tokens from other Pods.
+
+##RunAsUser and RunAsUsername
```
saifshaikh48, line 39: same comment here

PR #642, opened 2021-08-25
[wmco] Ensure HNS networks are removed

mansikulkarni96:
This PR introduces ensuring HNS Networks are removed by waiting and verifying the output of Get-HNSNetwork does not contain the removed network. If the network is not removed, it will retry the removal of network until it is removed.


saifshaikh48 (commented):
/lgtm

mansikulkarni96:
/retitle WINC-648: [wmco] Ensure HNS networks are removed

PR #148, opened 2020-10-29
WINC-523: [wmco] Update Node deletion during upgrade

VaishnaviHire:
Previously, during an upgrade, wmco defined minHealthy count. This count was
taken into consideration while deleting the nodes with mismatched version.
This commit updates the deletion process to take into account maxUnhealthy
count. This count ensures that we have only limited number of nodes that
are in Not Ready state during an upgrade

It also adds tests for workload availability during Upgrade process.

[WINC-523](https://issues.redhat.com/browse/WINC-523)

PR #645, opened 2021-08-30
Bug 1996615: Add pods eviction permission and logging for drainer helper

jrvaldes:
Add create permission to specific pods/eviction resource in the ConfigMap
Controller and config role.yaml.

Bundle manifests generated with:
`make bundle`

### DIFF
--- a/bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
@@ -210,6 +210,12 @@ spec:
         - apiGroups:
           - ""
           resources:
+          - pods/eviction
+          verbs:
+          - create
+        - apiGroups:
+          - ""
+          resources:
           - secrets
           verbs:
           - create
--- a/pkg/nodeconfig/nodeconfig.go
+++ b/pkg/nodeconfig/nodeconfig.go
@@ -147,7 +147,7 @@ func getClusterAddr(kubeAPIServerEndpoint string) (string, error) {
 
 // Configure configures the Windows VM to make it a Windows worker node
 func (nc *nodeConfig) Configure() error {
-	drainHelper := &drain.Helper{Ctx: context.TODO(), Client: nc.k8sclientset}
+	drainHelper := nc.newDrainHelper()
 	// If we find a node  it implies that we are reconfiguring and we should cordon the node
 	if err := nc.setNode(true); err == nil {
 		// Make a best effort to cordon the node until it is fully configured
@@ -356,6 +356,38 @@ func (nc *nodeConfig) configureCNI() error {
 	return nil
 }
 
+// ErrWriter is a wrapper to enable error-level logging inside kubectl drainer implementation
+type ErrWriter struct {
+	log logr.Logger
+}
+
+func (ew ErrWriter) Write(p []byte) (n int, err error) {
+	// log error
+	ew.log.Error(err, string(p))
+	return len(p), nil
+}
+
+// OutWriter is a wrapper to enable debug-level logging inside kubectl drainer implementation
+type OutWriter struct {
+	log logr.Logger
+}
+
+func (ow OutWriter) Write(p []byte) (n int, err error) {
+	// log debug
+	ow.log.V(1).Info(string(p))
+	return len(p), nil
+}
+
+// newDrainHelper returns new drain.Helper instance
+func (nc *nodeConfig) newDrainHelper() *drain.Helper {
+	return &drain.Helper{
+		Ctx:    context.TODO(),
+		Client: nc.k8sclientset,
+		ErrOut: &ErrWriter{nc.log},
+		Out:    &OutWriter{nc.log},
+	}
+}
+
 // Deconfigure removes the node from the cluster, reverting changes made by the Configure function
 func (nc *nodeConfig) Deconfigure() error {
 	// Set nc.node to the existing node
@@ -364,7 +396,7 @@ func (nc *nodeConfig) Deconfigure() error {
 	}
 
 	// Cordon and drain the Node before we interact with the instance
-	drainHelper := &drain.Helper{Ctx: context.TODO(), Client: nc.k8sclientset}
+	drainHelper := nc.newDrainHelper()
 	if err := drain.RunCordonOrUncordon(drainHelper, nc.node, true); err != nil {
 		return errors.Wrapf(err, "unable to cordon node %s", nc.node.GetName())
 	}
--- a/test/e2e/network_test.go
+++ b/test/e2e/network_test.go
@@ -344,6 +344,39 @@ func getAffinityForNode(node *v1.Node) (*v1.Affinity, error) {
 	}, nil
 }
 
+// getNodeAffinityForLabel returns a node affinity that matches the associated label key and values for the given
+// operator. `values` equals `nil` suppress the property, useful for `NodeSelectorOpDoesNotExist` operator.
+func getNodeAffinityForLabel(operator v1.NodeSelectorOperator, key string, values ...string) (*v1.Affinity, error) {
+	if operator == "" {
+		return nil, errors.New("operator cannot be empty")
+	}
+	if key == "" {
+		return nil, errors.New("key cannot be empty")
+	}
+	// build match expression
+	expression := v1.NodeSelectorRequirement{
+		Key:      key,
+		Operator: operator,
+	}
+	// use values, if any
+	if values != nil {
+		expression.Values = values
+	}
+	return &v1.Affinity{
+		NodeAffinity: &v1.NodeAffinity{
+			RequiredDuringSchedulingIgnoredDuringExecution: &v1.NodeSelector{
+				NodeSelectorTerms: []v1.NodeSelectorTerm{
+					{
+						MatchExpressions: []v1.NodeSelectorRequirement{
+							expression,
+						},
+					},
+				},
+			},
+		},
+	}, nil
+}
+
 // ensureNamespace checks if a namespace with the provided name exists and creates one if it does not
 func (tc *testContext) ensureNamespace(name string) error {
 	// Check if the namespace exists
--- a/test/e2e/upgrade_test.go
+++ b/test/e2e/upgrade_test.go
@@ -9,12 +9,12 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/require"
-	batchv1 "k8s.io/api/batch/v1"
 	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 
+	"github.com/openshift/windows-machine-config-operator/controllers"
 	"github.com/openshift/windows-machine-config-operator/pkg/metadata"
 )
 
@@ -31,15 +31,69 @@ const (
 	windowsWorkloadTesterJob = "windows-workload-tester"
 )
 
+// getNodeTypeAffinities generate node affinity for non-BYOH and BYOH nodes if present
+func getNodeTypeAffinities() ([]*v1.Affinity, error) {
+	var affinities []*v1.Affinity
+	if gc.numberOfMachineNodes >= 1 {
+		affinityNoByoh, err := getNodeAffinityForLabel(v1.NodeSelectorOpDoesNotExist, controllers.BYOHLabel)
+		if err != nil {
+			return nil, errors.Wrapf(err, "error creating node affinity without BYOH label")
+		}
+		affinities = append(affinities, affinityNoByoh)
+	}
+	if gc.numberOfBYOHNodes >= 1 {
+		affinityByoh, err := getNodeAffinityForLabel(v1.NodeSelectorOpIn, controllers.BYOHLabel, "true")
+		if err != nil {
+			return nil, errors.Wrapf(err, "error creating node affinity for BYOH label")
+		}
+		affinities = append(affinities, affinityByoh)
+	}
+	return affinities, nil
+}
+
+// generateLinuxWorkloadTesterCommand creates a Job to curl Windows webserver every 5 seconds.
+// If the webserver is not accessible for few minutes Curl Windows webserver
+// exit the pod with an error.
+func generateLinuxWorkloadTesterCommand(clusterIP string) []string {
+	return []string{
+		"bash",
+		"-c",
+		" while true;" +
+			" do curl " + clusterIP + ";" +
+			" if [ $? != 0 ]; then" +
+			" sleep 60;" +
+			" curl " + clusterIP + " || exit 1;" +
+			" fi; " +
+			"sleep 5; " +
+			"done"}
+}
+
 // upgradeTestSuite tests behaviour of the operator when an upgrade takes place.
 func upgradeTestSuite(t *testing.T) {
 	testCtx, err := NewTestContext()
 	require.NoError(t, err)
 
-	// test if Windows workloads are running by creating a Job that curls the workloads continuously.
-	testerJob, err := testCtx.deployWindowsWorkloadAndTester()
-	require.NoError(t, err, "error testing Windows workloads")
-	defer testCtx.deleteJob(testerJob.Name)
+	// generate node affinities
+	affinities, err := getNodeTypeAffinities()
+	require.NoError(t, err, "error getting node affinity")
+	// loop affinities to ensure deployed workloads will exclusively run based
+	// on the given affinity, e.g.  BYOH and/or Machine nodes
+	for _, affinity := range affinities {
+		// test if Windows workloads are running by creating a Job that curls the workloads continuously.
+		deployment, err := testCtx.deployWindowsWebServer("win-webserver", affinity)
+		require.NoErrorf(t, err, "error creating Windows Webserver deployment for upgrade test")
+		defer testCtx.deleteDeployment(deployment.Name)
+
+		// create a clusterIP service which can be used to reach the Windows webserver
+		intermediarySVC, err := testCtx.createService(deployment.Name, v1.ServiceTypeClusterIP, *deployment.Spec.Selector)
+		require.NoErrorf(t, err, "error creating service for deployment %s", deployment.Name)
+		defer testCtx.deleteService(intermediarySVC.Name)
+
+		// create a Job object that continuously curls the webserver every 5 seconds.
+		testerJob, err := testCtx.createLinuxJob(windowsWorkloadTesterJob, generateLinuxWorkloadTesterCommand(intermediarySVC.Spec.ClusterIP))
+		require.NoErrorf(t, err, "error creating linux job %s", windowsWorkloadTesterJob)
+		defer testCtx.deleteJob(testerJob.Name)
+	}
 
 	// apply configuration steps before running the upgrade tests
 	err = testCtx.configureUpgradeTest()
@@ -68,6 +122,7 @@ func testUpgradeVersion(t *testing.T) {
 	// Test if prometheus is reconfigured with ip addresses of newly configured nodes
 	testPrometheus(t)
 
+	// TODO: Fix matching label for jobs. See https://issues.redhat.com/browse/WINC-673
 	// Test if there was any downtime for Windows workloads by checking the failure on the Job pods.
 	pods, err := testCtx.client.K8s.CoreV1().Pods(testCtx.workloadNamespace).List(context.TODO(),
 		metav1.ListOptions{FieldSelector: "status.phase=Failed",
@@ -150,34 +205,4 @@ func (tc *testContext) scaleWMCODeployment(desiredReplicas int32) error {
 	})
 
 	return err
-}
-
-// deployWindowsWorkloadAndTester tests if the Windows Webserver deployment is available.
-// This is achieved by creating a Job object that continuously curls the webserver every 5 seconds.
-func (tc *testContext) deployWindowsWorkloadAndTester() (*batchv1.Job, error) {
-	// Create a Windows Webserver deployment
-	deployment, err := tc.deployWindowsWebServer("win-webserver", nil)
-	if err != nil {
-		return nil, errors.Wrap(err, "error creating Windows Webserver deployment for upgrade test")
-	}
-	defer tc.deleteDeployment(deployment.Name)
-
-	// Create a clusterIP service which can be used to reach the Windows webserver
-	intermediarySVC, err := tc.createService(deployment.Name, v1.ServiceTypeClusterIP, *deployment.Spec.Selector)
-	if err != nil {
-		return nil, errors.Wrap(err, "could not create service ")
-	}
-	defer tc.deleteService(intermediarySVC.Name)
-	// Create a Job to curl Windows webserver
-	// Curl Windows webserver every 5 seconds. If the webserver is not accessible for few minutes
-	// exit the pod with an error.
-	command := []string{"bash", "-c",
-		" while true; do curl " + intermediarySVC.Spec.ClusterIP + "; if [ $? != 0 ]; then sleep 60; curl " +
-			intermediarySVC.Spec.ClusterIP + "|| exit 1;" + " fi; sleep 5; done"}
-
-	job, err := tc.createLinuxJob(windowsWorkloadTesterJob, command)
-	if err != nil {
-		return nil, err
-	}
-	return job, nil
 }
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -71,6 +71,12 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - pods/eviction
+  verbs:
+  - create
+- apiGroups:
+  - ""
+  resources:
   - secrets
   verbs:
   - create
--- a/controllers/configmap_controller.go
+++ b/controllers/configmap_controller.go
@@ -48,6 +48,7 @@ import (
 //+kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups="",resources=configmaps/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups="",resources=configmaps/finalizers,verbs=update
+//+kubebuilder:rbac:groups="",resources=pods/eviction,verbs=create
 
 const (
 	// BYOHLabel is a label that should be applied to all Windows nodes not associated with a Machine.


test/e2e/upgrade_test.go
```diff
@@ -41,6 +43,18 @@ func upgradeTestSuite(t *testing.T) {
 	require.NoError(t, err, "error testing Windows workloads")
 	defer testCtx.deleteJob(testerJob.Name)
 
+	// check BYOH enabled
+	if gc.numberOfBYOHNodes >= 1 {
```
mansikulkarni96, line 47: consider adding a require to check number of byohnodes is 1 instead of if statement to make sure we are always deploying workloads.

test/e2e/upgrade_test.go
```diff
@@ -41,6 +43,18 @@ func upgradeTestSuite(t *testing.T) {
...
+		// ensure the presence of a running workload in BYOH node
+		byohDeployment, err := testCtx.deployByohWorkload()
+		require.NoError(t, err, "error deploying BYOH workload")
+		// cleanup workload
+		defer func(testCtx *testContext, name string) {
+			err := testCtx.deleteDeployment(name)
+			require.NoErrorf(t, err, "error deleting deployment %s", name)
```
mansikulkarni96, line 54: since deleteDeployment returns an error you can just check on the function and defer that
require.NoError(t, err, testCtx.deleteDeployment(name))

jrvaldes: Nice!

test/e2e/upgrade_test.go
jrvaldes: As per our review session, that's a good point. I feel the step of actually enforcing as a hard requirement the presence of at least 1 BYOH node in the e2e-upgrade tests should takes place elsewhere.

The question is: Must the e2e-upgrade test run with at least 1 BYOH node?

mansikulkarni96: I think it should, how else would we ensure this particular scenario of having a pod and upgrading is tested?

pkg/nodeconfig/nodeconfig.go
```diff
@@ -356,6 +356,30 @@ func (nc *nodeConfig) configureCNI() error {
...
 
+// ErrWriter is a wrapper to enable error-level logging inside kubectl drainer implementation
+type ErrWriter struct {
+	log logr.Logger
+}
+
+func (ew ErrWriter) Write(p []byte) (n int, err error) {
+	// log error
+	ew.log.Error(err, string(p))
+	// return
```
sebsoto, line 367: Comments not needed

pkg/nodeconfig/nodeconfig.go
```diff
@@ -356,6 +356,30 @@ func (nc *nodeConfig) configureCNI() error {
...
+
+// OutWriter is a wrapper to enable debug-level logging inside kubectl drainer implementation
+type OutWriter struct {
+	log logr.Logger
+}
+
+func (ow OutWriter) Write(p []byte) (n int, err error) {
+	// log debug
+	ow.log.V(1).Info(string(p))
+	// return
```
sebsoto, line 379: Comments not needed

pkg/nodeconfig/nodeconfig.go
```diff
@@ -364,7 +388,12 @@ func (nc *nodeConfig) Deconfigure() error {
 	}
 
 	// Cordon and drain the Node before we interact with the instance
-	drainHelper := &drain.Helper{Ctx: context.TODO(), Client: nc.k8sclientset}
+	drainHelper := &drain.Helper{
+		Ctx:    context.TODO(),
+		Client: nc.k8sclientset,
+		ErrOut: &ErrWriter{nc.log},
+		Out:    &OutWriter{nc.log},
+	}
```
sebsoto, line 396: The other instance of `drainHelper` in `nc.Configure()` should be changed as well.

Since the object initialization is a bit more complicated now, how about creating a new constructor:
```
func (nc *nodeConfig) newDrainHelper() *drain.Helper {
  return &drain.Helper{
		Ctx:    context.TODO(),
		Client: nc.k8sclientset,
		ErrOut: &ErrWriter{nc.log},
		Out:    &OutWriter{nc.log},
	}
}
```

test/e2e/upgrade_test.go
sebsoto: We should not fail the test on cleanup errors

test/e2e/upgrade_test.go
```diff
@@ -41,6 +43,18 @@ func upgradeTestSuite(t *testing.T) {
 	require.NoError(t, err, "error testing Windows workloads")
 	defer testCtx.deleteJob(testerJob.Name)
 
+	// check BYOH enabled
+	if gc.numberOfBYOHNodes >= 1 {
+		// ensure the presence of a running workload in BYOH node
+		byohDeployment, err := testCtx.deployByohWorkload()
+		require.NoError(t, err, "error deploying BYOH workload")
```
sebsoto, line 50: You're deploying a workload here, but what is actually being tested by doing so?

test/e2e/upgrade_test.go
```diff
@@ -78,6 +92,30 @@ func testUpgradeVersion(t *testing.T) {
 
 }
 
+// deployByohWorkload deploys a workload in a BYOH node, using BYOH-label affinity as NodeSelector
+func (tc *testContext) deployByohWorkload() (*appsv1.Deployment, error) {
```
sebsoto, line 96: Why does this need to be a separate function from the general `deployWindowsWorkloadAndTester()`?
 https://github.com/openshift/windows-machine-config-operator/blob/7c4ed1e3950757d077ba7482f39c4af68ff25429/test/e2e/upgrade_test.go#L157

That function could be modified to take an affinity.

test/e2e/upgrade_test.go
sebsoto: Take note of https://github.com/openshift/windows-machine-config-operator/blob/7c4ed1e3950757d077ba7482f39c4af68ff25429/test/e2e/upgrade_test.go#L71-L77

sebsoto: Please try and use the logic already present in the linked code and in deployWindowsWorkloadAndTester() instead of creating a new deployment.

sebsoto (changes requested):

test/e2e/network_test.go
```diff
@@ -344,6 +344,27 @@ func getAffinityForNode(node *v1.Node) (*v1.Affinity, error) {
 	}, nil
```
saifshaikh48, line 344: typo in commit message: `suit` => `suite`

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
```diff
@@ -207,6 +207,12 @@ spec:
           verbs:
```
saifshaikh48, line 207: Is this commit also part of the fix? If so, please add `fixes bz#____` to the commit message.

test/e2e/upgrade_test.go
saifshaikh48: What we could do is break this into a separate test function and then lead it off with a check+skip as we do in other places. Example: https://github.com/openshift/windows-machine-config-operator/blob/6ba37ed99a01fdebf124c5098b02b0dffc6537f0/test/e2e/reconfigure_test.go#L65-L68

test/e2e/upgrade_test.go
jrvaldes: Sounds good, will adjust.

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
jrvaldes: Actually, this commit does not `fixes` the bug, It's a prep work for https://github.com/openshift/windows-machine-config-operator/pull/645/commits/bf2a331e4d0043ffbfa422f233de741d15e5a119.

Thanks for calling this out, will add more context in the commit message.

pkg/nodeconfig/nodeconfig.go
jrvaldes: Should the `drain.Helper` instance be part of the [nodeConfig struct](https://github.com/openshift/windows-machine-config-operator/blob/master/pkg/nodeconfig/nodeconfig.go#L48)? So, we don't create a new instance on each constructor call.

test/e2e/upgrade_test.go
jrvaldes: What would be the specific test to skip, if no BYOH node is present?

mansikulkarni96: I think the suggestion is to skip the deployment on byoh node.

test/e2e/upgrade_test.go
jrvaldes: > You're deploying a workload here, but what is actually being tested by doing so?

As per comment: `// ensure the presence of a running workload in BYOH node`

The logic already present [deletes the deployment](https://github.com/openshift/windows-machine-config-operator/blob/master/test/e2e/upgrade_test.go#L163) before the [upgrade process is triggered](https://github.com/openshift/windows-machine-config-operator/blob/master/test/e2e/upgrade_test.go#L104), namely updating the version annotation.

So, the new workload is introduced to last during the upgrade process.

It does makes sense to accommodate the existing workload deployment [testCtx.deployWindowsWorkloadAndTester()](https://github.com/openshift/windows-machine-config-operator/blob/master/test/e2e/upgrade_test.go#L40) to persist during the upgrade.

test/e2e/upgrade_test.go
jrvaldes: Sure, we can move in that direction. The only show stopper is that the[ deployment get deleted ](https://github.com/openshift/windows-machine-config-operator/blob/master/test/e2e/upgrade_test.go#L163)before the upgrade process is triggered

mansikulkarni96: @jrvaldes to reuse logic you could  modify existing function deployWindowsWorkloadAndTester() to add the affinity parameter and call it in the upgrade test suite wherever required . One way to deal with the issue you are facing could be to can move the creation of deployment outside of the function and use that name to delete the deployment at the end of the suite.

test/e2e/upgrade_test.go
```diff
@@ -31,15 +31,71 @@ const (
 	windowsWorkloadTesterJob = "windows-workload-tester"
 )
 
+// getNodeAffinities generate node affinity for non-BYOH and BYOH nodes if present
+func getNodeAffinities() ([]*v1.Affinity, error) {
+	var affinities []*v1.Affinity
+	if gc.numberOfMachineNodes >= 1 {
+		affinityNoByoh, err := getNodeAffinityForLabel(v1.NodeSelectorOpDoesNotExist, controllers.BYOHLabel)
```
mansikulkarni96, line 38: I think we should name this variable different for byoh and no byoh scenario.

test/e2e/upgrade_test.go
```diff
@@ -31,15 +31,71 @@ const (
...
+
+			// create a clusterIP service which can be used to reach the Windows webserver
+			intermediarySVC, err := testCtx.createService(deployment.Name, v1.ServiceTypeClusterIP, *deployment.Spec.Selector)
+			require.NoErrorf(t, err, "error creating service for deployment %s", deployment.Name)
+			defer testCtx.deleteService(intermediarySVC.Name)
+
+			// create a Job object that continuously curls the webserver every 5 seconds.
+			testerJob, err := testCtx.createLinuxJob(windowsWorkloadTesterJob, getLinuxJobCommand(intermediarySVC.Spec.ClusterIP))
+			require.NoErrorf(t, err, "error creating linux job %s", windowsWorkloadTesterJob)
+			defer testCtx.deleteJob(testerJob.Name)
```
mansikulkarni96, line 96: This snippet could be a helper function.
I am also a little confused by this function, are the defer calls executed at the end of entire suite? Just making sure this implementation is solving the problem of not deleting deployments before running upgrade tests.

jrvaldes: Moving the loop content to a helper is similar to what it's been [removed](https://github.com/openshift/windows-machine-config-operator/pull/645/commits/3551877c5aca09efea84533bb131bb105505ff78#diff-fc8ea8c7c1a732fd9a425771cb3463d1cd143e5527acec751f32ba88f2df697aL157), in order to allow the deferred calls to execute in the current context, so the deployments do not get deleted before the upgrade process starts.

jrvaldes: > I am also a little confused by this function, are the defer calls executed at the end of entire suite?

No, they are executed just before [upgradeTestSuite](https://github.com/openshift/windows-machine-config-operator/pull/645/commits/3551877c5aca09efea84533bb131bb105505ff78#diff-fc8ea8c7c1a732fd9a425771cb3463d1cd143e5527acec751f32ba88f2df697aR72)() returns

jrvaldes: > Just making sure this implementation is solving the problem of not deleting deployments before running upgrade tests.

That's the main driver for bringing the [deployWindowsWorkloadAndTester()](https://github.com/openshift/windows-machine-config-operator/pull/645/commits/3551877c5aca09efea84533bb131bb105505ff78#diff-fc8ea8c7c1a732fd9a425771cb3463d1cd143e5527acec751f32ba88f2df697aL157) content here, to ensure deployments are running while running the upgrade test.

test/e2e/upgrade_test.go
jrvaldes: Good catch, missed this one.

jrvaldes:
/test vsphere-e2e-operator
due to network failures

test/e2e/upgrade_test.go
```diff
@@ -31,15 +31,70 @@ const (
...
+		require.NoErrorf(t, err, "error creating service for deployment %s", deployment.Name)
+
+		// create a Job object that continuously curls the webserver every 5 seconds.
+		testerJob, err := testCtx.createLinuxJob(windowsWorkloadTesterJob, getLinuxJobCommand(intermediarySVC.Spec.ClusterIP))
+		require.NoErrorf(t, err, "error creating linux job %s", windowsWorkloadTesterJob)
+
+		// stack the deferred calls in reverse to ensure the right execution order (LIFO)
+		defer testCtx.deleteJob(testerJob.Name)
+		defer testCtx.deleteService(intermediarySVC.Name)
+		defer testCtx.deleteDeployment(deployment.Name)
```
sebsoto, line 96: Please call these defer functions immediately after confirming the resource created successfully, like so:
```
		deployment, err := testCtx.deployWindowsWebServer("win-webserver", affinity)
		require.NoErrorf(t, err, "error creating Windows Webserver deployment for upgrade test")
 		defer testCtx.deleteDeployment(deployment.Name)
```

sebsoto: As it is right now there is a risk of leaking resources if one of the creations fails

test/e2e/upgrade_test.go
```diff
@@ -31,15 +31,70 @@ const (
...
+		}
+		affinities = append(affinities, affinityByoh)
+	}
+	return affinities, nil
+}
+
+// getLinuxJobCommand creates a Job to curl Windows webserver every 5 seconds.
+// If the webserver is not accessible for few minutes Curl Windows webserver
+// exit the pod with an error.
+func getLinuxJobCommand(clusterIP string) []string {
```
sebsoto, line 57: getLinuxJobCommand is very broad, how about `linuxWorkloadTesterCommand()`?

test/e2e/upgrade_test.go
```diff
@@ -151,33 +206,3 @@ func (tc *testContext) scaleWMCODeployment(desiredReplicas int32) error {
 
 	return err
 }
-
-// deployWindowsWorkloadAndTester tests if the Windows Webserver deployment is available.
-// This is achieved by creating a Job object that continuously curls the webserver every 5 seconds.
-func (tc *testContext) deployWindowsWorkloadAndTester() (*batchv1.Job, error) {
```
sebsoto, line 157: Please call out in the commit message that the defers were causing an issue, so you need to break this logic out of this function.

test/e2e/upgrade_test.go
```diff
@@ -31,15 +31,70 @@ const (
 	windowsWorkloadTesterJob = "windows-workload-tester"
 )
 
+// getNodeAffinities generate node affinity for non-BYOH and BYOH nodes if present
+func getNodeAffinities() ([]*v1.Affinity, error) {
```
sebsoto, line 35: This name made me think the function was returning a slice of affinities for each individual node, how about `getNodeTypeAffinities()` instead? Or something like that. Just something a little more specific :)

test/e2e/upgrade_test.go
```diff
@@ -31,15 +31,70 @@ const (
...
 	require.NoError(t, err)
 
-	// test if Windows workloads are running by creating a Job that curls the workloads continuously.
-	testerJob, err := testCtx.deployWindowsWorkloadAndTester()
-	require.NoError(t, err, "error testing Windows workloads")
-	defer testCtx.deleteJob(testerJob.Name)
+	// generate node affinities
+	affinities, err := getNodeAffinities()
+	require.NoError(t, err, "error getting node affinity")
+	// loop affinities
```
sebsoto, line 79: Comments explain what is happening but not why they are happening, would prefer something like:
```
// Deploy separate workloads which will exclusively run on either BYOH or Machine nodes
```

test/e2e/upgrade_test.go
```diff
@@ -31,15 +31,70 @@ const (
...
+
+		// create a Job object that continuously curls the webserver every 5 seconds.
+		testerJob, err := testCtx.createLinuxJob(windowsWorkloadTesterJob, getLinuxJobCommand(intermediarySVC.Spec.ClusterIP))
+		require.NoErrorf(t, err, "error creating linux job %s", windowsWorkloadTesterJob)
+
+		// stack the deferred calls in reverse to ensure the right execution order (LIFO)
+		defer testCtx.deleteJob(testerJob.Name)
+		defer testCtx.deleteService(intermediarySVC.Name)
+		defer testCtx.deleteDeployment(deployment.Name)
+	}
```
sebsoto, line 97: The existing methodology for checking that the workloads did not fail is flawed:
https://github.com/openshift/windows-machine-config-operator/blob/f24ca6c568285cdbd2f0e146a70a64adadb96071/test/e2e/upgrade_test.go#L71-L74

Its just checking that the number of found failed pods is 0. But if the label selector isn't actually picking up pods from the jobs being created that number will always be 0. I suspect that that is what is currently occurring.

To make things clear with an example:
Lets say you are creating 2 jobs `windows-workload-tester-job-one` and `windows-workload-tester-job-two`
The selector is looking for failed pods with the label: `job-name=windows-workload-tester-job`. It won't find any failed pods because it wouldnt find any pods with that label at all. It needs to be doing two seperate calls to find failed pods with the label `job-name=windows-workload-tester-job-one` and `job-name=windows-workload-tester-job-two`

Instead you could have a slice which keeps track of jobs
```
var workloadTesterJobs []*batch.jobs
```
add the jobs to it as you create them, and move the check for failed pods to this function, where you can use the generated job name `job.GetName()` to search for the pods. 
But you cant just search for failed pods, you need to list all the pods and ensure there is at least one pod, and then check if any of them are in the failed state.
```
require.NotEmpty(t, pods)
// then assert status.Phase != true
```

test/e2e/upgrade_test.go
saifshaikh48, line 96: `defer` calls are implicitly LIFO so I think this is the reverse order of what we want. Currently it will delete the deployment first, which is the first thing created.

sebsoto (changes requested):
Thanks @jrvaldes I understand what you're going for and its definitely on the right track

test/e2e/upgrade_test.go
saifshaikh48, line 97: I don't see why `deployWindowsWorkloadAndTester()` needed to be removed. My suggestion would be to keep the helper and just add an `affinity *v1.Affinity` parameter to it and leave it as a helper. So changes here would be reduced to:
 
```suggestion
	// generate node affinities
	affinities, err := getNodeAffinities()
	require.NoError(t, err, "error getting node affinity")
	// loop affinities
	for _, affinity := range affinities {
		// test if Windows workloads are running by creating a Job that curls the workloads continuously.
		testerJob, err := testCtx.deployWindowsWorkloadAndTester(affinity)
		require.NoError(t, err, "error testing Windows workloads")
		defer testCtx.deleteJob(testerJob.Name)
	}
```

saifshaikh48 (commented):
Thanks for working on this, a few suggestions:

test/e2e/upgrade_test.go
saifshaikh48: I still think this deferral ordering is not correct.
How it stands now:
- order of creation: deployment, service, job
- order of deletion: deployment, service, job

We want order of deletion: job, service, deployment so the calls would be: 
1. defer testCtx.deleteDeployment(deployment.Name)
2. defer testCtx.deleteService(intermediarySVC.Name)
3. defer testCtx.deleteJob(testerJob.Name)

[This suggestion](https://github.com/openshift/windows-machine-config-operator/pull/645/files#r701922247) fixes this anyway.

test/e2e/upgrade_test.go
jrvaldes: Created [WINC-673](https://issues.redhat.com/browse/WINC-673)

test/e2e/upgrade_test.go
jrvaldes: Addressed.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -356,6 +356,38 @@ func (nc *nodeConfig) configureCNI() error {
 	return nil
 }
 
+// ErrWriter is a wrapper to enable error-level logging inside kubectl drainer implementation
```
aravindhp, line 359: Such structs should not be added to the middle of a file but rather at the end. This will make it easier to find and maintain the file as it grows.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -356,6 +356,38 @@ func (nc *nodeConfig) configureCNI() error {
...
+}
+
+// OutWriter is a wrapper to enable debug-level logging inside kubectl drainer implementation
+type OutWriter struct {
+	log logr.Logger
+}
+
+func (ow OutWriter) Write(p []byte) (n int, err error) {
+	// log debug
+	ow.log.V(1).Info(string(p))
```
aravindhp, line 377: Ideally we need a INFO log when we are draining and evicting pods

test/e2e/upgrade_test.go
```diff
@@ -31,15 +31,69 @@ const (
...
-	// test if Windows workloads are running by creating a Job that curls the workloads continuously.
-	testerJob, err := testCtx.deployWindowsWorkloadAndTester()
-	require.NoError(t, err, "error testing Windows workloads")
-	defer testCtx.deleteJob(testerJob.Name)
+	// generate node affinities
+	affinities, err := getNodeTypeAffinities()
+	require.NoError(t, err, "error getting node affinity")
+	// loop affinities to ensure deployed workloads will exclusively run based
+	// on the given affinity, e.g.  BYOH and/or Machine nodes
+	for _, affinity := range affinities {
```
aravindhp, line 81: Do we really need this affinity business? Wouldn't having a `Deployment` with replicas set to n+1 nodes be sufficient to distribute the pods across the BYOH and Machine nodes?

test/e2e/upgrade_test.go
```diff
@@ -31,15 +31,69 @@ const (
...
+		}
+		affinities = append(affinities, affinityByoh)
+	}
+	return affinities, nil
+}
+
+// generateLinuxWorkloadTesterCommand creates a Job to curl Windows webserver every 5 seconds.
+// If the webserver is not accessible for few minutes Curl Windows webserver
+// exit the pod with an error.
+func generateLinuxWorkloadTesterCommand(clusterIP string) []string {
```
aravindhp, line 57: We don't want multiple ways calling the curler jobs. Please see [testWindowsExporter](https://github.com/openshift/windows-machine-config-operator/blob/master/test/e2e/metrics_test.go#L37). We ideally should have a single method that kicks of a curler job with different inputs.

test/e2e/upgrade_test.go
```diff
@@ -31,15 +31,69 @@ const (
...
+	// loop affinities to ensure deployed workloads will exclusively run based
+	// on the given affinity, e.g.  BYOH and/or Machine nodes
+	for _, affinity := range affinities {
+		// test if Windows workloads are running by creating a Job that curls the workloads continuously.
+		deployment, err := testCtx.deployWindowsWebServer("win-webserver", affinity)
+		require.NoErrorf(t, err, "error creating Windows Webserver deployment for upgrade test")
+		defer testCtx.deleteDeployment(deployment.Name)
+
+		// create a clusterIP service which can be used to reach the Windows webserver
+		intermediarySVC, err := testCtx.createService(deployment.Name, v1.ServiceTypeClusterIP, *deployment.Spec.Selector)
```
aravindhp, line 88: Not sure why this intermediate service is needed.

test/e2e/upgrade_test.go
```diff
@@ -31,15 +31,69 @@ const (
...
+		require.NoErrorf(t, err, "error creating service for deployment %s", deployment.Name)
+		defer testCtx.deleteService(intermediarySVC.Name)
+
+		// create a Job object that continuously curls the webserver every 5 seconds.
+		testerJob, err := testCtx.createLinuxJob(windowsWorkloadTesterJob, generateLinuxWorkloadTesterCommand(intermediarySVC.Spec.ClusterIP))
```
aravindhp, line 93: Please modularize and make this common across here and [testWindowsExporter](https://github.com/openshift/windows-machine-config-operator/blob/master/test/e2e/metrics_test.go#L37).

test/e2e/upgrade_test.go
jrvaldes: Same here

pkg/nodeconfig/nodeconfig.go
aravindhp: I am fine with that too.

pkg/nodeconfig/nodeconfig.go
aravindhp: I would like to see a log when we are draining and evicting pods and not for all situations. How that is achieved is up to you.

test/e2e/upgrade_test.go
aravindhp: Given this test code we want to keep the complexity as low as possible. This seems unnecessarily complex.

test/e2e/upgrade_test.go
aravindhp: Then why did you add it without understanding why this service is needed?

jrvaldes: It was added in https://github.com/openshift/windows-machine-config-operator/pull/148/commits/9d8001b47784c256bee666c72f6715cbdc2882e8

aravindhp: That was added for a different reason. You seemed have copied it over without figuring if it is needed in your scenario or not.

jrvaldes: AFAIK, it is the same scenario.

aravindhp: You are just wanting to ensure that pods get evicted. Do you need an intermediate service to ping?

jrvaldes: > You are just wanting to ensure that pods get evicted. Do you need an intermediate service to ping?

Correct, your statement is accurate. That was my initial implementation, but refactored a bit to reuse the [logic already present](https://github.com/openshift/windows-machine-config-operator/pull/645#discussion_r700390441)

aravindhp: I can even understand if all the code was made common. Here it seems you reproduced what was being done in the windows exporter test case.

aravindhp: Moreover, you could have potentially left the original workloads created in other tests running without having to create new ones here. And yes, there would be an ordering issue here. Not sure if that was not discussed before or not.

test/e2e/upgrade_test.go
jrvaldes: Created [WINC-677](https://issues.redhat.com/browse/WINC-677)

test/e2e/upgrade_test.go
jrvaldes: @aravindhp Thanks for highlighting this.  No, the ordering issue was not discussed during the review. I feel relying on previously created workloads by other tests, introduces a hard dependency for the `e2e-upgrade` test, and may justify the creation of new workloads here.

aravindhp: >I feel relying on previously created workloads by other tests, introduces a hard dependency for the e2e-upgrade test, and may justify the creation of new workloads here.

OK, that's a fair point and that's the statement I wanted to hear.

test/e2e/upgrade_test.go
aravindhp: This could have still remained a helper method that took affinity as an input. You could have then introduced another method that cleans up the resources created. We should always think more deeply before removing modularization and inlining a chunk of code.

jrvaldes: That's a fair point, +1 for modularization, but this is an outlier; 

Affinity is no longer needed, so it's out of the equation. Will remove the complex "affinity business" in follow-up PR;

The chunk of code inlined here is concise, and with the modularization of the [LinuxWorkloadTesterCommand](https://github.com/openshift/windows-machine-config-operator/pull/645/commits/c85df6109de0af3485ae3583b98b27c207f06f19#diff-fc8ea8c7c1a732fd9a425771cb3463d1cd143e5527acec751f32ba88f2df697aR57) generation. it aligns better in the parent execution thread, rather than passing in-context parameters back in order to defer the cleanup of the created resources.

aravindhp: >The chunk of code inlined here is concise

Well your definition of concise differs from mine 😄

aravindhp: >Affinity is no longer needed, so it's out of the equation. Will remove the complex "affinity business" in follow-up PR;

And given you are doing this, please reintroduce the function. It is always easier to read:
```go
	for x := range y {
		foobarstuff()
	}
```
rather than
```go
	for x := range y {
		// foo
		// bar
		// stuff
	}
```
Doesn't matter your definition of concise 😉

test/e2e/upgrade_test.go
jrvaldes: Adjusted in https://github.com/openshift/windows-machine-config-operator/pull/674

pkg/nodeconfig/nodeconfig.go
jrvaldes: Fixed in https://github.com/openshift/windows-machine-config-operator/pull/676

pkg/nodeconfig/nodeconfig.go
jrvaldes: Refactored in https://github.com/openshift/windows-machine-config-operator/pull/676/commits/febc428b765ed99b8928424f79d04079bb75f331

test/e2e/upgrade_test.go
jrvaldes: Agree, and again +1 for modularization. You may be missing the [main reason for the refactoring](https://github.com/openshift/windows-machine-config-operator/pull/645#discussion_r700542776): bring the content of `deployWindowsWorkloadAndTester()` to the parent thread given the content of the function is simple, to ensure the deletion of the workloads happen after the node upgrade is complete and tested.

In layman's terms, reads better

```go
		foo := Foo()
                defer close(foo)
		bar := Bar()
                defer close(bar)
```

rather than

```go
		foo, bar := FooBar()
                defer close(foo)
                defer close(bar)
```

where `foo` and `bar` are orthogonal.

aravindhp: OK, I understand now as to why you removed the function. However making the argument that I want use defer, so I will not moduralize is going to put us in a place where we will have ever growing functions. This seems to be case where the test needs to have a setup and teardown methods. Given the entities being created are finite, we don't need `foo, bar := FooBar()` but rather `createFooBar()` and `deleteFooBar()` given we know what is being created and what needs to be destroyed now.

aravindhp: However this fix does not explain how the changed code will expose errors like we were seeing with the bug in place. Don't we need to check after the upgrade is complete that the workload continues to run and function?

jrvaldes: Yes, the logic is there already, just need a minor [adjustment](https://github.com/openshift/windows-machine-config-operator/pull/645#discussion_r701947458).

jrvaldes: >Given the entities being created are finite, we don't need foo, bar := FooBar() but rather createFooBar() and deleteFooBar() given we know what is being created and what needs to be destroyed now.

Correct, entities are finite and created with randomly generated names. The later, may jeopardize the happy flow, since we need to keep track of the generated names, preferred over increasing the numbers of  API calls to fetch the resources.

aravindhp: I will let this go for now but I don't want to see this pattern of demodularizing repeated in the future.

jrvaldes: Agree, that's an antipattern.

PR #646, opened 2021-08-30
[hack] Support vSphere platform in machineset.sh 

jrvaldes:
Include the machineSet generation for vSphere platform
in machineset.sh script.


hack/machineset.sh
```diff
@@ -173,6 +173,74 @@ $(get_spec $infraID $az $provider)
...
+  cat <<EOF
+$(get_spec $infraID "" $provider)
+      providerSpec:
+        value:
+          apiVersion: vsphereprovider.openshift.io/v1beta1
+          credentialsSecret:
+            name: ${credentialsSecret}
+          diskGiB: 128
+          kind: VSphereMachineProviderSpec
+          memoryMiB: 8192
```
aravindhp, line 222: Please use 16384 as we are in our e2e test.

hack/machineset.sh
```diff
@@ -173,6 +173,74 @@ $(get_spec $infraID $az $provider)
...
+          metadata:
+            creationTimestamp: null
```
aravindhp, line 224: This is not needed. Yes we are doing this in our other templates and it is a bad pattern. PS: not asking you to change this in other providers in this PR.

hack/machineset.sh
```diff
@@ -173,6 +173,74 @@ $(get_spec $infraID $az $provider)
...
+
+  local infraID=$1
+  local provider=$2
+
+  # set golden image template name
+  # TODO: read from parameter
+  template="windows-golden-images/windows-server-2004-template"
+
+  # get vsphere providerSpec for any worker machine
+  # TODO: filter machines in running phase
```
aravindhp, line 191: The real TODO is to make just one `oc get machines` call and just pass the data around.  This is the 3rd call being introduced and can be avoided. PS: not asking for that change here.

aravindhp (changes requested):
Thanks for doing this, @jrvaldes. Just a couple of small changes requested.

hack/machineset.sh
jrvaldes: Good catch, will adjust.

hack/machineset.sh
```diff
@@ -173,6 +173,74 @@ $(get_spec $infraID $az $provider)
...
+          network:
+            devices:
+            - networkName: ${networkName}
+          numCPUs: 4
+          numCoresPerSocket: 1
+          snapshot: ""
+          template: ${template}
+          # TODO: remove; WMCO does not use userDataSecret in vSphere platform
+          userDataSecret:
+            name: windows-user-data
```
jrvaldes, line 234: Planning to remove `userDataSecret` since not used in vSphere platform.

Thoughts?

jrvaldes: Ended up removing the `userDataSecret` section for vSphere machineSet; 
a) YAGNI, the pre-reqs are met in the golden image, and
b) allows creating a machineSet before WMCO is deployed (BYOH use case). Otherwise, we depends on WMCO to create `windows-user-data`.

saifshaikh48 (commented):
/lgtm

selansen (approved):
/lgtm

sebsoto:
@saifshaikh48 tested that this works
/hold cancel
/approve

PR #581, opened 2021-08-11
[wmco] Update k8s version to 1.22.0-rc.0

aravindhp:
- [vendor] Update k8s libraries to v0.22.0-rc.0
- [wmco] Update k8s version to v1.22.0-rc.0
- [submodule][kubelet] Update to 9b1230e8847

PR #649, opened 2021-08-30
[config] Update the k8s version in the CSV

jrvaldes:
Update the k8s version to v1.22 in the CSV

Follow-up to https://github.com/openshift/windows-machine-config-operator/pull/581/commits/6698f0296c05ca2e534c1f72265945c0ffbe98a0


mansikulkarni96:
Thanks for fixing this @jrvaldes 
PR can be reframed/retitled since the change is part of the config and not the bundle directly.

aravindhp (approved):
/approve

PR #663, opened 2021-09-07
Bug 2000772: [wmco] Use sc stderr to determine 1060 error

sebsoto:
This commit changes the way we are checking sc commands for acceptable
1060 errors, by now checking the actual output of the command, and not
the returned error code. In some scenarios a 1060 error was not being
returned, and instead a generic error code of 1 was being returned. This
was causing the error to not be properly recognized.

Fixes BZ#2000772


saifshaikh48 (commented):
/lgtm

docs/byoh-instance-pre-requisites.md
```diff
@@ -4,11 +4,10 @@ The following pre-requisites must be fulfilled in order to add a Windows BYOH no
...
-  * The default shell for the SSH server must be set to [cmd.exe](https://docs.microsoft.com/en-us/windows-server/administration/openssh/openssh_server_configuration#configuring-the-default-shell-for-openssh-in-windows).
 * Port 10250 must be open in order for log collection to function.
 * An administrator user is present with the [private key used in the secret](/README.md#create-a-private-key-secret) set as an authorized SSH key.
 * The hostname of the instance must follow the [RFC 1123](https://datatracker.ietf.org/doc/html/rfc1123) DNS label standard:
   * Contain only lowercase alphanumeric characters or '-'.
   * Start with an alphanumeric character.
   * End with an alphanumeric character.
-* A PTR record must exist corresponding to the instance address which resolves to the instance hostname for successful reverse DNS lookups. 
\ No newline at end of file
+* A PTR record must exist corresponding to the instance address which resolves to the instance hostname for successful reverse DNS lookups.
```
selansen, line 13: looks like just a extra space got remove . can we undo this ?

selansen (approved):
/lgtm

pkg/windows/windows.go
```diff
@@ -191,12 +193,30 @@ func New(workerIgnitionEndpoint, vxlanPort string, instanceInfo *instance.Info,
 		return nil, errors.Wrapf(err, "unable to setup VM %s sshConnectivity", instanceInfo.Address)
 	}
 
+	// When in powershell, this will evaluate to an object containing information about the shell.
+	// When in cmd.exe, this will not evaluate to anything, and will be used literally. This can be used to determine
+	// if cmd or Powershell is being used.
+	powershellVersionTable := "($PSVersionTable)"
```
jrvaldes, line 199: Nice logic using `powershellVersionTable`.

pkg/windows/windows.go
```diff
@@ -191,12 +193,30 @@ func New(workerIgnitionEndpoint, vxlanPort string, instanceInfo *instance.Info,
...
+	out, err := conn.run("echo " + powershellVersionTable)
+	if err != nil {
+		return nil, errors.Wrap(err, "error determining default shell")
+	}
+	var powershellDefault bool
+	if strings.TrimSpace(out) == powershellVersionTable {
+		powershellDefault = false
+	} else if strings.Contains(out, "PSEdition") {
```
jrvaldes, line 207: Is the presence of `PSEdition` good enough to claim powershell as default?  

Nit; if you are  looking after `PSEdition` you can fetch and compare the value using `$PSVersionTable.PSEdition`

pkg/windows/windows.go
```diff
@@ -191,12 +193,30 @@ func New(workerIgnitionEndpoint, vxlanPort string, instanceInfo *instance.Info,
...
+		powershellDefault = true
+	} else {
+		return nil, errors.Wrapf(err, "unexpected output when determining default shell: %s", out)
```
jrvaldes, line 210: In case the default shell is other than cmd or powershell, should WMCO default to powershell instead? If so, why don't always default to powershell?

jrvaldes (commented):
@sebsoto  thanks for working on this, comments added PTAL

docs/byoh-instance-pre-requisites.md
jrvaldes: LGTM; IMO fixing a surrounding typo does not hurt.

pkg/windows/windows.go
aravindhp: This looks fragile and more like a hack to me. Shouldn't you be querying whatever registry entry that defines the default shell?

Moreover do we really need to figure out what the default shell is? Can't we always launch a PowerShell given the that cmd commands will also work in PowerShell?

test/e2e/validation_test.go
```diff
@@ -289,13 +288,31 @@ func (tc *testContext) sshSetup() error {
 	return nil
 }
 
-// runSSHJob creates and waits for a Kubernetes job to run. The command provided will be executed on the host specified
-// by the provided IP.
-func (tc *testContext) runSSHJob(name, command, ip string) (string, error) {
+// runPowershellSSHJob creates and waits for a Kubernetes job to run. The command provided will be executed through
+// powershell, on the host specified by the provided IP.
+func (tc *testContext) runPowershellSSHJob(name, command, ip string) (string, error) {
```
aravindhp, line 293: Why is this change required? Is this because the job could land on a node that has the PowerShell as default? I also suggest leaving the name as `runSSHJob`. The only reason to make it more verbose would be if we were also running command prompt jobs.

test/e2e/validation_test.go
```diff
@@ -289,13 +288,31 @@ func (tc *testContext) sshSetup() error {
...
+				"ssh -o StrictHostKeyChecking=no -i %s %s@%s ${CMD_PREFIX}' %s '${CMD_SUFFIX}",
+			filepath.Join(keyMountDir, secrets.PrivateKeySecretKey), tc.vmUsername(), ip,
+			filepath.Join(keyMountDir, secrets.PrivateKeySecretKey), tc.vmUsername(), ip, command)}
+
+	return tc.runToolsJob(name, sshCommand)
+}
+
+// runTools creates and waits for a Kubernetes job to run. The command provided will be executed on a linux worker,
+// using the tools image.
+func (tc *testContext) runToolsJob(name string, command []string) (string, error) {
```
aravindhp, line 313: Shouldn't this just be `runJob`? What is `tool` signifying here?

test/e2e/validation_test.go
```diff
@@ -289,13 +288,31 @@ func (tc *testContext) sshSetup() error {
...
-// by the provided IP.
-func (tc *testContext) runSSHJob(name, command, ip string) (string, error) {
+// runPowershellSSHJob creates and waits for a Kubernetes job to run. The command provided will be executed through
+// powershell, on the host specified by the provided IP.
+func (tc *testContext) runPowershellSSHJob(name, command, ip string) (string, error) {
+	keyMountDir := "/private-key"
+	sshCommand := []string{"bash", "-c",
+		fmt.Sprintf(
+			// first determine if the host has powershell or cmd as the default shell
+			"PSVERSION_TABLE=$(ssh -o StrictHostKeyChecking=no -i %s %s@%s 'echo ($PSVersionTable)');"+
```
aravindhp, line 298: This looks really fragile. There must be a cleaner way to figure out what the default shell is.

pkg/windows/connectivity.go
```diff
@@ -54,6 +54,9 @@ type sshConnectivity struct {
 	// sshClient is the client used to access the Windows VM via ssh
 	sshClient *ssh.Client
 	log       logr.Logger
+	// defaultShellPowershell indicates if the default SSH shell is powershell. If it is not, it is assumed cmd.exe is
+	// the default shell.
+	defaultShellPowershell bool
```
aravindhp, line 59: Why is this needed here and in the windows struct?

docs/byoh-instance-pre-requisites.md
aravindhp: While I am fine with this instance, fixing surrounding typos needs to be taken on a case by case basis. I am seeing other examples of more invasive changes and excuse is that I was in the area.

test/e2e/create_test.go
```diff
@@ -143,6 +143,29 @@ func (tc *testContext) testBYOHConfiguration(t *testing.T) {
...
 	require.NoError(t, err, "Machines did not reach expected state")
+
+	// Change the default shell to Powershell for the first BYOH VM. We expect there to be BYOH VMs with either
+	// cmd or powershell as the default shell, so getting a mix of both, either between different BYOH VMs, or with the
+	// VMs spun up for the Machine testing is important.
+
+	// This needs to be retried as it will only succeed once the VM is fully initialized
+	// This should not increase the overall test time, as WMCO would have to wait for the VM to be ready anyway.
+	err = wait.Poll(nodeRetryInterval, time.Minute*10, func() (done bool, err error) {
+		command := "New-ItemProperty -Path \"HKLM:\\SOFTWARE\\OpenSSH\" -Name DefaultShell " +
```
aravindhp, line 154: This changes the default shell for SSH to PowerShell. Shouldn't you be replacing the default `Command Prompt` with `PowerShell`? IOW, does changing the default prompt to PowerShell translate into it being the default in SSH or is that a separate step? Can I can have the Windows instance default to `PowerShell` but still use `cmd` in SSH?

test/e2e/create_test.go
```diff
@@ -143,6 +143,29 @@ func (tc *testContext) testBYOHConfiguration(t *testing.T) {
...
+			log.Printf("Machine %s does not have a valid address, retrying...", machines.Items[0].GetName())
+			return false, nil
+		}
+		_, err = tc.runPowershellSSHJob("change-default-shell", command, addr)
+		if err != nil {
+			log.Printf("failed to change %s default shell to Powershell, retrying...", machines.Items[0].GetName())
+			return false, nil
+		}
+		return true, nil
+	})
```
aravindhp, line 167: Please make this a separate helper method like `changeDefualtShell` or something to that effect.

aravindhp (changes requested):
Thanks for working on this @sebsoto. 
- I wondering if this PR is doing more than just fixing the bug. Is it possible to fix the bug with a smaller delta and then have a follow up "feature" PR that transitions WMCO to fully using PowerShell (For example: stop using sc.exe)?
- Please use `PowerShell`

test/e2e/create_test.go
sebsoto: I'm not sure what you mean. Are you saying if a user is physically accessing the VM? In that case if they had `Windows Server with Desktop Experience`  they would either open the CMD or PowerShell executables. If using a different remote management tool the shell would need to be configured for that tool.

test/e2e/validation_test.go
sebsoto: That it uses the using the tools image

test/e2e/validation_test.go
sebsoto: This function was required for the same reason we need to determine the default shell in the operator code, running `powershell.exe ... -Command "..."` within powershell, was causing parsing issues that I couldn't resolve.

I created a separate function for this to separate the logic of determining the shell and modifying the command based on that, from the logic of running the job and everything that comes with that.

pkg/windows/windows.go
sebsoto: In general defaulting to powershell would not be the way to go, if there would be a default it would using the `powershell -Command` wrapper. But I'd prefer the default case to error out early, and prevent undefined behavior later, which should reduce the amount of hard to solve bugs :)

pkg/windows/windows.go
sebsoto: Actually, looking at the registry adds a requirement that instances must be using OpenSSH as their SSH server... I'm not sure if we want to do that.

pkg/windows/windows.go
```diff
@@ -843,14 +843,15 @@ func (vm *windows) removeHNSNetwork(networkName string) error {
 
 // Generic helper methods
 
-// mkdirCmd returns the Windows command to create a directory if it does not exists
+// mkdirCmd returns the Powershell command to ensure a directory at the given path exists.
 func mkdirCmd(dirName string) string {
-	return "if not exist " + dirName + " mkdir " + dirName
+	// A trailing space is required here as paths ending in `\` will cause parsing issues for the go SSH library
+	return fmt.Sprintf("New-Item -ItemType Directory -Force -Path %s ", dirName)
```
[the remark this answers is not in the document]
sebsoto: You're correct though, ill make the minimum delta required, and keep mkdir

pkg/windows/windows.go
```diff
@@ -843,14 +843,15 @@ func (vm *windows) removeHNSNetwork(networkName string) error {
...
 }
 
-// rmDirCmd returns the Windows command to recursively remove a directory if it exists
+// rmDirCmd returns the Powershell command to recursively remove a directory if it exists.
 func rmDirCmd(dirName string) string {
-	return "if exist " + dirName + " rmdir " + dirName + " /s /q"
+	return fmt.Sprintf("if(Test-Path -Path %s){rm -r %s}", dirName, dirName)
```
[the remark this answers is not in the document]
sebsoto: When ran in powershell rmdir becomes an alias for Remove-Item and the passed in arguments no longer work
```
2021-09-30T15:19:03.845Z        ERROR   wc ip-10-0-158-244.ec2.internal error running   {"cmd": "powershell.exe -NonInteractive -ExecutionPolicy Bypass if(Test-Path -Path C:\\k\\){ rmdir C:\\k\\ /s /q }", "out": "Remove-Item : A positional parameter cannot be found that accepts argument '/s'.\r\nAt line:1 char:28\r\n+ if(Test-Path -Path C:\\k\\){ rmdir C:\\k\\ /s /q }\r\n+                            ~~~~~~~~~~~~~~~~~\r\n    + CategoryInfo          : InvalidArgument: (:) [Remove-Item], ParameterBindingException\r\n    + FullyQualifiedErrorId : PositionalParameterNotFound,Microsoft.PowerShell.Commands.RemoveItemCommand\r\n \r\n", "error": "Process exited with status 1"}
```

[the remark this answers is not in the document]
aravindhp: Have you considered using `cmd /c rmdir` all the time?

test/e2e/validation_test.go
sebsoto: Since we cannot check the default shell in the operator code through the registry, unless we want to force users to use OpenSSH, I'd prefer to keep the test code consistent with what we are doing there.

test/e2e/create_test.go
```diff
@@ -151,6 +159,27 @@ func (tc *testContext) testBYOHConfiguration(t *testing.T) {
 	})
 }
 
+// setPowerShellDefaultShell changes the instance backed by the given Machine to have a default SSH shell of PowerShell
+func (tc *testContext) setPowerShellDefaultShell(machine *mapi.Machine) error {
+	// This needs to be retried as it will only succeed once the VM is fully initialized
+	// This should not increase the overall test time, as WMCO would have to wait for the VM to be ready anyway.
+	return wait.Poll(nodeRetryInterval, time.Minute*10, func() (done bool, err error) {
```
saifshaikh48, line 166: Curious if there's a constant laying around we can use instead of just the `time.Minute*10` value?

test/e2e/create_test.go
```diff
@@ -151,6 +159,27 @@ func (tc *testContext) testBYOHConfiguration(t *testing.T) {
 	})
 }
 
+// setPowerShellDefaultShell changes the instance backed by the given Machine to have a default SSH shell of PowerShell
+func (tc *testContext) setPowerShellDefaultShell(machine *mapi.Machine) error {
+	// This needs to be retried as it will only succeed once the VM is fully initialized
```
saifshaikh48, line 164: I don't quite understand why the retry logic is needed. I thought that since we already `waitForWindowsMachines()` to be in `Provisioned` state, the machine should already have an address by the time `setPowerShellDefaultShell` is called?

test/e2e/validation_test.go
```diff
@@ -289,13 +288,31 @@ func (tc *testContext) sshSetup() error {
...
+func (tc *testContext) runPowerShellSSHJob(name, command, ip string) (string, error) {
+	keyMountDir := "/private-key"
+	sshCommand := []string{"bash", "-c",
+		fmt.Sprintf(
+			// first determine if the host has PowerShell or cmd as the default shell
+			"PSVERSION_TABLE=$(ssh -o StrictHostKeyChecking=no -i %s %s@%s 'echo ($PSVersionTable)');"+
+				"if [[ \"$PSVERSION_TABLE\" == *\"PSEdition\"* ]]; "+
+				"then CMD_PREFIX=\"\";CMD_SUFFIX=\"\";"+
+				// if PowerShell is not the default shell, explicitly run the command through PowerShell
+				"else CMD_PREFIX=\"powershell.exe -NonInteractive -ExecutionPolicy Bypass -Command \\\"\";CMD_SUFFIX=\"\\\"\";"+
```
saifshaikh48, line 302: I think a constant `remotePowerShellCmdPrefix` already exists in the test suite holding the same value as `CMD_PREFIX` here. Might make this a bit cleaner.

test/e2e/validation_test.go
```diff
@@ -289,13 +288,31 @@ func (tc *testContext) sshSetup() error {
...
+				"fi;"+
+				"ssh -o StrictHostKeyChecking=no -i %s %s@%s ${CMD_PREFIX}' %s '${CMD_SUFFIX}",
+			filepath.Join(keyMountDir, secrets.PrivateKeySecretKey), tc.vmUsername(), ip,
+			filepath.Join(keyMountDir, secrets.PrivateKeySecretKey), tc.vmUsername(), ip, command)}
+
+	return tc.runToolsJob(name, sshCommand)
+}
+
+// runTools creates and waits for a Kubernetes job to run. The command provided will be executed on a linux worker,
```
saifshaikh48, line 311: `runTools` --> `runToolsJob`

test/e2e/validation_test.go
aravindhp: I don't see any reason for the name change given we were always using the tools image.

aravindhp (changes requested):

pkg/windows/windows.go
```diff
@@ -250,7 +250,7 @@ func (vm *windows) Run(cmd string, psCmd bool) (string, error) {
 	if err != nil {
 		// Hack to not print the error log for "sc.exe qc" returning 1060 for non existent services
 		// and not print error when the command takes too long to return after removing HNS networks.
-		if !(strings.HasPrefix(cmd, serviceQueryCmd) && strings.HasSuffix(err.Error(), serviceNotFound)) &&
+		if !(strings.Contains(cmd, serviceQueryCmd) && strings.Contains(out, serviceNotFound)) &&
```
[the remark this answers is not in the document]
sebsoto: I've checked on all ways to run this (powershell, powershell with cmd, cmd, cmd with powershell), and the output always contained the `Failed 1060` error. I think it would be superfluous to check for  `status 1060` in the error message.

test/e2e/delete_test.go
```diff
@@ -81,12 +81,12 @@ func (tc *testContext) testBYOHRemoval(t *testing.T) {
 // checkDirsDoNotExist returns true if the required directories do not exist on the Windows instance with the given
 // address
 func (tc *testContext) checkDirsDoNotExist(address string) (bool, error) {
-	command := remotePowerShellCmdPrefix
```
saifshaikh48, line 84: You've removed all usages of `remotePowerShellCmdPrefix` but not [the constant itself](https://github.com/openshift/windows-machine-config-operator/pull/663/files#diff-cd50fad6da0839c40258ca356bf995a7ee5e1d5c270ee0af2cc5752e2f10a454R25)

test/e2e/validation_test.go
```diff
@@ -289,13 +288,31 @@ func (tc *testContext) sshSetup() error {
...
+func (tc *testContext) runPowerShellSSHJob(name, command, ip string) (string, error) {
+	keyMountDir := "/private-key"
+	sshCommand := []string{"bash", "-c",
+		fmt.Sprintf(
+			// first determine if the host has PowerShell or cmd as the default shell by running a simple PowerShell
+			// command. If it succeeds, then the host's default shell is PowerShell
+			"if ssh -o StrictHostKeyChecking=no -i %s %s@%s 'Get-Help';"+
+				"then CMD_PREFIX=\"\";CMD_SUFFIX=\"\";"+
+				// if PowerShell is not the default shell, explicitly run the command through PowerShell
+				"else CMD_PREFIX=\""+windows.RemotePowerShellCmdPrefix+" -Command \\\"\";CMD_SUFFIX=\"\\\"\";"+
```
saifshaikh48, line 302: nit: don't think the extra space before `-Command` is needed as the `windows.RemotePowerShellCmdPrefix` const already ends in a space.

test/e2e/validation_test.go
```diff
@@ -289,13 +288,31 @@ func (tc *testContext) sshSetup() error {
...
+				"fi;"+
+				"ssh -o StrictHostKeyChecking=no -i %s %s@%s ${CMD_PREFIX}' %s '${CMD_SUFFIX}",
```
saifshaikh48, line 304: potential spacing nit: are the spaces surrounding the 4th `%s` arg needed for powershell's sake? Just wondering

test/e2e/create_test.go
```diff
@@ -29,6 +29,10 @@ import (
 	"github.com/openshift/windows-machine-config-operator/test/e2e/clusterinfo"
 )
 
+// vmConfigurationTime is the maximum amount of time expected for a Windows VM to be fully configured and ready for WMCO
+// after the hardware is provisioned.
+const vmConfigurationTime = 8 * time.Minute
```
aravindhp, line 34: We use [10 minutes](https://github.com/openshift/windows-machine-config-operator/blob/master/pkg/retry/retry.go#L11) for SSH connectivity to go through in the WMCO code. Shouldn't we match that? PS: not asking to reuse the same const but to match the time.

aravindhp (approved):
Nice work on this @sebsoto. 

/approve

with the caveat that other review comments are addressed.

test/e2e/delete_test.go
sebsoto: Thanks, I forgot this constant existed. Using this as it is already in-use, instead of exporting the one from the `Windows` package.

mansikulkarni96:
LGTM, will let @saifshaikh48 give the final lgtm for his comments.

saifshaikh48:
/lgtm 

:rocket:

PR #664, opened 2021-09-07
Bug 2001547: Validate instance addresses before deconfiguring BYOH nodes

saifshaikh48:
The purpose of this PR is to properly validate and retrieve usable addresses 
for instances specified in the `windows-instances` ConfigMap. 

Each network address is checked to ensure it is either:
- an IPv4 address _or_
- a DNS name that resolves to an IPv4 address

This is needed to support instances that are added to the ConfigMap
using DNS names that do not match their hostnames.
Previously, doing direct comparisons using unresolved instance addresses
missed BYOH node associations, causing premature deconfiguration.

Fixes BZ#2001547


sebsoto:
/bugzilla refresh
changed target release to 4.10

sebsoto (changes requested):

saifshaikh48:
/hold

test env to reproduce

controllers/configmap_controller.go
```diff
@@ -211,8 +211,14 @@ func (r *ConfigMapReconciler) deconfigureInstances(instances []*instance.Info, n
 // hasAssociatedInstance returns true if the given node is associated with an instance in the given slice
 func hasAssociatedInstance(node *core.Node, instances []*instance.Info) bool {
 	for _, instanceInfo := range instances {
+		instanceAddress, err := instanceInfo.GetIPv4Address()
+		if err != nil {
+			// Current instance does not have a valid address to compare
+			continue
+		}
```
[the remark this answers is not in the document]
saifshaikh48: I like this approach, thanks for the suggestion.

controllers/configmap_controller.go
```diff
@@ -195,9 +195,14 @@ func (r *ConfigMapReconciler) deconfigureInstances(instances []*instance.Info, n
 		Namespace: r.watchNamespace}}
 	for _, node := range nodes.Items {
 		// Check for instances associated with this node
-		if hasEntry := hasAssociatedInstance(&node, instances); hasEntry {
+		hasEntry, err := hasAssociatedInstance(&node, instances)
+		if err != nil {
+			return errors.Wrapf(err, "failure checking for instances associated with node %s", node.GetName())
```
aravindhp, line 200: I would prefer `unable to find instance associated with node`

pkg/instance/instance.go
```diff
@@ -59,3 +80,29 @@ func (i *Info) UpgradeRequired() bool {
...
 }
+
+// GetIPv4Address returns an IPv4 address of the instance. An error will be thrown if the instance object has
+// an IPv6 address or a DNS address that does not resolve to an IPv4 network address.
+func (i *Info) GetIPv4Address() (string, error) {
+	if ip := net.ParseIP(i.Address); ip != nil {
+		// Address is either an IPv6 or IPv4 address
+		ipv4 := ip.To4()
+		if ipv4 == nil {
+			return "", newNoIPv4Err(errors.Errorf("error using IP %s: only IPv4 addresses are supported", ip.String()))
```
aravindhp, line 91: Would the final error read `No IPv4 address found: error using IP: only IPv4 addresses are supported`?

controllers/configmap_controller.go
```diff
@@ -208,16 +213,27 @@ func (r *ConfigMapReconciler) deconfigureInstances(instances []*instance.Info, n
...
+// hasAssociatedInstance returns true if the given node is associated with an instance in the given slice.
+// The instance's network address must be an IPv4 address or resolve to one.
+func hasAssociatedInstance(node *core.Node, instances []*instance.Info) (bool, error) {
 	for _, instanceInfo := range instances {
+		instanceAddress, err := instanceInfo.GetIPv4Address()
+		if err != nil {
+			var ipv4Err *instance.NoIPv4Error
+			if errors.As(err, &ipv4Err) {
+				// Current instance does not have a valid address to compare
+				continue
```
aravindhp, line 225: So are we silently ignoring this entry in the `ConfigMap`?

pkg/instance/instance.go
```diff
@@ -1,12 +1,33 @@
...
+	"github.com/pkg/errors"
 	core "k8s.io/api/core/v1"
 
 	"github.com/openshift/windows-machine-config-operator/pkg/metadata"
 	"github.com/openshift/windows-machine-config-operator/version"
 )
 
+// NoIPv4Error occurs when trying to retrieve an IPv4 address from an instance whose network address is IPv6 or a DNS
+// name that does not resolve to an IPv4 address
+type NoIPv4Error struct {
```
aravindhp, line 16: Do we really need the `Error` suffix?

controllers/configmap_controller.go
```diff
@@ -208,16 +213,27 @@ func (r *ConfigMapReconciler) deconfigureInstances(instances []*instance.Info, n
...
 
-// hasAssociatedInstance returns true if the given node is associated with an instance in the given slice
-func hasAssociatedInstance(node *core.Node, instances []*instance.Info) bool {
+// hasAssociatedInstance returns true if the given node is associated with an instance in the given slice.
+// The instance's network address must be an IPv4 address or resolve to one.
+func hasAssociatedInstance(node *core.Node, instances []*instance.Info) (bool, error) {
 	for _, instanceInfo := range instances {
+		instanceAddress, err := instanceInfo.GetIPv4Address()
+		if err != nil {
+			var ipv4Err *instance.NoIPv4Error
```
aravindhp, line 222: `noIPv4`

aravindhp (changes requested):
Thanks for working on this, @saifshaikh48. Please address my questions.

pkg/instance/instance.go
saifshaikh48: I will make the inner error more clear and less repetitive: `unable to use IPv6 address <ip>`

pkg/instance/instance.go
saifshaikh48: I see k8s errors do not have the `Error` suffix so I will remove follow that convention.

pkg/instance/instance.go
aravindhp: But that assumes the error is always because an IPv6 address is present. Could there be other reasons?

controllers/configmap_controller.go
saifshaikh48: My reasoning was that this helper method should not be concerned with invalid entries as its focus is to match any entries to the given node. The question is, should we avoid deconfiguring any nodes until an unrelated issue with the config map is resolved?

saifshaikh48: I do see that swallowing the error could hide any config map errors from the user. So I think it's best to return an error here instead of continuing.

aravindhp: >My reasoning was that this helper method should not be concerned with invalid entries as its focus is to match any entries to the given node. 

The how do we know that we have an invalid entry?

>The question is, should we avoid deconfiguring any nodes until an unrelated issue with the config map is resolved?

I don't follow this question.

pkg/windows/windows.go
```diff
@@ -205,27 +204,7 @@ func New(workerIgnitionEndpoint, vxlanPort string, instanceInfo *instance.Info,
 // Interface methods
 
 func (vm *windows) GetIPv4Address() (string, error) {
-	if ip := net.ParseIP(vm.instance.Address); ip != nil {
```
aravindhp, line 208: Please call out that you are doing this refactoring in the commit message and why you are doing so.

pkg/instance/instance.go
```diff
@@ -59,3 +62,29 @@ func (i *Info) UpgradeRequired() bool {
...
+		return ipv4.String(), nil
+	}
+
+	// DNS address in this case
+	ips, err := net.LookupIP(i.Address)
+	if err != nil {
+		return "", errors.Wrapf(err, "lookup of address %s failed", i.Address)
+	}
+	// Get first IPv4 address returned
+	for _, returnedIP := range ips {
```
aravindhp, line 84: This just occurred to me:
What if there are multiple IPv4 addresses and the first IPv4 address for whatever reason is not accessible but one of the others are? I guess this could also be a problem when the node is getting configured also and the question should have been asked where this was introduced.

aravindhp (changes requested):

aravindhp (changes requested):
>This is needed to properly support instances that are added
to the `windows-instances` ConfigMap using a DNS name, preventing
premature deconfiguration of BYOH nodes.

I also would like to see why the premature configuration is happening being explained in the commit and PR message.

pkg/instance/instance.go
```diff
@@ -59,3 +62,29 @@ func (i *Info) UpgradeRequired() bool {
 	// fully deconfigured before being configured by the current version.
 	return true
 }
+
+// GetIPv4Address returns an IPv4 address of the instance. An error will be thrown if the instance object has
+// an IPv6 address or a DNS address that does not resolve to an IPv4 network address.
+func (i *Info) GetIPv4Address() (string, error) {
+	if ip := net.ParseIP(i.Address); ip != nil {
+		// Address is either an IPv6 or IPv4 address
+		ipv4 := ip.To4()
```
mansikulkarni96, line 71: you can directly do:
if net.ParseIP(i.Address).To4()  != nil {
}

saifshaikh48: if `net.ParseIP(i.Address)` returns `nil` wouldn't calling `.To4()` on it cause a panic?

pkg/instance/instance.go
saifshaikh48: Addressed with new commit

pkg/instance/instance.go
mansikulkarni96: It does not, tried it [here](https://play.golang.org/p/-uCW6RvUGQg)

pkg/nodeutil/nodeutil.go
```diff
@@ -6,13 +6,13 @@ import (
 
 // FindByAddress returns a pointer to the node within the given list with an address matching the given address
 // and a bool indicating if the node was found or not.
```
sebsoto, line 8: comment needs to be changed

controllers/controllers.go
```diff
@@ -107,7 +107,11 @@ func (r *instanceReconciler) instanceFromNode(node *core.Node) (*instance.Info,
 		return nil, errors.Wrapf(err, "unable to decrypt username annotation for node %s", node.Name)
 	}
 
-	return instance.NewInfo(addr, username, "", false, node), nil
+	instanceInfo, err := instance.NewInfo(addr, username, "", false, node)
+	if err != nil {
+		return nil, err
+	}
+	return instanceInfo, nil
```
sebsoto, line 114: can just `return instance.NewInfo(addr, username, "", false, node)`

pkg/instance/instance.go
```diff
@@ -59,3 +70,29 @@ func (i *Info) UpgradeRequired() bool {
 	// fully deconfigured before being configured by the current version.
 	return true
 }
+
+// getIPv4Address returns an IPv4 address associated with the given address. An error will be thrown if given
+// an IPv6 address or a DNS address that does not resolve to an IPv4 network address.
+func getIPv4Address(address string) (string, error) {
```
sebsoto, line 76: nit: the naming makes it sound like a getter
consider something else like `resolveIPv4Address`
not a big deal considering this isnt a method of the `Info` struct.

sebsoto (changes requested):

controllers/controllers.go
aravindhp: +1

pkg/instance/instance.go
```diff
@@ -9,8 +12,11 @@ import (
 
 // Info represents a instance that is meant to be joined to the cluster
 type Info struct {
-	// Address is the network address of the instance
+	// Address is the network address of the instance as specified by the associated ConfigMap entry.
+	// Must be an IPv4 address or a DNS name that resolves to one.
 	Address string
+	// IPv4Address is the IPv4 address associated with the instance's given Address. May be the same value.
+	IPv4Address string
```
aravindhp, line 19: Do we need separate entries? Do we really need the DNS entry stored? IOW can we just use the IPv4 address everywhere? PS: this is a question as I have not looked at all the places where `Address` and whether it needs to be storied. If you find using the IPv4 address everywhere is involved, then we can leave it for a follow up PR. Otherwise just store the IPv4 address in `Address`.

pkg/instance/instance.go
aravindhp: However, as per the [docs](https://pkg.go.dev/net#ParseIP):
>If s is not a valid textual representation of an IP address, ParseIP returns nil.

pkg/instance/instance.go
```diff
@@ -59,3 +70,29 @@ func (i *Info) UpgradeRequired() bool {
...
 }
+
+// getIPv4Address returns an IPv4 address associated with the given address. An error will be thrown if given
+// an IPv6 address or a DNS address that does not resolve to an IPv4 network address.
+func getIPv4Address(address string) (string, error) {
+	if ip := net.ParseIP(address); ip != nil {
+		// Address is either an IPv6 or IPv4 address
+		ipv4 := ip.To4()
+		if ipv4 == nil {
+			return "", errors.Errorf("IPv6 network address not supported: %s", ip.String())
```
aravindhp, line 81: Nit: I would rather say not an IPv4 address given the [doc](https://pkg.go.dev/net#IP.To4) says:
>If ip is not an IPv4 address, To4 returns nil.

controllers/configmap_controller.go
```diff
@@ -208,16 +214,32 @@ func (r *ConfigMapReconciler) deconfigureInstances(instances []*instance.Info, n
...
 
-// hasAssociatedInstance returns true if the given node is associated with an instance in the given slice
-func hasAssociatedInstance(node *core.Node, instances []*instance.Info) bool {
-	for _, instanceInfo := range instances {
-		for _, nodeAddress := range node.Status.Addresses {
-			if instanceInfo.IPv4Address == nodeAddress.Address {
-				return true
+// hasAssociatedInstance returns true if the given node is associated with any instance in the given slice.
+// The instance's network address must be a valid IPv4 address or resolve to one.
+func hasAssociatedInstance(node *core.Node, instances []*instance.Info) (bool, error) {
```
aravindhp, line 219: Given you are changing this method, I suggest passing just `node.Status.Addresses` given you are not using anything else in the node object.

aravindhp (changes requested):

pkg/instance/instance.go
saifshaikh48: The two reservations I had with consolidating to 1 entry were
1. Error messages and logs would be hard to follow. For example, a DNS entry in the ConfigMap would be resolved to some IP, and only this IP would be present in all output messages which could be tricky to understand.
2. I am not sure how using the resolved IP address would affect [the CSR logic ](https://github.com/openshift/windows-machine-config-operator/blob/6901ac3ed2891b7627a1431ab66703358b902671/pkg/csr/csr.go#L320-L336)

Would suggest a follow-up PR.

saifshaikh48: Story created here https://issues.redhat.com/browse/WINC-675.

pkg/instance/instance.go
saifshaikh48: Agreed, will make a change to `resolveToIPv4Address`

pkg/instance/instance.go
saifshaikh48: I will opt to keep this as is; allows for more accurate error reporting.

pkg/instance/instance.go
saifshaikh48: I am not certain what you mean here. I assume you mean not specifying `IPv6` in the error message. After the `parseIP` call we know that `To4` is called on either an IPv4 or an IPv6 address (whether mapped or unmapped), so it is accurate to do so IMO.

pkg/instance/instance.go
aravindhp: More than accurate error reporting, `net.Parse()` can return null if the string is not a valid IP address.

pkg/instance/instance.go
aravindhp: Not an IPv4 address needn't imply a IPv6 address is the point I am making. So it is safer to say `not an IPv4 address`.

aravindhp (approved):
/approve

mansikulkarni96:
verified @sebsoto comments have been addressed as well.
/lgtm

saifshaikh48:
/hold

until PR fixing deprecated way of referencing images

aravindhp:
/retest-required

East-west networking flake

PR #669, opened 2021-09-14
[vsphere] create machineset spec based on machineset spec created at install time

rvanderp3:
The intent of this PR is to derive the Workspace for machinesets created during testing based on the compute machineset provisioned by the cluster at installation.  The reason for this change is that a vCenter in IBM Cloud is being also used to run CI jobs from a shared lease pool to increase test coverage in some key configurations.   Additionally, the vmTemplate can now be optionally defined as an environment variable in the [e2e test script](https://github.com/openshift/release/blob/master/ci-operator/step-registry/windows/e2e/operator/test/windows-e2e-operator-test-commands.sh).

### DIFF
--- a/test/e2e/providers/vsphere/vsphere.go
+++ b/test/e2e/providers/vsphere/vsphere.go
@@ -1,8 +1,10 @@
 package vsphere
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
+	"log"
 	"os"
 
 	"github.com/pkg/errors"
@@ -32,10 +34,24 @@ func New(clientset *clusterinfo.OpenShift) (*Provider, error) {
 }
 
 // newVSphereMachineProviderSpec returns a vSphereMachineProviderSpec generated from the inputs, or an error
-func newVSphereMachineProviderSpec(clusterID string) (*vsphere.VSphereMachineProviderSpec, error) {
+func (p *Provider) newVSphereMachineProviderSpec(clusterID string) (*vsphere.VSphereMachineProviderSpec, error) {
 	if clusterID == "" {
 		return nil, fmt.Errorf("clusterID is empty")
 	}
+	workspace, err := p.getWorkspaceFromExistingMachineSet(clusterID)
+	if err != nil {
+		return nil, err
+	}
+	log.Printf("creating machineset provider spec which targets %s\n", workspace.Server)
+
+	// The template is an image which has been properly sysprepped.  The image is derived from an environment variable
+	// defined in the job spec.
+	vmTemplate := os.Getenv("VM_TEMPLATE")
+	if vmTemplate == "" {
+		vmTemplate = "windows-golden-images/windows-server-2004-template"
+	}
+
+	log.Printf("creating machineset based on template %s\n", vmTemplate)
 
 	return &vsphere.VSphereMachineProviderSpec{
 		TypeMeta: meta.TypeMeta{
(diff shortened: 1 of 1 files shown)


rvanderp3:
Confirmed in both VMC and the IBM Cloud that the workspace properly populates based on the existing compute node workspace and that a windows machine is created.  Also confirmed that the `VM_TEMPLATE` environment variable is able to override the default template.

test/e2e/providers/vsphere/vsphere.go
```diff
@@ -1,8 +1,11 @@
 package vsphere
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
sebsoto, line 7: Looks like this is already imported as `meta`, below. You should use that one instead

test/e2e/providers/vsphere/vsphere.go
```diff
@@ -52,19 +66,44 @@ func newVSphereMachineProviderSpec(clusterID string) (*vsphere.VSphereMachinePro
...
-		// TODO: Find a way to automatically update this with latest image
-		Template: "windows-golden-images/windows-server-2004-template",
-		Workspace: &vsphere.Workspace{
-			Datacenter:   "SDDC-Datacenter",
-			Datastore:    "WorkloadDatastore",
-			Folder:       "/SDDC-Datacenter/vm/" + clusterID,
-			ResourcePool: "/SDDC-Datacenter/host/Cluster-1/Resources",
-			Server:       "vcenter.sddc-44-236-21-251.vmwarevmc.com",
-		},
+
```
sebsoto, line 69: extra newline can be removed

test/e2e/providers/vsphere/vsphere.go
```diff
@@ -52,19 +66,44 @@ func newVSphereMachineProviderSpec(clusterID string) (*vsphere.VSphereMachinePro
...
+		Template:  vmTemplate,
+		Workspace: workspace,
 	}, nil
 }
 
+// getWorkspaceFromExistingMachineSet returns Workspace from a machineset provisioned during installation
+func (p *Provider) getWorkspaceFromExistingMachineSet() (*vsphere.Workspace, error) {
+	clusterID, err := p.oc.GetInfrastructureID()
+	if err != nil {
+		return nil, fmt.Errorf("unable to get cluster id: %v", err)
```
sebsoto, line 79: We've been following the pattern of using errors.Wrap like so:
```
	clusterID, err := p.oc.GetInfrastructureID()
	if err != nil {
		return nil, errors.Wrap(err, "unable to get cluster id")
```

To stay consistent could you also follow this pattern? Applies to other lines as well

test/e2e/providers/vsphere/vsphere.go
```diff
@@ -52,19 +66,44 @@ func newVSphereMachineProviderSpec(clusterID string) (*vsphere.VSphereMachinePro
...
+		return nil, fmt.Errorf("unable to get machinesets: %v", err)
+	}
+
+	if len(machineSets.Items) == 0 {
+		return nil, fmt.Errorf("no matching machinesets found")
+	}
+
+	machineSet := machineSets.Items[0]
+	templateSpec := machineSet.Spec.Template.Spec
+	providerSpecRaw := templateSpec.ProviderSpec.Value
```
sebsoto, line 94: This can be simplified to `providerSpecRaw := machineSet.Spec.Template.Spec.ProviderSpec.Value`
Having the extra variable `templateSpec` is confusing as we are only interested in the ProviderSpec's value.

test/e2e/providers/vsphere/vsphere.go
```diff
@@ -88,7 +127,7 @@ func (p *Provider) GenerateMachineSet(withWindowsLabel bool, replicas int32) (*m
 	}
 
 	// create new machine provider spec for deploying Windows node
-	providerSpec, err := newVSphereMachineProviderSpec(clusterID)
+	providerSpec, err := p.newVSphereMachineProviderSpec()
```
sebsoto, line 130: I'm not sure about making this a method and no longer passing in `clusterID`, seems like this is going to result in an extra API call.

sebsoto (changes requested):
Thanks for the PR! Left a few requests

test/e2e/providers/vsphere/vsphere.go
rvanderp3: My rationale for that was that the cluster ID is no longer required in that method.  Since we are importing the workspace from an existing machineset, we no longer need the cluster ID to produce a valid workspace as it's already embedded in the workspace we are importing.

aravindhp: That is what I understood too. Are we missing something, @sebsoto? Where is the extra API call coming from?

saifshaikh48: I think @sebsoto is referring to the [this call in `getWorkspaceFromExistingMachineSet()`](https://github.com/openshift/windows-machine-config-operator/blob/47a9b7796075ee2f2961b4e37ad54c8f7a386bc9/test/e2e/providers/vsphere/vsphere.go#L74) despite [already getting the cluster ID here](https://github.com/openshift/windows-machine-config-operator/blob/47a9b7796075ee2f2961b4e37ad54c8f7a386bc9/test/e2e/providers/vsphere/vsphere.go#L120).

rvanderp3: @saifshaikh48 @sebsoto got it, I updated the PR to prevent the additional call.

rvanderp3:
/hold
adding hold pending passing vsphere job

PR #674, opened 2021-09-16
[e2e] Rollback node affinity for BYOH label

jrvaldes:
This commit revert the node affinity logic introduced, in order to keep
the e2e-upgrade test low in complexity.

To ensure the presence of a running workload in a BYOH node, if any, we trust
the kube-scheduler placement algorithm, that balance the number of workloads
across the available nodes that match taints and selectors.

The number of workloads is set to 3, representing the total number of
nodes (one BYOH and one WM) + 1, assuring the eviction of at least 1 pod from
the BYOH node during the de-configuration process.

### DIFF
--- a/test/e2e/upgrade_test.go
+++ b/test/e2e/upgrade_test.go
@@ -14,7 +14,6 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 
-	"github.com/openshift/windows-machine-config-operator/controllers"
 	"github.com/openshift/windows-machine-config-operator/pkg/metadata"
 )
 
@@ -30,26 +29,6 @@ const (
 	// windowsWorkloadTesterJob is the name of the job created to test Windows workloads
 	windowsWorkloadTesterJob = "windows-workload-tester"
 )
-
-// getNodeTypeAffinities generate node affinity for non-BYOH and BYOH nodes if present
-func getNodeTypeAffinities() ([]*v1.Affinity, error) {
-	var affinities []*v1.Affinity
-	if gc.numberOfMachineNodes >= 1 {
-		affinityNoByoh, err := getNodeAffinityForLabel(v1.NodeSelectorOpDoesNotExist, controllers.BYOHLabel)
-		if err != nil {
-			return nil, errors.Wrapf(err, "error creating node affinity without BYOH label")
-		}
-		affinities = append(affinities, affinityNoByoh)
-	}
-	if gc.numberOfBYOHNodes >= 1 {
-		affinityByoh, err := getNodeAffinityForLabel(v1.NodeSelectorOpIn, controllers.BYOHLabel, "true")
-		if err != nil {
-			return nil, errors.Wrapf(err, "error creating node affinity for BYOH label")
-		}
-		affinities = append(affinities, affinityByoh)
-	}
-	return affinities, nil
-}
 
 // generateLinuxWorkloadTesterCommand creates a Job to curl Windows webserver every 5 seconds.
 // If the webserver is not accessible for few minutes Curl Windows webserver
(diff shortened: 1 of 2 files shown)


pkg/nodeconfig/nodeconfig.go
```diff
@@ -77,8 +77,8 @@ type ErrWriter struct {
 }
 
 func (ow OutWriter) Write(p []byte) (n int, err error) {
-	// log debug
-	ow.log.V(1).Info(string(p))
+	// log info
+	ow.log.Info(string(p))
```
jrvaldes, line 81: Sample output below. There is only a minor cosmetic issue, it adds a [line break](https://github.com/openshift/windows-machine-config-operator/blob/master/vendor/k8s.io/kubectl/pkg/drain/drain.go#L292).

```
	INFO	controllers.configmap	processing	{"instances in": "windows-instances"}
	DEBUG	wc 172.31.251.202	initializing SSH connection
	INFO	nc 172.31.251.202	evicting pod default/win-webserver-wm-6756f7468f-scbm8

	INFO	nc 172.31.251.202	evicting pod default/win-webserver-wm-6756f7468f-zmnc4


```

test/e2e/upgrade_test.go
```diff
@@ -73,27 +52,20 @@ func upgradeTestSuite(t *testing.T) {
...
+
+	// create a clusterIP service which can be used to reach the Windows webserver
+	intermediarySVC, err := testCtx.createService(deployment.Name, v1.ServiceTypeClusterIP, *deployment.Spec.Selector)
+	require.NoErrorf(t, err, "error creating service for deployment %s", deployment.Name)
+	defer testCtx.deleteService(intermediarySVC.Name)
+
+	// create a Job object that continuously curls the webserver every 5 seconds.
+	testerJob, err := testCtx.createLinuxJob(windowsWorkloadTesterJob, generateLinuxWorkloadTesterCommand(intermediarySVC.Spec.ClusterIP))
+	require.NoErrorf(t, err, "error creating linux job %s", windowsWorkloadTesterJob)
+	defer testCtx.deleteJob(testerJob.Name)
```
aravindhp, line 68: Please revert back to how the code was before i.e. reintroduce `deployWindowsWorkloadAndTester()`. Sorry this is not concise code but belongs to a method that by looking at the signature one can understand what it does. I had called this out [earlier](https://github.com/openshift/windows-machine-config-operator/pull/645#discussion_r710261878).

aravindhp (changes requested):

test/e2e/upgrade_test.go
jrvaldes: Reverted in https://github.com/openshift/windows-machine-config-operator/pull/674/commits/d3b3c0c2bdc42999910ff263994ae8516d85044e

saifshaikh48 (commented):
/lgtm

test/e2e/upgrade_test.go
```diff
@@ -206,3 +168,31 @@ func (tc *testContext) scaleWMCODeployment(desiredReplicas int32) error {
...
+	testerJob, err := tc.createLinuxJob(windowsWorkloadTesterJob, generateLinuxWorkloadTesterCommand(intermediarySVC.Spec.ClusterIP))
+	if err != nil {
+		return nil, errors.Wrapf(err, "error creating linux job %s", windowsWorkloadTesterJob)
+	}
+	// return a cleanup func
+	return func() {
+		// ignore errors while deleting the objects
+		_ = tc.deleteDeployment(deployment.Name)
+		_ = tc.deleteService(intermediarySVC.Name)
+		_ = tc.deleteJob(testerJob.Name)
```
aravindhp, line 196: Do you really need `_ =`? It should be sufficient to leave that out given you have added the comment.

test/e2e/upgrade_test.go
```diff
@@ -206,3 +168,31 @@ func (tc *testContext) scaleWMCODeployment(desiredReplicas int32) error {
...
+func (tc *testContext) deployWindowsWorkloadAndTester() (func(), error) {
+	// create a Windows Webserver deployment
+	deployment, err := tc.deployWindowsWebServer("win-webserver", nil)
+	if err != nil {
+		return nil, errors.Wrap(err, "error creating Windows Webserver deployment for upgrade test")
+	}
+	// create a clusterIP service which can be used to reach the Windows webserver
+	intermediarySVC, err := tc.createService(deployment.Name, v1.ServiceTypeClusterIP, *deployment.Spec.Selector)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error creating service for deployment %s", deployment.Name)
```
aravindhp, line 184: By return `nil` here you will end up not deleting the deployment and leak resources. How are you going to handle that?
Yes, it will get cleaned up on cluster deletion but given we are attempting a clean up, we should do that consistently. If you are not doing so for any reason, then it needs to be called here in a comment.

test/e2e/upgrade_test.go
```diff
@@ -206,3 +168,31 @@ func (tc *testContext) scaleWMCODeployment(desiredReplicas int32) error {
...
+	}
+	// create a Job object that continuously curls the webserver every 5 seconds.
+	testerJob, err := tc.createLinuxJob(windowsWorkloadTesterJob, generateLinuxWorkloadTesterCommand(intermediarySVC.Spec.ClusterIP))
+	if err != nil {
+		return nil, errors.Wrapf(err, "error creating linux job %s", windowsWorkloadTesterJob)
```
aravindhp, line 189: Same comment here.

aravindhp (changes requested):

test/e2e/upgrade_test.go
jrvaldes: It can be removed, yes; it's a matter of style. `_ =`  is the `go` way of saying  "explicitly ignore the returned error", and the inline comment complements that.

test/e2e/upgrade_test.go
jrvaldes: Good catch, we should do that consistently and cleanup the already created objects. I missed that edge case, will adjust.

test/e2e/upgrade_test.go
jrvaldes: Fixed in https://github.com/openshift/windows-machine-config-operator/pull/674/commits/267e616e832d92bbae86788c2612b4224a78d5a8

test/e2e/upgrade_test.go
```diff
@@ -206,3 +168,34 @@ func (tc *testContext) scaleWMCODeployment(desiredReplicas int32) error {
...
+		return nil, errors.Wrap(err, "error creating Windows Webserver deployment for upgrade test")
+	}
+	// create a clusterIP service which can be used to reach the Windows webserver
+	intermediarySVC, err := tc.createService(deployment.Name, v1.ServiceTypeClusterIP, *deployment.Spec.Selector)
+	if err != nil {
+		_ = tc.deleteDeployment(deployment.Name)
+		return nil, errors.Wrapf(err, "error creating service for deployment %s", deployment.Name)
+	}
+	// create a Job object that continuously curls the webserver every 5 seconds.
+	testerJob, err := tc.createLinuxJob(windowsWorkloadTesterJob, generateLinuxWorkloadTesterCommand(intermediarySVC.Spec.ClusterIP))
```
aravindhp, line 188: This line goes beyond 120 columns

aravindhp (approved):
/approve

with the caveat you fix the long line.

test/e2e/upgrade_test.go
jrvaldes: Fixed.

PR #675, opened 2021-09-16
Bug 1992841: Do not react to Windows Node deletion

saifshaikh48:
This PR does not react/re-queue when a Node associated with a Windows machine
cannot be found (i.e. is deleted), leaving the previously configured Machine
in the `Running` state. 

This approach was chosen rather than deleting &
re-creating the unassociated Machine as optimizing machine management is not in
scope of WMCO responsibilities. In addition, not reacting is in line with the 
current functionality of Linux Machines, standardizing OpenShift functionality.
Also, the Machine cannot be reconfigured to create a new Node object since
the machine-api's nodelink_controller will not update any Machine's Node
reference, neither after deleting a Node nor after reconfiguring a Machine.

Fixes BZ#1992841

### DIFF
--- a/controllers/windowsmachine_controller.go
+++ b/controllers/windowsmachine_controller.go
@@ -247,6 +247,11 @@ func (r *WindowsMachineReconciler) Reconcile(ctx context.Context, request ctrl.R
 		err := r.client.Get(ctx, kubeTypes.NamespacedName{Namespace: machine.Status.NodeRef.Namespace,
 			Name: machine.Status.NodeRef.Name}, node)
 		if err != nil {
+			// Do not requeue if associated node cannot be found (i.e. deleted) for a running machine
+			if k8sapierrors.IsNotFound(err) {
+				log.Info("the node associated with this machine does not exist, no-op", "name", machine.GetName())
+				return ctrl.Result{}, nil
+			}
 			return ctrl.Result{}, errors.Wrapf(err, "could not get node associated with machine %s", machine.GetName())
 		}
 


controllers/windowsmachine_controller.go
```diff
@@ -337,6 +332,24 @@ func (r *WindowsMachineReconciler) Reconcile(ctx context.Context, request ctrl.R
 	return ctrl.Result{}, nil
 }
 
+// deleteMachineIfAllowed deletes the specified Machine if deletion is allowed.
+// Returns a boolean if the deletion attempt should result in a requeuing.
```
jrvaldes, line 336: Consider leaving the `requeuing` decision out of the `deleteMachineIfAllowed()` func.

`deleteMachineIfAllowed()` returns whether the machine was deleted or not, then the requeuing logic takes place.

```go
   machineDeleted, err := r.deleteMachineIfAllowed(machine)

   //  error handling ...

   // requeue based on machineDeleted ...
```

controllers/windowsmachine_controller.go
```diff
@@ -337,6 +332,24 @@ func (r *WindowsMachineReconciler) Reconcile(ctx context.Context, request ctrl.R
...
+func (r *WindowsMachineReconciler) deleteMachineIfAllowed(machine *mapi.Machine) (bool, error) {
```
jrvaldes, line 337: s/deleteMachineIfAllowed/deleteMachine

The `r.isAllowedDeletion(machine)` check is intrinsic of the machine deletion flow, no need to expose it in the function name.

controllers/windowsmachine_controller.go
```diff
@@ -337,6 +332,24 @@ func (r *WindowsMachineReconciler) Reconcile(ctx context.Context, request ctrl.R
...
+		return false, errors.Wrapf(err, "unable to determine if Machine can be deleted")
+	}
+	if !deletionAllowed {
+		r.log.Info("machine deletion restricted", "maxUnhealthyCount", maxUnhealthyCount)
+		r.recorder.Eventf(machine, core.EventTypeWarning, "MachineDeletionRestricted",
+			"Machine %v deletion restricted as the maximum unhealthy machines can`t exceed %v count",
+			machine.Name, maxUnhealthyCount)
+		return true, nil
+	}
+	return false, r.deleteMachine(machine)
```
jrvaldes, line 350: A positive direction in the execution flow, always reads better.  Consider:

```go
if fooAllowed() {
   doFoo()
   return
}

doNotFoo();
return
```

rather than

```go
if !fooAllowed() {
     doNotFoo();
     return
}
doFoo()
return

```

controllers/windowsmachine_controller.go
```diff
@@ -247,6 +247,12 @@ func (r *WindowsMachineReconciler) Reconcile(ctx context.Context, request ctrl.R
 		err := r.client.Get(ctx, kubeTypes.NamespacedName{Namespace: machine.Status.NodeRef.Namespace,
 			Name: machine.Status.NodeRef.Name}, node)
 		if err != nil {
+			// Delete a configured/running machine if no associated node can be determined (node object deleted)
+			if k8sapierrors.IsNotFound(err) {
+				log.Info("machine has no associated node", "name", machine.GetName())
```
jrvaldes, line 252: Shouldn't log `Error`,  given you are returning a `err != nil`  ?

controllers/windowsmachine_controller.go
```diff
@@ -247,6 +247,12 @@ func (r *WindowsMachineReconciler) Reconcile(ctx context.Context, request ctrl.R
 		err := r.client.Get(ctx, kubeTypes.NamespacedName{Namespace: machine.Status.NodeRef.Namespace,
 			Name: machine.Status.NodeRef.Name}, node)
 		if err != nil {
+			// Delete a configured/running machine if no associated node can be determined (node object deleted)
```
jrvaldes, line 250: Is `configured` a valid  phase of machine actuation?

jrvaldes (commented):
@saifshaikh48  thanks for working on this, added few comments. PTAL.

controllers/windowsmachine_controller.go
mansikulkarni96: I don't think log.Error is required here as we are trying to remediate the node not being found by deleting the machine.

controllers/windowsmachine_controller.go
```diff
@@ -247,6 +247,12 @@ func (r *WindowsMachineReconciler) Reconcile(ctx context.Context, request ctrl.R
...
+			if k8sapierrors.IsNotFound(err) {
+				log.Info("machine has no associated node", "name", machine.GetName())
+				requeueNeeded, err := r.deleteMachineIfAllowed(machine)
+				return ctrl.Result{Requeue: requeueNeeded}, err
```
mansikulkarni96, line 254: If the machine deletion is not allowed because it falls below the minHealthyCount, we will not requeue and return the error here. How do we re-mediate that?
Did we consider not re-queuing?

controllers/windowsmachine_controller.go
saifshaikh48: I like this suggestion. Thanks Jose

controllers/windowsmachine_controller.go
saifshaikh48: There is already another function called `deleteMachine` which is called from `deleteMachineIfAllowed` and [other places](https://github.com/openshift/windows-machine-config-operator/blob/7cc2d2972253c64b5869803caa2d1454900ef05d/controllers/windowsmachine_controller.go#L325) where a machine must be deleted.

controllers/windowsmachine_controller.go
saifshaikh48: It is not, I'll remove `configured` for clarity.

controllers/windowsmachine_controller.go
mansikulkarni96: Thanks for the explanation and that is the behavior I was expecting. I must have misread the function and thought we are returning error.

saifshaikh48:
/retitle Bug 1992841: Do not react to Windows Node deletion

controllers/windowsmachine_controller.go
```diff
@@ -247,6 +247,11 @@ func (r *WindowsMachineReconciler) Reconcile(ctx context.Context, request ctrl.R
 		err := r.client.Get(ctx, kubeTypes.NamespacedName{Namespace: machine.Status.NodeRef.Namespace,
 			Name: machine.Status.NodeRef.Name}, node)
 		if err != nil {
+			// Do not requeue if associated node cannot be found (i.e. deleted) for a running machine
+			if k8sapierrors.IsNotFound(err) {
+				log.Info("could not find node associated with machine", "name", machine.GetName())
```
mansikulkarni96, line 252: nit : should we add something like no-op to the log indicating we are not taking any action, similar to what machine-api does?
/lgtm

selansen (approved):
/lgtm

controllers/windowsmachine_controller.go
sebsoto: I agree that this log message could use some re-assurance that this is not an error on WMCO's side, something like:
`the Node associated with this machine does not exist, no action taken`? 
Or however you want to word it

saifshaikh48: Good points, taken care of

aravindhp (changes requested):
Thanks for fixing this, @saifshaikh48. 
- Did you mean MCO in the commit message? Shouldn't it be MAO? 
- Please call out why we decided not to delete the Machine in the PR / commit message

saifshaikh48:
@aravindhp I did mean MCO, as they do not react to node deletions for Linux machines just as this PR proposes WMCO does not react for Windows counterparts. The functionality (or, rather, lack thereof) for both MCO and WMCO is influenced by the MAOs underlying behavior of not updating a Machine's node reference.

aravindhp (approved):
/approve

PR #696, opened 2021-09-21
WINC-672: [build] Add containerd components

aravindhp:
- [submodules] Add containerd
- [submodules] Add hcsshim
- [build] Add containerd components

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -24,3 +24,7 @@
 	path = kube-proxy
 	url = https://github.com/openshift/kubernetes
 	branch = sdn-4.9-kubernetes-1.22.0-rc.0
+[submodule "containerd"]
+	path = containerd
+	url = https://github.com/openshift/containerd
+	branch = release/1.5
--- a/.gitmodules
+++ b/.gitmodules
@@ -28,3 +28,7 @@
 	path = containerd
 	url = https://github.com/openshift/containerd
 	branch = release/1.5
+[submodule "hcsshim"]
+	path = hcsshim
+	url = https://github.com/openshift/hcsshim
+	branch = master
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -26,6 +26,16 @@ WORKDIR /build/windows-machine-config-operator/windows_exporter/
 COPY windows_exporter/ .
 RUN GOOS=windows ./promu build -v
 
+# Build containerd
+WORKDIR /build/windows-machine-config-operator/containerd/
+COPY containerd/ .
+RUN GOOS=windows make
+
+# Build containerd shim
+WORKDIR /build/windows-machine-config-operator/hcsshim/
+COPY hcsshim/ .
+RUN GOOS=windows go build ./cmd/containerd-shim-runhcs-v1
+
 # Build kubelet
 WORKDIR /build/windows-machine-config-operator/kubelet/
 COPY kubelet/ .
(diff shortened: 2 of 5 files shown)


.gitmodules
```diff
@@ -24,3 +24,11 @@
...
 	url = https://github.com/openshift/kubernetes
 	branch = sdn-4.9-kubernetes-1.22.0-rc.0
+[submodule "containerd"]
+	path = containerd
+	url = https://github.com/openshift/containerd
+	branch = release/1.5
+[submodule "hcsshim"]
+	path = hcsshim
+	url = https://github.com/openshift/hcsshim
+	branch = master
```
selansen, line 34: are we not suppose to point to specific version instead of master ?

saifshaikh48: +1
Commit message mentions `v0.9.0-rc0`

build/Dockerfile
```diff
@@ -69,6 +79,9 @@ RUN make build
 #│   ├── win-bridge.exe
 #│   ├── win-overlay.exe
 #│   └── cni-conf-template.json
+#├── containerd
+#│   ├── containerd.exe
+#│   └── containerd-shim-runhcs-v1
```
saifshaikh48, line 84: `containerd-shim-runhcs-v1.exe`

.gitmodules
jrvaldes: > are we not suppose to point to specific version instead of master ?

I don't think so, there is no such branch as `v0.9.0-rc0` .

[v0.9.0-rc0](https://github.com/openshift/hcsshim/releases/tag/v0.9.0-rc0) is a tag.

aravindhp: hcsshim does not seem to follow branching. As @jrvaldes pointed out, v0.9.0-rc0 is a tag which I checked out.

selansen: Actually this is local RedHat Repo. we can ask the hcsshim owners to have some branching strategy.  I think we should not wait for it. lets go ahead and merge it.

aravindhp: PS: anything we use in OpenShift is forked into the OpenShift github organization but we try not to add any carry patches. We never reference or use an upstream repo directly. Sure we can ask the hcsshim owners for a branching strategy 🤞🏽

build/Dockerfile
```diff
@@ -86,6 +99,11 @@ LABEL stage=operator
 WORKDIR /payload/
 COPY --from=build /build/windows-machine-config-operator/windows-machine-config-bootstrapper/wmcb.exe .
 
+# Copy containerd.exe and containerd-shim-runhcs-v1.exe
+WORKDIR /payload/containerd/
+COPY --from=build /build/windows-machine-config-operator/containerd/bin/containerd.exe .
+COPY --from=build /build/windows-machine-config-operator/hcsshim/containerd-shim-runhcs-v1.exe .
+
```
mansikulkarni96, line 106: I might be missing something here, don't you need to change the WORKDIR back to /payload/ after this step, likewise for the other Dockerfile.

aravindhp: You are right. That is a good catch. It also tells me that if we are changing the Dockerfile, just ensuring that it builds locally is not sufficient but we need to actually deploy the operator in a dev cluster. My bad for not doing that.

aravindhp: I have now fixed this and confirmed that the operator deploys on my dev cluster.

PR #698, opened 2021-09-22
[windows] Rename getSourceVIP()

sebsoto:
Renames the function and related variables in order to be accurate to
what the function is actually doing. The function is asking for a new
HNS endpoint to be created and attached to the HNS network created by
hybrid overlay.


aravindhp (approved):

I would like us to call out the package we are making changes to in the commit message rather than using `[wmco]` all the time. The only time `[wmco]` is appropriate is when we are changing multiple packages in a PR.


/approve

PR #705, opened 2021-09-27
[windows] Add dependencies to Windows service struct

sebsoto:
This commit adds the new dependencies field to the service struct. This
is preferrable to having to add the `depend` field in each service's
arguments, as depend is not part of the binPath.

### DIFF
--- a/pkg/windows/service.go
+++ b/pkg/windows/service.go
@@ -10,16 +10,19 @@ type service struct {
 	name string
 	// args is the arguments that the binary will be ran with
 	args string
+	// dependencies is a list of the names of the services that this service is dependent on
+	dependencies []string
 }
 
 // newService initializes and returns a pointer to the service struct
-func newService(binaryPath, name, args string) (*service, error) {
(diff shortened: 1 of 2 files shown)


pkg/windows/service.go
```diff
@@ -10,16 +10,19 @@ type service struct {
 	name string
 	// args is the arguments that the binary will be ran with
 	args string
+	// dependencies is a list of the services that this service is dependant on
```
saifshaikh48, line 13: nit: `...list of the services' names...` since service is a struct
nit: `dependant` --> `dependent`

saifshaikh48 (commented):
/lgtm

Thanks for working doing this @sebsoto. Reduces some of the string formatting we've had to maintain :+1:

selansen:
I see all the places we pass only one service or nil in dependencies list. we are using  string array instead of just string because there may be more services can be added in the dependency list. Is my understanding correct ?

aravindhp (approved):
/approve

PR #747, opened 2021-10-07
[community 4.8] [docs] Add community operator supported platforms

alinaryan:
The purpose of this change was to call out
which platforms are available for trial by
our community operator.

### DIFF
--- a/docs/wmco-prerequisites.md
+++ b/docs/wmco-prerequisites.md
@@ -50,3 +50,18 @@ consistent across all supported cloud providers.
 
 ## vSphere Specific Requirements
 Please refer to [VMware vSphere pre-requisites](vsphere-prerequisites.md).
+
+## Supported Platforms for Community Operator
+| Cloud Provider            | Available for Trial |
+| -----------               | -----------         |
+| AWS IPI                   | Yes                 |
+| AWS IPI BYOH              | Yes                 |
+| AWS UPI BYOH              | Yes                 |
+| Azure IPI                 | Yes                 |
+| Azure IPI BYOH            | Yes                 |
+| Azure UPI BYOH            | Yes                 |
+| BareMetal BYOH            | Yes                 | 
+| Platform-Type = None      | Yes                 |
+| VMware vSphere IPI        | Yes                 |
+| VMware vSphere IPI BYOH   | Yes                 |
+| VMware vSphere UPI BYOH   | Yes                 |


selansen (approved):
/lgtm

docs/wmco-prerequisites.md
```diff
@@ -50,3 +50,18 @@ consistent across all supported cloud providers.
 
 ## vSphere Specific Requirements
 Please refer to [VMware vSphere pre-requisites](vsphere-prerequisites.md).
+
+## Supported Platforms for Community Operator
+| Cloud Provider            | Available for Trial |
```
aravindhp, line 55: - `Supported Platforms` conflicts with `Available for Trial`. Moreover, do you really need this column if it is defaulting to yes.
- Shouldn't this be in the community CSV?

aravindhp (changes requested):

docs/wmco-prerequisites.md
alinaryan: > Supported Platforms conflicts with Available for Trial. Moreover, do you really need this column if it is defaulting to yes.
I'll remove the Yes column and change table title to Community Operator Platforms Available for Trial
>Shouldn't this be in the community CSV?
Where does that live?

aravindhp: > > Supported Platforms conflicts with Available for Trial. Moreover, do you really need this column if it is defaulting to yes.
> 
> I'll remove the Yes column and change table title to Community Operator Platforms Available for Trial

I am not sure if "Trial" is correct in this context. Trial implies that after trying it out there is some way to "buy" the product. Please remind me what was the decision we made for community regarding Supported vs Tech Preview. That is what needs to be called out for the community operator.

> > Shouldn't this be in the community CSV?
> 
> Where does that live?

Please work with @sebsoto on that.

alinaryan: >Please remind me what was the decision we made for community regarding Supported vs Tech Preview.

We decided that we wouldn't explicitly declare support for these platforms for the release, but users are welcome to try it out with our community operator. Anand asked me to add this table to the community docs per this comment https://docs.google.com/document/d/1x74Bf7e-9DKxX_wltR9F340rhwk3xu9qSJ1YSI5ZeHY/edit?disco=AAAAPc8GM-4

>Please work with @sebsoto on that.
Will do!

aravindhp: Thanks for refreshing my memory. I now see where the trial verbiage comes from. I would rather say "works on". Everything on the community operator is preview by default. Don't want to give users the impression that they go from trial to supported on the community version.

alinaryan:
Closing this. Table will instead be added to the community CSV when the next community release is cut.

PR #748, opened 2021-10-13
Bug 1990573:   Display only encrypted string in node annotations

selansen:
This PR addresses 'describe node' output annotation issue. username
 displays extra new lines and tags as part of the value.

 Below code changes have been done to fix this issue.
 we change Encrypt and  Decrypt functions.
 1) Once we encrypt username we trim the encrypted string before we return back
    encrypted string to the caller function.
 2) When decrypt gets called with encrypted string, we insert back tags
    and new lines in the same order so that we get back original username
    back

Fixes BZ#1990573

### DIFF
--- a/pkg/crypto/crypto.go
+++ b/pkg/crypto/crypto.go
@@ -13,6 +13,9 @@ import (
 
 const (
 	wmcoMarker = "<wmcoMarker>"
+	tag        = "ENCRYPTED DATA"
+	startTag   = "-----BEGIN " + tag + "-----"
+	endTag     = "-----END " + tag + "-----"
 )
 
 // EncryptToJSONString returns the encrypted JSON text value of the given string
@@ -40,7 +43,6 @@ func encrypt(plaintext string, key []byte) (string, error) {
 	}
 
 	// Prepare PGP block with a wrapper tag around the encrypted data
-	tag := "ENCRYPTED DATA"
 	msgBuffer := bytes.NewBuffer(nil)
 	encoder, err := armor.Encode(msgBuffer, tag, nil)
 	if err != nil {
@@ -61,13 +63,27 @@ func encrypt(plaintext string, key []byte) (string, error) {
 	writer.Close()
 	encoder.Close()
 
-	return string(msgBuffer.Bytes()), nil
+	// Remove unnecessary characters and trim new lines before storing as
+	// part of node Annotations username.
+	encryptedStr := string(msgBuffer.Bytes())
+	encryptedStr = strings.TrimPrefix(encryptedStr, startTag)
+	encryptedStr = strings.TrimSuffix(encryptedStr, endTag)
+	encryptedStr = strings.Trim(encryptedStr, "\n")
+	return encryptedStr, nil
 }
 
 // decrypt converts encrypted contents to plaintext using the given key as a passphrase
 func decrypt(encrypted string, key []byte) (string, error) {
 	if key == nil {
 		return "", errors.New("decryption passphrase cannot be empty")
+	}
+	// As we trim and remove unnecessary characters during encrypt call, we need to
+	// insert them back to encrypted string to get original value.
+	// 1. add start tag 2. add new lines 3.encrypted string 4. add new line 5. add end tag
+	// For more details : https://tools.ietf.org/id/draft-ietf-openpgp-rfc4880bis-06.html#cleartext-signature-framework
+	// for upgrade case, we might still get old format string. if so we can skip below insert logic
+	if !strings.HasPrefix(encrypted, startTag) {
+		encrypted = startTag + "\n\n" + encrypted + "\n" + endTag
 	}
 
 	// Unwrap encoded block holding the message content
--- a/pkg/crypto/crypto_test.go
+++ b/pkg/crypto/crypto_test.go
@@ -6,6 +6,8 @@ import (
 	"crypto/rsa"
 	"crypto/x509"
 	"encoding/pem"
+	"golang.org/x/crypto/openpgp"
+	"golang.org/x/crypto/openpgp/armor"
 	"testing"
 
 	"github.com/pkg/errors"
@@ -126,6 +128,60 @@ func TestEncryptDecrypt(t *testing.T) {
 	})
 }
 
+func TestUpgradePathDecrypt(t *testing.T) {
+	simplePassphrase := []byte("secret key")
+	rsaPassphrase, _ := generatePrivateKey()
+
+	testCases := []struct {
+		name           string
+		input          string
+		encryptKey     []byte
+		decryptKey     []byte
+		expectedOut    string
+		expectedEncErr bool
+		expectedDecErr bool
+	}{
+		// happy path
+		{
+			name:           "decrypt previous version encrypted key",
+			input:          "Administrator",
+			encryptKey:     simplePassphrase,
+			decryptKey:     simplePassphrase,
+			expectedOut:    "Administrator",
+			expectedEncErr: false,
+			expectedDecErr: false,
+		},
+		// error case
+		{
+			name:           "decrypt attempt with wrong key",
+			input:          "Administrator",
+			encryptKey:     simplePassphrase,
+			decryptKey:     rsaPassphrase,
+			expectedOut:    "",
+			expectedEncErr: false,
+			expectedDecErr: true,
+		},
+	}
+
+	for _, test := range testCases {
+		t.Run(test.name, func(t *testing.T) {
+			cipherText, err := generateBuggyEncryptedString(test.input, test.encryptKey)
+			if test.expectedEncErr {
+				assert.Error(t, err)
+				return
+			}
+
+			out, err := DecryptFromJSONString(cipherText, test.decryptKey)
+			if test.expectedDecErr {
+				assert.Error(t, err)
+				return
+			}
+			require.NoError(t, err)
+			assert.Equal(t, test.expectedOut, out)
+		})
+	}
+}
+
 // generatePrivateKey generates a random RSA private key
 func generatePrivateKey() ([]byte, error) {
 	var keyData []byte
@@ -143,4 +199,39 @@ func generatePrivateKey() ([]byte, error) {
 		return nil, errors.Wrap(err, "error encoding generated private key")
 	}
 	return buf.Bytes(), nil
+}
+
+// generateBuggyEncryptedString encrypt given string and returns it without modifying it
+// In older WMCO version encrypted string has extra tags. When we try to decrypt older
+// version( WMCO 3.1.0 and below) encrypted string in newer version we need to make sure we are
+// handling it properly in our code. This function generates encrypted string with tags and
+// helps to validate this case.
+func generateBuggyEncryptedString(plaintext string, key []byte) (string, error) {
+	if key == nil {
+		return "", errors.New("encryption passphrase cannot be nil")
+	}
+
+	// Prepare PGP block with a wrapper tag around the encrypted data
+	tag := "ENCRYPTED DATA"
+	msgBuffer := bytes.NewBuffer(nil)
+	encoder, err := armor.Encode(msgBuffer, tag, nil)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to create armor block with tag %s", tag)
+	}
+
+	writer, err := openpgp.SymmetricallyEncrypt(encoder, key, nil, nil)
+	if err != nil {
+		return "", errors.Wrap(err, "failure during encryption")
+	}
+	_, err = writer.Write([]byte(plaintext))
+	if err != nil {
+		// Prevent leak in case of stream failure
+		encoder.Close()
+		return "", err
+	}
+	// Both writers must be closed before reading the bytes written to the buffer
+	writer.Close()
+	encoder.Close()
+
+	return string(msgBuffer.Bytes()), nil
 }


pkg/crypto/crypto.go
```diff
@@ -5,14 +5,17 @@ import (
 	"io"
 	"io/ioutil"
 	"strings"
-
+
```
sebsoto, line 8: Please revert

pkg/crypto/crypto.go
```diff
@@ -5,14 +5,17 @@ import (
...
 	"github.com/pkg/errors"
 	"golang.org/x/crypto/openpgp"
 	"golang.org/x/crypto/openpgp/armor"
 )
 
 const (
 	wmcoMarker = "<wmcoMarker>"
+	startTag = "-----BEGIN ENCRYPTED DATA-----"
+	endTag   = "-----END ENCRYPTED DATA-----"
+	newLine = "\n"
```
sebsoto, line 18: I'd prefer not to make this is a constant, to me it would be the same as defining `a` or `b` as a constant

pkg/crypto/crypto.go
```diff
@@ -61,17 +64,33 @@ func encrypt(plaintext string, key []byte) (string, error) {
...
 
 // decrypt converts encrypted contents to plaintext using the given key as a passphrase
 func decrypt(encrypted string, key []byte) (string, error) {
 	if key == nil {
 		return "", errors.New("decryption passphrase cannot be empty")
 	}
+	// As we trim and remove unnecessary characters during encrypt call, we need to
+	// insert them back to encrypted string to get original value.
+	// 1. add start tag 2. add new lines 3.encrypted string 4. add new line 5. add end tag
+    var strBld strings.Builder
```
sebsoto, line 84: Formatting issue

pkg/crypto/crypto.go
saifshaikh48, line 8: This will likely cause the linter to fail.

pkg/crypto/crypto.go
```diff
@@ -61,17 +64,33 @@ func encrypt(plaintext string, key []byte) (string, error) {
...
+	strBld.WriteString(startTag)
+	strBld.WriteString(newLine)
+	strBld.WriteString(newLine)
+	strBld.WriteString(encrypted)
+	strBld.WriteString(newLine)
+	strBld.WriteString(endTag)
```
sebsoto, line 90: I'd prefer avoiding use this pattern as I think it is unnecessarily verbose.
Could do this instead:
```
encrypted = fmt.Sprintf("%s\n\n%s\n%s", startTag, encrypted, endTag)
```

pkg/crypto/crypto.go
```diff
@@ -5,14 +5,17 @@ import (
...
-
+	
 	"github.com/pkg/errors"
 	"golang.org/x/crypto/openpgp"
 	"golang.org/x/crypto/openpgp/armor"
 )
 
 const (
 	wmcoMarker = "<wmcoMarker>"
+	startTag = "-----BEGIN ENCRYPTED DATA-----"
```
saifshaikh48, line 16: Do these constants relate to the [tag defined here](https://github.com/openshift/windows-machine-config-operator/pull/748/files#diff-8a03969770d7b59692a408c8f8cfcff4c2b4f21281ac81f36d94d4b39306c434R46)?

pkg/crypto/crypto.go
```diff
@@ -61,17 +64,33 @@ func encrypt(plaintext string, key []byte) (string, error) {
...
+	encryptedStr = strings.Trim(encryptedStr, "\n")
+	return encryptedStr, nil
 }
 
 // decrypt converts encrypted contents to plaintext using the given key as a passphrase
 func decrypt(encrypted string, key []byte) (string, error) {
 	if key == nil {
 		return "", errors.New("decryption passphrase cannot be empty")
 	}
+	// As we trim and remove unnecessary characters during encrypt call, we need to
```
sebsoto, line 81: How is this going to handle existing nodes from the previous WMCO version, which have the previous formatting, the start/end tags in the annotation?

pkg/crypto/crypto.go
```diff
@@ -61,17 +64,33 @@ func encrypt(plaintext string, key []byte) (string, error) {
 	writer.Close()
 	encoder.Close()
 
-	return string(msgBuffer.Bytes()), nil
+	// Remove unnecessary characters and trim new lines before storing as
+	// part of node Annotations username.
+	encryptedStr := string(msgBuffer.Bytes())
+	encryptedStr = strings.Replace(encryptedStr, startTag, "", -1)
+	encryptedStr = strings.Replace(encryptedStr, endTag, "", -1)
+	encryptedStr = strings.Trim(encryptedStr, "\n")
```
aravindhp, line 72: Can't `\n` also be part of an encrypted string i.e. not just near the start and end tags?

aravindhp (commented):
Thanks for working on this, @selansen 

Nit: `Encrypted` and `Annotation` should be lower cased.

pkg/crypto/crypto.go
selansen: It can, but we replace it with wmc_marker so that when we decrypt we can replace them back with `\n`

pkg/crypto/crypto.go
```diff
@@ -61,14 +66,25 @@ func encrypt(plaintext string, key []byte) (string, error) {
...
 // decrypt converts encrypted contents to plaintext using the given key as a passphrase
 func decrypt(encrypted string, key []byte) (string, error) {
 	if key == nil {
 		return "", errors.New("decryption passphrase cannot be empty")
 	}
+	// As we trim and remove unnecessary characters during encrypt call, we need to
+	// insert them back to encrypted string to get original value.
+	// 1. add start tag 2. add new lines 3.encrypted string 4. add new line 5. add end tag
+	// For more details : https://tools.ietf.org/id/draft-ietf-openpgp-rfc4880bis-06.html#cleartext-signature-framework
+	encrypted = startTag+ newLine + newLine + encrypted + newLine + endTag
```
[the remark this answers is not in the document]
selansen: if we are using different type we can use Sprintf.  All the variables are string here . String addition uses string builder internally which is efficient . Hence wanted to use + instead of Sprintf

pkg/crypto/crypto.go
```diff
@@ -15,6 +15,13 @@ const (
 	wmcoMarker = "<wmcoMarker>"
 )
 
+var (
+	tag      = "ENCRYPTED DATA"
+	startTag = "-----BEGIN " + tag + "-----"
+	endTag   = "-----END " + tag + "-----"
+	newLine  = "\n"
```
sebsoto, line 22: Still not sure why this needs to be a variable

pkg/crypto/crypto.go
aravindhp: In addition we could also check the annotation in e2e test that it does not have the tag.

pkg/crypto/crypto.go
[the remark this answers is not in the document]
sebsoto: For this code readability trumps maximum efficiency, this isn't used frequently so we don't need to prematurely optimize.
I don't have a strong opinion for either `sprintf` or `+` in this specific case, just pointing out that we should prioritize readability when we can.

pkg/crypto/crypto.go
```diff
@@ -61,14 +67,25 @@ func encrypt(plaintext string, key []byte) (string, error) {
...
 // decrypt converts encrypted contents to plaintext using the given key as a passphrase
 func decrypt(encrypted string, key []byte) (string, error) {
 	if key == nil {
 		return "", errors.New("decryption passphrase cannot be empty")
 	}
+	// As we trim and remove unnecessary characters during encrypt call, we need to
+	// insert them back to encrypted string to get original value.
+	// 1. add start tag 2. add new lines 3.encrypted string 4. add new line 5. add end tag
+	// For more details : https://tools.ietf.org/id/draft-ietf-openpgp-rfc4880bis-06.html#cleartext-signature-framework
+	encrypted = startTag + newLine + newLine + encrypted + newLine + endTag
```
sebsoto, line 88: In response to https://github.com/openshift/windows-machine-config-operator/pull/748#discussion_r728999370

For this code readability trumps maximum efficiency, this isn't used frequently so we don't need to prematurely optimize.
I don't have a strong opinion for either sprintf or + in this specific case, just pointing out that we should prioritize readability when we can.

pkg/crypto/crypto.go
aravindhp: Edit: Forgot a `not` in the previous comment

selansen: this last comment  `Forgot a not in the previous comment` confuses me. do you prefer tag and newLine to be contents or variables ?
I will remove '\n' from variable list.

pkg/crypto/crypto.go
selansen: I am not able to follow up the ask here. we already have function to test encrypt/decrypt .  Should I add similar one to test DecryptFromJSONString /EncryptToJSONString calls ?

pkg/crypto/crypto.go
aravindhp: I assume you meant constants and not contents. Yes, they should be constants.

pkg/crypto/crypto.go
```diff
@@ -61,14 +63,25 @@ func encrypt(plaintext string, key []byte) (string, error) {
...
 // decrypt converts encrypted contents to plaintext using the given key as a passphrase
 func decrypt(encrypted string, key []byte) (string, error) {
 	if key == nil {
 		return "", errors.New("decryption passphrase cannot be empty")
 	}
+	// As we trim and remove unnecessary characters during encrypt call, we need to
+	// insert them back to encrypted string to get original value.
+	// 1. add start tag 2. add new lines 3.encrypted string 4. add new line 5. add end tag
+	// For more details : https://tools.ietf.org/id/draft-ietf-openpgp-rfc4880bis-06.html#cleartext-signature-framework
+	encrypted = startTag + "\n" + "\n" + encrypted + "\n" + endTag
```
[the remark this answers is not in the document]
selansen: good point . Thanks.

pkg/crypto/crypto_test.go
```diff
@@ -126,6 +128,49 @@ func TestEncryptDecrypt(t *testing.T) {
...
+	testCases := []struct {
+		name           string
+		input          string
+		encryptKey     []byte
+		decryptKey     []byte
+		expectedOut    string
+		expectedEncErr bool
+		expectedDecErr bool
+	}{
+		// error cases
```
saifshaikh48, line 143: This is a happy path test case, not error case. Speaking of error cases, I think it would be beneficial to add one. Just a negative test case where you pass in a different key to decrypt and makes sure it returns an error.

pkg/crypto/crypto_test.go
```diff
@@ -126,6 +128,49 @@ func TestEncryptDecrypt(t *testing.T) {
...
+			decryptKey:     simplePassphrase,
+			expectedOut:    "Administrator",
+			expectedEncErr: false,
+			expectedDecErr: false,
+		},
+	}
+
+	for _, test := range testCases {
+		t.Run(test.name, func(t *testing.T) {
+			cipherText, err := getEncryptedString(test.input, test.encryptKey)
```
saifshaikh48, line 157: This could be simplified to remove the need for the helper `getEncryptedString`. I don't have a strong opinion either way, just worth thinking about
```suggestion
			cipherText, err := startTag + "\n\n" + encrypt(test.input, test.encryptKey) + "\n" + endTag 
```

pkg/crypto/crypto.go
```diff
@@ -61,14 +63,28 @@ func encrypt(plaintext string, key []byte) (string, error) {
...
 func decrypt(encrypted string, key []byte) (string, error) {
 	if key == nil {
 		return "", errors.New("decryption passphrase cannot be empty")
 	}
+	// As we trim and remove unnecessary characters during encrypt call, we need to
+	// insert them back to encrypted string to get original value.
+	// 1. add start tag 2. add new lines 3.encrypted string 4. add new line 5. add end tag
+	// For more details : https://tools.ietf.org/id/draft-ietf-openpgp-rfc4880bis-06.html#cleartext-signature-framework
+	// for upgrade case, we might still get old format string. if so we can skip below insert logic
+	if !strings.Contains(encrypted, startTag) {
```
saifshaikh48, line 85: Consider [`strings.HasPrefix`](https://pkg.go.dev/strings#HasPrefix) over `Contains`

pkg/crypto/crypto_test.go
```diff
@@ -126,6 +128,49 @@ func TestEncryptDecrypt(t *testing.T) {
 	})
 }
 
+func TestUpgradePathEncryptDecrypt(t *testing.T) {
```
aravindhp, line 131: Shouldn't this be `TestDecryptFromJSONString()`? The upgrade path is just one of the test cases.

pkg/crypto/crypto_test.go
aravindhp: I am a +1 if we can remove code 😃

pkg/crypto/crypto_test.go
aravindhp: +1

aravindhp (changes requested):
Thanks for working on this, @selansen. Please address my comments.

pkg/crypto/crypto.go
selansen: added new UT to cover this case.

pkg/crypto/crypto_test.go
selansen: Initially I had it simple like that. but Instead of creating old format by inserting values, i prefer to use same old way to generate code so that its guaranteed correct format so that our test case is 100% test the old format. 
would like to keep it this way. :)

aravindhp: Then please add a doc to `getEncryptedString` explaining this. Otherwise it looks like overly complicated code for no reason. I also suggest naming the function `generateBuggyEncryptedString()` in addition.

pkg/crypto/crypto_test.go
```diff
@@ -144,3 +200,38 @@ func generatePrivateKey() ([]byte, error) {
 	}
 	return buf.Bytes(), nil
 }
+
+// generateBuggyEncryptedString encrypt given string and returns it without modifying it
+// In older WMCO version encrypted string has extra tags. When we try to decrypt
+// older version encrypted string in newer version we need to make sure we are handling it
+// properly in our code. This function generates encrypted string with tags and helps to
+// validate this case.
+func generateBuggyEncryptedString(plaintext string, key []byte) (string, error) {
```
saifshaikh48, line 209: `generateBuggyEncryptedString` --> `generateUnmodifiedEncryptedString`?

pkg/crypto/crypto_test.go
aravindhp: If that's the case why don't you add this as a separate test case to the existing method?

pkg/crypto/crypto_test.go
```diff
@@ -144,3 +200,38 @@ func generatePrivateKey() ([]byte, error) {
 	}
 	return buf.Bytes(), nil
 }
+
+// generateBuggyEncryptedString encrypt given string and returns it without modifying it
+// In older WMCO version encrypted string has extra tags. When we try to decrypt
```
aravindhp, line 205: Please call out the exact version here.

pkg/crypto/crypto_test.go
aravindhp: So you are adding an encrypt test method and taking the output of that and testing DecryptFromJSONString(). IOW, you are really not testing any `encrypt` functionality. So `TestUpgradePathDecrypt()` might be more appropriate. However if we continue to add more unit test cases for `DecryptFromJSONString()`, they all should be collapsed into a `TestDecryptFromJSONString()` method.

selansen: Yes. TestUpgradePathDecrypt more appropriate one. 
agreed to the second comment. will change the test function name back to original .

pkg/crypto/crypto_test.go
```diff
@@ -144,3 +200,38 @@ func generatePrivateKey() ([]byte, error) {
 	}
 	return buf.Bytes(), nil
 }
+
+// generateBuggyEncryptedString encrypt given string and returns it without modifying it
+// In older WMCO version encrypted string has extra tags. When we try to decrypt older
+// version( WMCO 4.xx and below) encrypted string in newer version we need to make sure we are
```
aravindhp, line 206: This should read `WMCO 3.1.0 and below`

aravindhp:
@selansen please change your commit message to `[crypto] Display only encrypted string in node annotations`. You only fixed the PR.

pkg/crypto/crypto_test.go
selansen: we are targeting this fix in next release , right ? 4.x will be out as soon as we fix blocker issues. so I was kind of thinking this needs to go after 4.0 code is cut off.

aravindhp (approved):
Thanks for fixing this, @selansen 

/approve

PR #764, opened 2021-10-26
[bundle] Fix RBAC generation for OperatorCondition and CSRs

saifshaikh48:
This PR fixes the bundle image, giving WMCO proper RBAC 
permissions for each external resource it interacts with. 

CSV and associated role changes generated using `make bundle`.

Part of this PR is a follow-up to https://github.com/openshift/windows-machine-config-operator/pull/618. The RBAC policy for OperatorConditions 
is added directly to the condition package, as expected by OperatorSDK. 
Before this, the cluster role was getting overwritten with incorrect
permissions, preventing WMCO from properly managing OperatorCondition CR.

### DIFF
--- a/bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
@@ -308,6 +308,16 @@ spec:
           - get
           - list
         - apiGroups:
+          - operators.coreos.com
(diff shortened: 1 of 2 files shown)


saifshaikh48:
/cherry-pick release-4.9

Now that https://github.com/openshift/windows-machine-config-operator/pull/778 has merged

PR #779, opened 2021-11-01
Bug 2000592: [connectivity] Modify SSH connection retry logic

saifshaikh48:
This PR reduces the overall amount of time used while trying to
establish SSH client connections. 
Firstly, the total time to retry the connect attempt is reduced by half.
Secondly, a timeout is introduced to each Dial call, giving a short, few
second window of time between each attempt to reach a host.

The result of these changes is that WMCO picks up changes to the BYOH
ConfigMap faster -- if an unreachable IP is given, WMCO is stuck
retrying this invalid address for less time now.

Fixes BZ#2000592

### DIFF
--- a/pkg/windows/connectivity.go
+++ b/pkg/windows/connectivity.go
@@ -13,12 +13,16 @@ import (
 	"github.com/pkg/sftp"
 	"golang.org/x/crypto/ssh"
 	"k8s.io/apimachinery/pkg/util/wait"
-
-	"github.com/openshift/windows-machine-config-operator/pkg/retry"
 )
 
-// sshPort is the default SSH port
-const sshPort = "22"
+const (
+	// sshPort is the default SSH port
+	sshPort = "22"
+	// TCPConnTimeout is the maximum amount of time for a single TCP connection to establish
+	TCPConnTimeout = time.Second * 45
+	// SSHRetryTimeout is the total time we will retry to establish an SSH connection with a particular socket address
+	SSHRetryTimeout = time.Minute * 5
+)
 
 // AuthErr occurs when our authentication into the VM is rejected
 type AuthErr struct {
@@ -82,11 +86,12 @@ func (c *sshConnectivity) init() error {
 			ssh.PublicKeys(c.signer),
 		},
 		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
+		Timeout:         TCPConnTimeout,
 	}
 	var err error
 	var sshClient *ssh.Client
 	// Retry if we are unable to create a client as the VM could still be executing the steps in its user data
-	err = wait.PollImmediate(time.Minute, retry.Timeout, func() (bool, error) {
+	err = wait.PollImmediate(time.Minute, SSHRetryTimeout, func() (bool, error) {
 		sshClient, err = ssh.Dial("tcp", c.ipAddress+":"+sshPort, config)
 		if err == nil {
 			return true, nil


pkg/retry/retry.go
```diff
@@ -9,6 +9,10 @@ const (
 	Interval = 15 * time.Second
 	// Timeout is the total time we will wait for an event to occur.
 	Timeout = time.Minute * 10
+	// SSHConnTimeout is the time we will wait for a handshake to occur when trying to establish an SSH connection
```
jrvaldes: nit:  There may be others elements involved here, not only the `handshake`. For instances, name resolution, system timeouts, etc.
Consider:
`// SSHConnTimeout is the maximum amount of time for the SSH connection to establish.`

pkg/windows/connectivity.go
```diff
@@ -82,11 +82,12 @@ func (c *sshConnectivity) init() error {
 			ssh.PublicKeys(c.signer),
 		},
 		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
+		Timeout:         retry.SSHConnTimeout,
```
jrvaldes, line 85: Before there was no SSH connection timeout; given the default is zero; The retry was handled only by retry.Timeout.

Will the two newly introduced timeouts overlap or aggregate?

jrvaldes (commented):
@saifshaikh48  Thanks for working on this, LGTM. 
Just a few comments. PTAL

pkg/retry/retry.go
saifshaikh48: Thanks, reworded

pkg/windows/connectivity.go
saifshaikh48: They will overlap. So overall the time table looks like this
| Time  |  Action |
|---|---|
| 0s  | dial attempt 1  |
| 45s  | timeout 1  |
| 1m  | dial attempt 2  |
| 1m 45s  | timeout 2  |
| 2m  | dial attempt 3  |
| 2m 45s  | timeout 3  |
| 3m  | dial attempt 4  |
| 3m 45s  | timeout 4  |
| 4m  | dial attempt 5  |
| 4m 45s  | timeout 5  |
| 5m  | throw reconcile error  |

pkg/retry/retry.go
```diff
@@ -9,6 +9,10 @@ const (
 	Interval = 15 * time.Second
 	// Timeout is the total time we will wait for an event to occur.
 	Timeout = time.Minute * 10
+	// TCPConnTimeout is the maximum amount of time for a single TCP connection to establish
+	TCPConnTimeout = time.Second * 45
+	// SSHRetryTimeout is the total time we will retry to establish an SSH connection with a particular socket address
+	SSHRetryTimeout = time.Minute * 5
```
aravindhp, line 15: These constants seem specific to the connectivity package. I suggest moving them there.

pkg/retry/retry.go
```diff
@@ -9,6 +9,10 @@ const (
 	Interval = 15 * time.Second
 	// Timeout is the total time we will wait for an event to occur.
 	Timeout = time.Minute * 10
+	// TCPConnTimeout is the maximum amount of time for a single TCP connection to establish
+	TCPConnTimeout = time.Second * 45
```
aravindhp, line 13: How did you decide on 45 seconds?

pkg/retry/retry.go
aravindhp, line 15: The Windows networking stack takes a while to get configured especially on 1st boot. Are you certain that reducing this to 5 minutes won't introduce flakes that we might miss? i.e. the instance will get configured on the 2nd try.

aravindhp (changes requested):

pkg/retry/retry.go
selansen: I personally don't feel like changing this timeout logic without making sure we dont regress ( by creating any new flakyness)

pkg/retry/retry.go
saifshaikh48: A 45 second timeout gives 15 seconds of downtime before the next dial attempt. Consistent with our [interval between API calls](https://github.com/openshift/windows-machine-config-operator/blob/master/pkg/retry/retry.go#L8-L9)

pkg/retry/retry.go
saifshaikh48: I've tested this on different platforms and did not see any issue (tested 5 times total). Though I've also looked through some recent CI jobs in different PRs and have not seen any where this logic would cause a premature reconcile error.

@aravindhp @selansen Aside from this manual testing, is there recommended way to ensure we wouldn't regress?

mansikulkarni96: We might have to test this out in CI as well, before we can ensure this doesn't introduce flakes

saifshaikh48:
/close

The team has decided this fix, while potentially beneficial to user experience, may cause unnecessary flakes in the happy path (especially for Windows Machine object upgrades, which require re-creation and therefore a full configuration of the networking stack).

PR #787, opened 2021-11-02
WIP: [e2e] Generate private key instead of using CI key

aravindhp:
Given openshift-dev.pem will soon be retired, generate a key that will be used in the cloud-private-key secret for communicating with Windows instances.


aravindhp:
/close

This is not going to help with the BYOH case on vSphere where the golden image is preconfigured with the key. We will have to do create our own key and [add it to a CI secret](https://docs.ci.openshift.org/docs/how-tos/adding-a-new-secret-to-ci/)

PR #794, opened 2021-11-09
WINC-678: Upgrade to Operator SDK 1.15.0

alinaryan:
This PR updates Operator SDK in WMCO from v1.6.1 to latest v1.15.0. 

Followed upgrade steps from https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.15.0/


jrvaldes (commented):
/lgtm

sebsoto:
> Followed upgrade steps from https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.14.0/

@alinaryan  Please make sure you follow all upgrade instructions, from the current version, up to 1.14, not just the latest ones.

alinaryan:
I skipped https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.11.0/#gov3-ansiblev1-helmv1-increase-resource-limits-for-controller-manager because we don't have cpu and resources fields in the config/manager/manger.yaml, but should I increase those limits elsewhere?

config/rbac/leader_election_role.yaml
```diff
@@ -18,6 +16,10 @@ rules:
   - update
   - patch
   - delete
+  - apiGroups:
```
mansikulkarni96, line 19: The indentation here does not align with other apiGroups and resources

aravindhp: Which leads me to ask how was this tested?

selansen:
> I skipped https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.11.0/#gov3-ansiblev1-helmv1-increase-resource-limits-for-controller-manager because we don't have cpu and resources fields in the config/manager/manger.yaml, but should I increase those limits elsewhere?

I think we need WMCO to be running even if it exceeds cpu/mem utilization(without it we will not be able to have widnows node in kube cluster. Not sure if this is omitted intentionally by design or is this option introduced in later part of operator SDK version. If we are adding some numbers then we need to make sure we have enoughdata point and add buffer so that WMCO pod will not be evicted/removed.

aravindhp:
> > I skipped https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.11.0/#gov3-ansiblev1-helmv1-increase-resource-limits-for-controller-manager because we don't have cpu and resources fields in the config/manager/manger.yaml, but should I increase those limits elsewhere?
> 
> I think we need WMCO to be running even if it exceeds cpu/mem utilization(without it we will not be able to have widnows node in kube cluster. Not sure if this is omitted intentionally by design or is this option introduced in later part of operator SDK version. If we are adding some numbers then we need to make sure we have enoughdata point and add buffer so that WMCO pod will not be evicted/removed.

We need to be careful about the definition of "exceeds" as WMCO is not the most important pod on the node. I suggest we do some profiling and come up with the limits in a follow up PR.

config/rbac/leader_election_role.yaml
alinaryan: Fixed

alinaryan: I was going off the indentation in the upgrade guide. When I did a local build it did not catch this. Thanks Mansi.

selansen:
> > > I skipped https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.11.0/#gov3-ansiblev1-helmv1-increase-resource-limits-for-controller-manager because we don't have cpu and resources fields in the config/manager/manger.yaml, but should I increase those limits elsewhere?
> > 
> > 
> > I think we need WMCO to be running even if it exceeds cpu/mem utilization(without it we will not be able to have widnows node in kube cluster. Not sure if this is omitted intentionally by design or is this option introduced in later part of operator SDK version. If we are adding some numbers then we need to make sure we have enoughdata point and add buffer so that WMCO pod will not be evicted/removed.
> 
> We need to be careful about the definition of "exceeds" as WMCO is not the most important pod on the node. I suggest we do some profiling and come up with the limits in a follow up PR.

Agree. we should do some profiling with scale and measure CPU/Mem usage before come up with final number.
@alinaryan , let me know if you want me to create Jira ticket or you can create one and add it on our backlog .

alinaryan:
>>>>I skipped https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.11.0/#gov3-ansiblev1-helmv1-increase->>>>resource-limits-for-controller-manager because we don't have cpu and resources fields in the >>>>config/manager/manger.yaml, but should I increase those limits elsewhere?

>>>I think we need WMCO to be running even if it exceeds cpu/mem utilization(without it we will not be able to have widnows >>>node in kube cluster. Not sure if this is omitted intentionally by design or is this option introduced in later part of operator >>>SDK version. If we are adding some numbers then we need to make sure we have enoughdata point and add buffer so that >>>WMCO pod will not be evicted/removed.

>>We need to be careful about the definition of "exceeds" as WMCO is not the most important pod on the node. I suggest we >>do some profiling and come up with the limits in a follow up PR.

>Agree. we should do some profiling with scale and measure CPU/Mem usage before come up with final number.
>@alinaryan , let me know if you want me to create Jira ticket or you can create one and add it on our backlog .

Sounds good! I will create one

alinaryan:
/test aws-e2e-upgrade
Timed out provisioning Windows machine

saifshaikh48 (commented):
I don't think the config/ changes will be reflected in the bundle image as is. Try `make bundle`

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
```diff
@@ -11,7 +11,7 @@ metadata:
     olm.skipRange: '>=4.0.0 <5.0.0'
     operatorframework.io/cluster-monitoring: "true"
     operatorframework.io/suggested-namespace: openshift-windows-machine-config-operator
-    operators.operatorframework.io/builder: operator-sdk-v2.0.0+git
+    operators.operatorframework.io/builder: operator-sdk-v1.15.0+git
```
sebsoto, line 14: In the commit message `Update bundle image` isn't the correct terminology here, you're updating the bundle itself. We don't maintain the bundle image, that is something that a user can build using the bundle.

config/default/manager_auth_proxy_patch.yaml
```diff
@@ -18,6 +18,7 @@ spec:
         - "--v=10"
         ports:
         - containerPort: 8443
+          protocol: TCP
```
sebsoto, line 21: This change is going to require `make bundle` as well. I suggest moving this commit to before the `make bundle` commit.

sebsoto (changes requested):
Good work @alinaryan 
Just two small changes requested

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
sebsoto: To be clear `Update bundle directory` would be more accurate

config/default/manager_auth_proxy_patch.yaml
alinaryan: I reordered the commits, but nothing changed in the csv after running make bundle again. Should I have expected a change or was this change solely for proper commit ordering?

sebsoto: If no change happened, that makes sense. Looks like this isnt being used at the moment by the operator.
In general all changes to config/ should be followed up by a make bundle.

alinaryan: Makes sense! Thanks

alinaryan:
/retest
DNS failures on Azure and AWS

alinaryan:
\hold needs rebase

alinaryan:
\hold

PR #798, opened 2021-11-10
[docs] Update SAM verbiage in project volume doc

aravindhp:
...from https://github.com/kubernetes/website/pull/30366/

### DIFF
--- a/docs/projected-volume.md
+++ b/docs/projected-volume.md
@@ -4,10 +4,15 @@ A [projected volume](https://kubernetes.io/docs/concepts/storage/volumes/#projec
 volume sources into the same directory in the container. With the [proposal for file permission handling in projected
 service account volume](https://github.com/kubernetes/enhancements/pull/1598) enhancement, the projected files have
 the correct permissions set including container user ownership for Linux pods when `RunAsUser` is set in the Pod 
-`SecurityContext`. However, for Windows pods this is not the case when `RunAsUsername` is set, due to differences in 
-the way user accounts are managed in Windows. Windows stores and manages local user and group accounts in a database
-file called Security Account Manager (SAM). This database is not shared between the Windows container host and the
-running containers. This prevents the kubelet from setting correct ownership on the files in the projected volume.
+`SecurityContext`. In Windows pods that have a projected volume and `RunAsUsername` set in the Pod `SecurityContext`,
+the ownership is not enforced due to the way user accounts are managed in Windows. Windows stores and manages local
+user and group accounts in a database file called Security Account Manager (SAM). Each container maintains its own
+instance of the SAM database, to which the host has no visibility into while the container is running. Windows
+containers are designed to run the user mode portion of the OS in isolation from the host, hence the maintenance of a
+virtual SAM database. As a result, the kubelet running on the host does not have the ability to dynamically configure
+host file ownership for virtualized container accounts. If files on the host machine are to be shared with the
+container, they should be placed into their own volume mount outside of the `C:\` drive.
+
 This problem is tracked by [Windows Pod with RunAsUserName and a Projected Volume does not honor file
 permissions in the volume](https://github.com/kubernetes/kubernetes/issues/102849) issue.
 


docs/projected-volume.md
```diff
@@ -4,10 +4,15 @@ A [projected volume](https://kubernetes.io/docs/concepts/storage/volumes/#projec
...
-`SecurityContext`. However, for Windows pods this is not the case when `RunAsUsername` is set, due to differences in 
-the way user accounts are managed in Windows. Windows stores and manages local user and group accounts in a database
-file called Security Account Manager (SAM). This database is not shared between the Windows container host and the
-running containers. This prevents the kubelet from setting correct ownership on the files in the projected volume.
+`SecurityContext`. In Windows pods that have a projected volume and `RunAsUsername` set in the Pod `SecurityContext`,
+the ownership is not enforced due to the way user accounts are managed in Windows. Windows stores and manages local
+user and group accounts in a database file called Security Account Manager (SAM). Each container maintains its own
+instance of the SAM database, to which the host has no visibility into while the container is running. Windows
+containers are designed to run the user mode portion of the OS in isolation from the host, hence the maintenance of a
+virtual SAM database. As a result, the kublet running on the host does not have the ability to dynamically configure
```
sebsoto, line 12: kublet -> kubelet

docs/projected-volume.md
```diff
@@ -4,10 +4,15 @@ A [projected volume](https://kubernetes.io/docs/concepts/storage/volumes/#projec
...
+host file ownership for virtualized container accounts. It is recommended that if files on the host machine are to be
+shared with the container then they should be placed into their own volume mount outside of `C:\`.
```
sebsoto, line 14: Is this implying that there should be a separate letter drive for each set of host files needed to be shared with different containers?

docs/projected-volume.md
```diff
@@ -4,10 +4,15 @@ A [projected volume](https://kubernetes.io/docs/concepts/storage/volumes/#projec
...
 service account volume](https://github.com/kubernetes/enhancements/pull/1598) enhancement, the projected files have
 the correct permissions set including container user ownership for Linux pods when `RunAsUser` is set in the Pod 
-`SecurityContext`. However, for Windows pods this is not the case when `RunAsUsername` is set, due to differences in 
-the way user accounts are managed in Windows. Windows stores and manages local user and group accounts in a database
-file called Security Account Manager (SAM). This database is not shared between the Windows container host and the
-running containers. This prevents the kubelet from setting correct ownership on the files in the projected volume.
+`SecurityContext`. In Windows pods that have a projected volume and `RunAsUsername` set in the Pod `SecurityContext`,
+the ownership is not enforced due to the way user accounts are managed in Windows. Windows stores and manages local
+user and group accounts in a database file called Security Account Manager (SAM). Each container maintains its own
+instance of the SAM database, to which the host has no visibility into while the container is running. Windows
```
codyhoag, line 10:
```suggestion
user and group accounts in a database file called Security Account Manager (SAM). Each container maintains its own
instance of the SAM database; the host machine has no visibility into the SAM database while the container is running. Windows
```
This sentence sounded awkward to me. Does this suggestion mean the same thing?

docs/projected-volume.md
```diff
@@ -4,10 +4,15 @@ A [projected volume](https://kubernetes.io/docs/concepts/storage/volumes/#projec
...
+containers are designed to run the user mode portion of the OS in isolation from the host, hence the maintenance of a
+virtual SAM database. As a result, the kublet running on the host does not have the ability to dynamically configure
+host file ownership for virtualized container accounts. It is recommended that if files on the host machine are to be
```
codyhoag, line 13: Wording nit; perhaps this would be more concise?

```
If files on the host machine are to be shared with the container, they should be placed into
their own volume mount outside of the `C:\` drive.
```

codyhoag (commented):
Just a couple wording nits from my side; LGTM otherwise

docs/projected-volume.md
aravindhp: Looks like it. This is coming directly from Microsoft.

docs/projected-volume.md
aravindhp: This could imply the SAM database on the host. So it would be safer to leave this as is.

docs/projected-volume.md
aravindhp: Fixed

codyhoag:
LGTM

PR #802, opened 2021-11-12
Use more appropriate sample domain name for docs

dbaker-rh:
Avoid using "dns.com" which is owned by an actual company;  "example.com" is more appropriate for use here.


sebsoto:
Thanks for catching this @dbaker-rh 
/approve

mansikulkarni96:
Thanks for fixing this @dbaker-rh 
/lgtm

PR #803, opened 2021-11-12
[config] Add ServiceAccount manifest

Fedosin:
When I deploy the operator with `make deploy`, I get next error from WMCO deployment:

```txt
error looking up service account openshift-windows-machine-config-operator/windows-machine-config-operator:
      serviceaccount "windows-machine-config-operator" not found'
```

This commit adds a manifest for the missing service account. After that everything installs well.

### DIFF
--- a/config/rbac/service_account.yaml
+++ b/config/rbac/service_account.yaml
@@ -1,0 +1,5 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  namespace: system
+  name: windows-machine-config-operator
--- a/config/rbac/kustomization.yaml
+++ b/config/rbac/kustomization.yaml
@@ -1,4 +1,5 @@
 resources:
+- service_account.yaml
 # All RBAC will be applied under this service account in
 # the deployment namespace. You may comment out this resource
 # if your manager will use a service account that exists at


mansikulkarni96:
Thanks for fixing this @Fedosin 
This was removed as part of a service_account duplication manifest issue. Tested it locally, works well as long as we do not add the service account file to the manifests directory that gets created with this when we do `make bundle`

/lgtm

config/rbac/service_account.yaml
```diff
@@ -0,0 +1,7 @@
+
+---
```
aravindhp, line 2: This is only required if you are adding multiple "kinds".

aravindhp (changes requested):
Thanks for fixing this, @Fedosin. Please add a `[config]` prefix to the commit and PR headers.

sebsoto:
/approve
As this is manually tested and linting PR is in progress

aravindhp:
/hold

Azure cluster bring up is consistently failing

aravindhp:
/hold

until we figure out why the override for the cluster approver is not taking effect causing the CSR validation test to fail

PR #817, opened 2021-11-22
[wmco] Re-arrange import and add missing spaces

jrvaldes:
Fix formatting issue in comments without leading space and
re-arrange import order in file.


selansen (approved):
/lgtm

PR #827, opened 2021-11-23
[hack] Remove stale OSDK test code

sebsoto:
Removes stale code that was previously used to run the e2e tests.
Current implementation uses `go test` to run the tests, so the old
method is no longer needed.


aravindhp (approved):
/approve

PR #844, opened 2021-12-06
Bug 2032048: Fix CVO override and retry Update of CSRs approval status

saifshaikh48:
This PR fixes the group value when adding a CVO override for 
the cluster-machine-approver. Overrides were not taking effect 
due to an invalid value ([CVO docs](https://github.com/openshift/enhancements/blob/f97876821d3bb506d28fee565271d3bebbbc682c/dev-guide/cluster-version-operator/dev/clusterversion.md#setting-objects-unmanaged) are incorrect).This is required for testing 
BYOH CSR approval feature so that CSRs associated with 
BYOH instances are not approved by Cluster Machine Approver.

The PR also takes the recommended k8s approach to Updating CSRs.
If an update fails due to a conflict, we want to retrieve an updated
reference to the object before trying again to apply our own changes to it.
Otherwise, Update calls can result in failures as the CSR object can
become a stale reference (if some other entity changes the CSR resource
while WMCO is configuring for example). Lastly it brings the WMCO CSR 
approval condition's message and reason fields in line with 
the kubectl certificate approver and the Node CSR approver.

### DIFF
--- a/controllers/certificatesigningrequests_controller.go
+++ b/controllers/certificatesigningrequests_controller.go
@@ -23,6 +23,8 @@ import (
 	certificates "k8s.io/api/certificates/v1"
 	k8sapierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
@@ -33,6 +35,7 @@ import (
 	"github.com/openshift/windows-machine-config-operator/pkg/cluster"
 	"github.com/openshift/windows-machine-config-operator/pkg/condition"
 	"github.com/openshift/windows-machine-config-operator/pkg/csr"
+	"github.com/openshift/windows-machine-config-operator/pkg/retry"
 )
 
 const (
@@ -102,8 +105,34 @@ func (r *certificateSigningRequestsReconciler) Reconcile(ctx context.Context,
 	return ctrl.Result{}, r.reconcileCSR(certificateSigningRequest)
 }
 
-// reconcileCSR handles the CSR validation and approval
-func (r *certificateSigningRequestsReconciler) reconcileCSR(request *certificates.CertificateSigningRequest) error {
+// reconcileCSR kicks off CSR validation and approval. In case of update conflicts, the approval process is wrapped
+// with retry logic using the Update --> Get --> Retry Update pattern. Retry will occur until specified timeout
+func (r *certificateSigningRequestsReconciler) reconcileCSR(req *certificates.CertificateSigningRequest) error {
+	err := wait.Poll(retry.Interval, retry.ResourceChangeTimeout, func() (bool, error) {
+		err := r.validateAndApprove(req)
+		if err == nil {
+			return true, nil
+		}
+		if !k8sapierrors.IsConflict(err) {
+			// If the error is not an update conflict, return error immediately
+			return false, err
+		}
+
+		r.log.Info("Update conflict when attempting to approve CSR, retrying...", "name", req.Name,
+			"namespace", req.Namespace, "error", err.Error())
+		// Get a new object reference
+		err = r.client.Get(context.TODO(), types.NamespacedName{Namespace: req.Namespace, Name: req.Name}, req)
+		if err != nil {
+			return false, errors.Wrapf(err, "could not get CSR %s/%s", req.Name, req.Namespace)
+		}
+		// Retry CSR approval reconciliation with new object reference
+		return false, nil
+	})
+	return err
+}
+
+// validateAndApprove handles the CSR validation and approval
+func (r *certificateSigningRequestsReconciler) validateAndApprove(request *certificates.CertificateSigningRequest) error {
 	// If a CSR is approved/denied after being added to the queue, but before we reconcile it,
 	// trying to approve it again will result in an error and cause a loop.
 	// Return early if the CSR has been approved/denied externally.
--- a/pkg/csr/csr.go
+++ b/pkg/csr/csr.go
@@ -89,9 +89,9 @@ func (a *Approver) Approve() error {
 	a.csr.Status.Conditions = append(a.csr.Status.Conditions, certificates.CertificateSigningRequestCondition{
 		Type:           certificates.CertificateApproved,
 		Status:         "True",
-		Message:        "Approved by WMCO",
+		Message:        "This CSR was approved by the WMCO certificate Approver.",
 		LastUpdateTime: meta.Now(),
-		Reason:         "WMCOApproved",
+		Reason:         "WMCOApprove",
 	})
 
 	if _, err := a.k8sclientset.CertificatesV1().CertificateSigningRequests().UpdateApproval(context.Background(),
--- a/test/e2e/create_test.go
+++ b/test/e2e/create_test.go
@@ -133,7 +133,7 @@ func (tc *testContext) testBYOHConfiguration(t *testing.T) {
 
 	// Patch the CVO with overrides spec value for cluster-machine-approver deployment
 	// Doing so, stops CVO from creating/updating its deployment hereafter.
-	patchData := `[{"op":"add","path":"/spec/overrides","value":[{"kind":"Deployment","group":"apps/v1","name":"machine-approver","namespace":"openshift-cluster-machine-approver","unmanaged":true}]}]`
+	patchData := `[{"op":"add","path":"/spec/overrides","value":[{"kind":"Deployment","group":"apps","name":"machine-approver","namespace":"openshift-cluster-machine-approver","unmanaged":true}]}]`
 	_, err := tc.client.Config.ConfigV1().ClusterVersions().Patch(context.TODO(), "version", types.JSONPatchType, []byte(patchData), metav1.PatchOptions{})
 
 	// Scale the Cluster Machine Approver Deployment to 0
--- a/test/e2e/validation_test.go
+++ b/test/e2e/validation_test.go
@@ -439,13 +439,14 @@ func testCSRApproval(t *testing.T) {
 		for _, csr := range csrs {
 			isWMCOApproved := func() bool {
 				for _, c := range csr.Status.Conditions {
-					if c.Type == certificates.CertificateApproved && c.Reason == "WMCOApproved" {
+					if c.Type == certificates.CertificateApproved && c.Reason == "WMCOApprove" {
 						return true
 					}
 				}
 				return false
 			}()
-			assert.Equal(t, isWMCOApproved, true, "expected BYOH node CSR to be approved by WMCO CSR approver: %s", node.GetName())
+			assert.Equal(t, isWMCOApproved, true, "expected BYOH node %s CSR %s to be approved by WMCO CSR approver",
+				node.GetName(), csr.GetName())
 		}
 	}
 


pkg/csr/csr.go
```diff
@@ -86,18 +88,24 @@ func (a *Approver) Approve() error {
...
+	approvalCond := certificates.CertificateSigningRequestCondition{
 		Type:           certificates.CertificateApproved,
 		Status:         "True",
-		Message:        "Approved by WMCO",
+		Message:        "This CSR was approved by the WMCO certificate Approver.",
 		LastUpdateTime: meta.Now(),
-		Reason:         "WMCOApproved",
-	})
-
-	if _, err := a.k8sclientset.CertificatesV1().CertificateSigningRequests().UpdateApproval(context.Background(),
```
sebsoto, line 97: in most cases i'd think patching would be the way to go here, but considering there is an explicit `UpdateApproval()` method in the client, is that not the recommended way to approve a CSR?

Can you find any docs that support switching to the patch method?

pkg/csr/csr.go
```diff
@@ -91,9 +91,9 @@ func (a *Approver) Approve() error {
 	approvalCond := certificates.CertificateSigningRequestCondition{
 		Type:           certificates.CertificateApproved,
 		Status:         "True",
-		Message:        "Approved by WMCO",
+		Message:        "This CSR was approved by the WMCO certificate Approver.",
 		LastUpdateTime: meta.Now(),
-		Reason:         "WMCOApproved",
+		Reason:         "WMCOApprove",
```
mansikulkarni96, line 96: The test would need to be updated as well to look for this reason in the approved CSR's

pkg/csr/csr.go
saifshaikh48: Approval is a subresource of the CSR, so just as other resources have an `Update()` method upon them, CSRs have `Update()`, `UpdateApproval()`, and `Update<OtherSubresource>()`. Under the hood they are all just simple HTTP Put calls. Note they also all have `Patch()`.

In my mind this switch has the same justification as other switches to Patch over Update (e.g. https://github.com/openshift/windows-machine-config-operator/pull/518). 
`A patch is best used when
modifying parts of a resource, which fits the CSR approval case;
this is why a patch was favored over a Get -> Update loop.`

mansikulkarni96: The recommendation in the kubernetes docs is to use update 
https://v1-18.docs.kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/#authorization

saifshaikh48:
/retest

500 error from container registry

Fedosin (approved):
/lgtm

saifshaikh48:
@aravindhp any reason for the hold?

aravindhp:
Yes. I leave a review shortly to articulate why.

pkg/csr/csr.go
```diff
@@ -96,6 +96,18 @@ func (a *Approver) Approve() error {
...
+			// Get a new object reference and retry the approval
+			if err = a.client.Get(context.TODO(),
+				kubeTypes.NamespacedName{Namespace: a.csr.Namespace, Name: a.csr.Name}, a.csr); err != nil {
+				if apierrors.IsNotFound(err) {
+					// Request object not found, likely deleted by another entity. Nothing to approve, no action needed
+					return nil
+				}
+				return errors.Wrapf(err, "could not get CSR %s from namespace %s", a.csr.Name, a.csr.Namespace)
+			}
+			a.Approve()
```
selansen, line 109: This can lead to recursive calls and there is no way we can exit the function until stack overflow happens. we should handle it so that we will not get into an infinite loop

selansen (changes requested):
Thanks for fixing this @saifshaikh48, I suggest creating a Bugzilla for this issue so that QE can verify it.

pkg/csr/csr.go
```diff
@@ -96,6 +96,18 @@ func (a *Approver) Approve() error {
 
 	if _, err := a.k8sclientset.CertificatesV1().CertificateSigningRequests().UpdateApproval(context.Background(),
 		a.csr.Name, a.csr, meta.UpdateOptions{}); err != nil {
+		if apierrors.IsConflict(err) {
+			// Get a new object reference and retry the approval
+			if err = a.client.Get(context.TODO(),
+				kubeTypes.NamespacedName{Namespace: a.csr.Namespace, Name: a.csr.Name}, a.csr); err != nil {
+				if apierrors.IsNotFound(err) {
+					// Request object not found, likely deleted by another entity. Nothing to approve, no action needed
+					return nil
```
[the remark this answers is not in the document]
aravindhp: Yeah, this can be a problem if the CSR in question is one the WMCO is supposed to approve but was deleted by another entity. We will now silent fail without a `Ready` node.

pkg/csr/csr.go
aravindhp: +1

aravindhp (changes requested):
Thanks for working on this, @saifshaikh48.

- How did you test this fix?
- I agree with @selansen that this warrants a bug as we want QE to verify the fix. @jrvaldes given that you discovered the issue, I suggest you create the BZ.

pkg/csr/csr.go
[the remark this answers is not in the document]
saifshaikh48: My reasoning came from [our current logic](https://github.com/openshift/windows-machine-config-operator/blob/2a59a283e13f66febc7b38fc13c9fafa80842d6a/controllers/certificatesigningrequests_controller.go#L93-L96) in the higher level reconcile code. If the CSR associated with the event we are processing is not found, we silently fail still. Am I correct to say this is handled improperly in both locations?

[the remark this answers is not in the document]
aravindhp: In the current logic, it needn't be with a CSR that WMCO is supposed to approve whereas in the retry case it definitely was with a CSR that WCMO was supposed to approve and has now been deleted.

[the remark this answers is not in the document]
saifshaikh48: I see, I incorrectly assumed our predicates were a bit more thorough, filtering down to only CSRs WMCO should approve. Thanks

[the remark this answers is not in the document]
aravindhp: @mansikulkarni96 tried making the predicates more thorough but that would have involved too much logic in the predicates if I remember correctly.

jrvaldes:
>  @jrvaldes given that you discovered the issue, I suggest you create the BZ.

Sorry missed that. Thanks @saifshaikh48  for creating the BZ.

controllers/certificatesigningrequests_controller.go
```diff
@@ -102,8 +105,35 @@ func (r *certificateSigningRequestsReconciler) Reconcile(ctx context.Context,
...
 
-// reconcileCSR handles the CSR validation and approval
+// reconcileCSR kicks off CSR validation and approval. In case of update conflicts, the approval process is wrapped
+// with retry logic using the Update --> Get --> Retry Update pattern. Retry will occur until specified timeout
 func (r *certificateSigningRequestsReconciler) reconcileCSR(request *certificates.CertificateSigningRequest) error {
+	err := wait.Poll(retry.Interval, retry.ResourceChangeTimeout, func() (bool, error) {
+		err := r.validateAndApprove(request)
+		if err != nil {
+			if k8sapierrors.IsConflict(err) {
+				r.log.Info("Update conflict when attempting to approve CSR, retrying...", "name", request.Name,
```
aravindhp, line 115: Any reason to not print the namespace like you are doing when returning the error?

aravindhp (approved):

/approve

controllers/certificatesigningrequests_controller.go
saifshaikh48: I'll add this in, good call

saifshaikh48:
/hold

to investigate CI failures

saifshaikh48:
/retest azure-e2e-operator 

Not seeing these failures locally, AWS or vSphere, CI

saifshaikh48:
/test azure-e2e-operator 

Not seeing these failures locally, AWS or vSphere, CI

test/e2e/create_test.go
```diff
@@ -33,6 +32,12 @@ import (
 // after the hardware is provisioned.
 const vmConfigurationTime = 10 * time.Minute
 
+// CMAConfigName is the namespace of the Cluster Machine Approver which approves Node client CSRs
+const CMANamespace = "openshift-cluster-machine-approver"
+
+// CMAConfigName is the name of the ConfigMap used to disable the Cluster Machine Approver
+const CMAConfigName = "machine-approver-config"
+
```
aravindhp, line 40: Please make this a `const` block

test/e2e/create_test.go
```diff
@@ -131,14 +136,24 @@ func (tc *testContext) testBYOHConfiguration(t *testing.T) {
...
+	machineApproverConfig.Data["config.yaml"] = "nodeClientCert:\n  disabled: true"
+	_, err := tc.client.K8s.CoreV1().ConfigMaps(machineApproverConfig.Namespace).Create(context.TODO(),
+		machineApproverConfig, metav1.CreateOptions{})
+	require.NoErrorf(t, err, "unable to create ConfigMap %s/%s", machineApproverConfig.Namespace, machineApproverConfig.Name)
 
 	// Scale the Cluster Machine Approver Deployment to 0
-	// This is required for testing BYOH CSR approval feature so that BYOH instances
-	// CSR's are not approved by Cluster Machine Approver
+	// This is required as the machine-approver pods must restart for the node client CSR approval disabling
+	// to actually take effect. CVO will reset this replica count
```
aravindhp, line 156: Nit missing `.` at the end.

test/e2e/validation_test.go
```diff
@@ -450,10 +450,14 @@ func testCSRApproval(t *testing.T) {
 		}
 	}
 
-	// Scale the Cluster Machine Approver deployment back to 1.
-	expectedPodCount := int32(1)
+	// Re-enable the Cluster Machine Approver by deleting the config then kicking off a re-creation of the CMA deployment
+	err = testCtx.client.K8s.CoreV1().ConfigMaps(CMANamespace).Delete(context.TODO(), CMAConfigName, meta.DeleteOptions{})
+	require.NoErrorf(t, err, "unable to delete ConfigMap %s/%s", CMANamespace, CMAConfigName)
+
+	expectedPodCount := int32(0)
```
aravindhp, line 457: It is plausible that CVO doesn't reset the replicas in time or there is bug in CVO that doesn't allow it to take effect. It would be safer to scale the deployment down and scale it back up.

aravindhp (changes requested):

selansen:
still seeing many CIs failures.

selansen (approved):
/lgtm

controllers/certificatesigningrequests_controller.go
```diff
@@ -102,8 +105,34 @@ func (r *certificateSigningRequestsReconciler) Reconcile(ctx context.Context,
 	return ctrl.Result{}, r.reconcileCSR(certificateSigningRequest)
 }
 
-// reconcileCSR handles the CSR validation and approval
+// reconcileCSR kicks off CSR validation and approval. In case of update conflicts, the approval process is wrapped
+// with retry logic using the Update --> Get --> Retry Update pattern. Retry will occur until specified timeout
 func (r *certificateSigningRequestsReconciler) reconcileCSR(request *certificates.CertificateSigningRequest) error {
+	err := wait.Poll(retry.Interval, retry.ResourceChangeTimeout, func() (bool, error) {
+		err := r.validateAndApprove(request)
+		if err != nil {
```
sebsoto, line 113: This is causing a large portion of this function be indented. Can the inverse be done?
```
if err == nil {
    return true, nil
}

			if k8sapierrors.IsConflict(err) {
...
```

sebsoto (changes requested):
partial review

controllers/certificatesigningrequests_controller.go
```diff
@@ -102,8 +105,34 @@ func (r *certificateSigningRequestsReconciler) Reconcile(ctx context.Context,
...
+			if k8sapierrors.IsConflict(err) {
```
sebsoto, line 114: Same can be done here
```
if !k8sapierrors.IsConflict(err) {
  return false, err
  }
```

controllers/certificatesigningrequests_controller.go
```diff
@@ -102,8 +105,34 @@ func (r *certificateSigningRequestsReconciler) Reconcile(ctx context.Context,
...
+				r.log.Info("Update conflict when attempting to approve CSR, retrying...", "name", request.Name,
```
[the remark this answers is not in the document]
saifshaikh48: Sorry I am not sure the link is resolving properly, it's not clear to me how the name and namespace would already be present.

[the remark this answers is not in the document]
sebsoto: I was trying to link to this line:
```
	_ = r.log.WithValues(CSRController, req.NamespacedName)
```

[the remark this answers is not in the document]
saifshaikh48: Conclusion here is that the name and namespace will not be propagated to the log output because we are using the `logr.WithValues` method improperly. [Tech debt story](https://issues.redhat.com/browse/WINC-721) created as that is out of PR scope

PR #582, opened 2021-08-11
[release-4.8][submodule] Update 

sebsoto:
* [submodule][kube-proxy] Update to https://github.com/openshift/kubernetes/commit/43b818f2e5a3be1f85e796c5f7430c10427f826c
* [submodule][kubelet] Update to https://github.com/openshift/kubernetes/commit/98073871f173baaa04dc2bafab50effd62c308a6
* [submodule][containernetworking-plugins] Update to https://github.com/openshift/containernetworking-plugins/commit/71a8b3469df865daf68cb893bfe313328d4be992
* [submodule][ovn-kubernetes] Update to https://github.com/openshift/ovn-kubernetes/commit/7da97fcfffb1770d980a85b5667b78d5c7f67c95

Also includes cleaning up go.mod by removing necessary replace directives.

PR #847, opened 2021-12-06
[community-4.8][submodule] Remove replace directives and submodule update

alinaryan:
- [submodule][kubelet] Update to openshift/kubernetes@c180a7c
- [submodule][ovn-kubernetes] Update to openshift/ovn-kubernetes@46a77c4

Also includes cleaning up go.mod by removing necessary replace directives.

Manual cherry-pick of #582


sebsoto:
The submodule updates should not reference a commit that doesn't exist in this repo. If you're cherry-picking them from a local branch you should not use the `cherry-pick -x` option. If someone sees that and tries to find the original commit, they will not be able to, because it won't be present in any branch.

PR #850, opened 2021-12-08
[release-4.9] Revert containerd addition

aravindhp:
This is a manual revert of #701 as we have decided not to ship containerd with WMCO 4.y.z


aravindhp:
/hold
for 4.0.1 release

PR #869, opened 2022-01-13
[e2e] Scale down CVO when testing BYOH CSRs

saifshaikh48:
This PR scales down the CVO when testing WMCO's CSR approval functionality.
This is needed to prevent CVO from scaling the cluster-machine-operator
deployment back up, something we scale down ourselves so it does not try to
auto-approve CSRs for Windows BYOH nodes (in CI, our BYOH nodes are
Machine-backed but without the Windows version annotation for ease of testing).
We are seeing an issue where the previous solution, adding a CVO override to
make the cluster-machine-operator deployment unmanaged, is no longer sufficient.
This PR also refactors a method used to scale deployments in the e2e tests.

### DIFF
--- a/test/e2e/create_test.go
+++ b/test/e2e/create_test.go
@@ -2,7 +2,6 @@ package e2e
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"log"
 	"math"
@@ -18,14 +17,12 @@ import (
 	"k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 
 	"github.com/openshift/windows-machine-config-operator/controllers"
 	"github.com/openshift/windows-machine-config-operator/pkg/crypto"
 	"github.com/openshift/windows-machine-config-operator/pkg/metadata"
 	"github.com/openshift/windows-machine-config-operator/pkg/nodeconfig"
-	"github.com/openshift/windows-machine-config-operator/pkg/patch"
 	"github.com/openshift/windows-machine-config-operator/pkg/retry"
 	"github.com/openshift/windows-machine-config-operator/pkg/wiparser"
 	"github.com/openshift/windows-machine-config-operator/test/e2e/clusterinfo"
@@ -35,6 +32,10 @@ const (
 	// vmConfigurationTime is the maximum amount of time expected for a Windows VM to be fully configured and ready for WMCO
 	// after the hardware is provisioned.
 	vmConfigurationTime = 10 * time.Minute
+
+	clusterVersionOperatorNamespace   = "openshift-cluster-version"
+	clusterVersionOperatorDeployment  = "cluster-version-operator"
+	clusterVersionOperatorPodSelector = "k8s-app=cluster-version-operator"
 
 	machineApproverNamespace   = "openshift-cluster-machine-approver"
 	machineApproverDeployment  = "machine-approver"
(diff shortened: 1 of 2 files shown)


saifshaikh48:
/retitle [e2e] Scale down CVO when testing BYOH CSRs

test/e2e/create_test.go
```diff
@@ -15,7 +15,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
aravindhp: You make this change in this commit and then revert in the next. Please fix.

test/e2e/create_test.go
```diff
@@ -154,7 +154,7 @@ func (tc *testContext) testBYOHConfiguration(t *testing.T) {
 	// This is required for testing BYOH CSR approval feature so that BYOH instances
 	// CSR's are not approved by Cluster Machine Approver
 	expectedPodCount := int32(0)
-	err = tc.scaleMachineApproverDeployment(&expectedPodCount)
+	err = tc.scaleDeployment("openshift-cluster-machine-approver", "machine-approver", &expectedPodCount)
```
aravindhp, line 142: Please add constants for these strings given the use in multiple places. Same for CVO.

aravindhp (changes requested):
Mostly lgtm . Thanks for fixing this, @saifshaikh48

test/e2e/create_test.go
selansen: +1

selansen (commented):
/lgtm

test/e2e/create_test.go
```diff
@@ -455,29 +455,23 @@ func (tc *testContext) listFullyConfiguredWindowsNodes(isBYOH bool) ([]v1.Node,
...
-// scaleMachineApproverDeployment scales the Machine Approver deployment pods to the expectedPodCount
-func (tc *testContext) scaleMachineApproverDeployment(expectedPodCount *int32) error {
-	deployment, err := tc.client.K8s.AppsV1().Deployments("openshift-cluster-machine-approver").Get(context.TODO(),
-		"machine-approver", metav1.GetOptions{})
+// scaleDeployment scales the deployment associated with the given namespace and name to the expectedPodCount
+func (tc *testContext) scaleDeployment(namespace, name string, expectedPodCount *int32) error {
+	deployment, err := tc.client.K8s.AppsV1().Deployments(namespace).Get(context.TODO(), name, metav1.GetOptions{})
 	if err != nil {
-		return errors.Wrap(err, "error listing Cluster Machine Approver deployment")
+		return errors.Wrapf(err, "error listing deployment %s/%s", namespace, name)
```
sebsoto, line 447: this is a get not a list

test/e2e/create_test.go
```diff
@@ -455,29 +455,23 @@ func (tc *testContext) listFullyConfiguredWindowsNodes(isBYOH bool) ([]v1.Node,
...
-	retryTimeout := retry.Timeout
-	if err = wait.Poll(retryInterval, retryTimeout, func() (bool, error) {
-		if deployment.Spec.Replicas == expectedPodCount {
-			return true, nil
-		}
-		return false, nil
+
+	if err = wait.Poll(retry.Interval, retry.Timeout, func() (bool, error) {
+		return deployment.Spec.Replicas == expectedPodCount, nil
 	}); err != nil {
```
sebsoto, line 458: How is this poll working? Unless you redo the get of the deployment, the value is not going to change.

sebsoto: Shouldn't this also be looking at the status and not the spec...?
The spec is always going to be changed by the update, we need to wait for the actual count of replicas to reach zero.
[api reference](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#deployment-v1-apps)

Is this why the bug is occurring? :thinking:

sebsoto (changes requested):

test/e2e/create_test.go
saifshaikh48: I'll be honest, it looked off but I didn't want to touch the logic since in the past and through all my testing it seemed to work. The CMA deployment is being scaled down properly, the issue is that it comes back up despite our CVO override. So I doubt its causing the bug, but I will edit the logic with your suggestions.

test/e2e/create_test.go
saifshaikh48: My IDE loves to add this in when auto-formatting code :smile:

selansen (approved):
/lgtm

test/e2e/validation_test.go
```diff
@@ -468,8 +468,14 @@ func testCSRApproval(t *testing.T) {
...
-	err = testCtx.scaleMachineApproverDeployment(&expectedPodCount)
-	require.NoError(t, err, "failed to scale Cluster Machine Approver pods")
+	err = testCtx.scaleDeployment(machineApproverNamespace, machineApproverDeployment, machineApproverPodSelector,
+		&expectedPodCount)
+	require.NoError(t, err, "failed to scale up Cluster Machine Approver pods")
+
+	// Scale the Cluster Version Operator deployment back to 1.
+	err = testCtx.scaleDeployment(clusterVersionOperatorNamespace, clusterVersionOperatorDeployment,
+		clusterVersionOperatorPodSelector, &expectedPodCount)
+	require.NoError(t, err, "failed to scale up CVO pods")
```
sebsoto, line 478: I'm not a fan of how we are scaling up and down within different tests. Outside the scope of this PR though.

sebsoto (approved):

PR #872, opened 2022-01-18
WINC-617: Specify container runtime flag and transfer containerd files

mansikulkarni96:
This PR adds a flag through CSV args to specify the Windows host container runtime as docker or containerd. If the flag value is "containerd", the Containerd related files are transferred to the Windows VM.
This PR addresses both [WINC-617](https://issues.redhat.com/browse/WINC-617) and [WINC-694](https://issues.redhat.com/browse/WINC-694).

Ran
`make bundle`

### DIFF
--- a/controllers/controllers.go
+++ b/controllers/controllers.go
@@ -52,6 +52,8 @@ type instanceReconciler struct {
 	recorder record.EventRecorder
 	// platform indicates the cloud on which the cluster is running
 	platform config.PlatformType
+	// dockerRuntime indicates if the container runtime used is docker or containerd
+	dockerRuntime bool
 }
 
 // ensureInstanceIsUpToDate ensures that the given instance is configured as a node and upgraded to the specifications
@@ -71,7 +73,7 @@ func (r *instanceReconciler) ensureInstanceIsUpToDate(instanceInfo *instance.Inf
 	}
 
 	nc, err := nodeconfig.NewNodeConfig(r.k8sclientset, r.clusterServiceCIDR, r.vxlanPort, instanceInfo, r.signer,
-		labelsToApply, annotationsToApply, r.platform)
+		labelsToApply, annotationsToApply, r.platform, r.dockerRuntime)
 	if err != nil {
 		return errors.Wrap(err, "failed to create new nodeconfig")
 	}
@@ -139,7 +141,7 @@ func (r *instanceReconciler) deconfigureInstance(node *core.Node) error {
 	}
 
 	nc, err := nodeconfig.NewNodeConfig(r.k8sclientset, r.clusterServiceCIDR, r.vxlanPort, instance, r.signer,
-		nil, nil, r.platform)
+		nil, nil, r.platform, r.dockerRuntime)
 	if err != nil {
 		return errors.Wrap(err, "failed to create new nodeconfig")
 	}
--- a/main.go
+++ b/main.go
@@ -49,7 +49,10 @@ func init() {
 
 func main() {
 	var debugLogging bool
+	var dockerRuntime bool
+
 	flag.BoolVar(&debugLogging, "debugLogging", false, "Log debug messages")
+	flag.BoolVar(&dockerRuntime, "dockerRuntime", true, "Container runtime to be used")
 
 	// Add flags registered by imported packages (e.g. glog and
 	// controller-runtime)
@@ -164,7 +167,7 @@ func main() {
 	}
 
 	// Setup all Controllers
-	winMachineReconciler, err := controllers.NewWindowsMachineReconciler(mgr, clusterConfig, watchNamespace)
+	winMachineReconciler, err := controllers.NewWindowsMachineReconciler(mgr, clusterConfig, watchNamespace, dockerRuntime)
 	if err != nil {
 		setupLog.Error(err, "unable to create Windows Machine reconciler")
 		os.Exit(1)
@@ -183,7 +186,7 @@ func main() {
 		setupLog.Error(err, "error removing invalid annotations from Linux nodes")
 	}
 
-	configMapReconciler, err := controllers.NewConfigMapReconciler(mgr, clusterConfig, watchNamespace)
+	configMapReconciler, err := controllers.NewConfigMapReconciler(mgr, clusterConfig, watchNamespace, dockerRuntime)
 	if err != nil {
 		setupLog.Error(err, "unable to create ConfigMap reconciler")
 		os.Exit(1)
--- a/pkg/nodeconfig/payload/payload.go
+++ b/pkg/nodeconfig/payload/payload.go
@@ -20,6 +20,11 @@ const (
 	// KubeProxyPath contains the path of the kube-proxy binary. The container image should already have this binary
 	// mounted
 	KubeProxyPath = payloadDirectory + "/kube-node/kube-proxy.exe"
+	// ContainerdPath contains the path of the containerd binary. The container image should already have this binary
+	// mounted
+	ContainerdPath = payloadDirectory + "/containerd/containerd.exe"
+	//HcsshimPath contains the path of the hcsshim binary. The container image should already have this binary mounted
+	HcsshimPath = payloadDirectory + "/containerd/containerd-shim-runhcs-v1.exe"
 	// IgnoreWgetPowerShellPath contains the path of the powershell script which allows wget to ignore certs. The
 	// container image should already have this mounted
 	IgnoreWgetPowerShellPath = payloadDirectory + "/powershell/wget-ignore-cert.ps1"
--- a/pkg/windows/windows.go
+++ b/pkg/windows/windows.go
@@ -192,10 +192,13 @@ type windows struct {
 	defaultShellPowerShell bool
 	// platformType overrides default hostname in bootstrapper
 	platformType string
+	// dockerRuntime indicates if the container runtime used is docker or containerd
+	dockerRuntime bool
 }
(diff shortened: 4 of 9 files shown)


selansen (approved):
/lgtm

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
```diff
@@ -350,6 +350,7 @@ spec:
               containers:
               - args:
                 - --debugLogging
+                - --runtime=docker
```
sebsoto, line 353: This is a feature flag and should be a bool. We either want to use docker or we dont want to use docker, making it a string is adding complexity.

pkg/windows/windows.go
```diff
@@ -212,6 +215,7 @@ func New(workerIgnitionEndpoint, clusterDNS, vxlanPort string, instanceInfo *ins
 			log:                    log,
 			defaultShellPowerShell: defaultShellPowershell(conn),
 			platformType:           platformType,
+			runtime:                runtime,
```
sebsoto, line 218: Is this PR missing code? This isnt being used anywhere, is it?

sebsoto: In that case these commits should be within his PR. In most cases, variables and code in general should only be merged in when they are actually being used.

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
mansikulkarni96: The string flag is added keeping in mind the proposed enhancement to use --runtime option. And having a bool for runtime true/false does not help representing docker/containerd option.

sebsoto: Thanks for pointing that out. Since its in the enhancement thats fine.

controllers/controllers.go
```diff
@@ -52,6 +52,8 @@ type instanceReconciler struct {
 	recorder record.EventRecorder
 	// platform indicates the cloud on which the cluster is running
 	platform config.PlatformType
+	// runtime indicates the container runtime to be used for configuring nodes
+	runtime string
```
sebsoto, line 56: This doesn't need to be a string, a bool `dockerRuntime` or `containerdRuntime` makes more sense here. Moves the validation of the string to earlier in the configuration process.

pkg/windows/windows.go
```diff
@@ -550,6 +563,9 @@ func (vm *windows) changeHostName() error {
 
 // createDirectories creates directories required for configuring the Windows node on the VM
 func (vm *windows) createDirectories() error {
+	if vm.runtime == containerdRuntime {
+		RequiredDirectories = append(RequiredDirectories, containerdDir)
```
sebsoto, line 567: This wont be tested here: https://github.com/openshift/windows-machine-config-operator/blob/c7090d62ed3af48c5114a49324639a3639b71781/test/e2e/delete_test.go#L85

sebsoto (changes requested):
Thanks for the changes

The ticket calls out doc changes are required for this work, should those be updated as well?

main.go
```diff
@@ -49,7 +49,10 @@ func init() {
 
 func main() {
 	var debugLogging bool
+	var runtime string
+
 	flag.BoolVar(&debugLogging, "debugLogging", false, "Log debug messages")
+	flag.StringVar(&runtime, "runtime", "docker", "Container runtime to be used")
```
saifshaikh48, line 55: I am fine with this being a string rather than bool, but we should validate it is an expected value (`containerd` or `docker`).

pkg/windows/windows.go
```diff
@@ -192,10 +192,13 @@ type windows struct {
...
 	// platformType overrides default hostname in bootstrapper
 	platformType string
+	// container runtime to be used
+	runtime string
 }
 
 // New returns a new Windows instance constructed from the given WindowsVM
-func New(workerIgnitionEndpoint, clusterDNS, vxlanPort string, instanceInfo *instance.Info, signer ssh.Signer, platformType string) (Windows, error) {
+func New(workerIgnitionEndpoint, clusterDNS, vxlanPort string, instanceInfo *instance.Info, signer ssh.Signer, platformType,
+	runtime string) (Windows, error) {
```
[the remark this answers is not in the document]
aravindhp: I am assuming `runtime` will be removed in 6.0.0, so the method signature will shrink.

pkg/nodeconfig/payload/payload.go
```diff
@@ -20,6 +20,11 @@ const (
 	// KubeProxyPath contains the path of the kube-proxy binary. The container image should already have this binary
```
aravindhp, line 20: If this commit was co-authored with @selansen , please add `co-authored-by` to the commit message.

pkg/windows/windows.go
```diff
@@ -38,12 +38,18 @@ const (
...
+	// containerdDir is the directory for storing Containerd binary
+	containerdDir = k8sDir + "containerd\\"
 	// windowsExporterPath is the location of the windows_exporter.exe
 	windowsExporterPath = k8sDir + "windows_exporter.exe"
 	// azureCloudNodeManagerPath is the location of the azure-cloud-node-manager.exe
 	azureCloudNodeManagerPath = k8sDir + payload.AzureCloudNodeManager
 	// kubeProxyPath is the location of the kube-proxy exe
 	kubeProxyPath = k8sDir + "kube-proxy.exe"
+	// containerdPath is the location of the containerd exe
+	containerdPath = containerdDir + "containerd.exe"
```
aravindhp, line 50: Please move this below `containerdDir`

pkg/windows/windows.go
```diff
@@ -82,6 +88,8 @@ const (
 	cmdExitNoStatus = "command exited without exit status or exit signal"
 	// removeHNSCommand is the Windows command used to remove HNS network.
 	removeHNSCommand = "Remove-HnsNetwork"
+	//containerdRuntime is containerd runtime name
```
aravindhp, line 91: Won't this also be the name of the service?

aravindhp (changes requested):

pkg/windows/windows.go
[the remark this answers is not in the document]
selansen: I agree

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
selansen: Initially started with boolean in the proposal document and then moved string for more runtime options ( what if we plan to support CRIO in 6.0, just a thought )  and keep the implementation open to accommodate future change. But we know it will not happen. We can move it to boolean.

controllers/controllers.go
selansen: will make the change.

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
aravindhp: >what if we plan to support CRIO in 6.0, just a thought

Not a valid argument 😄 CRI-O does not have Windows support and there are no plans to add support for CRI-O as you called out in the enhancement. Anyhow this is a temporary flag, so whatever is easier.

main.go
selansen: I think either we skip --runtime so that default value will kick in or we should use --runtime=<value>.  while testing our initial draft was PR was just having --runtime. WMCO container didn't come up due to the parameter value being left out. We can test it one more time by removing it and make sure the default value kicks in

pkg/windows/windows.go
selansen: Yes. This will be the service name

pkg/windows/windows.go
mansikulkarni96: thanks for the catch, added the containerdDir as discussed.

main.go
```diff
@@ -49,7 +49,10 @@ func init() {
 
 func main() {
 	var debugLogging bool
+	var dockerRuntime bool
+
 	flag.BoolVar(&debugLogging, "debugLogging", false, "Log debug messages")
+	flag.BoolVar(&dockerRuntime, "dockerRuntime", true, "Container runtime to be used")
```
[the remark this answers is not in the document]
mansikulkarni96: yes you are right, was still testing this out.

test/e2e/delete_test.go
```diff
@@ -82,6 +82,10 @@ func (tc *testContext) testBYOHRemoval(t *testing.T) {
 // address
 func (tc *testContext) checkDirsDoNotExist(address string) (bool, error) {
 	command := ""
+
+	// TODO: Remove appending containerd directory when we default to using runtime containerd
+	windows.RequiredDirectories = append(windows.RequiredDirectories, windows.ContainerdDir)
```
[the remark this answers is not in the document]
mansikulkarni96: The test checks if the directories do not exist.
In case of docker, the directory would not have been created so the test should pass.

aravindhp (commented):
LGTM

saifshaikh48 (commented):
/lgtm

Please mention in the PR description this completes both [WINC-617](https://issues.redhat.com/browse/WINC-617) and [WINC-694](https://issues.redhat.com/browse/WINC-694). It isn't clear to me how WMCB will be made aware of the flag, but that is a upcoming story so no problem.

aravindhp (approved):

PR #891, opened 2022-02-02
WINC-695: Configure containerd service and start it.

mansikulkarni96:
This PR makes the required CNI directory changes to the default containerd config file and stores it in the payload directory. It also creates the Windows containerd service using the config file along with the logging options and starts the service on the Windows node.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,8 +3,7 @@
 ## Introduction
 The Windows Machine Config Operator configures Windows instances into nodes, enabling Windows container workloads to
 be ran within OKD/OCP clusters. Windows instances can be added either by creating a [MachineSet](https://docs.openshift.com/container-platform/4.5/machine_management/creating_machinesets/creating-machineset-aws.html#machine-api-overview_creating-machineset-aws),
-or by specifying existing instances through a ConfigMap. Through either method, the Windows instance must have the
-Docker container runtime installed. The operator will do all the necessary steps to configure the instance so that it
+or by specifying existing instances through a ConfigMap. The operator will do all the necessary steps to configure the instance so that it
 can join the cluster as a worker node.
 
 More design details can be explored in the [WMCO enhancement](https://github.com/openshift/enhancements/blob/master/enhancements/windows-containers/windows-machine-config-operator.md).
@@ -213,6 +212,9 @@ Cluster autoscaling is supported for Windows instances.
 - Define and deploy a [ClusterAutoscaler](https://docs.openshift.com/container-platform/latest/machine_management/applying-autoscaling.html#configuring-clusterautoscaler).
 - Create a Windows node through a MachineSet (see spec in [Usage section](https://github.com/openshift/windows-machine-config-operator#usage)).
 - Define and deploy a [MachineAutoscaler](https://docs.openshift.com/container-platform/latest/machine_management/applying-autoscaling.html#configuring-machineautoscaler), referencing a Windows MachineSet.
+
+### Container Runtime
+Windows instances brought up with WMCO are set up with the containerd container runtime.
 
 ## Development
 
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -91,6 +91,7 @@ RUN make build
 #├── containerd
 #│   ├── containerd.exe
 #│   └── containerd-shim-runhcs-v1.exe
+#|   └── containerd_conf.toml
 #├── hybrid-overlay-node.exe
 #├── kube-node
 #│   ├── kubelet.exe
@@ -117,10 +118,11 @@ COPY --from=build /build/windows-machine-config-operator/windows_exporter/window
 # Copy azure-cloud-node-manager.exe
 COPY --from=build /build/windows-machine-config-operator/cloud-provider-azure/azure-cloud-node-manager.exe .
 
-# Copy containerd.exe and containerd-shim-runhcs-v1.exe
+# Copy containerd.exe, containerd-shim-runhcs-v1.exe and containerd config containerd_conf.toml
 WORKDIR /payload/containerd/
 COPY --from=build /build/windows-machine-config-operator/containerd/bin/containerd.exe .
 COPY --from=build /build/windows-machine-config-operator/hcsshim/containerd-shim-runhcs-v1.exe .
+COPY pkg/internal/containerd_conf.toml .
 
 # Copy kubelet.exe and kube-proxy.exe
 WORKDIR /payload/kube-node/
--- a/config/manifests/bases/windows-machine-config-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/windows-machine-config-operator.clusterserviceversion.yaml
@@ -25,8 +25,7 @@ spec:
     The Windows Machine Config Operator configures Windows Machines into nodes, enabling Windows container workloads to
     be run on OCP clusters. Windows instances can be added either by creating a [MachineSet](https://docs.openshift.com/container-platform/latest/machine_management/creating_machinesets/creating-machineset-aws.html#machine-api-overview_creating-machineset-aws),
     or by specifying existing instances through a [ConfigMap](https://docs.openshift.com/container-platform/latest/windows_containers/configuring-byoh-windows-instance.html)
-    Through either method, the Windows instance to be configured must have the Docker container runtime installed. The operator
-    completes all the necessary steps to configure the underlying VM so that it can join the cluster as a worker node.
+    The operator completes all the necessary steps to configure the Windows instance so that it can join the cluster as a worker node.
 
     Usage of this operator requires a Red Hat OpenShift subscription. Users looking to deploy Windows containers workloads in
     production clusters should acquire a subscription before attempting to install this operator. Users without a subscription
--- a/docs/byoh-instance-pre-requisites.md
+++ b/docs/byoh-instance-pre-requisites.md
@@ -1,7 +1,6 @@
 # BYOH Instance Pre-requisites
 
 The following pre-requisites must be fulfilled in order to add a Windows BYOH node.
-* The Docker container runtime must be installed on the instance.
 * The instance must be on the same network as the Linux worker nodes in the cluster.
 * Port 22 must be open and running [an SSH server](https://docs.microsoft.com/en-us/windows-server/administration/openssh/openssh_install_firstuse).
 * Port 10250 must be open in order for log collection to function.
--- a/pkg/internal/cni-conf-template.json
+++ b/pkg/internal/cni-conf-template.json
@@ -2,7 +2,9 @@
 	"CniVersion":"0.2.0",
 	"Name":"OVNKubernetesHybridOverlayNetwork",
 	"Type":"win-overlay",
+	"apiVersion": 2,
 	"Capabilities":{
+		"portMappings": true,
 		"Dns":true
 	},
 	"Ipam":{
@@ -11,23 +13,36 @@
 	},
 	"Policies":[
 		{
-			"Name":"EndpointPolicy",
-			"Value":{
-				"Type":"OutBoundNAT",
-				"ExceptionList":[
-					"service_network_cidr"
-				],
-				"DestinationPrefix":"",
-				"NeedEncap":false
+			"Name": "EndpointPolicy",
+			"Value": {
+				"Type": "OutBoundNAT",
+				"Settings": {
+					"ExceptionList": [
+						"service_network_cidr"
+					],
+					"DestinationPrefix": "",
+					"NeedEncap": false
+				}
 			}
 		},
 		{
-			"Name":"EndpointPolicy",
-			"Value":{
-				"Type":"ROUTE",
-				"ExceptionList":[],
-				"DestinationPrefix":"service_network_cidr",
-				"NeedEncap":true
+			"Name": "EndpointPolicy",
+			"Value": {
+				"Type": "SDNROUTE",
+				"Settings": {
+					"ExceptionList": [],
+					"DestinationPrefix": "service_network_cidr",
+					"NeedEncap": true
+				}
+			}
+		},
+		{
+			"Name": "EndpointPolicy",
+			"Value": {
+				"Type": "ProviderAddress",
+				"Settings": {
+					"ProviderAddress": "ip_host"
+				}
 			}
 		}
 	]
--- a/pkg/internal/containerd_conf.toml
+++ b/pkg/internal/containerd_conf.toml
@@ -1,0 +1,174 @@
+disabled_plugins = []
+imports = []
+oom_score = 0
+plugin_dir = ""
+required_plugins = []
+root = "C:\\ProgramData\\containerd\\root"
+state = "C:\\ProgramData\\containerd\\state"
+version = 2
+
+[cgroup]
+  path = ""
+
+[debug]
+  address = ""
+  format = ""
+  gid = 0
+  level = ""
+  uid = 0
+
+[grpc]
+  address = "\\\\.\\pipe\\containerd-containerd"
+  gid = 0
+  max_recv_message_size = 16777216
+  max_send_message_size = 16777216
+  tcp_address = ""
+  tcp_tls_cert = ""
+  tcp_tls_key = ""
+  uid = 0
+
+[metrics]
+  address = ""
+  grpc_histogram = false
+
+[plugins]
+
+  [plugins."io.containerd.gc.v1.scheduler"]
+    deletion_threshold = 0
+    mutation_threshold = 100
+    pause_threshold = 0.02
+    schedule_delay = "0s"
+    startup_delay = "100ms"
+
+  [plugins."io.containerd.grpc.v1.cri"]
+    disable_apparmor = false
+    disable_cgroup = false
+    disable_hugetlb_controller = false
+    disable_proc_mount = false
+    disable_tcp_service = true
+    enable_selinux = false
+    enable_tls_streaming = false
+    ignore_image_defined_volumes = false
+    max_concurrent_downloads = 3
+    max_container_log_line_size = 16384
+    netns_mounts_under_state_dir = false
+    restrict_oom_score_adj = false
+    sandbox_image = "k8s.gcr.io/pause:3.5"
+    selinux_category_range = 0
+    stats_collect_period = 10
+    stream_idle_timeout = "4h0m0s"
+    stream_server_address = "127.0.0.1"
+    stream_server_port = "0"
+    systemd_cgroup = false
+    tolerate_missing_hugetlb_controller = false
+    unset_seccomp_profile = ""
+
+    [plugins."io.containerd.grpc.v1.cri".cni]
+      bin_dir = "C:\\Program Files\\containerd\\cni\\bin"
+      conf_dir = "C:\\Program Files\\containerd\\cni\\conf"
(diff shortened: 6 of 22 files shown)


pkg/internal/containerd_conf.toml
```diff
@@ -64,8 +64,8 @@ version = 2
     unset_seccomp_profile = ""
 
     [plugins."io.containerd.grpc.v1.cri".cni]
-      bin_dir = "C:\\Program Files\\containerd\\cni\\bin"
-      conf_dir = "C:\\Program Files\\containerd\\cni\\conf"
+      bin_dir = "C:\\k\\cni"
+      conf_dir = "C:\\k\\cni\\conf"
```
saifshaikh48, line 68: I think `conf` should be `config` in the `conf_dir` 
https://github.com/openshift/windows-machine-config-operator/blob/ce2d16caf8ba25c739b275bc6311617983cf44a0/pkg/windows/windows.go#L37-L40

build/Dockerfile
```diff
@@ -91,6 +91,7 @@ RUN make build
 #├── containerd
```
saifshaikh48, line 91: nit: typo in commit message: `tranfer` --> `transfer`

pkg/internal/containerd_conf.toml
mansikulkarni96: you are right, will fix this.

aravindhp: Things like this is what is concerning when merging the PR without any e2e testing.

pkg/internal/containerd_conf.toml
```diff
@@ -0,0 +1,174 @@
...
+    path = "C:\\ProgramData\\containerd\\root\\opt"
+
+  [plugins."io.containerd.internal.v1.restart"]
+    interval = "10s"
+
+  [plugins."io.containerd.metadata.v1.bolt"]
+    content_sharing_policy = "shared"
+
+  [plugins."io.containerd.runtime.v2.task"]
+    platforms = ["windows/amd64", "linux/amd64"]
```
jrvaldes, line 142: Is `linux/amd64` needed?

mansikulkarni96: This is the default containerd generated config file

pkg/internal/cni-conf-template.json
```diff
@@ -11,23 +13,36 @@
...
+				}
+			}
+		},
+		{
+			"Name": "EndpointPolicy",
+			"Value": {
+				"Type": "ProviderAddress",
+				"Settings": {
+					"ProviderAddress": "ip_host"
+				}
```
sebsoto, line 45: Are these changes compatible with dockershim?
If you are leaving the `dockerRuntime` flag within the code we have to test both paths, dockershim and containerd.

From our discussion, if we are targeting 6.0.0 which is incompatible with dockershim, the `dockerRuntime` flag should be removed, as by leaving it in we are include a broken code path.

pkg/windows/windows.go
```diff
@@ -656,6 +657,8 @@ func (vm *windows) runBootstrapper() error {
 	}
 	wmcbInitializeCmd += " --platform-type=" + vm.platformType
 
+	wmcbInitializeCmd += " --dockerRuntime=" + strconv.FormatBool(vm.dockerRuntime)
```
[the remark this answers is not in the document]
sebsoto: Are there existing e2e tests validating that the containerd runtime is being used?
If not, something must be added, could be as simple as checking the kubelet args.

sebsoto (changes requested):

build/Dockerfile.wmco
```diff
@@ -29,6 +29,9 @@ WORKDIR /payload/powershell/
 COPY pkg/internal/wget-ignore-cert.ps1 .
 COPY pkg/internal/hns.psm1 .
 
+WORKDIR /payload/cni/
```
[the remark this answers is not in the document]
mansikulkarni96: I think this got mixed up during one of the rebases, will fix this.

pkg/windows/windows.go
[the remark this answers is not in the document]
sebsoto: :thinking: 

I think thats fine, adding a test to make sure we are using containerd within this PR is going to be dead code soon anyway.
And looking at the [kubelet log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/891/pull-ci-openshift-windows-machine-config-operator-master-aws-e2e-ccm-install/1516540294080761856/artifacts/aws-e2e-ccm-install/windows-e2e-operator-test/artifacts/nodes/ip-10-0-144-53.ec2.internal/kubelet/kubelet.log) we can see that the flag is being passed down to WMCB properly

pkg/internal/cni-conf-template.json
mansikulkarni96: dockerRuntime flag usage has been removed in first commit.

sebsoto (changes requested):
Great work, almost LGTM

pkg/internal/containerd_conf.toml
```diff
@@ -53,7 +53,7 @@ version = 2
     max_container_log_line_size = 16384
     netns_mounts_under_state_dir = false
     restrict_oom_score_adj = false
-    sandbox_image = "k8s.gcr.io/pause:3.5"
+    sandbox_image = "k8s.gcr.io/pause:3.6"
```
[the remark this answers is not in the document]
selansen: From K8s doc:
--pod-infra-container-image string     Default: k8s.gcr.io/pause:3.6
--
  | Specified image will not be pruned by the image garbage collector. When container-runtime is set to docker, all containers in each pod will use the network/IPC namespaces from this image. Other CRI implementations have their own configuration to set this image.

So containerd pause image version will become the source of truth.

pkg/nodeconfig/network.go
```diff
@@ -15,6 +15,7 @@ import (
 // cniConf contains the structure of the CNI template
 type cniConf struct {
 	CNIVersion   string       `json:"cniVersion"`
+	ApiVersion   int          `json:"apiVersion"`
```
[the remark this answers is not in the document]
selansen: This is parsed in CNI and we are following the same acronym used by win-bridge to be in sync with them. 
if we modify it here, we will have to change it in upstream CNI code as well. Prefer to keep it in the same way as win-bridge CNI.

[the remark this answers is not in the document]
mansikulkarni96: @selansen is referring to this field in upstream https://github.com/containernetworking/plugins/blob/0c3933576590734ed53768ecd84ea3e0545158a5/pkg/hns/netconf_windows.go#L37

pkg/internal/containerd_conf.toml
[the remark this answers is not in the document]
mansikulkarni96: @selansen I think we can remove the pause image specification from WMCB side as part of the kubelet flag cleanup.

sebsoto:
LGTM, will let others who reviewed leave the final lgtm

pkg/internal/containerd_conf.toml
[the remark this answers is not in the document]
jrvaldes: Why not use `mcr.microsoft.com/oss/kubernetes/pause:3.6` that is built from the same source as the Kubernetes maintained image (k8s.gcr.io/pause) but all of the Windows binaries are [authenticode signed](https://docs.microsoft.com/en-us/windows-hardware/drivers/install/authenticode) by Microsoft?

Taken form k8s [docs](https://kubernetes.io/docs/setup/production-environment/windows/intro-windows-in-kubernetes/#pause-container).

pkg/nodeconfig/network.go
```diff
@@ -124,11 +131,14 @@ func (nw *network) populateCniConfig(serviceCIDR string, templatePath string) (s
 }
 
 // populateCfgPolicies populates the policies in cniConf struct with serviceCIDR information
-func populateCfgPolicies(cniCfgPolicies *policies, serviceCIDR string) error {
-	if len(*cniCfgPolicies) < 2 || len((*cniCfgPolicies)[0].Value.ExceptionList) == 0 || (*cniCfgPolicies)[1].Value.DestinationPrefix == "" {
+func populateCfgPolicies(cniCfgPolicies *policies, serviceCIDR string, ipAddress string) error {
+	if len(*cniCfgPolicies) < 2 || len((*cniCfgPolicies)[0].Value.Settings.ExceptionList) == 0 ||
+		(*cniCfgPolicies)[1].Value.Settings.DestinationPrefix == "" ||
+		(*cniCfgPolicies)[2].Value.Settings.ProviderAddress == "" {
```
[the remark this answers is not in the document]
jrvaldes: In addition, would be nice to [unit test](https://github.com/openshift/windows-machine-config-operator/pull/891/commits/43289a2b11b2df6d2ecf106261f82ec548c0643f#diff-fe7716bbf35bcd81a5898b2ef4c92b8a2b7fed03bd52b6dae57b0d7dc005e188R17) all these cases.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -412,7 +412,7 @@ func (nc *nodeConfig) configureCNI() error {
 		return errors.Wrap(err, "error populating host subnet in node network")
 	}
 	// populate the CNI config file with the host subnet and the service network CIDR
```
jrvaldes, line 414: Update the comment to account for the addition of `nc.GetIPv4Address()`

jrvaldes (changes requested):
@mansikulkarni96  Thanks for working on this, added a few comments. PTAL

README.md
```diff
@@ -214,6 +213,9 @@ Cluster autoscaling is supported for Windows instances.
 - Create a Windows node through a MachineSet (see spec in [Usage section](https://github.com/openshift/windows-machine-config-operator#usage)).
 - Define and deploy a [MachineAutoscaler](https://docs.openshift.com/container-platform/latest/machine_management/applying-autoscaling.html#configuring-machineautoscaler), referencing a Windows MachineSet.
 
+### Containerd Runtime
```
jrvaldes, line 216: I'd rather name this section `Container Runtime`.

README.md
```diff
@@ -214,6 +213,9 @@ Cluster autoscaling is supported for Windows instances.
...
+Windows instances brought up with WMCO are set up with Containerd Container runtime.
```
jrvaldes, line 217: Consider including a link to `Containerd` main page.

config/manifests/bases/windows-machine-config-operator.clusterserviceversion.yaml
```diff
@@ -25,8 +25,7 @@ spec:
     The Windows Machine Config Operator configures Windows Machines into nodes, enabling Windows container workloads to
     be run on OCP clusters. Windows instances can be added either by creating a [MachineSet](https://docs.openshift.com/container-platform/latest/machine_management/creating_machinesets/creating-machineset-aws.html#machine-api-overview_creating-machineset-aws),
     or by specifying existing instances through a [ConfigMap](https://docs.openshift.com/container-platform/latest/windows_containers/configuring-byoh-windows-instance.html)
-    Through either method, the Windows instance to be configured must have the Docker container runtime installed. The operator
-    completes all the necessary steps to configure the underlying VM so that it can join the cluster as a worker node.
+    The operator completes all the necessary steps to configure the underlying VM so that it can join the cluster as a worker node.
```
jrvaldes, line 28:
```suggestion
    The operator completes all the necessary steps to configure the Windows instance so that it can join the cluster as a worker node.
```

docs/byoh-instance-pre-requisites.md
```diff
@@ -1,7 +1,6 @@
 # BYOH Instance Pre-requisites
 
 The following pre-requisites must be fulfilled in order to add a Windows BYOH node.
-* The Docker container runtime must be installed on the instance.
```
jrvaldes, line 4: The `containers feature` must be enabled in the Windows instance, is that documented somewhere?

README.md
```diff
@@ -3,8 +3,7 @@
 ## Introduction
```
jrvaldes, line 3: [Set up the container runtime](https://github.com/openshift/windows-machine-config-operator/blob/6e5858b8bb64eee88928f62da8da6033df485a52/docs/vsphere-golden-image.md#4-set-up-the-container-runtime) section may need an update too?

pkg/internal/containerd_conf.toml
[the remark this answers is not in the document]
selansen: That is good point @jrvaldes. @mansikulkarni96 can you pls quick make that change?

pkg/windows/windows.go
```diff
@@ -374,13 +384,42 @@ func (vm *windows) Configure() error {
...
 	return vm.runBootstrapper()
 }
 
+// configureContainerd starts the Windows containerd service
+func (vm *windows) configureContainerd() error {
+	// set Windows defender exclusions for containerd
+	setExclusionCmd := "Add-MpPreference -ExclusionProcess " + containerdPath
+	out, err := vm.Run(setExclusionCmd, true)
+	if err != nil {
+		vm.log.Info("setting Windows defender exclusion failed", "command", setExclusionCmd,
```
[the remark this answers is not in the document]
selansen: yes. This is kind of a must until containerd fix "downloadtimeexcced timeout"  issue. Without this, image download will take a much longer time. So we can make this an error message even though we don't stop containerd configuration.

README.md
selansen: I am not suer if we are referring other submodules links like this. Ex contianernetworkruntime, kubeproxy and kubelet. Is there any specific reason you want to add containerd webpage link here?

docs/byoh-instance-pre-requisites.md
selansen: That is part of 2022 server. we should address that as part of 2022 server support.

README.md
selansen: Good catch @jrvaldes I think we need to remove that section. From now on WMCO will manage the runtime so we don't need that part.

pkg/internal/containerd_conf.toml
```diff
@@ -64,8 +64,8 @@ version = 2
     unset_seccomp_profile = ""
 
     [plugins."io.containerd.grpc.v1.cri".cni]
-      bin_dir = "C:\\Program Files\\containerd\\cni\\bin"
-      conf_dir = "C:\\Program Files\\containerd\\cni\\conf"
+      bin_dir = "C:\\k\\cni"
+      conf_dir = "C:\\k\\cni\\config"
```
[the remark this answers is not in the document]
mansikulkarni96: That commit is not doing anything apart from generating the default config file. This has been kept here so it is easier to review and one can skip reviewing earlier commit. All the changes done by us are kept in this commit.

pkg/windows/windows.go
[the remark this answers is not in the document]
jrvaldes: >Not sure error logging would helpful for the user here, the error is being returned later which should lead to reconcile here.

Given the error is being returned later, I suggest printing the command information with a lower verbosity level (debug), otherwise no logging at all.

pkg/windows/windows.go
```diff
@@ -374,13 +384,42 @@ func (vm *windows) Configure() error {
...
+	containerdServiceArgs := "--config " + containerdConfPath + " --log-file " + ContainerdLogDir + "containerd.log" +
+		" --log-level=info" + " --run-service"
+
+	containerdService, err := newService(containerdPath, containerdServiceName, containerdServiceArgs, nil)
+	if err != nil {
+		return errors.Wrapf(err, "error creating %s service object", containerdServiceName)
+	}
+
+	if err := vm.ensureServiceIsRunning(containerdService); err != nil {
+		return errors.Wrapf(err, "error ensuring %s Windows service has started running", containerdServiceName)
```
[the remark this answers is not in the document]
jrvaldes: Sure thing and I'm fine with that, my two cents:

The already existing constant should be used while [creating](https://github.com/openshift/windows-machine-config-operator/pull/891/commits/3331ed3c44b224d222bc9714d05dde9543083238#diff-4d8fe3bb66a872f0e4a277f820d60d2efa722f09aa89516be74b98c06b57f3bdR461) the service. Now, after the service is created, as a best practice, we should rely on the returned instance (`containerdService`). 

For example; in this case `newService()` return a service where the passed name is used as-is. In a different scenario, where the `newService()` adds a prefix or uses a generated name, using the constant in the log won't be accurate.

pkg/nodeconfig/network.go
```diff
@@ -124,11 +131,14 @@ func (nw *network) populateCniConfig(serviceCIDR string, templatePath string) (s
 }
 
 // populateCfgPolicies populates the policies in cniConf struct with serviceCIDR information
-func populateCfgPolicies(cniCfgPolicies *policies, serviceCIDR string) error {
-	if len(*cniCfgPolicies) < 2 || len((*cniCfgPolicies)[0].Value.ExceptionList) == 0 || (*cniCfgPolicies)[1].Value.DestinationPrefix == "" {
+func populateCfgPolicies(cniCfgPolicies *policies, serviceCIDR string, ipAddress string) error {
```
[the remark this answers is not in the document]
jrvaldes: Right, a comment always helps.

pkg/nodeconfig/network.go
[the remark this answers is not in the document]
jrvaldes: I don't see full coverage for all conditional statements. For instance, [mockInvalidPolicies()](https://github.com/openshift/windows-machine-config-operator/pull/891/commits/54c3342602b9dfd7a39198a780c97859aab2c9e9#diff-fe7716bbf35bcd81a5898b2ef4c92b8a2b7fed03bd52b6dae57b0d7dc005e188L79) will always returns a collection with  three (3) policies, so the length check will never fail.

```
	if len(*cniCfgPolicies) < numberOfPolicies {
		return errors.Errorf (...)
	}
```

jrvaldes:
LGTM, and will let others review.

pkg/nodeconfig/network.go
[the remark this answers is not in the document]
mansikulkarni96: addressed

selansen (approved):
/lgtm

README.md
```diff
@@ -214,6 +213,9 @@ Cluster autoscaling is supported for Windows instances.
 - Create a Windows node through a MachineSet (see spec in [Usage section](https://github.com/openshift/windows-machine-config-operator#usage)).
 - Define and deploy a [MachineAutoscaler](https://docs.openshift.com/container-platform/latest/machine_management/applying-autoscaling.html#configuring-machineautoscaler), referencing a Windows MachineSet.
 
+### Container Runtime
+Windows instances brought up with WMCO are set up with Containerd Container runtime.
```
sebsoto, line 217: Based on their docs, as well as the [k8s docs](https://kubernetes.io/docs/setup/production-environment/container-runtimes/#containerd) `containerd` should be all lowercase
additionally i think it might be more correct to say `the containerd container runtime`

sebsoto (changes requested):

mansikulkarni96:
cc: @mburke5678

PR #894, opened 2022-02-02
[csr_controller] RetryOnConflict in CSR approvals

saifshaikh48:
This PR consolidates the retry logic introduced in 223c8b95202a17b7072743bea44010ce3f5700c9 using [RetryOnConflict](https://github.com/kubernetes/client-go/blob/v0.23.3/util/retry/util.go#L103).

### DIFF
--- a/controllers/certificatesigningrequests_controller.go
+++ b/controllers/certificatesigningrequests_controller.go
@@ -24,8 +24,8 @@ import (
 	k8sapierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
+	k8sretry "k8s.io/client-go/util/retry"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/event"
@@ -35,7 +35,6 @@ import (
 	"github.com/openshift/windows-machine-config-operator/pkg/cluster"
 	"github.com/openshift/windows-machine-config-operator/pkg/condition"
 	"github.com/openshift/windows-machine-config-operator/pkg/csr"
-	"github.com/openshift/windows-machine-config-operator/pkg/retry"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


aravindhp (commented):
Thanks for addressing this. LGTM

controllers/certificatesigningrequests_controller.go
```diff
@@ -102,54 +100,35 @@ func (r *certificateSigningRequestsReconciler) Reconcile(ctx context.Context,
...
-	err := wait.Poll(retry.Interval, retry.ResourceChangeTimeout, func() (bool, error) {
-		err := r.validateAndApprove(req)
-		if err == nil {
-			return true, nil
-		}
-		if !k8sapierrors.IsConflict(err) {
-			// If the error is not an update conflict, return error immediately
-			return false, err
+// validateAndApprove handles the CSR validation and approval. Process wrapped in retry logic case of update conflicts
+func (r *certificateSigningRequestsReconciler) validateAndApprove(req *certificates.CertificateSigningRequest) error {
```
sebsoto, line 107: naming `req` overloads the naming `req ctrl.Request` passed into all reconcile functions.

controllers/certificatesigningrequests_controller.go
```diff
@@ -102,54 +100,35 @@ func (r *certificateSigningRequestsReconciler) Reconcile(ctx context.Context,
...
+	err := k8sretry.RetryOnConflict(k8sretry.DefaultBackoff, func() error {
```
sebsoto, line 108: When you are doing this retry you need to fetch the CSR each retry, else you are acting on stale information. see function documentation [here](https://pkg.go.dev/k8s.io/client-go/util/retry#RetryOnConflict)

I would think you'd need to pass in the namespaced name to this function.
the get being done within the `Reconcile` func should be removed and that logic should also be captured within the function passed into `RetryOnConflict`

sebsoto (changes requested):

controllers/certificatesigningrequests_controller.go
saifshaikh48: Great catch, definitely deleted then ignored the re-fetching. I'll rework this.

saifshaikh48: Done below

aravindhp: Nice catch @sebsoto 👍🏽

controllers/certificatesigningrequests_controller.go
```diff
@@ -90,66 +89,49 @@ func (r *certificateSigningRequestsReconciler) Reconcile(ctx context.Context,
...
-func (r *certificateSigningRequestsReconciler) validateAndApprove(request *certificates.CertificateSigningRequest) error {
-	// If a CSR is approved/denied after being added to the queue, but before we reconcile it,
-	// trying to approve it again will result in an error and cause a loop.
-	// Return early if the CSR has been approved/denied externally.
-	if !isPending(request) {
-		r.log.Info("CSR is already approved/denied", "Name", request.Name)
+		if err := csrApprover.Approve(); err != nil {
+			// have to return err itself here (not wrapped inside another error) so RetryOnConflict can identify it
+			return err
+		}
```
sebsoto, line 128: `return csrApprover.Approve()` would suffice here, would it not?

sebsoto (changes requested):

jrvaldes:
/hold

for backports. @saifshaikh48  add the corresponding `cherry-pick` commands.

alinaryan:
Nice work,
/lgtm

saifshaikh48:
4.9 backport should go in after https://github.com/openshift/windows-machine-config-operator/pull/907

saifshaikh48:
/cherry-pick release-4.9

#907 merged so this should go in cleanly

PR #897, opened 2022-02-08
WINC-724: Managed service string

sebsoto:


pkg/windows/windows.go
```diff
@@ -650,6 +653,10 @@ func (vm *windows) ensureServiceIsRunning(svc *service) error {
 	if err := vm.startService(svc); err != nil {
 		return errors.Wrapf(err, "error starting %s Windows service", svc.name)
 	}
+	if err := vm.setServiceDescription(svc.name); err != nil {
```
aravindhp, line 656: Shouldn't this be done before the service is started? In addition:
- You should be augmenting `vm.createService()` to do this and not doing it after the fact. 
- `vm.ServiceExists()` should also be ensuring the service exists with the expected description and add it if it does not

PS: I know all this code will be removed in 6.0 but we are going to ship with this code for 5.0 and it is best it is written correctly.

pkg/windows/windows.go
```diff
@@ -647,9 +650,13 @@ func (vm *windows) ensureServiceIsRunning(svc *service) error {
 			return errors.Wrapf(err, "error creating %s Windows service", svc.name)
 		}
 	}
+	if err := vm.setServiceDescription(svc.name); err != nil {
+		return errors.Wrapf(err, "error setting %s Windows service description", svc.name)
```
jrvaldes, line 654:
```suggestion
		return errors.Wrapf(err, "error setting description in Windows service: %s ", svc.name)
```

test/e2e/validation_test.go
```diff
@@ -373,33 +379,36 @@ func (tc *testContext) runJob(name string, command []string) (string, error) {
...
-		fields := strings.Split(line, ",")
-		if len(fields) != 2 {
+		// Split into 3 substrings, Name, State, Description. The description can contain a comma, so SplitN is required
+		fields := strings.SplitN(line, ",", 3)
+		if len(fields) != 3 {
 			return nil, errors.New("expected comma separated values, found: " + line)
 		}
-		services[strings.Trim(fields[0], "\"")] = strings.Trim(fields[1], "\"")
+		services[strings.Trim(fields[0], "\"")] = winService{state: strings.Trim(fields[1], "\""),
+			description: strings.Trim(fields[2], "\"")}
```
jrvaldes, line 411: To improve readability,  consider doing the cleanup first, then the assignation.
For example:
```
name := strings.Trim(fields[0], "\"")
state := strings.Trim(fields[1], "\"")
...
services[name] = winService{state: state, ... }

```

test/e2e/validation_test.go
```diff
@@ -373,33 +379,36 @@ func (tc *testContext) runJob(name string, command []string) (string, error) {
...
-	// This command returns CR+newline separated quoted CSV entries consisting of service name and status. For example:
-	// "kubelet","Running"\r\n"VaultSvc","Stopped"
-	command := "Get-Service | Select-Object -Property Name,Status | ConvertTo-Csv -NoTypeInformation"
+// getWinServices returns a map of Windows services from the instance with the given address, the map key being the
+// service's name
+func (tc *testContext) getWinServices(addr string) (map[string]winService, error) {
+	// This command returns CR+newline separated quoted CSV entries consisting of service name, state and description.
+	// For example: "kubelet","Running","OpenShift managed kubelet"\r\n"VaultSvc","Stopped",
+	command := "Get-CimInstance -ClassName Win32_Service | Select-Object -Property Name,State,Description | " +
+		"ConvertTo-Csv -NoTypeInformation"
```
jrvaldes, line 388: This command returns all Windows services and `testExpectedServicesRunning()` only checks Windows services managed by WMCO (i.e. `windows.RequiredServices`). 

To reduce the result set, consider adding a filter in this command to return only the services that contain the `windows.ManagedTag` in the description field.

jrvaldes (commented):
@sebsoto  
Thanks for working on this, comments added. PTAL

pkg/windows/windows.go
sebsoto: I moved the the setServiceDescription() call to before the service is started

Keeping setServiceDescription() separate from the createService func allows us to set ensure the correctness of the service description in both code paths, when the service already exists and where it needs to be created.

Additionally, it makes more sense to me to just set the service, as it would require a separate ssh command to check the value of the description. So always setting it to the proper value simplifies the code and reduces the ssh command count

aravindhp: >Keeping setServiceDescription() separate from the createService func allows us to set ensure the correctness of the service description in both code paths, when the service already exists and where it needs to be created.

You can do this otherwise also. You can update `ensureServiceExists` to also check for the description and only call `setServiceDescription` only if it is not present. But I think you would need to move away from `sc.exe qe` for that which IMO is a good thing. 

>Additionally, it makes more sense to me to just set the service, as it would require a separate ssh command to check the value of the description. So always setting it to the proper value simplifies the code and reduces the ssh command count

I can make the argument that you are making an unnecessary extra SSH call all the time in the case of the service existing. Moreover you shouldn't need a separate command if you move away from `sc.exe`. Is it that complicated to use a PowerShell cmdlet instead?

sebsoto: Handling this by moving away from sc.exe and editing CimInstance objects is something I considered, but I felt is out of scope. Moving away from sc.exe is deceptively challenging, as we have ensure parsing compatibility with instances using powershell as the default SSH shell, and instances using cmd. Its possible things might have changed since I did that investigation, but at that point getting powershell commands to parse properly in both cases was a huge time sink.

As a note, [Use powershell consistently throughout WMCO](https://issues.redhat.com/browse/WINC-636) has been in our backlog for a long time.

aravindhp: OK, given what you are saying I will let this through but I will be a hard no on using `sc,exe` when installing WICD 😄 So please don't restate this argument in that context. 

>As a note, [Use powershell consistently throughout WMCO](https://issues.redhat.com/browse/WINC-636) has been in our backlog for a long time.

And it is going to stay that way now that we are most probably going to use the native service library in WICD, so the only place PowerShell should be used is wrt to managing WICD.

test/e2e/validation_test.go
aravindhp: While it applies in this case, please refrain from adding single use variables in general for the sake of "readability".

pkg/windows/windows.go
selansen: Considering we will be using service description during the upgrade, shouldn't we retry the setting in case if set call fails?
Also, it is good to have some comments on the function description. The current description just talks about checking service and creating it if doesn't exist.

pkg/windows/windows.go
```diff
@@ -647,9 +650,13 @@ func (vm *windows) ensureServiceIsRunning(svc *service) error {
 			return errors.Wrapf(err, "error creating %s Windows service", svc.name)
 		}
 	}
+	if err := vm.setServiceDescription(svc.name); err != nil {
+		return errors.Wrapf(err, "error setting description of the %s Windows service", svc.name)
+	}
 	if err := vm.startService(svc); err != nil {
 		return errors.Wrapf(err, "error starting %s Windows service", svc.name)
 	}
+
```
selansen, line 659: nit: remove the extra line

test/e2e/validation_test.go
selansen: Good point @jrvaldes

pkg/windows/windows.go
sebsoto: > Considering we will be using service description during the upgrade, shouldn't we retry the setting in case if set call fails?

Retries are built into the reconciliation loop, if the description setting fails, the reconciliation will not move past that point until it succeeds.

>  Also, it is good to have some comments on the function description. The current description just talks about checking service and creating it if doesn't exist.

Changed to make the function description explicit

test/e2e/validation_test.go
```diff
@@ -35,6 +35,12 @@ import (
 	"github.com/openshift/windows-machine-config-operator/pkg/windows"
 )
 
+// winService contains information regarding a Windows service's current state
+type winService struct {
```
jrvaldes, line 39: What about exposing and extending the existing [service](https://github.com/openshift/windows-machine-config-operator/blob/22ad459356cb6514095986f6a776870f0a29d57a/pkg/windows/service.go#L6) struct, instead of defining a new one?

jrvaldes (commented):
Minor suggestion. Otherwise, LGTM

aravindhp (approved):
/approve
/hold

Waiting on lgtm from @selansen

test/e2e/validation_test.go
aravindhp: Or I could be wrong given you have to apply it in the context of the VM object.

aravindhp: I revisited this. `svc.setDescription()` is not doable given it needs a reference to the VM object. Passing in the VM object would actually be introducing the inverted dependency. Sorry for the noise.

aravindhp:
/hold

@sebsoto @selansen please investigate [vsphere-e2e-operator](https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-openshift-windows-machine-config-operator-master-vsphere-e2e-operator) failure.

selansen:
Not sure why dos2unix install fails only in wsphere platform.  Logs in was target show it eventually installs dos2unix rpm. But it errors out in vsphere.

aravindhp:
That seems to be a different issue on your PR. This is the [failure](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/897/pull-ci-openshift-windows-machine-config-operator-master-vsphere-e2e-operator/1492183935256367104) I am seeing which seems to be infra related:

```
INFO[2022-02-11T17:14:20Z] ci-operator version v20220211-a67395004-dirty 
INFO[2022-02-11T17:14:20Z] Loading configuration from https://config.ci.openshift.org/ for openshift/windows-machine-config-operator@master 
2022/02/11 17:14:20 [DEBUG] GET https://config.ci.openshift.org/config?branch=master&org=openshift&repo=windows-machine-config-operator
INFO[2022-02-11T17:14:22Z] Resolved source https://github.com/openshift/windows-machine-config-operator to master@691358f1, merging: #897 e3d1aa7b @sebsoto 
INFO[2022-02-11T17:14:23Z] Building release initial from a snapshot of ocp/4.11 
INFO[2022-02-11T17:14:23Z] Building release latest from a snapshot of ocp/4.11 
INFO[2022-02-11T17:14:23Z] Using namespace https://console-openshift-console.apps.build01-us-west-2.vmc.ci.openshift.org/k8s/cluster/projects/ci-op-s2qsiyd5 
INFO[2022-02-11T17:14:23Z] Running [input:root], [input:ocp_builder_rhel-8-golang-1.17-openshift-4.10], [input:origin-centos-8], [release-inputs:initial], [release-inputs:latest], src, windows-machine-config-operator-test, [output:stable:windows-machine-config-operator-test], [images], [release:latest], vsphere-e2e-operator 
INFO[2022-02-11T17:14:38Z] Ran for 16s                                  
ERRO[2022-02-11T17:14:38Z] Some steps failed:                           
ERRO[2022-02-11T17:14:38Z] 
  * could not initialize namespace: failed to wait for authentication cache to warm up after 15s: projects.project.openshift.io "ci-op-s2qsiyd5" is forbidden: User "system:serviceaccount:ci:ci-operator" cannot get resource "projects" in API group "project.openshift.io" in the namespace "ci-op-s2qsiyd5"
```

selansen:
Even though the current one seems to be an infra issue, @sebsoto's older CI's in the same PR, I am seeing the same issue that I am running into
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/897/pull-ci-openshift-windows-machine-config-operator-master-vsphere-e2e-operator/1492164289153208320

aravindhp:
Sure but you have to get over the infra issue first.

PR #898, opened 2022-02-08
Document BYOH instance reusability

sebsoto:


sebsoto:
Closing as unsure if this is the correct messaging

PR #892, opened 2022-02-02
WINC-602: [wmco] Support Windows Server 2022

saifshaikh48:
This PR removes SAC Windows Server 2004 (EOL as of 2021-12-14) from our CI and docs,
and replaces it with LTSC Windows Server 2022 
(support announced here kubernetes/kubernetes#104438, EOL is 2026-10-13).
It also improves documentation around building the vSphere golden image using Packer.

Windows Server 2022 is tested in the vSphere environments. 
The golden image is built with a private networking OS fix supplied my Microsoft, 
so official support is on hold until they release the patch publicly.
Support on AWS and Azure is not included in this work.

PR #920, opened 2022-02-11
[hack] Support specifying Windows Server versions in machineset.sh

saifshaikh48:
Allows for easy MachineSet creation on all supported cloud providers
with multiple Windows Server versions.

### DIFF
--- a/hack/machineset.sh
+++ b/hack/machineset.sh
@@ -1,21 +1,28 @@
 #!/usr/bin/env bash
 
-# machineset.sh - create a Windows MachineSet for Windows Machine Config Operator
-# Creates a MachineSet.yaml file and apply/delete the MachineSet if optional action is provided
-#
-# USAGE
-#    machineset.sh
-# OPTIONS
-#    $1      Action       (Optional) apply/delete the MachineSet
-# PREREQUISITES
-#    oc                   to fetch cluster info and apply/delete MachineSets on the cluster(cluster should be logged in)
-#    aws                  to fetch Windows AMI id for AWS platform (only required for clusters running on AWS)
 set -euo pipefail
+
+function help() {
+    echo "USAGE: machineset.sh [OPTIONS]... (apply|delete)"
+    echo "Creates a Windows MachineSet.yaml file for Windows Machine Config Operator on an OpenShift cluster"
+    echo ""
+    echo "OPTIONS:"
+    echo "-v=        Windows Server version of VMs associated with the resulting MachineSet. Defaults to 2022"
+    echo "-a         create or apply changes to the MachineSet resource"
+    echo "-d         delete the MachineSet resource"
+    echo ""
+    echo "PREREQUISITES:"
+    echo "oc         to fetch cluster info and apply/delete MachineSets on the cluster(cluster should be logged in)"
+    echo "aws        to fetch Windows AMI id for AWS platform (only required for clusters running on AWS)"
+    echo ""
+    echo "Examples:"
+    echo "machineset.sh                 # create Windows Server 2022 MachineSet yaml"
+    echo "machineset.sh -a              # create Windows Server 2022 MachineSet yaml and apply changes to k8s resource"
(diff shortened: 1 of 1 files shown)


saifshaikh48:
/hold

should go in after https://github.com/openshift/windows-machine-config-operator/pull/892

hack/machineset.sh
```diff
@@ -184,8 +223,22 @@ get_vsphere_ms() {
...
+  case "${WIN_SRV,,}" in
+    2022)
+      template="windows-golden-images/windows-server-2022-template"
+    ;;
+    2019)
+      echo "Note: Windows Server ${WIN_SRV} is not supported for platform ${provider}"
+      template="windows-golden-images/vm-winsrv-1909-golden-image"
+    ;;
+    20h2)
+      template="jvaldes/windows-server-20h2-template"
```
jrvaldes, line 236: templates should be in the folder `windows-golden-images`, otherwise the pruner will delete them

hack/machineset.sh
```diff
@@ -6,7 +6,8 @@
 # USAGE
 #    machineset.sh
 # OPTIONS
-#    $1      Action       (Optional) apply/delete the MachineSet
+#    $1    WIN_SRV   (Optional) Windows Server version of VMs associated with the resulting MachineSet. Defaults to 2022
+#    $2    Action    (Optional) apply/delete the MachineSet
```
jrvaldes, line 10: This is a breaking change.
Given Windows Server option defaults to 2022, consider keeping the action as the first option ($1), and just extend with the Windows Server option.

hack/machineset.sh
```diff
@@ -78,10 +80,28 @@ get_aws_ms() {
...
+    ;;
+    2019)
+      image_name="Windows_Server-2019*English*Full*Containers"
+    ;;
+    20h2)
+      echo "Note: Windows Server ${WIN_SRV} is not supported for platform ${provider}"
+      image_name="Windows_Server-20H2*English*Core*Containers"
+    ;;
+    *)
+      error-exit "Windows Server version (1st parameter) must be \"2022\", \"2019\", or \"20h2\" (case insensitive)."
```
jrvaldes, line 97:
```suggestion
      error-exit "Unsupported Windows Server version: ${WIN_SRV}"
```

hack/machineset.sh
```diff
@@ -135,7 +155,26 @@ get_azure_ms() {
...
+    ;;
+    2019)
+      sku="2019-Datacenter-with-Containers"
+    ;;
+    20h2)
+      echo "Note: Windows Server ${WIN_SRV} is not supported for platform ${provider}"
+      sku="datacenter-core-20h2-with-containers-smalldisk"
+    ;;
+    *)
+      error-exit "Windows Server version (1st parameter) must be \"2022\", \"2019\", or \"20h2\" (case insensitive)."
```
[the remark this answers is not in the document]
saifshaikh48: I don't see the suggestion as super beneficial. The function would have to re-check the cloud provider which is already known at these call sites, so it would be bunch of unnecessary nested clauses. Modularization just to avoid the repeating the error message does more harm than good.

hack/machineset.sh
saifshaikh48: That wouldn't quite work since `Action` is optional. Reworked it in a breaking but hopefully better manner :slightly_smiling_face:

hack/machineset.sh
saifshaikh48: The only 20h2 template we have is one you created under `jvaldes` folder. Given 20H2 going out of support in May, I'd rather not get too hung up on this as it's only for dev testing.

sebsoto: If there is no `permanent` template available, I suggest we just drop 20h2 from this PR, considering the focus on 2022 moving forward

saifshaikh48:
/close

closing while we figure out aws/azure support of Windows Server 2022

PR #936, opened 2022-02-25
Master submodule update 02 25

mansikulkarni96:
submodule update using:
`./hack/update_submodules.sh`


mansikulkarni96:
aws-e2e-upgrade test is consistently failing when trying to install the operator using the upgrade workflow with:
`ClusterServiceVersion was never created `

PR #942, opened 2022-02-28
Stop requiring kube-proxy ERROR and WARNING logs

sebsoto:
This PR drops the requirement of there being .ERROR and .WARNING
logs for kube-proxy. These are only present if there was logging at
those log levels


mansikulkarni96:
Thanks for taking care of this @sebsoto 
/lgtm

PR #944, opened 2022-03-01
WINC-714: Introduce no-op WICD

alinaryan:
This PR handles the 'framework' creation of Windows Instance Config Daemon(WICD) by creating a no-op WICD binary and adding a new Makefile target that builds windows-instance-config-daemon.exe using GOOS=windows.


config/rbac/role.yaml
```diff
@@ -130,6 +130,14 @@ rules:
   - signers
```
sebsoto, line 130: this change is unrelated, and should be dropped from this PR

build/build.sh
```diff
@@ -31,6 +31,7 @@ if [[ -z "$OUTPUT_DIR" ]] || [[ -z "$WMCO_SEMVER" ]] ; then
 fi
 
 PACKAGE="github.com/openshift/windows-machine-config-operator"
```
sebsoto, line 33: this is no longer being used and can be removed

build/build.sh
```diff
@@ -31,6 +31,7 @@ if [[ -z "$OUTPUT_DIR" ]] || [[ -z "$WMCO_SEMVER" ]] ; then
...
+MAIN_PACKAGE="${PACKAGE}/cmd/operator"
```
sebsoto, line 34: MAIN_PACKAGE is a bit of a misnomer, this can be renamed to something like WMCO_CMD_DIR

Makefile
```diff
@@ -92,6 +92,9 @@ build: fmt vet
 run: manifests generate fmt vet ## Run a controller from your host.
 	go run cmd/operator/main.go
 
+daemon: fmt vet
+	go run cmd/daemon/main.go
+
```
sebsoto, line 97: WICD is meant to be a windows exe, instead of having the endpoint use `go run`, please make this compile an executable called `windows-instance-config-daemon.exe`, you will need to cross compile for Windows by setting `GOOS`. the exe should be saved to the same directory that the WMCO binary is outputted to.

mansikulkarni96: You can use how we [build](https://github.com/openshift/windows-machine-config-bootstrapper/blob/38466b767564bddeeb68a6a9fab708dd73a937ce/Makefile#L8) the wmcb.exe as a reference.

alinaryan: Thanks! Fixed

cmd/daemon/main.go
```diff
@@ -0,0 +1,9 @@
+package main
```
selansen, line 1: should we add some file header info like below? I see it in many files but this is not followed in all files. Since this is new file, should we consider it?
``
/*
Copyright 2021.
Licensed under the Apache License, Version 2.0 (the "License");
you may not use this file except in compliance with the License.
You may obtain a copy of the License at
    http://www.apache.org/licenses/LICENSE-2.0
.....
*/

selansen:
@aravindhp Could you pls quickly review and let us know your comments? As this is going to be our base PR to create a new WICD module, your input would greatly help

Makefile
```diff
@@ -89,6 +90,10 @@ endif
 build: fmt vet
```
aravindhp, line 90: Shouldn't doing `make build` also result in building WICD? Otherwise you will need to add another explicit entry to build this in the release repo. Adding `build-wicd` will achieve that here which will also give you a CI signal as to whether it compiles or not.

cmd/daemon/main.go
```diff
@@ -0,0 +1,9 @@
...
+
+import (
+	"fmt"
+)
+
+func main() {
+	fmt.Printf("WICD placeholder...")
```
aravindhp, line 8: While this is fine to do for a start, we need to get the logging story right for WICD the 2nd time around. So please think through what logging mechanism will be used. 

@sebsoto if you have some particular logger in mind, I suggest to start using that from the beginning.

Makefile
```diff
@@ -89,6 +90,10 @@ endif
...
 	build/build.sh ${OUTPUT_DIR} ${WMCO_VERSION} ${GO_MOD_FLAGS}
 
+.PHONY: build-wicd
+build-wicd:
```
aravindhp, line 94: `build-wicd` or `build-daemon`? 😅

aravindhp (changes requested):
Thanks for working on this, @alinaryan. Please address my comments.

The commit `[wicd] add WICD main.go placeholder` should just be `[wicd] add main.go placeholder`. This leads me to if this should be referred to as `daemon` or `wicd`? We don't want a repeat of WMCB naming. 

The same needs to be called for the PR header itself. `[wmco] no-op WICD` is a bit strange as WICD and WMCO are independent. They just happen to live in the same repository. This PR should be something like `Introduce no-op WICD`

Makefile
alinaryan: Good point. I will build the daemon under `build` instead of creating a new target.

cmd/daemon/main.go
alinaryan: Added

Makefile
alinaryan: Added

aravindhp: > Good point. I will build the daemon under `make build` instead of creating a new target.

It is fine to have a separate target. That has the added flexibility of being able to compile just the daemon. In fact we might want to add a separate target to build the operator too.

cmd/daemon/main.go
sebsoto: I would suggest we use klog in order to maintain uniformity with the other on-node programs

Makefile
saifshaikh48: +1 to separate make target

cmd/daemon/main.go
alinaryan: @sebsoto should I start using that in this PR?

Makefile
sebsoto: I dont see a reason to have an entry in the release repo for this

cmd/daemon/main.go
sebsoto: Personally I don't think that's necessary as the ticket only requires a no-op WICD and nothing else

Makefile
sebsoto: daemon is preferred for me

selansen: I prefer daemon too :)

Makefile
selansen: why do we need to have an explicit entry in the release repo? Lets keep release repo change as minimal as possible.

cmd/daemon/main.go
selansen: because this is simple PR, I suggest adding some basic logs like "process running or something" ( in that way we know log files getting created and stored in the right place). A simple one to knock it along with this PR.

alinaryan: Added

Makefile
```diff
@@ -77,6 +77,7 @@ vet: ## Run go vet against code.
 ##@ Build
 
 OUTPUT_DIR="build/_output"
+DAEMON_DIR="cmd/daemon/main.go"
```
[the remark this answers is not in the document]
alinaryan: Thanks, I got rid of the var.

Makefile
alinaryan: I don't think it is needed either
Screenshot from 2022-03-08 10-32-24
. When running `make build-daemon` the binary is saved to the _output/bin directory.

cmd/daemon/main.go
```diff
@@ -0,0 +1,34 @@
...
+
+	"k8s.io/klog/v2"
+)
+
+func main() {
+	klog.InitFlags(nil)
+	// By default klog writes to stderr. Setting logtostderr to false makes klog
+	// write to a log file.
+	flag.Set("logtostderr", "false")
+	flag.Set("log_file", "daemonFile.log")
```
sebsoto, line 30: Lets not overwrite these flags at the moment, the defaults are fine for this purpose.
WMCO will be setting the flags when it starts WICD

sebsoto (changes requested):

alinaryan:
/test vsphere-e2e-operator
Networking failure

alinaryan:
/test vsphere-e2e-operator
Destroy cluster failure

PR #979, opened 2022-03-10
[vsphere_ci] Use valid hostname for PTR records

jrvaldes:
Use a valid hostname as the PTR record value, so that WMCO's CSR approver
success with the reverse IP lookup.

Follow-up to https://github.com/openshift/windows-machine-config-operator/commit/79434fe590bd44500a15d07fccb11864c7b4d802


selansen:
Checked the cherry pick list and looks good. 
/lgtm

sebsoto (approved):

PR #1002, opened 2022-03-16
WINC-715: WMCO creates and maintains Windows services ConfigMap

saifshaikh48:
This PR creates the data representation of a service ConfigMap which provides WICD with the specifications for each Windows service that must be created on a Windows instance.
WMCO will then create and maintain a service ConfigMap which provides WICD with the specifications for each Windows service that must be created on a Windows instance. To begin with maintain a no-op ConfigMap, and in follow-up PRs we will move services one by one to be included in the ConfigMap.

- Services ConfigMap is a singleton within the namespace
- The ConfigMap is immutable
- If the ConfigMap is deleted it is recreated
- If a ConfigMap with incorrect values is found, WMCO will delete and recreate it with the proper values
  - this applies both on operator start and during reconciling
- The Services ConfigMap will be created on operator startup
- Services ConfigMap follows the naming pattern `windows-services-<WMCOMajorVersion>-<WMCOMinorVersion>-<WMCOPatchVersion>-<Commit>`

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -3,12 +3,14 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-# returns the operator version in `Version+GitHash` format
+# returns the operator version in `Version-GitHash` format
 # Takes a semver as an argument. This should be the version present in the WMCO CSV
 add_git_data_to_version() {
   local WMCO_SEMVER=$1
   GIT_COMMIT=$(git rev-parse --short HEAD)
-  FULL_VERSION="${WMCO_SEMVER}+${GIT_COMMIT}"
+  # The commit hash retreived varies in length, so we will standardize to length received from the operator image
+  GIT_COMMIT="${GIT_COMMIT:0:7}"
+  FULL_VERSION="${WMCO_SEMVER}-${GIT_COMMIT}"
 
   # If any files that affect the building of the operator binary have been changed, append "-dirty" to the version.
   if [ -n "$(git status version tools.go go.mod go.sum vendor Makefile build main.go hack pkg controllers --porcelain)" ]; then
--- a/controllers/configmap_controller.go
+++ b/controllers/configmap_controller.go
@@ -44,6 +44,7 @@ import (
 	"github.com/openshift/windows-machine-config-operator/pkg/metrics"
 	"github.com/openshift/windows-machine-config-operator/pkg/nodeconfig"
 	"github.com/openshift/windows-machine-config-operator/pkg/secrets"
+	"github.com/openshift/windows-machine-config-operator/pkg/servicescm"
 	"github.com/openshift/windows-machine-config-operator/pkg/signer"
 	"github.com/openshift/windows-machine-config-operator/pkg/wiparser"
 )
@@ -121,19 +122,42 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context,
 	}
 
 	// Fetch the ConfigMap. The predicate will have filtered out any ConfigMaps that we should not reconcile
-	// so it is safe to assume that all ConfigMaps being reconciled describe hosts that need to be present in the
-	// cluster. This also handles the case when the reconciliation is kicked off by the InstanceConfigMap being deleted.
-	// In the deletion case, an empty InstanceConfigMap will be reconciled now resulting in all existing BYOH nodes
-	// being deleted.
+	// so it is safe to assume that all ConfigMaps being reconciled are one of:
+	// 1. windows-instances, describing hosts that need to be present in the cluster.
+	// 2. windows-services, describing expected configuration of WMCO-managed services on all Windows instances
 	configMap := &core.ConfigMap{}
 	if err := r.client.Get(ctx, req.NamespacedName, configMap); err != nil {
 		if !k8sapierrors.IsNotFound(err) {
 			// Error reading the object - requeue the request.
 			return ctrl.Result{}, err
 		}
+		if req.NamespacedName.Name == servicescm.Name {
+			// Create and retrieve the services ConfigMap as it is not present
+			if configMap, err = r.createServicesConfigMap(ctx); err != nil {
+				return ctrl.Result{}, err
+			}
+		}
 	}
 
-	return ctrl.Result{}, r.reconcileNodes(ctx, configMap)
+	r.log.V(1).Info("Reconciling", "ConfigMap", req.NamespacedName)
+	// At this point configMap will be set properly
+	switch req.NamespacedName.Name {
+	case servicescm.Name:
+		return ctrl.Result{}, r.reconcileServices(ctx, configMap)
+	case wiparser.InstanceConfigMap:
+		return ctrl.Result{}, r.reconcileNodes(ctx, configMap)
+	default:
+		// Unexpected configmap, log and return no error so we don't requeue
+		r.log.Error(errors.New("Unexpected resource triggered reconcile"), "ConfigMap", req.NamespacedName)
+	}
+	return ctrl.Result{}, nil
+}
+
+// reconcileServices uses the data within the services ConfigMap to ensure WMCO-managed Windows services on
+// Windows Nodes have the expected configuration and are in the expected state
+func (r *ConfigMapReconciler) reconcileServices(ctx context.Context, windowsServices *core.ConfigMap) error {
+	// TODO: actually react to changes to the services ConfigMap
+	return nil
 }
 
 // reconcileNodes corrects the discrepancy between the "expected" instances, and the "actual" Node list
@@ -288,5 +312,20 @@ func (r *ConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 // isValidConfigMap returns true if the ConfigMap object is the InstanceConfigMap
 func (r *ConfigMapReconciler) isValidConfigMap(o client.Object) bool {
-	return o.GetNamespace() == r.watchNamespace && o.GetName() == wiparser.InstanceConfigMap
+	return o.GetNamespace() == r.watchNamespace &&
+		(o.GetName() == wiparser.InstanceConfigMap || o.GetName() == servicescm.Name)
+}
+
+// createServicesConfigMap creates a valid ServicesConfigMap and returns it
+func (r *ConfigMapReconciler) createServicesConfigMap(ctx context.Context) (*core.ConfigMap, error) {
+	windowsServices, err := servicescm.Generate(servicescm.Name, r.watchNamespace)
+	if err != nil {
+		return nil, err
+	}
+
+	if err = r.client.Create(ctx, windowsServices); err != nil {
+		return nil, err
+	}
+	r.log.Info("Created", "ConfigMap", kubeTypes.NamespacedName{Namespace: r.watchNamespace, Name: servicescm.Name})
+	return windowsServices, nil
 }
--- a/controllers/configmap_controller.go
+++ b/controllers/configmap_controller.go
@@ -339,3 +339,45 @@ func (r *ConfigMapReconciler) createServicesConfigMap(ctx context.Context) (*cor
 	r.log.Info("Created", "ConfigMap", kubeTypes.NamespacedName{Namespace: r.watchNamespace, Name: servicescm.Name})
 	return windowsServices, nil
 }
+
+// createServicesConfigMapOnBootup creates a valid ServicesConfigMap
+// ConfigMapReconciler.createServicesConfigMap() cannot be used in its stead as the cache has not been
+// populated yet, which is why the typed client is used here as it calls the API server directly.
+func (r *ConfigMapReconciler) createServicesConfigMapOnBootup() error {
+	windowsServices, err := servicescm.Generate(servicescm.Name, r.watchNamespace)
+	if err != nil {
+		return err
+	}
+	cm, err := r.k8sclientset.CoreV1().ConfigMaps(r.watchNamespace).Create(context.TODO(), windowsServices,
+		meta.CreateOptions{})
+	if err != nil {
+		return err
+	}
+	r.log.Info("Created", "ConfigMap", kubeTypes.NamespacedName{Namespace: cm.Namespace, Name: cm.Name})
+	return nil
+}
+
+// EnsureServicesConfigMapExists ensures that the ServicesConfigMap is present and valid on operator bootup
+func (r *ConfigMapReconciler) EnsureServicesConfigMapExists() error {
+	windowsServices, err := r.k8sclientset.CoreV1().ConfigMaps(r.watchNamespace).Get(context.TODO(), servicescm.Name,
+		meta.GetOptions{})
+	if err != nil {
+		if k8sapierrors.IsNotFound(err) {
+			// If ConfigMap is not found, create it and return
+			return r.createServicesConfigMapOnBootup()
+		}
+		return err
+	}
+
+	// If a ConfigMap with incorrect values is found, WMCO will delete and recreate it with the proper values
+	if _, err := servicescm.Parse(windowsServices.Data); err != nil {
+		if err = r.k8sclientset.CoreV1().ConfigMaps(r.watchNamespace).Delete(context.TODO(), windowsServices.Name,
+			meta.DeleteOptions{}); err != nil {
+			return err
+		}
+		r.log.Info("Deleted invalid resource", "ConfigMap",
+			kubeTypes.NamespacedName{Namespace: r.watchNamespace, Name: servicescm.Name}, "Error", err.Error())
+		return r.createServicesConfigMapOnBootup()
+	}
+	return nil
+}
--- a/controllers/configmap_controller.go
+++ b/controllers/configmap_controller.go
@@ -156,6 +156,16 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context,
 // reconcileServices uses the data within the services ConfigMap to ensure WMCO-managed Windows services on
 // Windows Nodes have the expected configuration and are in the expected state
 func (r *ConfigMapReconciler) reconcileServices(ctx context.Context, windowsServices *core.ConfigMap) error {
+	// If a ConfigMap with invalid values is found, WMCO will delete and recreate it with proper values
+	if _, err := servicescm.Parse(windowsServices.Data); err != nil {
+		// Deleting will trigger an event for the configmap_controller, which will re-create a proper ConfigMap
+		if err = r.client.Delete(ctx, windowsServices); err != nil {
+			return err
+		}
+		r.log.Info("Deleted invalid resource", "ConfigMap",
+			kubeTypes.NamespacedName{Namespace: r.watchNamespace, Name: windowsServices.Name}, "Error", err.Error())
+		return nil
+	}
 	// TODO: actually react to changes to the services ConfigMap
 	return nil
 }
--- a/pkg/servicescm/servicescm.go
+++ b/pkg/servicescm/servicescm.go
@@ -1,0 +1,24 @@
+package servicescm
+
+import (
+	"fmt"
+
+	"github.com/openshift/windows-machine-config-operator/version"
+)
+
+// NamePrefix is the prefix of all Windows services ConfigMap names
+const NamePrefix = "windows-services-"
+
+// Name is the full name of the Windows services ConfigMap, detailing the service config for a specific WMCO version
+var Name string
+
+// init runs once, initializing global variables
+func init() {
+	Name = getName()
+}
+
+// getName returns the name of the ConfigMap, using the following naming convention:
+// windows-services-<WMCOFullVersion>
+func getName() string {
+	return fmt.Sprintf("%s%s", NamePrefix, version.Get())
+}
--- a/pkg/servicescm/servicescm.go
+++ b/pkg/servicescm/servicescm.go
@@ -17,6 +17,52 @@ func init() {
 	Name = getName()
 }
 
+// NodeCmdArg describes a Windows command variable and how its value can be populated
+type NodeCmdArg struct {
+	// Name is the variable name as it appears in commands
+	Name string `json:"name"`
+	// NodeObjectJsonPath is the JSON path of a field within an instance's Node object.
+	// The value of this field is the value of the variable
+	NodeObjectJsonPath string `json:"nodeObjectJsonPath"`
+}
+
+// PowershellCmdArg describes a PowerShell variable and how its value can be populated
+type PowershellCmdArg struct {
+	// Name is the variable name as it appears in commands
+	Name string `json:"name"`
+	// Path is the location of a PowerShell script whose output is the value of the variable
+	Path string `json:"path"`
+}
+
+// Service represents the configuration spec of a Windows service
+type Service struct {
+	// Name is the name of the Windows service
+	Name string `json:"name"`
+	// Command is the command that will launch the Windows service. This could potentially include strings whose values
+	// will be derived from NodeVariablesInCommand and PowershellVariablesInCommand.
+	// Before the command is run on an instance, all node and PowerShell variables will be replaced by their values
+	Command string `json:"path"`
+	// NodeVariablesInCommand holds all variables in the service command whose values are sourced from a node object
+	NodeVariablesInCommand []NodeCmdArg `json:"nodeVariablesInCommand,omitempty"`
+	// PowershellVariablesInCommand holds all variables in the command whose values are sourced from a PowerShell script
+	PowershellVariablesInCommand []PowershellCmdArg `json:"powershellVariablesInCommand,omitempty"`
+	// Dependencies is a list of service names that this service is dependent on
+	Dependencies []string `json:"dependencies,omitempty"`
+	// Bootstrap is a boolean flag indicating whether this service should be handled as part of node bootstrapping
+	Bootstrap bool `json:"bootstrap"`
+	// Priority is a non-negative integer that will be used to order the creation of the services.
+	// Priority 0 is created first
+	Priority uint `json:"priority"`
+}
+
+// FileInfo contains the path and checksum of a file copied to an instance by WMCO
+type FileInfo struct {
+	// Path is the filepath of a file on an instance
+	Path string `json:"path"`
+	// Checksum is the checksum of the file specified at Path. It is used to validate that a file has not been changed
+	Checksum string `json:"checksum"`
+}
+
 // getName returns the name of the ConfigMap, using the following naming convention:
 // windows-services-<WMCOFullVersion>
 func getName() string {
--- a/pkg/servicescm/servicescm.go
+++ b/pkg/servicescm/servicescm.go
@@ -1,16 +1,33 @@
 package servicescm
 
 import (
+	"encoding/json"
 	"fmt"
+
+	core "k8s.io/api/core/v1"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/openshift/windows-machine-config-operator/version"
 )
 
-// NamePrefix is the prefix of all Windows services ConfigMap names
-const NamePrefix = "windows-services-"
+const (
+	// NamePrefix is the prefix of all Windows services ConfigMap names
+	NamePrefix = "windows-services-"
+	// servicesKey is a required key in the services ConfigMap. The value for this key is a Service object JSON array.
+	servicesKey = "services"
+	// filesKey is a required key in the services ConfigMap. The value for this key is a FileInfo object JSON array.
+	filesKey = "files"
+)
 
-// Name is the full name of the Windows services ConfigMap, detailing the service config for a specific WMCO version
-var Name string
+var (
+	// Name is the full name of the Windows services ConfigMap, detailing the service config for a specific WMCO version
+	Name string
+	//TODO: Fill in required content as services and files are added to the ConfigMap definition
+	// requiredServices is the source of truth for expected service configuration on a Windows Node
+	requiredServices = &[]Service{}
+	// requiredFiles is the source of truth for files that are expected to exist on a Windows Node
+	requiredFiles = &[]FileInfo{}
+)
 
 // init runs once, initializing global variables
 func init() {
@@ -61,6 +78,51 @@ type FileInfo struct {
 	Path string `json:"path"`
 	// Checksum is the checksum of the file specified at Path. It is used to validate that a file has not been changed
 	Checksum string `json:"checksum"`
+}
+
+// data represents the Data field of a `windows-services` ConfigMap resource, which is all the required information to
+// configure a Windows instance as a Node
+type data struct {
+	// Services contains information required to start all required Windows services with proper arguments and order
+	Services []Service `json:"services"`
+	// Files contains the path and checksum of all the files copied to a Windows VM by WMCO
+	Files []FileInfo `json:"files"`
+}
+
+// newData returns a new 'data' object with the given services and files.
+func newData(services *[]Service, files *[]FileInfo) *data {
+	cmData := &data{*services, *files}
+	return cmData
+}
+
+// Generate creates an immutable service ConfigMap which provides WICD with the specifications
+// for each Windows service that must be created on a Windows instance.
+func Generate(name, namespace string) (*core.ConfigMap, error) {
+	immutable := true
+	servicesConfigMap := &core.ConfigMap{
+		ObjectMeta: meta.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		},
+		Immutable: &immutable,
+		Data:      make(map[string]string),
+	}
+
+	cmData := newData(requiredServices, requiredFiles)
+
+	jsonServices, err := json.Marshal(cmData.Services)
+	if err != nil {
+		return nil, err
+	}
+	servicesConfigMap.Data[servicesKey] = string(jsonServices)
+
+	jsonFiles, err := json.Marshal(cmData.Files)
+	if err != nil {
+		return nil, err
+	}
+	servicesConfigMap.Data[filesKey] = string(jsonFiles)
+
+	return servicesConfigMap, nil
 }
 
 // getName returns the name of the ConfigMap, using the following naming convention:
--- a/pkg/servicescm/servicescm.go
+++ b/pkg/servicescm/servicescm.go
@@ -3,11 +3,14 @@ package servicescm
 import (
 	"encoding/json"
 	"fmt"
+	"reflect"
+	"sort"
 
 	core "k8s.io/api/core/v1"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/openshift/windows-machine-config-operator/version"
+	"github.com/pkg/errors"
 )
 
 const (
@@ -89,10 +92,13 @@ type data struct {
 	Files []FileInfo `json:"files"`
 }
 
-// newData returns a new 'data' object with the given services and files.
-func newData(services *[]Service, files *[]FileInfo) *data {
+// newData returns a new 'data' object with the given services and files. Validates given object contents on creation.
+func newData(services *[]Service, files *[]FileInfo) (*data, error) {
 	cmData := &data{*services, *files}
-	return cmData
+	if err := cmData.validate(); err != nil {
+		return nil, errors.Wrap(err, "unable to create services ConfigMap data object")
+	}
+	return cmData, nil
 }
 
 // Generate creates an immutable service ConfigMap which provides WICD with the specifications
@@ -108,7 +114,10 @@ func Generate(name, namespace string) (*core.ConfigMap, error) {
 		Data:      make(map[string]string),
 	}
 
-	cmData := newData(requiredServices, requiredFiles)
+	cmData, err := newData(requiredServices, requiredFiles)
+	if err != nil {
+		return nil, err
+	}
 
 	jsonServices, err := json.Marshal(cmData.Services)
 	if err != nil {
@@ -123,6 +132,152 @@ func Generate(name, namespace string) (*core.ConfigMap, error) {
 	servicesConfigMap.Data[filesKey] = string(jsonFiles)
 
 	return servicesConfigMap, nil
+}
+
+// Parse converts ConfigMap data into the objects representing a Windows services ConfigMap schema
+// Returns error if the given data is invalid in structure or contents.
+func Parse(dataFromCM map[string]string) (*data, error) {
+	if len(dataFromCM) != 2 {
+		return nil, errors.New("services ConfigMap should have exactly 2 keys")
+	}
+
+	value, ok := dataFromCM[servicesKey]
+	if !ok {
+		return nil, errors.Errorf("expected key %s does not exist", servicesKey)
+	}
+	services := &[]Service{}
+	if err := json.Unmarshal([]byte(value), services); err != nil {
+		return nil, err
+	}
+
+	value, ok = dataFromCM[filesKey]
+	if !ok {
+		return nil, errors.Errorf("expected key %s does not exist", filesKey)
+	}
+	files := &[]FileInfo{}
+	if err := json.Unmarshal([]byte(value), files); err != nil {
+		return nil, err
+	}
+
+	return newData(services, files)
+}
+
+// validate ensures the given object represents a valid services ConfigMap, made up of only required services and files.
+// The validation also ensures bootstrap services are defined to always start before controller services.
+func (cmData *data) validate() error {
+	if err := validateRequiredContent(cmData.Services, cmData.Files); err != nil {
+		return err
+	}
+	if err := validateDependencies(cmData.Services); err != nil {
+		return err
+	}
+	return validatePriorities(cmData.Services)
+}
+
+// validateRequiredContent ensures that the given slices are comprised of all the required services/files and only these
+func validateRequiredContent(services []Service, files []FileInfo) error {
+	// Validate services
+	if len(services) != len(*requiredServices) {
+		return errors.New("Unexpected number of services")
+	}
+	for _, requiredSvc := range *requiredServices {
+		if !requiredSvc.isPresentAndCorrect(services) {
+			return errors.Errorf("Required service %s is not present with expected configuration", requiredSvc.Name)
+		}
+	}
+	// Validate files
+	if len(files) != len(*requiredFiles) {
+		return errors.New("Unexpected number of files")
+	}
+	for _, requiredFile := range *requiredFiles {
+		if !requiredFile.isPresentAndCorrect(files) {
+			return errors.Errorf("Required file %s is not present as expected", requiredFile.Path)
+		}
+	}
+	return nil
+}
+
+// isPresentAndCorrect checks if the required service exists as expected within the given services slice
+func (s Service) isPresentAndCorrect(services []Service) bool {
+	for _, service := range services {
+		if reflect.DeepEqual(s, service) {
+			return true
+		}
+	}
+	return false
+}
+
+// isPresentAndCorrect checks if the required file exists as expected within the given files slice
+// TODO: When we move to go1.18, consolodate this helper with the above Service.isPresentAndCorrect using generics
+func (f FileInfo) isPresentAndCorrect(files []FileInfo) bool {
+	for _, file := range files {
+		if reflect.DeepEqual(f, file) {
+			return true
+		}
+	}
+	return false
+}
+
+// validateDependencies ensures that no bootstrap service depends on a non-bootstrap service
+func validateDependencies(services []Service) error {
+	bootstrapServices := []Service{}
+	nonBootstrapServices := []Service{}
+	for _, svc := range services {
+		if svc.Bootstrap {
+			bootstrapServices = append(bootstrapServices, svc)
+		} else {
+			nonBootstrapServices = append(nonBootstrapServices, svc)
+		}
+	}
+
+	for _, bootstrapSvc := range bootstrapServices {
+		if bootstrapSvc.hasDependency(nonBootstrapServices) {
+			return errors.Errorf("bootstrap service %s cannot depend on non-bootstrap service", bootstrapSvc.Name)
+		}
+	}
+	return nil
+}
+
+// hasDependency checks if a service is dependent on any services in the given slice
+func (s *Service) hasDependency(possibleDependencies []Service) bool {
+	for _, dependency := range s.Dependencies {
+		for _, possibleDependency := range possibleDependencies {
+			if dependency == possibleDependency.Name {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// validatePriorities ensures that each service that has the bootstrap flag set as true has a higher priority than all
+// non-bootstrap services. There should be no overlap in the priorities of bootstrap services and controller services.
+func validatePriorities(services []Service) error {
+	// sort services in ascending priority, bootstrap services towards the front of slice
+	sort.Slice(services, func(i, j int) bool {
+		return services[i].Priority < services[j].Priority
+	})
+
+	// ensure no bootstrap service appears after a controller service in the ordered list
+	nonBootstrapSeen := false
+	lastBootstrapPriority := uint(0)
+	for _, svc := range services {
+		if svc.Bootstrap {
+			if nonBootstrapSeen {
+				return errors.Errorf("bootstrap service %s priority must be higher than all controller services",
+					svc.Name)
+			}
+			lastBootstrapPriority = svc.Priority
+		} else {
+			// corner case if two adjacent bootstrap and controller services have the same priority
+			if svc.Priority == lastBootstrapPriority {
+				return errors.Errorf("controller service %s priority must not overlap with any bootstrap service",
+					svc.Name)
+			}
+			nonBootstrapSeen = true
+		}
+	}
+	return nil
 }
 
 // getName returns the name of the ConfigMap, using the following naming convention:
--- a/pkg/servicescm/servicescm_test.go
+++ b/pkg/servicescm/servicescm_test.go
@@ -1,0 +1,472 @@
+package servicescm
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestParse(t *testing.T) {
+	// The following cases test only the structure of the services ConfigMap.
+	// Content validtion is tested below through test methods upon the helpers.
+	testCases := []struct {
+		name        string
+		input       map[string]string
+		expectedErr bool
+	}{
+		{
+			name: "both expected keys",
+			input: map[string]string{
+				servicesKey: "[]",
+				filesKey:    "[]",
+			},
+			expectedErr: false,
+		},
+		{
+			name:        "no keys",
+			input:       map[string]string{},
+			expectedErr: true,
+		},
+		{
+			name: "only 1 of the expected keys",
+			input: map[string]string{
+				servicesKey: "[]",
+			},
+			expectedErr: true,
+		},
+		{
+			name: "correct number but incorrect key",
+			input: map[string]string{
+				filesKey:  "[]",
+				"testKey": "[]",
+			},
+			expectedErr: true,
+		},
+		{
+			name: "too many keys",
+			input: map[string]string{
+				servicesKey: "[]",
+				filesKey:    "[]",
+				"testKey":   "[]",
+			},
+			expectedErr: true,
+		},
+	}
+
+	for _, test := range testCases {
+		t.Run(test.name, func(t *testing.T) {
+			cmData, err := Parse(test.input)
+			if test.expectedErr {
+				assert.Error(t, err)
+				return
+			}
+			require.NoError(t, err)
+			assert.NotNil(t, cmData.Services)
+			assert.NotNil(t, cmData.Files)
+		})
+	}
+}
+
+func TestGenerate(t *testing.T) {
+	// Ensure that the ConfigMap we generate internally as the source of truth passes our own validation functions
+	configMap, err := Generate(Name, "testNamespace")
+	require.NoError(t, err)
+
+	_, err = Parse(configMap.Data)
+	require.NoError(t, err)
+}
+
+func TestValidateDependencies(t *testing.T) {
+	testCases := []struct {
+		name        string
+		input       []Service
+		expectedErr bool
+	}{
+		{
+			name:        "empty services list",
+			input:       []Service{},
+			expectedErr: false,
+		},
+		{
+			name: "no dependencies",
+			input: []Service{
+				{
+					Name:    "new-bootstrap-service",
+					Command: "C:\new-service --variable-arg1=NODE_NAME --variable-arg2=NETWORK_IP",
+					NodeVariablesInCommand: []NodeCmdArg{
+						{
+							Name:               "NODE_NAME",
+							NodeObjectJsonPath: "metadata.name",
+						},
+					},
+					PowershellVariablesInCommand: []PowershellCmdArg{
+						{
+							Name: "NETWORK_IP",
+							Path: "C:\\k\\scripts\\get_net_ip.ps",
+						},
+					},
+					Dependencies: []string{},
+					Bootstrap:    true,
+					Priority:     0,
+				},
+				{
+					Name:         "test-controller-service",
+					Command:      "C:\\test-controller-service",
+					Dependencies: []string{},
+					Bootstrap:    false,
+					Priority:     1,
+				},
+			},
+			expectedErr: false,
+		},
+		{
+			name: "valid dependencies",
+			input: []Service{
+				{
+					Name:    "new-bootstrap-service",
+					Command: "C:\\new-service --variable-arg1=NODE_NAME --variable-arg2=NETWORK_IP",
+					NodeVariablesInCommand: []NodeCmdArg{
+						{
+							Name:               "NODE_NAME",
+							NodeObjectJsonPath: "metadata.name",
+						},
+					},
+					PowershellVariablesInCommand: []PowershellCmdArg{
+						{
+							Name: "NETWORK_IP",
+							Path: "C:\\k\\scripts\\get_net_ip.ps",
+						},
+					},
+					Dependencies: []string{},
+					Bootstrap:    true,
+					Priority:     0,
+				},
+				{
+					Name:         "test-controller-service",
+					Command:      "C:\\test-controller-service",
+					Dependencies: []string{"new-bootstrap-service"},
+					Bootstrap:    false,
+					Priority:     1,
+				},
+				{
+					Name:         "test-controller-service-2",
+					Command:      "C:\\test-controller-service-2",
+					Dependencies: []string{"test-controller-service-2"},
+					Bootstrap:    false,
+					Priority:     2,
+				},
+			},
+			expectedErr: false,
+		},
+		{
+			name: "bootstrap depends on all non-bootstrap services",
+			input: []Service{
+				{
+					Name:    "new-bootstrap-service",
+					Command: "C:\\new-service --variable-arg1=NODE_NAME --variable-arg2=NETWORK_IP",
+					NodeVariablesInCommand: []NodeCmdArg{
+						{
+							Name:               "NODE_NAME",
+							NodeObjectJsonPath: "metadata.name",
+						},
+					},
+					PowershellVariablesInCommand: []PowershellCmdArg{
+						{
+							Name: "NETWORK_IP",
+							Path: "C:\\k\\scripts\\get_net_ip.ps",
+						},
+					},
+					Dependencies: []string{"test-controller-service", "test-controller-service-2"},
+					Bootstrap:    true,
+					Priority:     0,
+				},
+				{
+					Name:         "test-controller-service",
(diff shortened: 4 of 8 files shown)


pkg/configmap/wsvalidator/wsvalidator.go
```diff
@@ -0,0 +1,23 @@
...
+// init runs once, initializing global variables
+func init() {
+	ServicesConfigMap = getServicesConfigMapName()
+}
+
+// getServicesConfigMapName returns the ConfigMap with the naming scheme:
+// windows-services-<MajorVersion>-<MinorVersion>-<PatchVersion>-<CommitHash>
+func getServicesConfigMapName() string {
+	// resource names cannot have `+` in them
+	sanitizedVersion := strings.ReplaceAll(version.Get(), "+", "-")
```
jrvaldes, line 188: Does it makes sense to use `-` instead of `+` while [building the full version](https://github.com/openshift/windows-machine-config-operator/blob/34b44fba96e4129c41db67f9ccd802000f63034d/build/build.sh#L11)?

We are using `-` anyway for the `-dirty` part.

pkg/configmap/wsvalidator/wsvalidator.go
```diff
@@ -0,0 +1,23 @@
+package wsvalidator
```
jrvaldes, line 1: _A parsing failure is also a validation error._

Consider `wsparser` instead of `wsvalidator` to follow the existing `wiparser` naming convention.

pkg/configmap/wsvalidator/wsvalidator.go
```diff
@@ -0,0 +1,23 @@
...
+func init() {
+	ServicesConfigMap = getServicesConfigMapName()
+}
+
+// getServicesConfigMapName returns the ConfigMap with the naming scheme:
+// windows-services-<MajorVersion>-<MinorVersion>-<PatchVersion>-<CommitHash>
+func getServicesConfigMapName() string {
+	// resource names cannot have `+` in them
+	sanitizedVersion := strings.ReplaceAll(version.Get(), "+", "-")
+	return fmt.Sprintf("windows-services-%s", sanitizedVersion)
```
jrvaldes, line 189: Consider introducing a constant for the configMap name.

```
const ServicesConfigMap = "windows-services"
```

test/e2e/create_test.go
```diff
@@ -15,19 +15,19 @@ import (
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
jrvaldes, line 18: s/v1/core

pkg/configmap/wsvalidator/wsvalidator.go
```diff
@@ -92,6 +94,93 @@ func GenerateWindowsServiceConfigMap(name, namespace string) *core.ConfigMap {
 	return cm
 }
 
+// ParseAndValidate converts ConfigMap data into the objects representing a Windows services ConfigMap and
+// ensures the data adheres to the expected schema
+func ParseAndValidate(data map[string]string) (*[]FileInfo, *[]Service, error) {
```
jrvaldes, line 99: Consider splitting this behavior in two functions:

- `func GetServices(data map[string]string) (*[]Service, error) {...}`
- `func GetFiles(data map[string]string) (*[]FileInfo, error) {...}`

test/e2e/validation_test.go
```diff
@@ -464,6 +465,64 @@ func testExpectedServicesRunning(t *testing.T) {
...
 }
 
+// testServicesConfigMap tests 2 aspects of expected functionality for the services ConfigMap
+// 1. It exists on operator startup and 2. It is recreated when deleted
+func testServicesConfigMap(t *testing.T) {
+	tc, err := NewTestContext()
+	require.NoError(t, err)
+
+	err = tc.validateWindowsServicesConfigMap()
+	require.NoError(t, err, "error validating windows-instances ConfigMap")
```
jrvaldes, line 475:
```suggestion
	require.NoError(t, err, "error validating windows-services ConfigMap")
```

test/e2e/validation_test.go
```diff
@@ -464,6 +465,64 @@ func testExpectedServicesRunning(t *testing.T) {
...
+// validateWindowsServicesConfigMap ensures the windows-services ConfigMap exists in the cluster and has valid contents
+func (tc *testContext) validateWindowsServicesConfigMap() error {
+	windowsServices, err := tc.client.K8s.CoreV1().ConfigMaps(tc.namespace).Get(context.TODO(),
+		wsvalidator.ServicesConfigMap, meta.GetOptions{})
+	if err != nil {
+		return errors.Wrapf(err, "error retrieving ConfigMap: %s", wsvalidator.ServicesConfigMap)
+	}
+
+	if _, _, err := wsvalidator.ParseAndValidate(windowsServices.Data); err != nil {
+		return errors.Wrapf(err, "invalid services ConfigMap data")
```
jrvaldes, line 490:
```suggestion
		return errors.Wrap(err, "invalid windows-services ConfigMap data")
```

jrvaldes (changes requested):
@saifshaikh48   Thanks for working on this, comments added. PTAL

pkg/configmap/wsvalidator/wsvalidator.go
saifshaikh48: Makes sense. I can make the change when building the version (assuming it doesn't break anything elsewhere, if so I think it would warrant separate PR)

controllers/configmap_controller.go
```diff
@@ -292,3 +302,23 @@ func (r *ConfigMapReconciler) isValidConfigMap(o client.Object) bool {
...
+	windowsServices := &core.ConfigMap{}
+	var err error
+	if err = r.client.Get(ctx, namespacedName, windowsServices); err != nil {
+		if k8sapierrors.IsNotFound(err) {
+			windowsServices = wsvalidator.GenerateWindowsServiceConfigMap(namespacedName.Name, namespacedName.Namespace)
+			if err = r.client.Create(ctx, windowsServices); err != nil {
+				return nil, err
+			}
+			r.log.Info("Created", "ConfigMap", namespacedName)
+			if err = r.client.Get(ctx, namespacedName, windowsServices); err != nil {
```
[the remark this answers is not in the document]
saifshaikh48: Yes, the returned CM is then validated (the next commit) and later will be used to pass needed info to WICD (coming stories).

pkg/configmap/wsvalidator/wsvalidator.go
```diff
@@ -14,6 +19,79 @@ func init() {
...
+const (
+	// servicesKey contains all data required to configure the required Windows services on any instance
+	// that is to be added to the cluster as a Node. The value for this key is a Service object JSON array.
+	servicesKey = "services"
+	// filesKey is a key within the services ConfigMap. The value for this key is a FileInfo object JSON array.
+	filesKey = "files"
+)
+
+// nodeVariable is a variable whose value is sourced from the node object associated with a Windows instance
+type NodeVariable struct {
```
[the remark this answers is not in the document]
saifshaikh48: Sure, I can use `NodeCmdArg` and `PowershellCmdArg`

pkg/configmap/wsvalidator/wsvalidator.go
saifshaikh48: I was thinking that returning the Service and FileInfo slices will be useful in the future, but I can now see the angle that this function is doing too much. 
I may split this into 3 functions: the two Jose mentioned so someone can get each component if needed, and then a top level ParseAndValidate which uses getServices() and getFiles() as helpers

controllers/configmap_controller.go
```diff
@@ -334,3 +336,50 @@ func (r *ConfigMapReconciler) ensureExists(ctx context.Context, namespacedName k
 	}
 	return windowsServices, err
 }
+
+// EnsureServicesConfigMapExists ensures that the ServicesConfigMap is present and valid on the cluster during operator
```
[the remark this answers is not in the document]
saifshaikh48: Good question. When I tested locally, the configmap reconciler wasn't responsive yet. I can do some more investigation though.

sebsoto (changes requested):

pkg/configmap/wsvalidator/wsvalidator.go
```diff
@@ -0,0 +1,23 @@
...
+
+	"github.com/openshift/windows-machine-config-operator/version"
+)
+
+// ServicesConfigMap is the name of the ConfigMap detailing service configuration for a specific WMCO version
+var ServicesConfigMap string
+
+// init runs once, initializing global variables
+func init() {
+	ServicesConfigMap = getServicesConfigMapName()
```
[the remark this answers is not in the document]
saifshaikh48: Agreed, I'll change it. Wanted to stay in line with the [InstanceConfigMap](https://github.com/openshift/windows-machine-config-operator/blob/master/pkg/wiparser/wiparser.go#L19)

saifshaikh48:
Splitting this PR up into 2:
- first PR includes the definition of the structs and the unit tested validation code (https://github.com/openshift/windows-machine-config-operator/pull/1004 )
- second PR is WMCO creating and maintaining the CM (will be this PR)

saifshaikh48:
/hold 

must merge after https://github.com/openshift/windows-machine-config-operator/pull/1004

selansen:
@aravindhp Could you pls review this one?

controllers/configmap_controller.go
```diff
@@ -39,13 +39,14 @@ import (
 
 	"github.com/openshift/windows-machine-config-operator/pkg/cluster"
 	"github.com/openshift/windows-machine-config-operator/pkg/condition"
+	"github.com/openshift/windows-machine-config-operator/pkg/configmap/wiparser"
```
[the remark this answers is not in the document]
selansen: Any reason why are we moving it under configmap?  if we are doing restructuring, can we pls keep it in a separate commit?

pkg/configmap/wsvalidator/wsvalidator.go
selansen: winInstParser & winSerParser ??

pkg/configmap/wsvalidator/wsvalidator.go
```diff
@@ -0,0 +1,23 @@
+package wsvalidator
+
+import (
+	"fmt"
+
+	"github.com/openshift/windows-machine-config-operator/version"
+)
+
+// ServicesConfigMap is the name of the ConfigMap detailing service configuration for a specific WMCO version
```
selansen, line 16: service - services ( this will handle multiple services )

pkg/configmap/wsvalidator/wsvalidator.go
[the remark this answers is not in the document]
selansen: +1

pkg/configmap/wsvalidator/wsvalidator.go
selansen: I prefer ParseAndValidate function just returns bool. Other functionalities can be split into different functions.

pkg/configmap/wsvalidator/wsvalidator.go
```diff
@@ -14,6 +19,79 @@ func init() {
...
+type FileInfo struct {
+	// Path is the filepath of a file on an instance
+	Path string `json:"path"`
+	// Checksum is used to validate that a file has not been changed
+	Checksum string `json:"checksum"`
+}
+
+// GenerateWindowsServiceConfigMap creates an immutable service ConfigMap which provides WICD with the specifications
+// for each Windows service that must be created on a Windows instance.
+func GenerateWindowsServiceConfigMap(name, namespace string) *core.ConfigMap {
```
aravindhp, line 77: Now a validator / parser package is also generating the ConfigMap. Shouldn't the package be name more generically? Maybe servicescm? And then have call sites like `servicescm.Generate()` etc.

test/e2e/validation_test.go
```diff
@@ -464,6 +465,64 @@ func testExpectedServicesRunning(t *testing.T) {
...
+
+// validateWindowsServicesConfigMap ensures the windows-services ConfigMap exists in the cluster and has valid contents
+func (tc *testContext) validateWindowsServicesConfigMap() error {
+	windowsServices, err := tc.client.K8s.CoreV1().ConfigMaps(tc.namespace).Get(context.TODO(),
+		wsvalidator.ServicesConfigMap, meta.GetOptions{})
+	if err != nil {
+		return errors.Wrapf(err, "error retrieving ConfigMap: %s", wsvalidator.ServicesConfigMap)
+	}
+
+	if _, _, err := wsvalidator.ParseAndValidate(windowsServices.Data); err != nil {
```
aravindhp, line 489: What is the advantage in calling this function given this will be called by the operator code also? The real check should be if the operator is not throwing an error.

aravindhp (changes requested):

build/build.sh
```diff
@@ -8,6 +8,8 @@ set -o pipefail
 add_git_data_to_version() {
   local WMCO_SEMVER=$1
   GIT_COMMIT=$(git rev-parse --short HEAD)
+  # The commit hash retreived varies in length, so we will standardize to length received from the operator image
+  GIT_COMMIT="${GIT_COMMIT:0:7}"
```
saifshaikh48, line 12: `7cee9a5e` len of 8 when run locally
`7cee9a5` len of 7 when run in operator image

controllers/configmap_controller.go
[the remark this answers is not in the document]
saifshaikh48: Your suggestion worked, thanks Mansi.

test/e2e/validation_test.go
```diff
@@ -464,6 +465,68 @@ func testExpectedServicesRunning(t *testing.T) {
...
+			}
+		}
+		// retry until timeout if the error is a resource NotFound error
+		return err == nil, nil
+	})
+	if err != nil {
+		return errors.Wrapf(err, "error waiting for ConfigMap %s/%s to be re-created", tc.namespace, cmName)
+	}
+
+	if _, err := servicescm.Parse(cm.Data); err != nil {
```
saifshaikh48, line 524: @aravindhp you'd [previously commented here](https://github.com/openshift/windows-machine-config-operator/pull/1002#discussion_r831502015)

> What is the advantage in calling this function given this will be called by the operator code also? The real check should be if the operator is not throwing an error.

Are you suggesting to remove this `Parse()` call? The goal here is to ensure it is recreated in valid manner

saifshaikh48 (commented):
Tried to address all comments from before and from #1004 in this PR, as we are consolidating this work to one location.

controllers/configmap_controller.go
```diff
@@ -121,19 +125,39 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context,
...
-	// so it is safe to assume that all ConfigMaps being reconciled describe hosts that need to be present in the
-	// cluster. This also handles the case when the reconciliation is kicked off by the InstanceConfigMap being deleted.
-	// In the deletion case, an empty InstanceConfigMap will be reconciled now resulting in all existing BYOH nodes
-	// being deleted.
-	configMap := &core.ConfigMap{}
-	if err := r.client.Get(ctx, req.NamespacedName, configMap); err != nil {
-		if !k8sapierrors.IsNotFound(err) {
-			// Error reading the object - requeue the request.
-			return ctrl.Result{}, err
+	// so it is safe to assume that all ConfigMaps being reconciled are 1 of 2 types
```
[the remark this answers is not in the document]
saifshaikh48: The comment line above should help explain, predicate filters to specific CM resources (instances or services CM)

test/e2e/validation_test.go
aravindhp: Correct. Rather than using `Parse()` you should check for the signal or lack of that an incorrect ConfigMap is present.

pkg/servicescm/servicescm.go
```diff
@@ -14,6 +14,52 @@ func init() {
...
+	// Priority is a non-negative integer that will be used to order the creation of the services.
+	// Priority 0 is created first
+	Priority uint `json:"priority"`
+}
+
+// FileInfo contains the path and checksum of a file copied to an instance by WMCO
+type FileInfo struct {
+	// Path is the filepath of a file on an instance
+	Path string `json:"path"`
+	// Checksum is used to validate that a file has not been changed
```
aravindhp, line 72: Checksum is actually the checksum of the file in Path and that is what needs to be called out here.

controllers/configmap_controller.go
```diff
@@ -121,19 +122,34 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context,
...
+			// If the services ConfigMap is not present, create and retrieve it
+			if configMap, err = r.createServicesConfigMap(ctx, req.NamespacedName); err != nil {
+				return ctrl.Result{}, nil
+			}
+		}
 	}
 
-	return ctrl.Result{}, r.reconcileNodes(ctx, configMap)
+	// at this point configMap will be set properly
+	if req.NamespacedName.Name == servicescm.ServicesConfigMapName {
```
aravindhp, line 143: Can a `switch` statement be used here? That might look clearer.

controllers/configmap_controller.go
```diff
@@ -258,9 +274,14 @@ func hasAssociatedInstance(nodeAddresses []core.NodeAddress, instances []*instan
 
 // mapToConfigMap fulfills the MapFn type, while always returning a request to the windows-instance ConfigMap
```
aravindhp, line 284: Please update the comment

pkg/servicescm/servicescm.go
```diff
@@ -0,0 +1,21 @@
+package servicescm
+
+import (
+	"fmt"
+
+	"github.com/openshift/windows-machine-config-operator/version"
+)
+
+// ServicesConfigMapName is the name of the ConfigMap detailing service configuration for a specific WMCO version
+var ServicesConfigMapName string
```
aravindhp, line 16: This can just be `Name`, given the use sites will have this as `servicescm.Name`

controllers/configmap_controller.go
```diff
@@ -288,5 +309,25 @@ func (r *ConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+// createServicesConfigMap creates a valid ServicesConfigMap and returns it
+func (r *ConfigMapReconciler) createServicesConfigMap(ctx context.Context, namespacedName kubeTypes.NamespacedName) (
+	*core.ConfigMap, error) {
+	windowsServices, err := servicescm.Generate(namespacedName.Name, namespacedName.Namespace)
+	if err != nil {
+		return nil, err
+	}
+
+	if err = r.client.Create(ctx, windowsServices); err != nil {
```
aravindhp, line 333: Won't this fail if the `ConfigMap` already exists say on a Pod restart?

pkg/servicescm/servicescm.go
```diff
@@ -60,6 +71,54 @@ type FileInfo struct {
...
+// configure a Windows instance as a Node
+type data struct {
+	// Services contains information required to start all required Windows services with proper arguments and order
+	Services []Service `json:"services"`
+	// Files contains the path and checksum of all the files copied to a Windows VM by WMCO
+	Files []FileInfo `json:"files"`
+}
+
+// newServicesCMData returns a new servicesCMData. Validates given object contents on creation.
+func newServicesCMData(services *[]Service, files *[]FileInfo) *data {
```
aravindhp: To go with coding conventions, this should be `newData()`,

pkg/servicescm/servicescm.go
```diff
@@ -119,6 +127,106 @@ func Generate(name, namespace string) (*core.ConfigMap, error) {
 	return servicesConfigMap, nil
 }
 
+// Parse converts ConfigMap data into the objects representing a Windows services ConfigMap schema
+// Returns error if the given data is invalid in structure or contents.
+func Parse(data map[string]string) (*data, error) {
```
aravindhp, line 132: Input and output are both called `data`. I suggest using `dataFromCM` as the function parameter.

I also suggest squashing the two validation commits, given they are actually one logical unit i.e. introduce `Parse()` and use it in the same commit.

controllers/configmap_controller.go
```diff
@@ -141,6 +141,15 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context,
 
 	// at this point configMap will be set properly
 	if req.NamespacedName.Name == servicescm.ServicesConfigMapName {
+		// If a ConfigMap with invalid values is found, WMCO will delete and recreate it with proper values
+		if _, err := servicescm.Parse(configMap.Data); err != nil {
+			// Deleting will trigger an event for the configmap_controller, which will re-create a proper ConfigMap
+			if err = r.client.Delete(ctx, configMap); err != nil {
+				return ctrl.Result{}, err
+			}
+			r.log.V(1).Info("Deleted", "ConfigMap", kubeTypes.NamespacedName{Namespace: req.Namespace, Name: req.Name})
```
aravindhp, line 150: This might actually be a useful log. I suggest making this just `Info`. But in addition to just saying `Deleted` please say why.

aravindhp (changes requested):

controllers/configmap_controller.go
saifshaikh48: This is only called if the ConfigMap does not already exist (see Reconcile method above)

controllers/configmap_controller.go
saifshaikh48: I have a question if I am understanding this change correctly. We feed `mapToConfigMap` into the `EnqueueRequestsFromMapFunc` function. Does returning a slice with 2 `Requests` in it mean that, for every event on either ConfigMap, 2 Requests will be enqueued?

controllers/configmap_controller.go
aravindhp: The comment was not clear in the Reconcile function. I have called that out.

controllers/configmap_controller.go
aravindhp: I would think so. You might need to make this mapping function more smart. BTW, what do you see in your testing?

controllers/configmap_controller.go
```diff
@@ -340,3 +340,51 @@ func (r *ConfigMapReconciler) createServicesConfigMap(ctx context.Context, names
...
 	return windowsServices, nil
 }
+
+// EnsureServicesConfigMapExists ensures that the ServicesConfigMap is present and valid on the cluster during operator
+// bootup. ConfigMapReconciler.createServicesConfigMap() cannot be called in its stead as the cache has not been
+// populated yet, which is why the typed client is used here as it calls the API server directly.
+func (r *ConfigMapReconciler) EnsureServicesConfigMapExists() error {
+	var windowsServices *core.ConfigMap
+	var err error
+	if windowsServices, err = r.k8sclientset.CoreV1().ConfigMaps(r.watchNamespace).Get(context.TODO(),
```
sebsoto, line 350: This should be turned into
```
windowsServices, err := r.k8sclientset.CoreV1().ConfigMaps(r.watchNamespace).Get(context.TODO(),
		servicescm.ServicesConfigMapName, meta.GetOptions{})
if err != nil {
...
```

as the types for windowsService and err don't need to be specified

sebsoto (changes requested):

build/build.sh
jrvaldes: This change should go in a different PR

test/e2e/validation_test.go
```diff
@@ -464,6 +465,68 @@ func testExpectedServicesRunning(t *testing.T) {
...
+// 1. It exists on operator startup and 2. It is recreated when deleted
+func testServicesConfigMap(t *testing.T) {
+	tc, err := NewTestContext()
+	require.NoError(t, err)
+
+	operatorVersion, err := getWMCOVersion()
+	require.NoError(t, err)
+	// There is no easy way to populate ldflags when running e2e tests so that version.Get() will return
+	// anything other than "" without having to maintain the WMCO version in another location
+	servicesConfigMapName := servicescm.ServicesConfigMapName + operatorVersion
```
jrvaldes, line 478: Do you need the `-` between `servicescm.ServicesConfigMapName` and `operatorVersion`?

test/e2e/validation_test.go
```diff
@@ -464,6 +465,68 @@ func testExpectedServicesRunning(t *testing.T) {
...
 
+// testServicesConfigMap tests 2 aspects of expected functionality for the services ConfigMap
+// 1. It exists on operator startup and 2. It is recreated when deleted
+func testServicesConfigMap(t *testing.T) {
+	tc, err := NewTestContext()
+	require.NoError(t, err)
+
+	operatorVersion, err := getWMCOVersion()
+	require.NoError(t, err)
+	// There is no easy way to populate ldflags when running e2e tests so that version.Get() will return
```
jrvaldes, line 476: As a complete e2e test, can you fetch all ConfigMaps where the name starts with `windows-services` and validates that there is one, and only one,  that matches `"windows-services"  + "-" + operatorVersion`, then use the content for further validation.

test/e2e/validation_test.go
jrvaldes, line 478: Doesn't `servicescm.ServicesConfigMapName` include the full operator version already?

You may want to introduce an exported constant with the [name prefix](https://github.com/openshift/windows-machine-config-operator/pull/1002#discussion_r831401542)

test/e2e/validation_test.go
```diff
@@ -464,6 +465,68 @@ func testExpectedServicesRunning(t *testing.T) {
...
+	// anything other than "" without having to maintain the WMCO version in another location
+	servicesConfigMapName := servicescm.ServicesConfigMapName + operatorVersion
+
+	err = tc.validateWindowsServicesConfigMap(servicesConfigMapName)
+	require.NoError(t, err, "error validating windows-services ConfigMap")
```
jrvaldes, line 481:
```suggestion
	require.NoErrorf(t, err, "error validating ConfigMap: %s", servicesConfigMapName)
```

Applies everywhere.

controllers/configmap_controller.go
```diff
@@ -340,3 +340,51 @@ func (r *ConfigMapReconciler) createServicesConfigMap(ctx context.Context, names
 	}
 	return windowsServices, nil
 }
+
+// EnsureServicesConfigMapExists ensures that the ServicesConfigMap is present and valid on the cluster during operator
+// bootup. ConfigMapReconciler.createServicesConfigMap() cannot be called in its stead as the cache has not been
+// populated yet, which is why the typed client is used here as it calls the API server directly.
+func (r *ConfigMapReconciler) EnsureServicesConfigMapExists() error {
+	var windowsServices *core.ConfigMap
+	var err error
```
jrvaldes, line 349: You can include the return variable (`err`) explicitly in the signature of the function.

controllers/configmap_controller.go
```diff
@@ -121,19 +122,34 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context,
...
 	configMap := &core.ConfigMap{}
 	if err := r.client.Get(ctx, req.NamespacedName, configMap); err != nil {
 		if !k8sapierrors.IsNotFound(err) {
 			// Error reading the object - requeue the request.
 			return ctrl.Result{}, err
 		}
+		if req.NamespacedName.Name == servicescm.ServicesConfigMapName {
+			// If the services ConfigMap is not present, create and retrieve it
+			if configMap, err = r.createServicesConfigMap(ctx, req.NamespacedName); err != nil {
+				return ctrl.Result{}, nil
```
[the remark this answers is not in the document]
saifshaikh48: right, good catch

build/build.sh
```diff
@@ -3,12 +3,12 @@ set -o errexit
...
 set -o pipefail
 
-# returns the operator version in `Version+GitHash` format
+# returns the operator version in `Version-GitHash` format
 # Takes a semver as an argument. This should be the version present in the WMCO CSV
 add_git_data_to_version() {
   local WMCO_SEMVER=$1
   GIT_COMMIT=$(git rev-parse --short HEAD)
-  FULL_VERSION="${WMCO_SEMVER}+${GIT_COMMIT}"
+  FULL_VERSION="${WMCO_SEMVER}-${GIT_COMMIT}"
```
jrvaldes, line 13: This should go in a separate (prep work) PR. 

Is there a "simple" way to unit test it?

pkg/servicescm/servicescm.go
```diff
@@ -0,0 +1,21 @@
...
+var ServicesConfigMapName string
+
+// init runs once, initializing global variables
+func init() {
+	ServicesConfigMapName = servicesConfigMapName()
+}
+
+// servicesConfigMapName returns the ConfigMap with the naming scheme:
+// windows-services-<WMCOFullVersion>
+func servicesConfigMapName() string {
```
jrvaldes, line 232: Consider `func getName() string { `

pkg/servicescm/servicescm.go
jrvaldes, line 16: Consider `var Name string`

jrvaldes (commented):
@saifshaikh48  Thanks for working on this, comments added. PTAL

build/build.sh
saifshaikh48: Cannot think of a way to unit test this.

test/e2e/validation_test.go
saifshaikh48: Since we don't delete any misnamed configmaps yet, I don't think fetching all `windows-services-*` CMs is needed. Your suggestion is good to keep in mind when we do delete outdated configmaps/handle upgrades in coming stories.

pkg/servicescm/servicescm.go
```diff
@@ -60,6 +71,54 @@ type FileInfo struct {
...
+// newServicesCMData returns a new servicesCMData. Validates given object contents on creation.
+func newServicesCMData(services *[]Service, files *[]FileInfo) *data {
+	cmData := &data{*services, *files}
+	return cmData
+}
+
+// Generate creates an immutable service ConfigMap which provides WICD with the specifications
+// for each Windows service that must be created on a Windows instance.
+func Generate(name, namespace string) (*core.ConfigMap, error) {
+	immutable := true
```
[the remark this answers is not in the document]
saifshaikh48: Yep it's required haha

pkg/servicescm/servicescm.go
```diff
@@ -119,6 +127,106 @@ func Generate(name, namespace string) (*core.ConfigMap, error) {
...
+func (cmData *data) validate() error {
+	if err := validateDependencies(cmData.Services); err != nil {
+		return err
+	}
+	return validatePriorities(cmData.Services)
+}
+
+// validateDependencies ensures that no bootstrap service depends on a non-bootstrap service
+func validateDependencies(services []Service) error {
+	boostrapServices := []Service{}
```
alinaryan, line 170: s/boostrap/bootstrap

pkg/servicescm/servicescm.go
```diff
@@ -119,6 +127,106 @@ func Generate(name, namespace string) (*core.ConfigMap, error) {
...
+	nonBoostrapServices := []Service{}
```
alinaryan, line 171: same comment applies here

pkg/servicescm/servicescm.go
```diff
@@ -119,6 +127,106 @@ func Generate(name, namespace string) (*core.ConfigMap, error) {
...
+// non-bootstrap services. There should be no overlap in the priorities of bootstrap services and controller services.
+func validatePriorities(services []Service) error {
+	// sort services in ascending priority, bootstrap services towards the front of slice
+	sort.Slice(services, func(i, j int) bool {
+		return services[i].Priority < services[j].Priority
+	})
+
+	// ensure no bootstrap service appears after a controller service in the ordered list
+	nonBoostrapSeen := false
+	lastBoostrapPriority := uint(0)
```
alinaryan, line 210: and here

controllers/configmap_controller.go
saifshaikh48: I don't see 2 Requests being processed in my testing with this change. 
Triggering an event on the services configmap (deleting it) results in a single request processed :
```
1.6480679313255267e+09 INFO controllers.configmap Created {"ConfigMap": "openshift-windows-machine-config-operator/windows-services-5.0.0-9419065-dirty"}
1.648067931325568e+09 DEBUG controllers.configmap Reacting to SERVICES event...
```

Likewise, triggering an event on the instances configmap (adding a bogus entry) results in a single request processed: 
```
1.648068484640567e+09 DEBUG controllers.configmap Reacting to INSTANCES event...
1.6480684846406677e+09 ERROR controller.configmap Reconciler error {"reconciler group": "", "reconciler kind": "ConfigMap", "name": "windows-instances", "namespace": "openshift-windows-machine-config-operator", "error": "unable to parse instances from ConfigMap: unable to get username for example.property.1: data has an incorrect format",...
```

Suffice to say I am confused what the purpose of the `[handler.EnqueueRequestsFromMapFunc(r.mapToConfigMap)](https://github.com/openshift/windows-machine-config-operator/pull/1002/commits/24988bcd31186aa43836d186827a99d492d4f922#diff-f3ec2e06fa75f3747dab6fac751809ce87a96e02263980287e9d9477830049aaR306)` call really is.

saifshaikh48: Roping @sebsoto into this convo -- I am not convinced [dropping the change](https://github.com/openshift/windows-machine-config-operator/pull/1002#discussion_r833554476) to `mapToConfigMap` is sufficient

alinaryan (commented):
Thanks for working on this, just a few comments

controllers/configmap_controller.go
sebsoto: The purpose is triggering a windows instance configmap reconciliation upon BYOH Node events

sebsoto: [Here](https://github.com/openshift/windows-machine-config-operator/pull/1002/files#r833713522) you're not seeing reconciliations for both because you are not hitting this function. this function is only called when BYOH Windows node events occur

controllers/configmap_controller.go
saifshaikh48: Fixed up that comment

pkg/servicescm/servicescm.go
```diff
@@ -14,6 +14,52 @@ func init() {
 	ServicesConfigMapName = servicesConfigMapName()
 }
 
+// NodeCmdArg describes a Windows command variable and how its value can be populated
+type NodeCmdArg struct {
+	// Name is the variable name as it appears in commands
+	Name string `json:"name"`
+	// NodeObjectJsonPath is the JSON path of a field within an instance's Node object.
+	// The value of this field is the value of the variable
+	NodeObjectJsonPath string `json:"nodeObjectJsonPath"`
```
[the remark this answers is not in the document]
saifshaikh48: I prefer the more descriptive name. In my mind, Path implies a file path

pkg/servicescm/servicescm.go
```diff
@@ -82,10 +84,13 @@ type data struct {
 	Files []FileInfo `json:"files"`
 }
 
-// newData returns a new 'data' object with the given services and files.
-func newData(services *[]Service, files *[]FileInfo) *data {
+// newData returns a new 'data' object with the given services and files. Validates given object contents on creation.
```
aravindhp, line 87: Going beyond 120 columns.

controllers/configmap_controller.go
```diff
@@ -334,3 +334,48 @@ func (r *ConfigMapReconciler) createServicesConfigMap(ctx context.Context, names
...
+
+		windowsServices, err = servicescm.Generate(servicescm.Name, r.watchNamespace)
+		if err != nil {
+			return err
+		}
+		if _, err = r.k8sclientset.CoreV1().ConfigMaps(r.watchNamespace).Create(context.TODO(), windowsServices,
+			meta.CreateOptions{}); err != nil {
+			return err
+		}
+		r.log.Info("Created", "ConfigMap", kubeTypes.NamespacedName{Namespace: r.watchNamespace, Name: servicescm.Name})
```
aravindhp, line 378: Will it be obvious from the log where this message is coming from? Asking as we have a similar log in `createServicesConfigMap()`

controllers/configmap_controller.go
```diff
@@ -334,3 +334,48 @@ func (r *ConfigMapReconciler) createServicesConfigMap(ctx context.Context, names
...
+		// If ConfigMap is not found, create it and return
+		windowsServices, err = servicescm.Generate(servicescm.Name, r.watchNamespace)
+		if err != nil {
+			return err
+		}
+		if _, err = r.k8sclientset.CoreV1().ConfigMaps(r.watchNamespace).Create(context.TODO(), windowsServices,
+			meta.CreateOptions{}); err != nil {
+			return err
+		}
+		r.log.Info("Created", "ConfigMap", kubeTypes.NamespacedName{Namespace: r.watchNamespace, Name: servicescm.Name})
```
aravindhp, line 357: This is repeated code. Please modularize. Or even better you can set a flag `createConfigMap` here and below which indicates the ConfigMap needs to be created; then only create the ConfigMap if the flag is set.

build/build.sh
aravindhp, line 12: Shouldn't this change go into the first commit?

test/e2e/validation_test.go
```diff
@@ -464,6 +465,64 @@ func testExpectedServicesRunning(t *testing.T) {
...
+func (tc *testContext) validateWindowsServicesConfigMap(cmName string) error {
+	windowsServices, err := tc.client.K8s.CoreV1().ConfigMaps(tc.namespace).Get(context.TODO(),
+		cmName, meta.GetOptions{})
+	if err != nil {
+		return errors.Wrapf(err, "error retrieving ConfigMap: %s", cmName)
+	}
+
+	if _, err := servicescm.Parse(windowsServices.Data); err != nil {
+		return errors.Wrapf(err, "invalid %s ConfigMap data", cmName)
+	}
```
aravindhp, line 498: I again see no point  calling this function here. You are using the same code to test if the ConfigMap is created correctly. There is no way this can be incorrect. I think checking if the ConfigMap is present is sufficient. 

The more interesting test would be to stop the operator, create an invalid ConfigMap and ensure it is recreated correctly.

controllers/configmap_controller.go
```diff
@@ -142,6 +142,16 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context,
 	// At this point configMap will be set properly
 	switch req.NamespacedName.Name {
 	case servicescm.Name:
+		// If a ConfigMap with invalid values is found, WMCO will delete and recreate it with proper values
```
aravindhp, line 145: What is the source of truth of the ConfigMap? At the moment the ConfigMap is empty but if I manually create a ConfigMap with some entries as `Parse()` excepts then that ConfigMap will persist. You need stronger validation for invalid data here.

aravindhp (changes requested):

controllers/configmap_controller.go
```diff
@@ -121,19 +122,46 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context,
...
+		if _, err := servicescm.Parse(configMap.Data); err != nil {
+			// Deleting will trigger an event for the configmap_controller, which will re-create a proper ConfigMap
+			if err = r.client.Delete(ctx, configMap); err != nil {
+				return ctrl.Result{}, err
+			}
+			r.log.Info("Deleted invalid resource", "ConfigMap",
+				kubeTypes.NamespacedName{Namespace: req.Namespace, Name: req.Name}, "Error", err.Error())
+			return ctrl.Result{}, errors.Wrap(err, "Invalid ConfigMap schema")
+		}
```
jrvaldes, line 154: This block can be modularized into a function. e.g.`r.reconcileWindowsServices(ctx, configMap)` where the `switch` will read:

```
	switch req.NamespacedName.Name {
	case servicescm.Name:
              return ctrl.Result{}, r.reconcileWindowsServices(ctx, configMap)

        ...
```

controllers/configmap_controller.go
```diff
@@ -121,19 +122,36 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context,
...
+			}
+		}
 	}
 
-	return ctrl.Result{}, r.reconcileNodes(ctx, configMap)
+	// At this point configMap will be set properly
+	switch req.NamespacedName.Name {
+	case servicescm.Name:
+		// TODO: actually react to changes to the services ConfigMap. Currently do nothing but log
+		r.log.V(1).Info("Reacting to event...", "ConfigMap", req.NamespacedName)
```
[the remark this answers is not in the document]
jrvaldes: +1

I suggest `r.log.V(1).Info("Reconciling", "ConfigMap", req.NamespacedName)` and move it up to line [#142](https://github.com/openshift/windows-machine-config-operator/pull/1002/files#diff-f3ec2e06fa75f3747dab6fac751809ce87a96e02263980287e9d9477830049aaR142), so it will also log the _Unexpected_ case.

controllers/configmap_controller.go
```diff
@@ -121,19 +122,46 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context,
...
+			return ctrl.Result{}, errors.Wrap(err, "Invalid ConfigMap schema")
+		}
+		// TODO: actually react to changes to the services ConfigMap. Currently do nothing but log
+		r.log.V(1).Info("Reacting to event...", "ConfigMap", req.NamespacedName)
+	case wiparser.InstanceConfigMap:
+		return ctrl.Result{}, r.reconcileNodes(ctx, configMap)
+	default:
+		// Unexpected configmap, return no error so we don't requeue
+		r.log.Error(errors.New("Unexpected resource triggered reconcile"), "ConfigMap", req.NamespacedName)
+		return ctrl.Result{}, nil
```
jrvaldes, line 162: No need to return here, it' covered in line [164](https://github.com/openshift/windows-machine-config-operator/pull/1002/files#diff-f3ec2e06fa75f3747dab6fac751809ce87a96e02263980287e9d9477830049aaR164). Otherwise, line 164 is an unreachable code.

controllers/configmap_controller.go
saifshaikh48: Explored both options. Modularizing lead to a much cleaner flow (since we only want to Parse() if we were able to successfully retrieve the configmap).

controllers/configmap_controller.go
saifshaikh48: WMCO is the source of truth. In my view, we have sufficient effort to ensure this, as `Parse()` also does service validation. If someone were to create a configmap adhering to the versioned resource naming structure and all data schema requirements, within our operator namespace while the operator is not running, IMO that's on the user.

If your ask is to have a `requiredServices` and `requiredFiles` set to check against the ConfigMap's data, I think that should be implemented when we start moving services into WICD's scope.

controllers/configmap_controller.go
```diff
@@ -156,6 +156,16 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context,
...
 func (r *ConfigMapReconciler) reconcileServices(ctx context.Context, windowsServices *core.ConfigMap) error {
+	// If a ConfigMap with invalid values is found, WMCO will delete and recreate it with proper values
+	if _, err := servicescm.Parse(windowsServices.Data); err != nil {
+		// Deleting will trigger an event for the configmap_controller, which will re-create a proper ConfigMap
+		if err = r.client.Delete(ctx, windowsServices); err != nil {
+			return err
+		}
+		r.log.Info("Deleted invalid resource", "ConfigMap",
+			kubeTypes.NamespacedName{Namespace: r.watchNamespace, Name: windowsServices.Name}, "Error", err.Error())
+		return errors.Wrap(err, "Invalid ConfigMap schema")
```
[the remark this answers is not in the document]
saifshaikh48: Good catch. This should not return error.

pkg/servicescm/servicescm.go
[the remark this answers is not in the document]
saifshaikh48: The CM's `Immutable` field expects a bool pointer. There's unfortunately no way to declare+instantiate a _true_ *bool inline.

pkg/servicescm/servicescm.go
```diff
@@ -125,6 +134,150 @@ func Generate(name, namespace string) (*core.ConfigMap, error) {
...
 
+// Parse converts ConfigMap data into the objects representing a Windows services ConfigMap schema
+// Returns error if the given data is invalid in structure or contents.
+func Parse(dataFromCM map[string]string) (*data, error) {
+	if len(dataFromCM) != 2 {
+		return nil, errors.New("services ConfigMap should have exactly 2 keys")
+	}
+
+	services := &[]Service{}
+	files := &[]FileInfo{}
```
aravindhp, line 145: Please define closer to the scope of use i.e. before each `Unmarshal`

pkg/servicescm/servicescm.go
```diff
@@ -125,6 +134,150 @@ func Generate(name, namespace string) (*core.ConfigMap, error) {
...
+	for _, requiredSvc := range *requiredServices {
+		// isPresent checks if the required service exists within the given services slice
+		isPresent := func() error {
+			for _, svc := range services {
+				if reflect.DeepEqual(requiredSvc, svc) {
+					return nil
+				}
+			}
+			return errors.Errorf("Required service %s is not present with expected configuration", requiredSvc.Name)
+		}
```
aravindhp, line 192: You are defining this function per required service. Won't it be more efficient to make this a helper?

pkg/servicescm/servicescm.go
```diff
@@ -125,6 +134,150 @@ func Generate(name, namespace string) (*core.ConfigMap, error) {
...
+	}
+	for _, requiredFile := range *requiredFiles {
+		// isPresent checks if the required file exists within the given files slice
+		isPresent := func() error {
+			for _, file := range files {
+				if reflect.DeepEqual(requiredFile, file) {
+					return nil
+				}
+			}
+			return errors.Errorf("Required file %s is not present as expected", requiredFile.Path)
```
aravindhp, line 210: Same comment here.

PS: with Go 1.18, you could use a generics and make a common method across files and services.

controllers/configmap_controller.go
```diff
@@ -339,3 +339,44 @@ func (r *ConfigMapReconciler) createServicesConfigMap(ctx context.Context) (*cor
 	r.log.Info("Created", "ConfigMap", kubeTypes.NamespacedName{Namespace: r.watchNamespace, Name: servicescm.Name})
 	return windowsServices, nil
 }
+
+// createServicesCMWithTypedClient creates a valid ServicesConfigMap
+// ConfigMapReconciler.createServicesConfigMap() cannot be used in its stead as the cache has not been
+// populated yet, which is why the typed client is used here as it calls the API server directly.
+func (r *ConfigMapReconciler) createServicesCMWithTypedClient() error {
```
aravindhp, line 346: I suggest calling this `createServiceConfigMapDuringBootup()`

test/e2e/validation_test.go
```diff
@@ -464,6 +465,95 @@ func testExpectedServicesRunning(t *testing.T) {
...
+	_, err = tc.client.K8s.CoreV1().ConfigMaps(tc.namespace).Get(context.TODO(),
+		servicesConfigMapName, meta.GetOptions{})
+	require.NoErrorf(t, err, "error retrieving ConfigMap: %s", servicesConfigMapName)
+
+	// Delete and ensure re-creation
+	err = tc.client.K8s.CoreV1().ConfigMaps(tc.namespace).Delete(context.TODO(), servicesConfigMapName,
+		meta.DeleteOptions{})
+	require.NoErrorf(t, err, "error deleting ConfigMap: %s", servicesConfigMapName)
+	_, err = tc.waitForConfigMap(servicesConfigMapName)
+	require.NoErrorf(t, err, "error ensuring ConfigMap %s is re-created when deleted", servicesConfigMapName)
```
aravindhp, line 491: For clarity this can be converted in to a subtest i.e. `testServicesCMRecreation()`

pkg/servicescm/servicescm_test.go
```diff
@@ -0,0 +1,463 @@
+package servicescm
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestParse(t *testing.T) {
```
aravindhp, line 10: I just realized that there is no test that ensures the output of Generate() passes through Parse(). Won't that be a useful unit test? Otherwise I think we would only catch that during e2e testing.

aravindhp (changes requested):
Thanks for adding the extra validation. Please remove `wsvalidator` from commit message headers. You can also just use `servicescm` as a prefix if it predominantly deals with it.

jrvaldes (commented):
Nice work @saifshaikh48, PTAL at my comments.

pkg/servicescm/servicescm.go
```diff
@@ -125,6 +134,150 @@ func Generate(name, namespace string) (*core.ConfigMap, error) {
...
+func validatePriorities(services []Service) error {
+	// sort services in ascending priority, bootstrap services towards the front of slice
+	sort.Slice(services, func(i, j int) bool {
+		return services[i].Priority < services[j].Priority
+	})
+
+	// ensure no bootstrap service appears after a controller service in the ordered list
+	nonBootstrapSeen := false
+	lastBootstrapPriority := uint(0)
+	for _, svc := range services {
```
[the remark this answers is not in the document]
jrvaldes: Moreover, if the sorted array is not used or needed elsewhere you could classify the `services` per `Priority` using a priority queue or hashmap and then validate, both will perform better than sorting the whole array (nLog(n) best case).

pkg/servicescm/servicescm.go
saifshaikh48: Yup I noticed this after opening for review so didn't fix it right away :). I'll pull these out into helper on the struct types.

pkg/servicescm/servicescm.go
saifshaikh48: We've not yet moved to 1.18 so I will leave a TODO to consolidate the helpers when we do.

pkg/servicescm/servicescm.go
```diff
@@ -125,6 +134,150 @@ func Generate(name, namespace string) (*core.ConfigMap, error) {
...
+func validateRequiredContent(services []Service, files []FileInfo) error {
+	// Validate services
+	if len(services) != len(*requiredServices) {
+		return errors.New("Unexpected number of services")
+	}
+	for _, requiredSvc := range *requiredServices {
+		// isPresent checks if the required service exists within the given services slice
+		isPresent := func() error {
+			for _, svc := range services {
+				if reflect.DeepEqual(requiredSvc, svc) {
```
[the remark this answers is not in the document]
saifshaikh48: From my reading `DeepEqual` is the recommended way to go. I was a bit skeptical too but the docs on [DeepEqual](https://pkg.go.dev/reflect#DeepEqual) are quite thorough, I recommend taking a look.
I also wrote up a quick demo for my own confidence: https://go.dev/play/p/YmFbShiTX78

test/e2e/validation_test.go
```diff
@@ -464,6 +465,95 @@ func testExpectedServicesRunning(t *testing.T) {
...
+
+	// Invalid as data is missing (services and files)
+	invalidServicesCM := &core.ConfigMap{
+		ObjectMeta: meta.ObjectMeta{
+			Name:      cmName,
+			Namespace: tc.namespace,
+		},
+		Data: make(map[string]string, 2),
+	}
+	if _, err := tc.client.K8s.CoreV1().ConfigMaps(tc.namespace).Create(context.TODO(), invalidServicesCM,
```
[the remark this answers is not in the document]
saifshaikh48: The services CM is immutable so unfortunately no.

test/e2e/validation_test.go
```diff
@@ -464,6 +465,95 @@ func testExpectedServicesRunning(t *testing.T) {
...
+		meta.CreateOptions{}); err != nil {
+		return err
+	}
+
+	// Restart the operator pod
+	if err := tc.scaleWMCODeployment(1); err != nil {
+		return err
+	}
+	// Try to retreive newly created ConfigMap and validate its contents
+	windowsServices, err := tc.waitForConfigMap(cmName)
```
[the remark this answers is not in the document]
saifshaikh48: I agree with everything except `there is a high probability for this test to fail`. By the time scaleDeployment() is done WMCO has deleted the invalid services CM (which is done on operator bootup in main), the way it's currently written has not failed locally.
I will convert the internals of `waitForConfigMap` to `waitForValidWindowsServicesConfigMap`

test/e2e/validation_test.go
```diff
@@ -464,6 +465,95 @@ func testExpectedServicesRunning(t *testing.T) {
...
+	configMap := &core.ConfigMap{}
+	err := wait.Poll(retry.Interval, retry.ResourceChangeTimeout, func() (bool, error) {
+		var err error
+		configMap, err = tc.client.K8s.CoreV1().ConfigMaps(tc.namespace).Get(context.TODO(), cmName, meta.GetOptions{})
+		if err != nil {
+			if !apierrors.IsNotFound(err) {
+				return false, errors.Wrapf(err, "error retrieving ConfigMap: %s", cmName)
+			}
+		}
+		// Retry if the Get() results in a IsNotFound error
```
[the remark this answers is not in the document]
saifshaikh48: Good suggestion. Though btw, in the IsNotFound case, the return value should be `false, nil` to trigger a retry

controllers/configmap_controller.go
```diff
@@ -339,3 +339,44 @@ func (r *ConfigMapReconciler) createServicesConfigMap(ctx context.Context) (*cor
 	r.log.Info("Created", "ConfigMap", kubeTypes.NamespacedName{Namespace: r.watchNamespace, Name: servicescm.Name})
 	return windowsServices, nil
 }
+
+// createServicesCMWithTypedClient creates a valid ServicesConfigMap
+// ConfigMapReconciler.createServicesConfigMap() cannot be used in its stead as the cache has not been
```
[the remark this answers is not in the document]
saifshaikh48: I can see this being possible by trying with the cached client, then depending on the error, try with the typed client. Parsing the error to see if the cache has not been populated yet is proving to be tricky though.

pkg/servicescm/servicescm.go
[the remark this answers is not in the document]
selansen: I thought priority queue will serve the requirement better

pkg/servicescm/servicescm.go
```diff
@@ -125,6 +134,152 @@ func Generate(name, namespace string) (*core.ConfigMap, error) {
...
+	for _, service := range services {
+		if reflect.DeepEqual(s, service) {
+			return true
+		}
+	}
+	return false
+}
+
+// isPresent checks if the required file exists within the given files slice
+// TODO: When we move to go1.18, consolodate this helper with the above Service.isPresent using generics
```
alinaryan, line 211: Why do we have to wait until 1.18 to consolidate?

pkg/servicescm/servicescm.go
```diff
@@ -125,6 +134,152 @@ func Generate(name, namespace string) (*core.ConfigMap, error) {
...
+		return err
+	}
+	if err := validateDependencies(cmData.Services); err != nil {
+		return err
+	}
+	return validatePriorities(cmData.Services)
+}
+
+// validateRequiredContent ensures that the given slices are comprised of all the required services/files and only these
+func validateRequiredContent(services []Service, files []FileInfo) error {
```
alinaryan, line 178: Would it make sense to write this once with a more generic param and call it twice (once for services, once for files) to avoid repeated code?

alinaryan (commented):
Great work! Just a couple questions:

pkg/servicescm/servicescm.go
saifshaikh48: Only 1.18+ supports generic types

pkg/servicescm/servicescm.go
saifshaikh48: Good suggestion, I think when generics are usable this could be done well though.

pkg/servicescm/servicescm.go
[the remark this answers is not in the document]
saifshaikh48: Had fun revisiting this, we could bring it down to O(n) with some more elaborate sorting (something like radix sort which is O(n) then the O(n) traversal to find the first nonBootstrap service). 
AFAIK there is no standard implementation of priority queue in go and I don't see our use case calling for a custom priority queue implementation, so I will leave it as is.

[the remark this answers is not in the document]
selansen: I agree. we need to implement a few functions. I am good with nlogn also. 
BTW I am assuming a max of 5 to 6 services now and expecting it might reach 10-15 to the max. So sorting shouldn't impact performance that much.

selansen:
good work @saifshaikh48 


/lgtm

test/e2e/validation_test.go
```diff
@@ -464,6 +465,107 @@ func testExpectedServicesRunning(t *testing.T) {
...
+
+	err = tc.testServicesCMRecreation(servicesConfigMapName)
+	require.NoErrorf(t, err, "error ensuring ConfigMap %s is re-created when deleted", servicesConfigMapName)
+
+	err = tc.testInvalidServicesCM(servicesConfigMapName)
+	require.NoError(t, err, "error testing handling of invalid ConfigMap")
+}
+
+// testServicesCMRecreation tests that if the services ConfigMap is deleted, a valid one is re-created in its place
+func (tc *testContext) testServicesCMRecreation(cmName string) error {
```
sebsoto, line 494: Does our configmap need recreation time? :laughing: 
I would say consider renaming this, but at the moment I cant think of a fitting word that means the same

test/e2e/validation_test.go
```diff
@@ -464,6 +465,107 @@ func testExpectedServicesRunning(t *testing.T) {
...
+	if err := tc.client.K8s.CoreV1().ConfigMaps(tc.namespace).Delete(context.TODO(), cmName, meta.DeleteOptions{}); err != nil {
+		return err
+	}
```
sebsoto, line 497: Remind me, is this going to return only after the object is fully deleted?
Want to make sure we aren't getting the same object that is actively being deleted

sebsoto (changes requested):
Mostly LGTM

test/e2e/validation_test.go
saifshaikh48: I'm pretty sure it returns after getting a response from the server, in the code I see no guarantee it waits for the resource has been fully deleted. However, that's the reasoning behind `waitForValidWindowsServicesConfigMap()` - if for some reason we get the same object that should be deleted, [the test will not pass until we get a valid (re-created) one](https://github.com/saifshaikh48/windows-machine-config-operator/blob/maintain-services-configmap/test/e2e/validation_test.go#L558-L561).

test/e2e/validation_test.go
saifshaikh48: Lol, I will change to `testServicesCMRegeneration`

test/e2e/validation_test.go
```diff
@@ -464,6 +465,107 @@ func testExpectedServicesRunning(t *testing.T) {
...
+func testServicesConfigMap(t *testing.T) {
+	tc, err := NewTestContext()
+	require.NoError(t, err)
+
+	operatorVersion, err := getWMCOVersion()
+	require.NoError(t, err)
+	// There is no easy way to populate ldflags when running e2e tests without having to maintain the WMCO version in
+	// another location, so that version.Get() will return something other than "". Because of this,
+	// we cannot use servicescm.Name and must build to the services CM name by retrieving the version in another manner.
+	servicesConfigMapName := servicescm.NamePrefix + operatorVersion
```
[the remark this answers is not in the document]
saifshaikh48: I see your point. I'll expand the doc of the helper `getWMCOVersion()`

pkg/servicescm/servicescm.go
```diff
@@ -125,6 +134,152 @@ func Generate(name, namespace string) (*core.ConfigMap, error) {
...
+	for _, requiredFile := range *requiredFiles {
+		if !requiredFile.isPresent(files) {
+			return errors.Errorf("Required file %s is not present as expected", requiredFile.Path)
+		}
+	}
+	return nil
+}
+
+// isPresent checks if the required service exists within the given services slice
+func (s Service) isPresent(services []Service) bool {
```
aravindhp, line 201: This is much stronger than `isPresent()` as not only are you checking if it is present but also that it has the expected configuration. This is more like `isPresentAndCorrect()`

pkg/servicescm/servicescm.go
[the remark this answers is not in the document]
aravindhp: Glad everyone is thinking about complexity. Given we are never going to hit 1000s of services, I am happy that you are keeping this simple.

aravindhp (commented):
Nice work, @saifshaikh48. LGTM.

pkg/servicescm/servicescm.go
aravindhp: Or maybe `isValid()`

saifshaikh48: > not only are you checking if it is present but also that it has the expected configuration

I lean towards `isPresentAndCorrect()` because of this

saifshaikh48:
/retest 

504 building operator image

PR #1004, opened 2022-03-17
WINC-715: Introduce Windows services ConfigMap schema

saifshaikh48:
This PR covers the data representation of a service ConfigMap which provides WICD with the specifications for each Windows service that must be created on a Windows instance.

- Services ConfigMap follows the naming pattern `windows-services-<WMCOMajorVersion>-<WMCOMinorVersion>-<WMCOPatchVersion>-<Commit>`

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -3,12 +3,12 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-# returns the operator version in `Version+GitHash` format
+# returns the operator version in `Version-GitHash` format
 # Takes a semver as an argument. This should be the version present in the WMCO CSV
 add_git_data_to_version() {
   local WMCO_SEMVER=$1
   GIT_COMMIT=$(git rev-parse --short HEAD)
-  FULL_VERSION="${WMCO_SEMVER}+${GIT_COMMIT}"
+  FULL_VERSION="${WMCO_SEMVER}-${GIT_COMMIT}"
 
   # If any files that affect the building of the operator binary have been changed, append "-dirty" to the version.
   if [ -n "$(git status version tools.go go.mod go.sum vendor Makefile build main.go hack pkg controllers --porcelain)" ]; then
--- a/pkg/servicescm/servicescm.go
+++ b/pkg/servicescm/servicescm.go
@@ -1,0 +1,21 @@
+package servicescm
+
+import (
+	"fmt"
+
+	"github.com/openshift/windows-machine-config-operator/version"
+)
+
+// ServicesConfigMapName is the name of the ConfigMap detailing service configuration for a specific WMCO version
+var ServicesConfigMapName string
+
+// init runs once, initializing global variables
+func init() {
+	ServicesConfigMapName = servicesConfigMapName()
+}
+
+// servicesConfigMapName returns the ConfigMap with the naming scheme:
+// windows-services-<WMCOFullVersion>
+func servicesConfigMapName() string {
+	return fmt.Sprintf("windows-services-%s", version.Get())
+}
--- a/pkg/servicescm/servicescm.go
+++ b/pkg/servicescm/servicescm.go
@@ -14,6 +14,49 @@ func init() {
 	ServicesConfigMapName = servicesConfigMapName()
 }
 
+// NodeCmdArg is a variable in a Windows command whose value is sourced from the node object associated with a instance
+type NodeCmdArg struct {
+	// Name is the variable name as it appears in commands
+	Name string `json:"name"`
+	// NodeObjectJsonPath is the JSON path of a field within an instance's Node object
+	NodeObjectJsonPath string `json:"nodeObjectJsonPath"`
+}
+
+// PowershellCmdArg is a variable in a Windows command whose value is determined from the output of a PowerShell script
+type PowershellCmdArg struct {
+	// Name is the variable name as it appears in commands
+	Name string `json:"name"`
+	// Path is the location of the PowerShell script to be run
+	Path string `json:"path"`
+}
+
+// Service represents the configuration spec of a Windows service
+type Service struct {
+	// Name is the name of the Windows service
+	Name string `json:"name"`
+	// Command is command that will be executed. This could potentially include strings whose values will be derived
+	// from NodeVariablesInCommand and PowershellVariablesInCommand.
+	Command string `json:"path"`
+	// Before a command is run on a Windows instance, all node and PowerShell variables will be replaced by their values
+	NodeVariablesInCommand       []NodeCmdArg       `json:"nodeVariablesInCommand,omitempty"`
+	PowershellVariablesInCommand []PowershellCmdArg `json:"powershellVariablesInCommand,omitempty"`
+	// Dependencies is a list of service names that this service is dependent on
+	Dependencies []string `json:"dependencies,omitempty"`
+	// Bootstrap is a boolean flag indicating whether this service should be handled as part of node bootstrapping
+	Bootstrap bool `json:"bootstrap"`
+	// Priority is a non-negative integer that will be used to order the creation of the services.
+	// Priority 0 is created first
+	Priority uint `json:"priority"`
+}
+
+// FileInfo contains the path and checksum of a file copied to an instance by WMCO
+type FileInfo struct {
+	// Path is the filepath of a file on an instance
+	Path string `json:"path"`
+	// Checksum is used to validate that a file has not been changed
+	Checksum string `json:"checksum"`
+}
+
 // servicesConfigMapName returns the ConfigMap with the naming scheme:
 // windows-services-<WMCOFullVersion>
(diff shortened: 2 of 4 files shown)


pkg/configmap/services/services.go
```diff
@@ -85,6 +87,29 @@ func NewServicesConfigMapData(services *[]Service, files *[]FileInfo) (*services
 	return cmData, nil
 }
 
+// GenerateWindowsServicesConfigMap creates an immutable service ConfigMap which provides WICD with the specifications
+// for each Windows service that must be created on a Windows instance.
+func GenerateWindowsServicesConfigMap(name, namespace string) *core.ConfigMap {
```
sebsoto, line 92: This commit seems like it belongs in the other PR, where this would actually be used

test/e2e/create_test.go
```diff
@@ -15,19 +15,19 @@ import (
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"k8s.io/api/core/v1"
+	core "k8s.io/api/core/v1"
```
sebsoto, line 18: this is an unrelated change

pkg/configmap/services/services.go
```diff
@@ -57,6 +68,119 @@ type FileInfo struct {
...
+
+	services := &[]Service{}
+	files := &[]FileInfo{}
+
+	value, ok := data[servicesKey]
+	if !ok {
+		return nil, errors.Errorf("expected `%s` key to exist", servicesKey)
+	}
+	if err := json.Unmarshal([]byte(value), services); err != nil {
+		return nil, errors.Wrapf(err, "`%s` must represent list of Service", servicesKey)
```
sebsoto, line 128: a list of Services

sebsoto: could just return error without wrapping it as i dont think you're adding much info here

pkg/configmap/services/services.go
```diff
@@ -57,6 +68,119 @@ type FileInfo struct {
...
+	}
+
+	value, ok = data[filesKey]
+	if !ok {
+		return nil, errors.Errorf("expected `%s` key to exist", filesKey)
+	}
+	if err := json.Unmarshal([]byte(value), files); err != nil {
+		return nil, errors.Wrapf(err, "`%s` must represent list of FileInfo", filesKey)
```
sebsoto, line 136: same as above

pkg/configmap/services/services.go
```diff
@@ -57,6 +68,119 @@ type FileInfo struct {
...
+
+	for _, bootstrapSvc := range boostrapServices {
+		for _, dependency := range bootstrapSvc.Dependencies {
+			for _, nonBootstrapSvc := range nonBoostrapServices {
+				if dependency == nonBootstrapSvc.Name {
+					return errors.Errorf("bootstrap service %s cannot depend on non-boostrap service %s",
+						bootstrapSvc.Name, nonBootstrapSvc.Name)
+				}
+			}
+		}
```
sebsoto, line 175: I think this should be moved into a separate function that takes in a bootstrap service and the slice of nonBootstrap services. The logic is good, just a bit hard to read. example
```
func hasDependency(s Service, possibleDependencies []Service) bool{
		for _, dependency := range s.Dependencies {
			for _, possibleDependency := range possibleDependencies {
				if dependency == possibleDependency.Name {
					return true
				}
			}
			return false
}
```

pkg/configmap/services/services.go
```diff
@@ -57,6 +68,119 @@ type FileInfo struct {
...
+	}
+	return nil
+}
+
+// validatePriorities ensures that each service that has the bootstrap flag set as true has a higher priority than all
+// non-bootstrap services. There should be no overlap in the priorities of bootstrap services and controller services.
+func validatePriorities(services *[]Service) error {
+	// sort services to ascending priority, bootstrap services towards the front of slice
```
sebsoto, line 183: services to ascending -> services in ascending

pkg/configmap/services/services.go
```diff
@@ -57,6 +68,119 @@ type FileInfo struct {
...
+				}
+			}
+		}
+	}
+	return nil
+}
+
+// validatePriorities ensures that each service that has the bootstrap flag set as true has a higher priority than all
+// non-bootstrap services. There should be no overlap in the priorities of bootstrap services and controller services.
+func validatePriorities(services *[]Service) error {
```
sebsoto, line 182: I could be wrong, but aren't slices already a pointer?
passing in a pointer to slice is redundant

sebsoto (changes requested):

pkg/configmap/services/services.go
saifshaikh48: True. I kept going back and forth whether include it or not haha

pkg/configmap/services/services.go
saifshaikh48: Thanks, I used your snippet as is.

pkg/configmap/services/services.go
```diff
@@ -0,0 +1,21 @@
...
+// ServicesConfigMapName is the name of the ConfigMap detailing service configuration for a specific WMCO version
+var ServicesConfigMapName string
+
+// init runs once, initializing global variables
+func init() {
+	ServicesConfigMapName = servicesConfigMapName()
+}
+
+// servicesConfigMapName returns the ConfigMap with the naming scheme:
+// windows-services-<MajorVersion>-<MinorVersion>-<PatchVersion>-<CommitHash>
```
jrvaldes, line 188:
```suggestion
// windows-services-<WMCO Full Version>
```


The WMCO full version schema is maintained [elsewhere](https://github.com/openshift/windows-machine-config-operator/blob/master/build/build.sh#L11). Here, we just consume it.

pkg/configmap/services/services.go
```diff
@@ -14,6 +14,49 @@ func init() {
 	ServicesConfigMapName = servicesConfigMapName()
```
jrvaldes, line 18: Consider squashing this commit with the previous one. (https://github.com/openshift/windows-machine-config-operator/pull/1004/commits/f523a455c16269b5304ea47d1b1b3c534c7888f9)

build/build.sh
```diff
@@ -8,7 +8,7 @@ set -o pipefail
 add_git_data_to_version() {
   local WMCO_SEMVER=$1
   GIT_COMMIT=$(git rev-parse --short HEAD)
-  FULL_VERSION="${WMCO_SEMVER}+${GIT_COMMIT}"
+  FULL_VERSION="${WMCO_SEMVER}-${GIT_COMMIT}"
```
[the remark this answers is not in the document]
saifshaikh48: Given the change is directly relevant to the work in this PR, I would like to keep this change here. As far as testing, the rest of the changes will have no effect on existing functionality (the CM data schema is not plugged in to reconcile logic yet) so we'll know exactly what's causing issues if they arise.

pkg/servicescm/servicescm.go
```diff
@@ -57,6 +68,123 @@ type FileInfo struct {
...
+	for _, bootstrapSvc := range boostrapServices {
+		if hasDependency(bootstrapSvc, nonBoostrapServices) {
+			return errors.Errorf("bootstrap service %s cannot depend on non-boostrap service", bootstrapSvc.Name)
+		}
+	}
+	return nil
+}
+
+// hasDependency checks if a service is dependent on any services in the given slice
+func hasDependency(s Service, possibleDependencies []Service) bool {
```
aravindhp, line 147: This can be method of the `Service` struct.

aravindhp (changes requested):

I found this a bit hard to review as it is unclear how the functions being introduced here are going to be used which makes me wonder if this is the correct way to split up the PRs.

pkg/servicescm/servicescm.go
```diff
@@ -0,0 +1,21 @@
+package servicescm
```
alinaryan, line 1: For simplicity, how about referring to this just as `services` in the package name, commit message etc?

alinaryan (changes requested):
Nice work so far, just one comment:

pkg/servicescm/servicescm.go
```diff
@@ -57,6 +68,123 @@ type FileInfo struct {
...
+	cmData := &servicesCMData{*services, *files}
+	if err := cmData.validate(); err != nil {
+		return nil, errors.Wrap(err, "unable to create services ConfigMap data object")
+	}
+	return cmData, nil
+}
+
+// Parse converts ConfigMap data into the objects representing a Windows services ConfigMap schema
+// Returns error if the given data is invalid in structure or contents.
+func Parse(data map[string]string) (*servicesCMData, error) {
```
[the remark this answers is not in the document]
selansen: why do we need hash Map here? won't a simple string slice work?

PR #1007, opened 2022-03-21
[e2e] Fix WMCO binary path

saifshaikh48:
This PR fixes an old bug that occurs only when running WMCO e2e tests locally.
Since the test files are not within the top level WMCO directory and instead
nested in the `test/e2e` directories, this must be reflected when referencing
the WMCO binary. Otherwise, things like [getting the operator version](https://github.com/openshift/windows-machine-config-operator/blob/master/test/e2e/validation_test.go#L195-L196) will fail.


mansikulkarni96:
@saifshaikh48 does this need to be backported?

PR #1013, opened 2022-03-28
WINC-687: Windows instances react to kubelet CA rotation

jrvaldes:
- Introduces the certificates package with common operations and constants
- Adds a new watcher to the ConfigMap controller to react on changes in the
serving CA ConfigMap from the Kube API Server
- Extends the instanceReconciler with a new capability to reconcile the
kubelet CA rotation
- Adds a function to update the kubelet CA file in a Windows instance. The
location and name of the file is fixed to `C:\k\kubelet-ca.crt` by WMCB while
setting the kubelet configuration.

The initial Kubelet CA certificate is valid for 1 year, from the date of the
cluster installation. Usually, the first rotation of the kubelet CA certificate
is generated by the [API Server Operator](https://github.com/openshift/cluster-kube-apiserver-operator)
at 80% for the CA certificate lifecycle (1 year), which is 292 days, and removes
the previous CA certificate after 365 days.

Alternatively, a cluster administrator can trigger the kubelet CA rotation
on-demand at any time with the following command:
```
oc patch secret kube-apiserver-to-kubelet-signer \
-n openshift-kube-apiserver-operator \
-p='{"metadata": {"annotations": {"auth.openshift.io/certificate-not-after": null}}}'
```

The proposed implementation covers both of the above scenarios and does not require
node drain or node reboot. When the new CA bundle (kubelet-ca.crt) is updated
in the worker node, kubelet detects the change in the file system, reads the
certificates from the file, and updates the client configuration.

Related MCO PRs:
- https://github.com/openshift/machine-config-operator/pull/626
- https://github.com/openshift/machine-config-operator/pull/965


jrvaldes:
/hold 

Depends on submodule update with https://github.com/openshift/windows-machine-config-bootstrapper/pull/344

saifshaikh48 (commented):
I am not familiar with the initial CA configmap -- my main concern (might be a stupid one) can be summarized as follows: if the initial kubelet CA is only valid for one year, then do we have a code path here that breaks after a year?

jrvaldes:
> I am not familiar with the initial CA configmap -- my main concern (might be a stupid one) can be summarized as follows: if the initial kubelet CA is only valid for one year, then do we have a code path here that breaks after a year?

Correct, it is a time bomb<sup>[1]</sup>

<sup>[1]</sup> https://github.com/openshift/machine-config-operator/pull/965#issuecomment-513924977

controllers/controllers.go
```diff
@@ -132,6 +136,90 @@ func (r *instanceReconciler) instanceFromNode(node *core.Node) (*instance.Info,
...
+	bundleCAConfigMap *core.ConfigMap) error {
+	r.log.Info("reconciling kubelet CA client certificates with", "ConfigMap", bundleCAConfigMap.Name)
+	// extract CAs form initial CA ConfigMap
+	initialKubeAPIServerServingCABytes, err := certificates.GetCAsFromConfigMap(initialCAConfigMap,
+		certificates.CABundleKey)
+	if err != nil {
+		return err
+	}
+	// extract CAs from new bundle CA ConfigMap
+	kubeAPIServerServingCABytes, err := certificates.GetCAsFromConfigMap(bundleCAConfigMap, certificates.CABundleKey)
```
selansen, line 151: when you name the previous one initialKubeAPI.., you can name this new or current-kubeAPIser...

controllers/controllers.go
```diff
@@ -132,6 +136,90 @@ func (r *instanceReconciler) instanceFromNode(node *core.Node) (*instance.Info,
...
+	if err != nil {
+		return err
+	}
+	// merge initial certificates with new bundle
+	kubeAPIServerServingCABytes = certificates.MergeCertWithCABundle(initialKubeAPIServerServingCABytes,
+		kubeAPIServerServingCABytes, "kube-apiserver-to-kubelet-signer")
+
+	// TODO: Improve windows.sshConnectivity.transfer() to avoid below IO operations.
+	// We already have the content of the the bundle CA in memory, but we are forced to dump it into a file because
+	// is the only mechanism accepted by windows.sshConnectivity.transfer(). The later IO operations are expensive and
```
selansen, line 161: because it is

controllers/controllers.go
```diff
@@ -132,6 +136,89 @@ func (r *instanceReconciler) instanceFromNode(node *core.Node) (*instance.Info,
...
+	// could be avoided by overloading .transfer() to accept a byte array.
```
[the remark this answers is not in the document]
selansen: This is going to happen once a year or once in 6 months. I don't think people are going to keep rotating it.  I do understand IO operation is expensive but we don't need to worry about it in this situation.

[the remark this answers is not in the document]
selansen: But interesting to know  data is read from file and is the only way available now

controllers/controllers.go
```diff
@@ -132,6 +136,90 @@ func (r *instanceReconciler) instanceFromNode(node *core.Node) (*instance.Info,
...
+	}(kubeletCAFilePath)
+
+	// fetch all Windows nodes (Machine and BYOH instances)
+	winNodes := &core.NodeList{}
+	if err = r.client.List(ctx, winNodes, client.MatchingLabels{core.LabelOSStable: "windows"}); err != nil {
+		return errors.Wrap(err, "error listing Windows nodes")
+	}
+	r.log.V(1).Info("processing", "node count", len(winNodes.Items))
+	// loop Windows nodes and trigger kubelet CA update
+	for _, winNode := range winNodes.Items {
```
selansen, line 186: Assume if we have 3 Windows nodes and the second one fails due to some error condition, aren't we going to try the 3rd node?

pkg/certificates/bundle.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/pkg/errors"
+	core "k8s.io/api/core/v1"
+	kubeTypes "k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// CABundleKey is the key in the Kube API Server CA ConfigMap where the CA bundle is located
+	CABundleKey = "ca-bundle.crt"
+	// KubeApiServerOperatorNamespace is the name of the namespace of the ConfigMap that contains the CA for the kubelet
```
selansen, line 20: namespace of the ConfigMap - ConfigMap's namespace

pkg/certificates/bundle.go
```diff
@@ -0,0 +1,99 @@
...
+	// to recognize the kube-apiserver client certificate.
+	KubeApiServerOperatorNamespace = "openshift-kube-apiserver-operator"
+	// KubeAPIServerServingCAConfigMapName is the name of the ConfigMap that contains the CA for the kubelet to
+	// recognize the kube-apiserver client certificate.
+	KubeAPIServerServingCAConfigMapName = "kube-apiserver-to-kubelet-client-ca"
+
+	// configNamespace is the name of the namespace of the ConfigMap that contains the initial kubelet client CA.
```
selansen, line 27: namespace of the ConfigMap - ConfigMap's namespace

pkg/certificates/bundle.go
```diff
@@ -0,0 +1,99 @@
...
+	mergedBytes := []byte{}
+	for len(initialBundle) > 0 {
+		b, next := pem.Decode(initialBundle)
+		if b == nil {
+			break
+		}
+		c, err := x509.ParseCertificate(b.Bytes)
+		if err != nil {
+			// glog.Warningf("Could not parse initial bundle certificate: %v", err)
+			continue
```
selansen, line 46: if we keep hitting in this part, can this become an infinite loop?. The for loop is running as long as `len(initialBundle) > 0 `

test/e2e/validation_test.go
```diff
@@ -464,6 +464,11 @@ func testExpectedServicesRunning(t *testing.T) {
 	}
 }
 
+// testKubeletCARotation test the kubelet CA rotation
+func testKubeletCARotation(t *testing.T) {
+	t.Skip("Not implemented yet")
```
selansen, line 469: Is there going to be a different ticket to track this one?

controllers/configmap_controller.go
```diff
@@ -21,6 +21,7 @@ import (
 	"net"
```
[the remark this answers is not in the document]
jrvaldes: Agree.

controllers/controllers.go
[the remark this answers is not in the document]
jrvaldes: Right, that is my take too. I don't feel we should track it in the backlog (for now).

controllers/controllers.go
jrvaldes: No prefix means current.

controllers/controllers.go
jrvaldes: Good question. In case the 3rd node fails, the reconciler will reconcile the failed event where nodes 1 and 2 are up to date, and only the 3rd node will get the file update.

pkg/certificates/bundle.go
jrvaldes: Correct, sounds like a good case for unit testing.

test/e2e/validation_test.go
jrvaldes: Included in [e163351 ](https://github.com/openshift/windows-machine-config-operator/commit/e163351929fa3facd6ca7be1ffc02d722204b353)

sebsoto (changes requested):

controllers/controllers.go
```diff
@@ -92,29 +92,43 @@ func (r *instanceReconciler) ensureInstanceIsUpToDate(instanceInfo *instance.Inf
...
 // instanceFromNode returns an instance object for the given node. Requires a username that can be used to SSH into the
 // instance to be annotated on the node.
 func (r *instanceReconciler) instanceFromNode(node *core.Node) (*instance.Info, error) {
+	// get default username
+	username := r.getDefaultUsername()
+	// check node for username annotation (BYOH case)
 	usernameAnnotation := node.Annotations[UsernameAnnotation]
-	if usernameAnnotation == "" {
-		return nil, errors.New("node is missing valid username annotation")
+	if usernameAnnotation != "" {
```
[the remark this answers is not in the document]
jrvaldes: Agree. I thought about that too. It's more nosy (and complex) so I went this route instead.

Happy to chat more about this.

jrvaldes:
> /hold
> 
> Depends on submodule update with [openshift/windows-machine-config-bootstrapper#344](https://github.com/openshift/windows-machine-config-bootstrapper/pull/344)

/hold cancel

https://github.com/openshift/windows-machine-config-operator/pull/1015 has merge

controllers/controllers.go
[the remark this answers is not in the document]
saifshaikh48: I am leaning towards including the Windows annotation on all nodes too, but I understand that may be out of scope of this story. Another alternative seems to be splitting this into 2 functions upon the respective Reconciler structs, one for machine nodes and one for BYOH?

test/e2e/certificates_test.go
```diff
@@ -0,0 +1,207 @@
...
+	tc, err := NewTestContext()
+	require.NoError(t, err)
+
+	// TODO: Validate no workload disruption during certificate rotation/deletion.
+	// Deploy a Windows workload with WebServer, a service endpoint, and a job
+	// that continuously curls the service.
+	// After below tests pass, check for errors in the job.
+
+	t.Run("Kubelet CA rotation", tc.testKubeletCARotation)
+	t.Run("Kubelet CA deletion", tc.testKubeletCARemoval)
```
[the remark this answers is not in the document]
jrvaldes: @saifshaikh48  good questions

> Would things break if we were to call these functions multiple times in a row?

No, the `kube-apiserver-to-kubelet-client-ca` ConfigMap will always contain the latest CA certificate


> at this point does initialCAConfigMap get updated by some entity to hold cert 2?

No. This is [static and effectively frozen in time](https://github.com/openshift/machine-config-operator/blob/5516bd8911ce3e53cfc20924609c129d2eabc604/pkg/operator/sync.go#L239).

> Then at day 657, cert 2 is rotated and cert 3 is added to CA bundle -- everything still works fine?

Yes,  `kube-apiserver-to-kubelet-client-ca` ConfigMap will contains both CA 

> Then at day 730, cert 2 is removed from CA bundle -- everything still works fine?

Yes,  `kube-apiserver-to-kubelet-client-ca` ConfigMap just contains the latest CA.


Now, WRT to
> ...the role of initialCAConfigMap is useful after the first year.

I'm a bit concerned about other CAs existing in the initialCAConfigMap, that are not related to the kubelet CA.

test/e2e/certificates_test.go
```diff
@@ -0,0 +1,207 @@
...
+// checkJournalLogInWindowsNodes checks journal logs in Windows nodes are available. If error, retries every retry.Interval
+// until retry.ResourceChangeTimeout is reached.
+func checkJournalLogInWindowsNodes() error {
+	// check logs first and retry if unavailable
+	err := wait.PollImmediate(retry.Interval, retry.ResourceChangeTimeout, func() (bool, error) {
+		// query last line of journal log for each Windows node
+		_, err := getNodeJournalLog(windowsJournalUnits, len(gc.allNodes()))
+		if err != nil {
+			// retry
+			log.Printf("error checking node journal logs, retrying in %s...", retry.Interval)
```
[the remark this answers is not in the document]
jrvaldes: I  did and preferred something simpler, similar to this [one](https://github.com/openshift/windows-machine-config-operator/blob/9818df05639982d690475a3f7743cea9afb73153/test/e2e/create_test.go#L253).

test/e2e/logs_test.go
```diff
@@ -63,3 +63,37 @@ func retrieveLog(nodeName, srcPath, destDir string) error {
...
+//
+// units filter the log entries for the specified unit(s)
+//
+// tailCount filter up to this many lines (not more than 100k) from the end of the log. For all logs set tailCount=0.
+func getNodeJournalLog(units []string, tailCount int, names ...string) ([]byte, error) {
+	// process units if any
+	unitsArg := ""
+	if len(units) > 0 {
+		// double-quote value for safer PowerShell parsing
+		unitsArg = fmt.Sprintf("%s=%q", "--unit", strings.Join(units, ","))
```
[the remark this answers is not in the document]
jrvaldes: Good point, will use single quotes.

pkg/certificates/bundle.go
```diff
@@ -0,0 +1,101 @@
...
+	// configNamespace is the namespace that contains the initial kubelet CA ConfigMap
+	configNamespace = "openshift-config"
+	// kubeAPIServerInitialCAConfigMapName is the name of the ConfigMap that contains the initial CA certificates
+	// created during the cluster installation, where the initial kubelet CA is valid only for the first year.
+	kubeAPIServerInitialCAConfigMapName = "initial-kube-apiserver-server-ca"
+)
+
+// MergeCertWithCABundle merges the content of the initial bundle with new bundle for the specified subject
+// Adapted from https://github.com/openshift/machine-config-operator/blob/1a9f70f333a2287c4a8f2e75cb37b94c7c7b2a20/pkg/operator/sync.go#L939
+func MergeCertWithCABundle(initialBundle, newBundle []byte, subject string) []byte {
```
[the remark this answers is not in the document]
jrvaldes: It ignores any invalid certificate that fails, the result will be an empty bundle (worst case, where are certs are invalid).

controllers/controllers.go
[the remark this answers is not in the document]
saifshaikh48: Not that involved from what I can tell:
- call `crypto.EncryptToJSONString` to get encrypted value of Machine [username](https://github.com/openshift/windows-machine-config-operator/blob/master/controllers/windowsmachine_controller.go#L401)
- pass in encrypted username annotation to ` r.ensureInstanceIsUpToDate` [call in wm_controller](https://github.com/openshift/windows-machine-config-operator/blob/master/controllers/windowsmachine_controller.go#L412)
- remove [this if/else block in secret_controller](https://github.com/openshift/windows-machine-config-operator/blob/master/controllers/secret_controller.go#L204-L219), leaving only the code from within the `if`
- for e2e tests, add UsernameAnnotation to `annotations` and remove the `if BYOH` check [here](https://github.com/openshift/windows-machine-config-operator/blob/master/test/e2e/create_test.go#L367-L371)
- remove the `isBYOH` condition from [this if statement](https://github.com/openshift/windows-machine-config-operator/blob/master/test/e2e/create_test.go#L450) in create_test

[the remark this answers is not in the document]
jrvaldes: @saifshaikh48  thanks for highlighting the implementation details.

Opened:
- https://github.com/openshift/windows-machine-config-operator/pull/1021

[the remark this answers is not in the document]
jrvaldes: > remove [this if/else block in secret_controller](https://github.com/openshift/windows-machine-config-operator/blob/master/controllers/secret_controller.go#L204-L219), leaving only the code from within the if

It makes sense, but that is a change of behavior. I don't think we should introduce it as part of this effort. So, far Windows instances configured by MachineController reconcile the annotation by "deletion".

jrvaldes:
/hold

Depends on https://github.com/openshift/windows-machine-config-operator/pull/1021

test/e2e/certificates_test.go
```diff
@@ -0,0 +1,207 @@
...
+	initialCABytes, err := certificates.GetCAsFromConfigMap(initialCAConfigMap, certificates.CABundleKey)
+	require.NoError(t, err)
+	initialKubeApiServerSignerCert := findCertInCABundle(initialCABytes, "kube-apiserver-to-kubelet-signer")
+
+	// remove initial Kubelet CA certificate from kube-apiserver-to-kubelet-client-ca ConfigMap
+	err = tc.removeCertFromKubeAPIServerCA(initialKubeApiServerSignerCert)
+	require.NoError(t, err)
+
+	// wait 5 seconds for the Kube API server to process the certificate removal
+	time.Sleep(5 * time.Second)
```
saifshaikh48, line 83: Will this introduce flakiness? Is there a way to `waitFor()` this to happen instead?

test/e2e/certificates_test.go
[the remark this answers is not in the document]
saifshaikh48: > other CAs existing in the initialCAConfigMap

If I understand correctly, we could potentially remove those other CAs within this test (which may or may not cause issues), but this is not something that affects a real world scenario?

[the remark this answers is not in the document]
jrvaldes: > we could potentially remove those other CAs within this test 

Not sure. There are other certs there with different expiration dates, for example, `kubelet-bootstrap-kubeconfig-signer` is valid for 10 years.

test/e2e/certificates_test.go
jrvaldes: Yes, we could track all the chain of events from kube-api-operator to MCO. However, it will add more complexity to the e2e test and a few extra API calls.

I ran several tests locally, and worst case it takes 5 seconds to apply the changes in vSphere and AWS (I did not test Azure) so I feel it's safer and simpler to just 🛌🏼 .

saifshaikh48: I suggest making it slightly longer (seconds seems like a short window though that is just based on my speculation). Consider using one of our [retry constants](https://github.com/openshift/windows-machine-config-operator/blob/master/pkg/retry/retry.go#L9)

jrvaldes:
> /hold
> 
> Depends on #1021

/hold cancel

#1021 has merge

pkg/certificates/bundle.go
```diff
@@ -0,0 +1,101 @@
+package certificates
```
sebsoto, line 1: Please rename this to certificates.go

unless a new complex object is being introduced we should try to keep file names == package names

pkg/certificates/bundle.go
sebsoto: I would like to see unit tests around this functionality

sebsoto (changes requested):
Good work @jrvaldes thanks for putting in the effort to research this
Please see my comments

controllers/configmap_controller.go
```diff
@@ -304,7 +309,8 @@ func (r *ConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			return r.isValidConfigMap(e.Object)
 		},
 		UpdateFunc: func(e event.UpdateEvent) bool {
-			return r.isValidConfigMap(e.ObjectNew)
+			return r.isValidConfigMap(e.ObjectNew) ||
+				r.isKubeAPIServerServingCAConfigMap(e.ObjectNew)
```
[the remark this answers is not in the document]
jrvaldes: Is not that simple, it also [accounts](https://github.com/openshift/windows-machine-config-operator/blob/3a2c73a59041d7be5682e6af68cd908a1161ba45/controllers/configmap_controller.go#L326) for the `windows-services` ConfigMap.

controllers/controllers.go
```diff
@@ -118,6 +122,93 @@ func (r *instanceReconciler) instanceFromNode(node *core.Node) (*instance.Info,
...
+		return err
+	}
+	// extract CAs from new bundle CA ConfigMap
+	kubeAPIServerServingCABytes, err := certificates.GetCAsFromConfigMap(bundleCAConfigMap, certificates.CABundleKey)
+	if err != nil {
+		return err
+	}
+	// merge initial certificates with new bundle
+	kubeAPIServerServingCABytes = certificates.MergeCertWithCABundle(initialKubeAPIServerServingCABytes,
+		kubeAPIServerServingCABytes, "kube-apiserver-to-kubelet-signer")
```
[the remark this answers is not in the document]
jrvaldes: Makes sense.

controllers/controllers.go
```diff
@@ -118,6 +122,93 @@ func (r *instanceReconciler) instanceFromNode(node *core.Node) (*instance.Info,
...
+	if err != nil {
+		return errors.Wrapf(err, "error creating file: %s", kubeletCAFilePath)
+	}
+	// cleanup auxiliary certificate file
+	defer func(name string) {
+		r.log.V(1).Info("deleting kubelet CA certificate", "file", name)
+		if err := os.Remove(name); err != nil {
+			r.log.V(1).Error(err, "error deleting kubelet CA certificate", "file", name)
+		}
+	}(kubeletCAFilePath)
```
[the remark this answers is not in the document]
jrvaldes: I tend to agree with you, but tech debts are not that bad. The refactoring you are suggesting is out of the scope of this PR.

I can remove the `TODO:`.

controllers/configmap_controller.go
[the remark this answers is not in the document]
jrvaldes: Because: 
- need to react only to `Update` events
- the KubeAPIServer CA ConfigMap is a resource outside of the watched namespace (`r.watchNamespace`)

pkg/certificates/bundle.go
```diff
@@ -0,0 +1,101 @@
...
+		} else {
+			// merge the original cert
+			mergedBytes = append(mergedBytes, pem.EncodeToMemory(b)...)
+		}
+	}
+	return mergedBytes
+}
+
+// GetCAsFromConfigMap extracts the given key from the ConfigMap object
+// Taken from https://github.com/openshift/machine-config-operator/blob/1a9f70f333a2287c4a8f2e75cb37b94c7c7b2a20/pkg/operator/sync.go#L874
```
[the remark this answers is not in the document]
jrvaldes: No, they are not exported.

pkg/certificates/bundle.go
jrvaldes: Added in https://github.com/openshift/windows-machine-config-operator/pull/1013/commits/6498dc09131d0b61dd01bfef860f36ed60a4e74d

controllers/controllers.go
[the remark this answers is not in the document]
jrvaldes: Refactored the file opening in https://github.com/openshift/windows-machine-config-operator/pull/1013/commits/1d7f9bdb843461b7e8a3543e18895fa937024ee2

controllers/controllers.go
```diff
@@ -118,6 +122,93 @@ func (r *instanceReconciler) instanceFromNode(node *core.Node) (*instance.Info,
...
+		r.log.Info("updating kubelet CA client certificates in", "node", winNode.Name)
+		err = nodeConfig.Windows.UpdateKubeletClientCA(kubeletCAFilePath)
+		if err != nil {
+			return errors.Wrapf(err, "error updating kubelet CA certificate in node %s", winNode.Name)
+		}
+	}
+	return nil
+}
+
+// createFile creates a file with the given path and content, and returns the path of the created file.
```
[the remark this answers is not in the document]
jrvaldes: Removed. With the improvements in the connectivity transfer, this function is no longer needed.

controllers/configmap_controller.go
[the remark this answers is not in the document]
saifshaikh48: I suggest isValidConfigMap --> isWatchNamespaceConfigMap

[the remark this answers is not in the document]
jrvaldes: Agree 💯 , but I'm skeptical to bring more unrelated changes into this PR.

saifshaikh48 (commented):
The refactoring looks good to me. I am fine with the decision to drop the kubelet CA removal test since the rotation test covers the same code path (though could you confirm this @jrvaldes?)

pkg/certificates/certificates.go
```diff
@@ -0,0 +1,168 @@
...
+		if strings.Contains(c.Subject.String(), subjectLike) && c.NotAfter.After(maxNotAfter) {
+			// update max
+			maxNotAfter = c.NotAfter
+			// update block with the longest validity
+			maxBlock = b
+		}
+	}
+	if maxNotAfter.IsZero() {
+		// not found
+		return nil, errors.Errorf("cannot find certificate in bundle with suject like: %s", subjectLike)
```
saifshaikh48, line 135: nit spelling: suject --> subject

jrvaldes: Good catch, updated ✅

jrvaldes:
> The refactoring looks good to me. I am fine with the decision to drop the kubelet CA removal test since the rotation test covers the same code path (though could you confirm this @jrvaldes?)

Thanks for highlighting this, and the answer depends on what "code path" are you referring to.

The initially proposed kubelet CA removal test attempted to cover the scenario where the expired certificated is removed from the ConfigMap.

On the other hand, the kubelet CA rotation test triggers a CA rotation and creates a new certificate, leaving two valid (not expired yet) certificates in the ConfigMap.

pkg/certificates/certificates.go
```diff
@@ -0,0 +1,168 @@
...
+	// created during the cluster installation, where the initial kubelet CA is valid only for the first year.
+	kubeAPIServerInitialCAConfigMapName = "initial-kube-apiserver-server-ca"
+)
+
+// ReconcileCAsConfigMaps merges the given CA ConfigMaps for the specified subject
+func ReconcileCAsConfigMaps(initialCAConfigMap, currentCAConfigMap *core.ConfigMap, subject string) ([]byte, error) {
+	if subject == "" {
+		return nil, errors.New("subject cannot be empty")
+	}
+	// extract CAs form initial CA ConfigMap
```
alinaryan, line 40: s/form/from/

alinaryan (commented):
Thanks Jose! PTAL at my comments:

controllers/configmap_controller.go
```diff
@@ -381,3 +387,10 @@ func (r *ConfigMapReconciler) EnsureServicesConfigMapExists() error {
 	}
 	return nil
 }
+
+// isKubeAPIServerServingCAConfigMap returns true if the provided object is the ConfigMap that contains the
+// CA for the kubelet to recognize the kube-apiserver client certificate
+func (r *ConfigMapReconciler) isKubeAPIServerServingCAConfigMap(obj client.Object) bool {
```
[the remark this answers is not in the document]
jrvaldes: No, it isn't implied. Could be any resource, but we leverage the namespace/name combination as a one-to-one mapping.

pkg/certificates/certificates.go
jrvaldes: Fixed

controllers/controllers.go
```diff
@@ -118,6 +119,50 @@ func (r *instanceReconciler) instanceFromNode(node *core.Node) (*instance.Info,
...
+	if err != nil {
+		return errors.Wrapf(err, "error creating instance for node %s", node.Name)
+	}
+	nodeConfig, err := nodeconfig.NewNodeConfig(r.k8sclientset, r.clusterServiceCIDR, r.vxlanPort, winInstance,
+		r.signer, nil, nil, r.platform, r.dockerRuntime)
+	if err != nil {
+		return errors.Wrapf(err, "error creating nodeConfig for instance %s", winInstance.Address)
+	}
+	r.log.Info("updating kubelet CA client certificates in", "node", node.Name)
+	return nodeConfig.Windows.UpdateKubeletClientCA(contents)
```
sebsoto, line 134: I'm on the fence about the instanceReconciler directly calling a `Windows` function
but its probably fine :)

sebsoto (changes requested):

pkg/windows/windows.go
```diff
@@ -264,9 +265,17 @@ func (vm *windows) EnsureFile(file *payload.FileInfo, remoteDir string) error {
...
+		return errors.Wrapf(err, "error opening %s file to be transferred", file.Path)
+	}
+	defer func() {
+		if err := f.Close(); err != nil {
+			vm.log.Error(err, "error closing local file", "file", file.Path)
+		}
+	}()
 	vm.log.V(1).Info("copy", "local file", file.Path, "remote dir", remoteDir)
-	if err := vm.interact.transfer(file.Path, remoteDir); err != nil {
+	if err := vm.interact.transfer(f, file.Path, remoteDir); err != nil {
```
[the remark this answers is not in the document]
jrvaldes: Addressed.

pkg/certificates/certificates.go
```diff
@@ -0,0 +1,168 @@
...
+
+	// configNamespace is the namespace that contains the initial kubelet CA ConfigMap
+	configNamespace = "openshift-config"
+	// kubeAPIServerInitialCAConfigMapName is the name of the ConfigMap that contains the initial CA certificates
+	// created during the cluster installation, where the initial kubelet CA is valid only for the first year.
+	kubeAPIServerInitialCAConfigMapName = "initial-kube-apiserver-server-ca"
+)
+
+// ReconcileCAsConfigMaps merges the given CA ConfigMaps for the specified subject
+func ReconcileCAsConfigMaps(initialCAConfigMap, currentCAConfigMap *core.ConfigMap, subject string) ([]byte, error) {
```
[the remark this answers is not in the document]
jrvaldes: Makes sense, renamed to `MergeCAsConfigMaps()`

controllers/controllers.go
jrvaldes: Can you add more color to that?

pkg/certificates/certificates_test.go
```diff
@@ -0,0 +1,598 @@
...
+
+import (
+	"reflect"
+	"testing"
+
+	core "k8s.io/api/core/v1"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestReconcileCAsConfigMaps(t *testing.T) {
```
sebsoto, line 11: name doesnt match function being tested

pkg/windows/windows.go
```diff
@@ -507,6 +539,22 @@ func (vm *windows) ConfigureAzureCloudNodeManager(nodeName string) error {
 	return nil
 }
 
+// UpdateKubeletClientCA updates the kubelet client CA certificate file in the Windows node. No service restart or
+// reboot required, kubelet detects the changes in the file system and use the new CA certificate. The file is replaced
+// if and only if it does not exist or there is a checksum mismatch.
+func (vm *windows) UpdateKubeletClientCA(contents []byte) error {
```
sebsoto, line 545: To me, this needs to be a NodeConfig method as it is not performing a generic Windows VM operation, and instead performing an operation directly related to ensuring the health of a Windows Node.

sebsoto (changes requested):

controllers/controllers.go
sebsoto: See https://github.com/openshift/windows-machine-config-operator/pull/1013#discussion_r852109570
I put more detail in that comment, and thinking about it more i think the change is necessary

pkg/windows/windows.go
```diff
@@ -152,8 +152,9 @@ type Windows interface {
 	// to the Windows VM if it is not present or if it has the incorrect contents. The remote directory is created if it
 	// does not exist.
 	EnsureFile(*payload.FileInfo, string) error
-	// FileExists returns true if a specific file exists at the given path on the Windows VM
-	FileExists(string) (bool, error)
+	// FileExists returns true if a specific file exists at the given path and checksum on the Windows VM. Set an
+	// empty checksum (checksum == "") to disable checksum check.
+	FileExists(string, string) (bool, error)
```
selansen, line 166: Any reason to add checksum validation here? Can't we have a separate API for that? Function name and what it does is not matching.

controllers/configmap_controller.go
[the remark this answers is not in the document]
selansen: I am fine with keeping the name as is it:  Just don't want to have too many changes that are not related to actual change.
I already see filename exist changes that can be avoided from the previous commits.

pkg/windows/windows.go
jrvaldes: Roger that!

pkg/windows/windows.go
jrvaldes: > Function name and what it does is not matching.

The function name describes the goal and the function arguments the configuration, with `path` and `checksum.


> Can't we have a separate API for that? 

I don't anticipate any use case for it.


> Any reason to add checksum validation here? 

Yes, in this context,  the `checksum` validation is tightly-coupled with the existence of the file in the given `path`, ensuring the expected content is present in the file.

test/e2e/certificates_test.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// windowsJournalUnits holds the list of units used to fetch the journal logs in Windows.
+// Must be valid `LogName` for Get-WinEvent. In Windows, use the `Get-WinEvent -ListLog *` cmdlet gets information about
+// available logs. In this case, an arbitrary selection was made.
+var windowsJournalUnits = []string{"Microsoft-Windows-WMI"}
+
+// checkJournalLogInWindowsNodes checks journal logs in Windows nodes are available. If error, retries every retry.Interval
+// until retry.Timeout is reached.
+func checkJournalLogInWindowsNodes() error {
```
sebsoto, line 72: please change this to EventLog as journald is not a concept in windows

test/e2e/certificates_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// wait 5 seconds for the new certificate and add 3 seconds per node to allow
+	// WMCO to update the kubelet-ca.crt file on each node
+	nodeCount := len(gc.allNodes())
+	time.Sleep(time.Duration(5+3*nodeCount) * time.Second)
+
+	// node logs must be available after rotation
+	err = checkJournalLogInWindowsNodes()
+	assert.NoError(t, err, "error checking journal logs in Windows nodes after kubelet CA rotation")
+}
```
sebsoto, line 52: I don't understand how this test works
how does checking that the Windows event log is readable indicate that the cert rotation correctly occurred?

sebsoto: I feel that this test is not directly testing the functionality being added, and could potentially lead to false positives/negatives
i'd prefer just checking the certificate file on the Windows VM to ensure that the new cert is present within it.

test/e2e/certificates_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+// isCertificateInCABundle returns True when the CA bundle file in the given Windows node contains certificate content,
+// False otherwise.
+func (tc *testContext) isCertificateInCABundle(node *core.Node, content string) (bool, error) {
+	addr, err := controllers.GetAddress(node.Status.Addresses)
+	if err != nil {
+		return false, err
+	}
+	// CA bundle location in Windows node. i.e. windows.GetK8sDir() + nodeconfig.KubeletClientCAFilename
+	caBundlePath := "C:\\k\\kubelet-ca.crt"
```
saifshaikh48, line 91: Can we make this string from the constants described in the comment so we don’t have to maintain it in another place?

test/e2e/certificates_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/openshift/windows-machine-config-operator/controllers"
+	"github.com/openshift/windows-machine-config-operator/pkg/certificates"
```
sebsoto, line 11: These should be in a separate import group

test/e2e/certificates_test.go
```diff
@@ -0,0 +1,102 @@
...
+// able to use it.
+func (tc *testContext) testKubeletCARotation(t *testing.T) {
+	// force CA rotation
+	err := tc.forceKubeApiServerCertificateRotation()
+	require.NoError(t, err)
+
+	// wait 5 seconds for the new certificate and add 5 seconds per node to allow
+	// WMCO to update the kubelet-ca.crt file on each node
+	nodeCount := len(gc.allNodes())
+	time.Sleep(time.Duration(5+5*nodeCount) * time.Second)
```
sebsoto, line 47: This is going to be prone to flaking, I'd prefer to see a retry loop when checking each of the Nodes

as in a retry within the sub test `t.Run("node/"+winNode.Name, func(t *testing.T) {`

test/e2e/certificates_test.go
```diff
@@ -0,0 +1,102 @@
...
+	caBundlePath := "C:\\k\\kubelet-ca.crt"
+	// PowerShell command to find the given content in the file
+	command := fmt.Sprintf("Get-Content -Raw -Path %s | Select-String -SimpleMatch -Quiet -Pattern \\\"%s\\\"",
+		caBundlePath, content)
+	// invoke command
+	out, err := tc.runPowerShellSSHJob("kubelet-ca-bundle-content", command, addr)
+	if err != nil {
+		return false, err
+	}
+	// trim spaces and return True iff last command output is "True"
```
sebsoto, line 100: iff -> if

test/e2e/certificates_test.go
```diff
@@ -0,0 +1,102 @@
...
+func (tc *testContext) isCertificateInCABundle(node *core.Node, content string) (bool, error) {
+	addr, err := controllers.GetAddress(node.Status.Addresses)
+	if err != nil {
+		return false, err
+	}
+	// CA bundle location in Windows node. i.e. windows.GetK8sDir() + nodeconfig.KubeletClientCAFilename
+	caBundlePath := "C:\\k\\kubelet-ca.crt"
+	// PowerShell command to find the given content in the file
+	command := fmt.Sprintf("Get-Content -Raw -Path %s | Select-String -SimpleMatch -Quiet -Pattern \\\"%s\\\"",
+		caBundlePath, content)
```
sebsoto, line 94: I'd prefer we make these powershell commands as simple as possible, as they are harder to tweak and maintain.
Lets just grab the file contents using powershell and do the string comparison here in go.

sebsoto (changes requested):

test/e2e/certificates_test.go
jrvaldes: Done.

test/e2e/certificates_test.go
jrvaldes: Correct, I missed that.

test/e2e/certificates_test.go
jrvaldes: Addressed.

test/e2e/certificates_test.go
jrvaldes: Indeed, I had brought the comparison to this execution thread already as a walkaround for the PowerShell issues I was facing.

test/e2e/certificates_test.go
```diff
@@ -0,0 +1,120 @@
...
+				node.Name, addr, retry.Interval, err)
+			return false, nil
+		}
+		// check certificate content in the CA bundle
+		if strings.Contains(output, content) {
+			// found
+			return true, nil
+		}
+		// certificate not found, invalid CA bundle
+		return false, errors.Errorf("invalid CA bundle in node %s with address %s", node.Name, addr)
```
sebsoto, line 116: I would think you would want to retry here wouldn't you?
Its possible WMCO has not been able to copy the cert over yet for various reasons

Otherwise this function is only retrying in the case that the ca file wasnt fetchable, which it always should be, except for when there are network errors

sebsoto (changes requested):
Thanks for the changes!
Last things, otherwise LGTM!

test/e2e/certificates_test.go
saifshaikh48: Agreed. Retry should be here and not in above case.

jrvaldes: Agreed ^ 2. Thanks for pointing this out. I refactored to `waitForKubeletCACertificateInNode(node *core.Node) error `. PTAL

test/e2e/certificates_test.go
```diff
@@ -0,0 +1,112 @@
...
+	command := fmt.Sprintf("Get-Content -Raw -Path %s", caBundlePath)
+	// wait retry.Interval and verify the CA bundle content, try if needed
+	return wait.Poll(retry.Interval, retry.Timeout, func() (bool, error) {
+		// invoke command
+		bundleContent, err := tc.runPowerShellSSHJob("kubelet-ca-bundle-content", command, addr)
+		if err != nil {
+			// retry
+			log.Printf("error fetching CA bundle in node %s with address %s, retrying in %s. %v",
+				node.Name, addr, retry.Interval, err)
+			return false, nil
```
saifshaikh48, line 90: Is there a reason to retry if there's a powershell job failure? The command is simple enough that it is expected to work, and if it doesn't we should return error.
```suggestion
			return false, errors.Wrapf(err, "error fetching CA bundle in node %s with address %s", node.Name, addr)
```

sebsoto (changes requested):

test/e2e/certificates_test.go
```diff
@@ -0,0 +1,112 @@
...
+	addr, err := controllers.GetAddress(node.Status.Addresses)
+	if err != nil {
+		return err
+	}
+	// CA bundle location in Windows node. i.e. "C:\k\kubelet-ca.crt"
+	caBundlePath := windows.GetK8sDir() + nodeconfig.KubeletClientCAFilename
+	// PowerShell command to fetch content in the file
+	command := fmt.Sprintf("Get-Content -Raw -Path %s", caBundlePath)
+	// wait retry.Interval and verify the CA bundle content, try if needed
+	return wait.Poll(retry.Interval, retry.Timeout, func() (bool, error) {
```
sebsoto, line 83: wait.PollImmediate should be used here if possible

test/e2e/certificates_test.go
jrvaldes: A network/connectivity issue, for instance. A retry may be helpful.

test/e2e/certificates_test.go
```diff
@@ -0,0 +1,112 @@
...
+			certificates.KubeAPIServerServingCAConfigMapName, meta.GetOptions{})
+		if err != nil {
+			return false, errors.Wrap(err, "error getting kubelet client CA ConfigMap")
+		}
+		// parse bundle from ConfigMap
+		kubeletCABytes, err := certificates.GetCAsFromConfigMap(cm, certificates.CABundleKey)
+		if err != nil {
+			return false, errors.Wrap(err, "error parsing CA bundle from ConfigMap")
+		}
+		kubeletCAString := strings.TrimSpace(string(kubeletCABytes))
```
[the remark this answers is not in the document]
jrvaldes: It's a trade-off, to fetch the kubelet CA ConfigMap first and pass it along, we need to wait for the kube-apiserver-operator to update the ConfigMap with the new certificate (the one resulting from the rotation with the longest validity). I do have an "estimation" for that, about 5 seconds. It may and will vary depending on many factors, cluster's overload, etc.

To reduce any possible [flakes](https://github.com/openshift/windows-machine-config-operator/pull/1013#discussion_r854534019), I decided to move this code here. Yes, It adds an extra API call per Windows node but is worth it compared to relying on reference data that may not be the latest.

test/e2e/certificates_test.go
jrvaldes: Not sure about that, in this case. I'm counting on the initial `retry.Interval` wait, for the kube-apiserver-operator to update the ConfigMap with the new certificate and WMCO to copy the new certificate over the Windows node.

test/e2e/certificates_test.go
saifshaikh48: Fair enough, fine with me

test/e2e/certificates_test.go
sebsoto: If it fails on the first try, it will just retry.
but in the case where it succeeds on the first try it will save our CI an amount of time == retry.Interval * nodecount

jrvaldes: There is a high probability to fail the first attempt because the file is not up to date, that is number one. 

Second, the first SSH call to fetch the content of `C:\k\kubelet-ca.crt` may block WMCO from performing the I/O operation required to copy the new certificate, and/or the kubelet service running in the node from dynamically loading the file.

To optimize the amount of time CI is idle, I feel we can reduce the retry time (currently `retry.Interval` => 15s)  in half, down to 7s. WDYT?

selansen (commented):
/lgtm

PR #1020, opened 2022-04-05
WINC-716: WMCO removes unneeded service ConfigMaps

saifshaikh48:
This PR adds the responsibility of removing outdated services ConfigMaps.
This will be particularly useful during upgrades -- the operator will clean up old 
services configs to ensure the source of truth is tied to the current operator version.

### DIFF
--- a/controllers/configmap_controller.go
+++ b/controllers/configmap_controller.go
@@ -316,7 +316,7 @@ func (r *ConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&core.ConfigMap{}, builder.WithPredicates(configMapPredicate)).
 		Watches(&source.Kind{Type: &core.Node{}}, handler.EnqueueRequestsFromMapFunc(r.mapToConfigMap),
-			builder.WithPredicates(windowsNodePredicate(true))).
+			builder.WithPredicates(outdatedWindowsNodePredicate(true))).
 		Complete(r)
 }
 
--- a/controllers/configmap_controller.go
+++ b/controllers/configmap_controller.go
@@ -290,10 +290,17 @@ func hasAssociatedInstance(nodeAddresses []core.NodeAddress, instances []*instan
 	return false, nil
 }
 
-// mapToConfigMap fulfills the MapFn type, while always returning a request to the windows-instance ConfigMap
-func (r *ConfigMapReconciler) mapToConfigMap(_ client.Object) []reconcile.Request {
+// mapToInstancesConfigMap fulfills the MapFn type, while always returning a request to the windows-instance ConfigMap
+func (r *ConfigMapReconciler) mapToInstancesConfigMap(_ client.Object) []reconcile.Request {
 	return []reconcile.Request{{
 		NamespacedName: kubeTypes.NamespacedName{Namespace: r.watchNamespace, Name: wiparser.InstanceConfigMap},
+	}}
+}
+
+// mapToServicesConfigMap fulfills the MapFn type, while always returning a request to the windows-services ConfigMap
+func (r *ConfigMapReconciler) mapToServicesConfigMap(_ client.Object) []reconcile.Request {
+	return []reconcile.Request{{
+		NamespacedName: kubeTypes.NamespacedName{Namespace: r.watchNamespace, Name: servicescm.Name},
 	}}
 }
 
@@ -315,8 +322,10 @@ func (r *ConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	}
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&core.ConfigMap{}, builder.WithPredicates(configMapPredicate)).
-		Watches(&source.Kind{Type: &core.Node{}}, handler.EnqueueRequestsFromMapFunc(r.mapToConfigMap),
+		Watches(&source.Kind{Type: &core.Node{}}, handler.EnqueueRequestsFromMapFunc(r.mapToInstancesConfigMap),
 			builder.WithPredicates(outdatedWindowsNodePredicate(true))).
+		Watches(&source.Kind{Type: &core.Node{}}, handler.EnqueueRequestsFromMapFunc(r.mapToServicesConfigMap),
+			builder.WithPredicates(windowsNodeVersionChangePredicate())).
 		Complete(r)
 }
 
--- a/controllers/configmap_controller.go
+++ b/controllers/configmap_controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	"net"
+	"strings"
 
 	config "github.com/openshift/api/config/v1"
 	"github.com/pkg/errors"
@@ -47,6 +48,7 @@ import (
 	"github.com/openshift/windows-machine-config-operator/pkg/servicescm"
 	"github.com/openshift/windows-machine-config-operator/pkg/signer"
 	"github.com/openshift/windows-machine-config-operator/pkg/wiparser"
+	"github.com/openshift/windows-machine-config-operator/version"
 )
 
 //+kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;patch;delete
@@ -156,6 +158,10 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context,
 // reconcileServices uses the data within the services ConfigMap to ensure WMCO-managed Windows services on
 // Windows Nodes have the expected configuration and are in the expected state
 func (r *ConfigMapReconciler) reconcileServices(ctx context.Context, windowsServices *core.ConfigMap) error {
+	if err := r.removeOutdatedServicesConfigMaps(ctx); err != nil {
+		return err
+	}
+
 	// If a ConfigMap with invalid values is found, WMCO will delete and recreate it with proper values
 	if _, err := servicescm.Parse(windowsServices.Data); err != nil {
 		// Deleting will trigger an event for the configmap_controller, which will re-create a proper ConfigMap
@@ -168,6 +174,62 @@ func (r *ConfigMapReconciler) reconcileServices(ctx context.Context, windowsServ
 	}
 	// TODO: actually react to changes to the services ConfigMap
 	return nil
+}
+
+// removeOutdatedServicesConfigMaps deletes any outdated services ConfigMaps, if all nodes have moved past that version
+func (r *ConfigMapReconciler) removeOutdatedServicesConfigMaps(ctx context.Context) error {
+	nodes := &core.NodeList{}
+	if err := r.client.List(ctx, nodes, client.MatchingLabels{core.LabelOSStable: "windows"}); err != nil {
+		return err
+	}
+	versionAnnotations := getVersionAnnotations(nodes.Items)
+
+	servicesConfigMaps, err := r.listServiceConfigMaps(ctx)
+	if err != nil {
+		return errors.Wrapf(err, "unable to retrieve list of services ConfigMaps")
+	}
+	for _, cm := range servicesConfigMaps {
+		cmVersion := strings.TrimPrefix(cm.Name, servicescm.NamePrefix)
+		if isTiedToRelevantVersion(cmVersion, versionAnnotations) {
+			continue
+		}
+		// Remove any services ConfigMap tied to a WMCO version that no Windows nodes are at anymore
+		if err := r.client.Delete(ctx, &cm); err != nil {
+			return errors.Wrapf(err, "could not delete outdated services ConfigMap %s", cm.Name)
+		}
+		r.log.Info("Deleted outdated resource", "ConfigMap",
+			kubeTypes.NamespacedName{Namespace: r.watchNamespace, Name: cm.Name})
+	}
+	return nil
+}
+
+// isTiedToRelevantVersion checks if the given version is the current WMCO version or is in the given map of versions
+func isTiedToRelevantVersion(v string, versions map[string]struct{}) bool {
+	if v == version.Get() {
+		// The current WMCO version is always considered relevant
+		return true
+	}
+	for version := range versions {
+		if v == version {
+			return true
+		}
+	}
+	return false
+}
+
+// listServiceConfigMaps returns a list of all windows-services ConfigMaps in the operator namespace
+func (r *ConfigMapReconciler) listServiceConfigMaps(ctx context.Context) ([]core.ConfigMap, error) {
+	watchNamespaceCMs := &core.ConfigMapList{}
+	if err := r.client.List(ctx, watchNamespaceCMs, &client.ListOptions{Namespace: r.watchNamespace}); err != nil {
+		return nil, err
+	}
+	servicesConfigMaps := []core.ConfigMap{}
+	for _, cm := range watchNamespaceCMs.Items {
+		if strings.HasPrefix(cm.Name, servicescm.NamePrefix) {
+			servicesConfigMaps = append(servicesConfigMaps, cm)
+		}
+	}
+	return servicesConfigMaps, nil
 }
 
 // reconcileNodes corrects the discrepancy between the "expected" instances, and the "actual" Node list
--- a/controllers/controllers.go
+++ b/controllers/controllers.go
@@ -214,6 +214,17 @@ func outdatedWindowsNodePredicate(byoh bool) predicate.Funcs {
 
 }
 
+// getVersionAnnotations returns a map whose keys are the WMCO versions that have configured any Windows nodes
+func getVersionAnnotations(nodes []core.Node) map[string]struct{} {
+	versions := make(map[string]struct{})
+	for _, node := range nodes {
+		if versionAnnotation, present := node.Annotations[metadata.VersionAnnotation]; present {
+			versions[versionAnnotation] = struct{}{}
+		}
+	}
+	return versions
+}
+
 // isValidWindowsNode returns true if the node object has the Windows label and the BYOH
 // label present for only the BYOH nodes based on the value of the byoh boolean parameter.
 func isValidWindowsNode(o client.Object, byoh bool) bool {
--- a/controllers/controllers.go
+++ b/controllers/controllers.go
@@ -155,6 +155,34 @@ func (r *instanceReconciler) deconfigureInstance(node *core.Node) error {
 	return nil
 }
 
+// windowsNodeVersionChangePredicate returns a predicate whose filter catches Windows nodes that indicate a version
+// change either through deletion away from an old version or creation/update to the latest WMCO version
+func windowsNodeVersionChangePredicate() predicate.Funcs {
+	return predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			// catch Machine-backed Windows node upgrades as they are re-created
+			return e.Object.GetLabels()[core.LabelOSStable] == "windows" &&
+				e.Object.GetAnnotations()[metadata.VersionAnnotation] == version.Get()
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			// catch BYOH Windows node upgrades to the current WMCO version as they are re-configured in place
+			return e.ObjectNew.GetLabels()[core.LabelOSStable] == "windows" &&
+				(e.ObjectOld.GetAnnotations()[metadata.VersionAnnotation] !=
+					e.ObjectNew.GetAnnotations()[metadata.VersionAnnotation]) &&
+				e.ObjectNew.GetAnnotations()[metadata.VersionAnnotation] == version.Get()
+		},
+		GenericFunc: func(e event.GenericEvent) bool {
+			return e.Object.GetLabels()[core.LabelOSStable] == "windows" &&
+				e.Object.GetAnnotations()[metadata.VersionAnnotation] == version.Get()
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			// catch if a node stuck at an older WMCO version is deleted
+			return e.Object.GetLabels()[core.LabelOSStable] == "windows" &&
+				e.Object.GetAnnotations()[metadata.VersionAnnotation] != version.Get()
+		},
+	}
+}
+
 // outdatedWindowsNodePredicate returns a predicate which filters out all node objects that are not up-to-date Windows
 // nodes. Up-to-date refers to the version annotation and public key hash annotations.
 // If BYOH is true, only BYOH nodes will be allowed through, else no BYOH nodes will be allowed.
--- a/controllers/controllers.go
+++ b/controllers/controllers.go
@@ -155,9 +155,10 @@ func (r *instanceReconciler) deconfigureInstance(node *core.Node) error {
 	return nil
 }
 
-// windowsNodePredicate returns a predicate which filters out all node objects that are not Windows nodes.
+// outdatedWindowsNodePredicate returns a predicate which filters out all node objects that are not up-to-date Windows
+// nodes. Up-to-date refers to the version annotation and public key hash annotations.
 // If BYOH is true, only BYOH nodes will be allowed through, else no BYOH nodes will be allowed.
-func windowsNodePredicate(byoh bool) predicate.Funcs {
+func outdatedWindowsNodePredicate(byoh bool) predicate.Funcs {
 	return predicate.Funcs{
 		CreateFunc: func(e event.CreateEvent) bool {
 			return isValidWindowsNode(e.Object, byoh) &&
--- a/test/e2e/upgrade_test.go
+++ b/test/e2e/upgrade_test.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -15,6 +16,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 
 	"github.com/openshift/windows-machine-config-operator/pkg/metadata"
+	"github.com/openshift/windows-machine-config-operator/pkg/servicescm"
 )
 
 const (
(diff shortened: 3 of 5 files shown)


controllers/configmap_controller.go
```diff
@@ -156,6 +159,10 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context,
 // reconcileServices uses the data within the services ConfigMap to ensure WMCO-managed Windows services on
 // Windows Nodes have the expected configuration and are in the expected state
 func (r *ConfigMapReconciler) reconcileServices(ctx context.Context, windowsServices *core.ConfigMap) error {
+	if err := r.removeOutdatedServicesConfigMaps(ctx); err != nil {
```
selansen, line 162: Should we have any checks before calling this function as this will be called every time reconcile function gets called? Anyways the upgrade doesn't happen frequently, right?

controllers/configmap_controller.go
```diff
@@ -170,6 +177,53 @@ func (r *ConfigMapReconciler) reconcileServices(ctx context.Context, windowsServ
...
+		// Nothing to clean up
+		return nil
+	}
+
+	nodes := &core.NodeList{}
+	if err := r.client.List(ctx, nodes, client.MatchingLabels{core.LabelOSStable: "windows"}); err != nil {
+		return err
+	}
+	// Check if there are any nodes that require upgrade
+	if allNodesUpToDate(nodes) {
```
selansen, line 203: Are we not going to have any else condition? At least logging some info would help us to debug if there is an issue with particular nodes that are not up to date.

controllers/configmap_controller.go
saifshaikh48: What kind of check are you thinking?

This has to be done every time we reconcile the services ConfigMap, which is relatively very infrequently since it cannot be updated, and it is only created by WMCO on startup/upgrade.

saifshaikh48: Note that we don't actually delete anything unless both:
- there is at least one ConfigMap in watchNamespace that has the name "windows-services-*" where * != current WMCO version
- all Windows nodes are up to date
And these two conditions are lazy evaluated in `removeOutdatedServicesConfigMaps()`

controllers/configmap_controller.go
```diff
@@ -170,6 +177,53 @@ func (r *ConfigMapReconciler) reconcileServices(ctx context.Context, windowsServ
...
+	return nil
+}
+
+// allNodesUpToDate returns true if all nodes in the given list have been configured by the current operator version
+func allNodesUpToDate(nodes *core.NodeList) bool {
+	for _, node := range nodes.Items {
+		if _, present := node.Annotations[metadata.VersionAnnotation]; present {
+			// If the version annotation doesn't match the current operator version, the machine needs upgrade
+			if node.Annotations[metadata.VersionAnnotation] != version.Get() {
+				return false
```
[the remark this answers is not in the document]
saifshaikh48: made this method unit testable and return error instead of bool so we know which node is not up to date

controllers/configmap_controller.go
```diff
@@ -170,6 +176,43 @@ func (r *ConfigMapReconciler) reconcileServices(ctx context.Context, windowsServ
...
+// removeOutdatedServicesConfigMaps deletes any outdated services ConfigMaps, if all nodes have moved past that version
+func (r *ConfigMapReconciler) removeOutdatedServicesConfigMaps(ctx context.Context) error {
+	watchNamespaceCMs := &core.ConfigMapList{}
+	if err := r.client.List(ctx, watchNamespaceCMs, &client.ListOptions{Namespace: r.watchNamespace}); err != nil {
+		return err
+	}
+	configMapsToDelete := []core.ConfigMap{}
+	for _, cm := range watchNamespaceCMs.Items {
+		// A possibly outdated services ConfigMap has the name "windows-services-*" where * != current WMCO version
+		if strings.HasPrefix(cm.Name, servicescm.NamePrefix) && cm.Name != servicescm.Name {
```
jrvaldes, line 188: Looks like the `servicescm.NamePrefix` part of the `cm.Name` slice is evaluated twice. Consider using `strings.HasPrefix()` and `strings.HasSuffix()` instead.

controllers/configmap_controller.go
```diff
@@ -170,6 +176,43 @@ func (r *ConfigMapReconciler) reconcileServices(ctx context.Context, windowsServ
...
+	if len(configMapsToDelete) == 0 {
+		// Nothing to clean up
+		return nil
+	}
+
+	nodes := &core.NodeList{}
+	if err := r.client.List(ctx, nodes, client.MatchingLabels{core.LabelOSStable: "windows"}); err != nil {
+		return err
+	}
+	// Check if there are any nodes that require upgrade to make sure configMapsToDelete are actually outdated
```
jrvaldes, line 201: I feel this logic is out of the scope of `removeOutdatedServicesConfigMaps()`, first we should check if all the preconditions are met, then invoke the cleanup

jrvaldes (commented):
Nice work @saifshaikh48. Comments added. PTAL

test/e2e/validation_test.go
```diff
@@ -478,15 +478,21 @@ func testServicesConfigMap(t *testing.T) {
 	servicesConfigMapName := servicescm.NamePrefix + operatorVersion
 
 	// Ensure the windows-services ConfigMap exists in the cluster
-	_, err = tc.client.K8s.CoreV1().ConfigMaps(tc.namespace).Get(context.TODO(),
-		servicesConfigMapName, meta.GetOptions{})
-	require.NoErrorf(t, err, "error retrieving ConfigMap: %s", servicesConfigMapName)
+	t.Run("Services ConfigMap existance", func(t *testing.T) {
+		_, err = tc.client.K8s.CoreV1().ConfigMaps(tc.namespace).Get(context.TODO(), servicesConfigMapName,
+			meta.GetOptions{})
+		require.NoErrorf(t, err, "error retrieving ConfigMap: %s", servicesConfigMapName)
```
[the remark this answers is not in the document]
saifshaikh48: The error would get printed out in test results anyway, example:
```
        	Error Trace:	validation_test.go:484
        	Error:      	       	            		...
        	Test:       	TestWMCO/upgrade/Services_ConfigMap_existance
        	Messages:   	error retrieving ConfigMap: windows-instances-5.0.0-f21e022-dirty
```

controllers/configmap_controller.go
```diff
@@ -170,6 +176,43 @@ func (r *ConfigMapReconciler) reconcileServices(ctx context.Context, windowsServ
...
+	if err := allNodesUpToDate(nodes.Items, version.Get()); err != nil {
+		r.log.V(1).Info("Waiting for all nodes to be up-to-date", "error", err.Error())
+		return nil
+	}
```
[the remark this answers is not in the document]
saifshaikh48: NVM, the current WMCO version will always be the one configuring

controllers/controllers.go
```diff
@@ -198,6 +198,22 @@ func isValidWindowsNode(o client.Object, byoh bool) bool {
 	return true
 }
 
+// allNodesUpToDate checks if all nodes in the given list have been configured by the given operator version
+func allNodesUpToDate(nodes []core.Node, expectedVersion string) error {
```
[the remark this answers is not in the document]
saifshaikh48: Realized after refactoring the logic that this func isn't needed. Will delete after a round of reviews.

controllers/controllers.go
```diff
@@ -198,6 +198,15 @@ func isValidWindowsNode(o client.Object, byoh bool) bool {
 	return true
 }
 
+// getWindowsNodeVersions returns a list of all the WMCO versions that have last configured any nodes
+func getWindowsNodeVersions(nodes []core.Node) []string {
+	versions := []string{}
+	for _, node := range nodes {
+		versions = append(versions, node.Annotations[metadata.VersionAnnotation])
+	}
+	return versions
```
saifshaikh48, line 207: opted to return a string slice here instead of a map since it may be useful to see how many nodes have the version annotation, e.g. if len(versions) != len(windoesNodes) then we know a node is currently being configured.

saifshaikh48:
/test aws-e2e-upgrade

services CM tests pass locally (both creation and upgrade suites)

controllers/controllers.go
```diff
@@ -198,6 +198,15 @@ func isValidWindowsNode(o client.Object, byoh bool) bool {
 	return true
 }
 
+// getWindowsNodeVersions returns a list of all the WMCO versions that have last configured any nodes
+func getWindowsNodeVersions(nodes []core.Node) []string {
```
jrvaldes, line 202: I don't think this is necesary (the loop). However, I can see the benefit of the logic encapsulation for extracting the WMCO version from a node. 

In addition, `getWindowsNodeVersions` may read as "returns the Windows Server versions for all nodes".

Consider: 
```suggestion
func extractWMCOVersion(node []core.Node) string {
```

controllers/configmap_controller.go
```diff
@@ -170,6 +176,60 @@ func (r *ConfigMapReconciler) reconcileServices(ctx context.Context, windowsServ
 	return nil
 }
 
+// removeOutdatedServicesConfigMaps deletes any outdated services ConfigMaps, if all nodes have moved past that version
+func (r *ConfigMapReconciler) removeOutdatedServicesConfigMaps(ctx context.Context) error {
+	nodes := &core.NodeList{}
+	if err := r.client.List(ctx, nodes, client.MatchingLabels{core.LabelOSStable: "windows"}); err != nil {
+		return err
+	}
+	nodeVersions := getWindowsNodeVersions(nodes.Items)
```
jrvaldes, line 185: No need to loop the nodes/version twice. You can move this into `isTiedToRelevantVersion(cm.Name, nodes)`

test/e2e/validation_test.go
```diff
@@ -545,22 +545,27 @@ func (tc *testContext) testInvalidServicesCM(cmName string) error {
...
-// waitForValidWindowsServicesConfigMap returns a reference to the ConfigMap that matches the given name.
-// If a ConfigMap with valid contents is not found within the time limit, an error is returned.
-func (tc *testContext) waitForValidWindowsServicesConfigMap(cmName string) (*core.ConfigMap, error) {
+// waitForValidWindowsServicesConfigMap waits for a ConfigMap by the given name to either be created or deleted,
+// depending on the value of the `deletion` flag. If the flag is false, it tries to retrieve and return a ConfigMap with
+// valid contents. Returns an error if it cannot be not found by the time limit. If the flag is true, it ensures that
+// the ConfigMap cannot be found within the WMCO namespace. Returns an error if it is still present at the time limit.
+func (tc *testContext) waitForValidWindowsServicesConfigMap(cmName string, deletion bool) (*core.ConfigMap, error) {
 	configMap := &core.ConfigMap{}
 	err := wait.Poll(retry.Interval, retry.ResourceChangeTimeout, func() (bool, error) {
```
jrvaldes, line 554: Is there a reason to wait `retry.Interval` and then get the ConfigMap? Why not `wait.PollImmediate()`?

test/e2e/validation_test.go
```diff
@@ -545,22 +545,27 @@ func (tc *testContext) testInvalidServicesCM(cmName string) error {
...
 }
 
-// waitForValidWindowsServicesConfigMap returns a reference to the ConfigMap that matches the given name.
-// If a ConfigMap with valid contents is not found within the time limit, an error is returned.
-func (tc *testContext) waitForValidWindowsServicesConfigMap(cmName string) (*core.ConfigMap, error) {
+// waitForValidWindowsServicesConfigMap waits for a ConfigMap by the given name to either be created or deleted,
+// depending on the value of the `deletion` flag. If the flag is false, it tries to retrieve and return a ConfigMap with
+// valid contents. Returns an error if it cannot be not found by the time limit. If the flag is true, it ensures that
+// the ConfigMap cannot be found within the WMCO namespace. Returns an error if it is still present at the time limit.
+func (tc *testContext) waitForValidWindowsServicesConfigMap(cmName string, deletion bool) (*core.ConfigMap, error) {
```
jrvaldes, line 552: Not sure about the introduction of the `deletion bool` here, sounds like you need a separated function to ensure the ConfigMap is not present.

controllers/controllers.go
saifshaikh48: I can get rid of this function/loop wholly by moving the logic into `isTiedToRelevantVersion()` as you suggested. Extracting the WMCO version from a node is a simple map access `node.Annotations[metadata.VersionAnnotation]` so no need to encapsulate

test/e2e/validation_test.go
saifshaikh48: Fair enough

test/e2e/validation_test.go
saifshaikh48: No reason, we can use `wait.PollImmediate()`

test/e2e/upgrade_test.go
```diff
@@ -8,13 +8,15 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
sebsoto, line 13: revert this change please

test/e2e/validation_test.go
```diff
@@ -549,7 +549,7 @@ func (tc *testContext) testInvalidServicesCM(cmName string) error {
 // If a ConfigMap with valid contents is not found within the time limit, an error is returned.
 func (tc *testContext) waitForValidWindowsServicesConfigMap(cmName string) (*core.ConfigMap, error) {
 	configMap := &core.ConfigMap{}
-	err := wait.Poll(retry.Interval, retry.ResourceChangeTimeout, func() (bool, error) {
+	err := wait.PollImmediate(retry.Interval, retry.ResourceChangeTimeout, func() (bool, error) {
```
sebsoto, line 552: this seems unrelated, please move to separate commit

sebsoto (changes requested):

controllers/configmap_controller.go
```diff
@@ -170,6 +177,58 @@ func (r *ConfigMapReconciler) reconcileServices(ctx context.Context, windowsServ
...
+			// Do not delete the services ConfigMap tied to the current WMCO version
+			continue
+		}
+		// Remove any services ConfigMap tied to a WMCO version that no Windows nodes are at anymore
+		if !isTiedToRelevantVersion(cm.Name, nodes.Items) {
+			if err := r.client.Delete(ctx, &cm); err != nil {
+				return errors.Wrapf(err, "could not delete outdated services ConfigMap %s", cm.Name)
+			}
+			r.log.Info("Deleted outdated resource", "ConfigMap",
+				kubeTypes.NamespacedName{Namespace: r.watchNamespace, Name: cm.Name})
```
sebsoto, line 201: Instead of doing this, which can result in going through all the node objects multiple times i suggest
going through the nodes once, and getting a map[string]struct{} indicating all the current versions
you can then use `strings.TrimPrefix()` to get the CM version, and check the map for a match.

test/e2e/validation_test.go
```diff
@@ -570,6 +570,26 @@ func (tc *testContext) waitForValidWindowsServicesConfigMap(cmName string) (*cor
...
+		_, err := tc.client.K8s.CoreV1().ConfigMaps(tc.namespace).Get(context.TODO(), cmName, meta.GetOptions{})
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				return true, nil
+			}
+			return false, errors.Wrapf(err, "error retrieving ConfigMap: %s", cmName)
+		}
+		// Retry if the resource is found
+		return false, nil
```
sebsoto, line 585: nit: identation can be reduced by doing the inverse error check
```
 _, err := tc.client.K8s.CoreV1().ConfigMaps(tc.namespace).Get(context.TODO(), cmName, meta.GetOptions{})
 if err == nil {
 	return false,nil
 }
 ...
 ```

File: controllers/configmap_controller.go [hunk 18]
```diff
@@ -374,11 +381,25 @@ func (r *ConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	}
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&core.ConfigMap{}, builder.WithPredicates(configMapPredicate)).
-		Watches(&source.Kind{Type: &core.Node{}}, handler.EnqueueRequestsFromMapFunc(r.mapToConfigMap),
+		Watches(&source.Kind{Type: &core.Node{}}, handler.EnqueueRequestsFromMapFunc(r.mapToInstancesConfigMap),
 			builder.WithPredicates(windowsNodePredicate(true))).
```
sebsoto, line 385: please rename this to something like `outdatedWindowsNodePredicate`

controllers/configmap_controller.go
```diff
@@ -374,11 +381,25 @@ func (r *ConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		Watches(&source.Kind{Type: &core.Node{}}, handler.EnqueueRequestsFromMapFunc(r.mapToServicesConfigMap),
+			// cannot use windowsNodePredicate as we want to map events Node regardless of the version annotation
+			builder.WithPredicates(anyWindowsNodePredicate())).
 		Complete(r)
 }
 
+// anyWindowsNodePredicate returns a predicate which filters out all node objects that are not Windows nodes.
+func anyWindowsNodePredicate() predicate.Funcs {
```
sebsoto, line 393: `anyWindowsNodePredicate` could probably just be named `windowsNodePredicate`

sebsoto (changes requested):

controllers/configmap_controller.go
```diff
@@ -170,6 +176,61 @@ func (r *ConfigMapReconciler) reconcileServices(ctx context.Context, windowsServ
...
+			return true
+		}
+	}
+	return false
+}
+
+// listServiceConfigMaps returns a list of all windows-services ConfigMaps in the operator namespace
+func (r *ConfigMapReconciler) listServiceConfigMaps(ctx context.Context) ([]core.ConfigMap, error) {
+	watchNamespaceCMs := &core.ConfigMapList{}
+	if err := r.client.List(ctx, watchNamespaceCMs, &client.ListOptions{Namespace: r.watchNamespace}); err != nil {
```
jrvaldes, line 222: Can you use [FieldSelector](https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/) in `client.ListOptions` prepare a query that fetches all `windows-services` ConfigMap?

controllers/controllers.go
```diff
@@ -155,9 +155,35 @@ func (r *instanceReconciler) deconfigureInstance(node *core.Node) error {
 	return nil
 }
 
-// windowsNodePredicate returns a predicate which filters out all node objects that are not Windows nodes.
+// windowsNodeUpgradePredicate returns a predicate whose filter catches only Windows node objects that signal upgrade
+// to the latest WMCO version through a change in the node version annotation
+func windowsNodeUpgradePredicate() predicate.Funcs {
```
jrvaldes, line 160: Something doesn't sound right. 

There is no indication in the implementation of this function that the predicates are targeting a Node object. Any resource with the `windows` label and the `metadata.VersionAnnotation` annotation satisfies the filter conditions, hence the function name `windowsNodeUpgradePredicate` is not quite accurate. 

Not quite sure, but adding `o.GetObjectKind().GroupVersionKind().Kind == "Node"` may be helpful.


Now, if the given conditions are guaranteed to uniquely identify a Windows node you should state that in the func description.

test/e2e/validation_test.go
```diff
@@ -478,15 +478,21 @@ func testServicesConfigMap(t *testing.T) {
...
+		_, err = tc.client.K8s.CoreV1().ConfigMaps(tc.namespace).Get(context.TODO(), servicesConfigMapName,
+			meta.GetOptions{})
+		assert.NoErrorf(t, err, "error ensuring ConfigMap %s exists", servicesConfigMapName)
+	})
 
-	err = tc.testServicesCMRegeneration(servicesConfigMapName)
-	require.NoErrorf(t, err, "error ensuring ConfigMap %s is re-created when deleted", servicesConfigMapName)
+	t.Run("Services ConfigMap re-creation", func(t *testing.T) {
+		err = tc.testServicesCMRegeneration(servicesConfigMapName)
+		assert.NoErrorf(t, err, "error ensuring ConfigMap %s is re-created when deleted", servicesConfigMapName)
```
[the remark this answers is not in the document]
saifshaikh48: Good catch, the test above it should also be a `require` as without the ConfigMap existing we can't test recreation, deletion, etc.

controllers/controllers.go
```diff
@@ -155,9 +155,35 @@ func (r *instanceReconciler) deconfigureInstance(node *core.Node) error {
...
+			return e.Object.GetLabels()[core.LabelOSStable] == "windows"
+		},
+	}
+}
+
+// outdatedWindowsNodePredicate returns a predicate which filters out all node objects that are not up-to-date Windows
+// nodes. Up-to-date refers to the version annotation and public key hash annotations.
 // If BYOH is true, only BYOH nodes will be allowed through, else no BYOH nodes will be allowed.
-func windowsNodePredicate(byoh bool) predicate.Funcs {
+func outdatedWindowsNodePredicate(byoh bool) predicate.Funcs {
```
jrvaldes, line 186: Did you considered extending  [isValidWindowsNode()](https://github.com/openshift/windows-machine-config-operator/blob/61e24e9f8fcb31a451ea1b2408c7382821f4a78f/controllers/controllers.go#L235)   instead?

controllers/configmap_controller.go
jrvaldes: Ideally, we should filter on the server-side. What about introducing an [annotation](https://github.com/openshift/windows-machine-config-operator/blob/61e24e9f8fcb31a451ea1b2408c7382821f4a78f/pkg/windows/windows.go#L92)?

controllers/controllers.go
saifshaikh48: I don't follow, isValidWindowsNode is a helper to this method

controllers/controllers.go
saifshaikh48: I see the idea behind this but given it the predicate is used only in context of watching node objects (and we only apply the version annotation to Windows node objects) I don't see it necessary to add a check on the object type. We follow the same principle in the [existing predicate function](https://github.com/openshift/windows-machine-config-operator/blob/master/controllers/controllers.go#L158-L160).

controllers/configmap_controller.go
jrvaldes: Correct, I'm fine with looping the result set.

alinaryan: I'm ok with how it is

controllers/controllers.go
saifshaikh48: Gotcha. We cannot extend `isValidWindowsNode()` as it's purpose is independent of the version annotation. See how it is used in this method a few line below (`VersionAnnotation != version.Get()` whereas the newly introduced predicate is looking for the inverse case, `VersionAnnotation == version.Get()`)

test/e2e/validation_test.go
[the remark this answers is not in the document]
saifshaikh48: Thought about this a bit more and since these are separate functions, a require will not stop the parent function `testServicesConfigMap`, only the subtests (each `t.Run` block). Though I may be mistaken.

> In case the ConfigMap re-creation fails, can WMCO still configure the instance without it?
Yep. Since WMCO does not need the services CM yet, I don't think we need to stop the e2e suite on failure here.

test/e2e/validation_test.go
```diff
@@ -481,12 +481,12 @@ func testServicesConfigMap(t *testing.T) {
...
 		_, err = tc.client.K8s.CoreV1().ConfigMaps(tc.namespace).Get(context.TODO(), servicesConfigMapName,
 			meta.GetOptions{})
-		assert.NoErrorf(t, err, "error ensuring ConfigMap %s exists", servicesConfigMapName)
+		require.NoErrorf(t, err, "error ensuring ConfigMap %s exists", servicesConfigMapName)
 	})
 
 	t.Run("Services ConfigMap re-creation", func(t *testing.T) {
 		err = tc.testServicesCMRegeneration(servicesConfigMapName)
-		assert.NoErrorf(t, err, "error ensuring ConfigMap %s is re-created when deleted", servicesConfigMapName)
+		require.NoErrorf(t, err, "error ensuring ConfigMap %s is re-created when deleted", servicesConfigMapName)
```
saifshaikh48, line 489: I will drop these assert --> require changes, not needed

sebsoto: You will drop them back to requires?
They are needed as you've changed the context they are being used in, as these are now separate subtests

test/e2e/validation_test.go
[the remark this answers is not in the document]
saifshaikh48: I think we can make it mandatory in the test suite once we move the first service definition into the services ConfigMap. I'll leave the question of _should we continue e2e tests if the services ConfigMap does not exist in cluster_ to be revisited when it's applicable. I left a comment on the relevant story [WINC-738: Kubelet service is described in services ConfigMap](https://issues.redhat.com/browse/WINC-738) so we don't forget.

controllers/configmap_controller.go
```diff
@@ -290,13 +290,20 @@ func hasAssociatedInstance(nodeAddresses []core.NodeAddress, instances []*instan
 	return false, nil
 }
 
-// mapToConfigMap fulfills the MapFn type, while always returning a request to the windows-instance ConfigMap
-func (r *ConfigMapReconciler) mapToConfigMap(_ client.Object) []reconcile.Request {
+// mapToInstancesConfigMap fulfills the MapFn type, while always returning a request to the windows-instance ConfigMap
```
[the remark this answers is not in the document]
saifshaikh48: Correct, now 2 separate map functions one for `instances` ConfigMap and one for `services` ConfigMap

controllers/controllers.go
```diff
@@ -155,6 +155,31 @@ func (r *instanceReconciler) deconfigureInstance(node *core.Node) error {
 	return nil
 }
 
+// windowsNodeUpgradePredicate returns a predicate whose filter catches only Windows node objects that signal upgrade
+// to the latest WMCO version through a change in the node version annotation
+func windowsNodeUpgradePredicate() predicate.Funcs {
```
sebsoto, line 160: This isnt actually catching upgrades
this is just filtering out outdated nodes
any time a fully upgraded node changes state this predicate will allow the event

please rename this function

sebsoto:
/hold
until CI issues resolved

saifshaikh48:
/retest-required
/unhold

https://github.com/openshift/installer/pull/5808 has merged

PR #1021, opened 2022-04-05
WINC-687: Add Username annotation for all Windows nodes

jrvaldes:
This PR ensures the Username annotation is present in nodes configured
by the machine controller. Similar to BYOH nodes, the Username annotation
contains the encrypted value of the username used to SSH into the Windows
instance.


saifshaikh48:
/lgtm

looks good, thanks for opening this.

sebsoto (approved):

selansen:
Curious Question : why didn't we have Username annotation on Machine set before and have it only for BYOH?

PR #1029, opened 2022-04-11
WINC-730: Windows service unit testable interfaces

sebsoto:
This commits adds interfaces to interact with the Windows service API.
A unit tested implementation has been added, along with unit tests to
ensure the proper behavior is being maintained. These interfaces should
be used when developing WICD, to enable large portions of the program to
be unit testable.

### DIFF
--- a/hack/wicd-unit.sh
+++ b/hack/wicd-unit.sh
@@ -3,8 +3,6 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-# Placeholder stub for WICD unit tests
-# Currently requiring Windows VM env variables to be set, in order to validate release job functionality
 if [ -z "$INSTANCE_ADDRESS" ]; then
     echo "Windows VM address not provided"
     return 1
@@ -17,8 +15,11 @@ if [ -z "$KUBE_SSH_KEY_PATH" ]; then
     echo "env KUBE_SSH_KEY_PATH not found"
     return 1
 fi
+WICD_UNIT_EXE="wicd_unit.exe"
+# Build unit tests for Windows. Specifically targetting packages used by WICD.
+GOOS=windows GOFLAGS=-v go test -c ./pkg/winsvc/... -o $WICD_UNIT_EXE
 
-# Ensure the image used in the release job has ssh installed, as this will be used to remotely run tests against a
-# Windows VM. Running a dummy command to ensure connection to the VM is possible
-ssh -o StrictHostKeyChecking=no -i $KUBE_SSH_KEY_PATH $INSTANCE_USERNAME@$INSTANCE_ADDRESS "dir"
+# Run the unit tests against the Windows host
+scp -o StrictHostKeyChecking=no -i $KUBE_SSH_KEY_PATH $WICD_UNIT_EXE $INSTANCE_USERNAME@$INSTANCE_ADDRESS:$WICD_UNIT_EXE
+ssh -o StrictHostKeyChecking=no -i $KUBE_SSH_KEY_PATH $INSTANCE_USERNAME@$INSTANCE_ADDRESS "$WICD_UNIT_EXE"
 
--- a/pkg/winsvc/winsvc.go
+++ b/pkg/winsvc/winsvc.go
@@ -1,0 +1,176 @@
+//go:build windows
+
+package winsvc
+
+import (
+	"errors"
+	"fmt"
+	"sync"
+
+	"golang.org/x/sys/windows/svc"
+	"golang.org/x/sys/windows/svc/mgr"
+)
+
+type Mgr interface {
+	CreateService(string, string, mgr.Config, ...string) (Service, error)
+	ListServices() ([]string, error)
+	OpenService(string) (Service, error)
+}
+
+// fakeServiceList mocks out the state of all services on a Windows instance
+type fakeServiceList struct {
+	m    *sync.Mutex
+	svcs map[string]Service
+}
+
+// write overwrites the given service to the svcs map
+func (l *fakeServiceList) write(name string, svc Service) {
+	l.m.Lock()
+	defer l.m.Unlock()
+	l.svcs[name] = svc
+}
+
+// read returns the entry with the given name, and a bool indicating if it exists or not
+func (l *fakeServiceList) read(name string) (Service, bool) {
+	l.m.Lock()
+	defer l.m.Unlock()
+	service, exists := l.svcs[name]
+	return service, exists
+}
+
+// listServiceNames returns a slice of all service names
+func (l *fakeServiceList) listServiceNames() []string {
+	l.m.Lock()
+	defer l.m.Unlock()
+	var names []string
+	for svcName := range l.svcs {
+		names = append(names, svcName)
+	}
+	return names
+}
+
+// remove deletes the entry with the given name, throwing an error if it doesn't exist
+func (l *fakeServiceList) remove(name string) error {
+	l.m.Lock()
+	defer l.m.Unlock()
+	_, exists := l.svcs[name]
+	if !exists {
+		return errors.New("service does not exist")
+	}
+	delete(l.svcs, name)
+	return nil
+}
+
+func newFakeServiceList() *fakeServiceList {
+	return &fakeServiceList{
+		m:    &sync.Mutex{},
+		svcs: make(map[string]Service),
+	}
+}
+
+type testMgr struct {
+	svcList *fakeServiceList
+}
+
+// CreateService installs new service name on the system.
+// The service will be executed by running exepath binary.
+// Use config c to specify service parameters.
+// Any args will be passed as command-line arguments when
+// the service is started; these arguments are distinct from
+// the arguments passed to Service.Start or via the "Start
+// parameters" field in the service's Properties dialog box.
+func (t testMgr) CreateService(name, exepath string, config mgr.Config, args ...string) (Service, error) {
+	// Throw an error if the service already exists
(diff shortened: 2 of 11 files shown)


hack/wicd-unit.sh
```diff
@@ -18,7 +16,10 @@ if [ -z "$KUBE_SSH_KEY_PATH" ]; then
     return 1
 fi
 
-# Ensure the image used in the release job has ssh installed, as this will be used to remotely run tests against a
-# Windows VM. Running a dummy command to ensure connection to the VM is possible
-ssh -o StrictHostKeyChecking=no -i $KUBE_SSH_KEY_PATH $INSTANCE_USERNAME@$INSTANCE_ADDRESS "dir"
+# Build unit tests for Windows. Specifically targetting packages used by WICD.
+GOOS=windows GOFLAGS=-v go test -c ./pkg/winsvc/... -o wicd_unit.exe
```
jrvaldes, line 20: Consider creating a constant for the binary name `wicd_unit.exe`, so that can be re-used in the script.

hack/wicd-unit.sh
```diff
@@ -18,7 +16,10 @@ if [ -z "$KUBE_SSH_KEY_PATH" ]; then
...
+
+# Run the unit tests against the Windows host
+scp -o StrictHostKeyChecking=no -i $KUBE_SSH_KEY_PATH wicd_unit.exe $INSTANCE_USERNAME@$INSTANCE_ADDRESS:wicd_unit.exe
+ssh -o StrictHostKeyChecking=no -i $KUBE_SSH_KEY_PATH $INSTANCE_USERNAME@$INSTANCE_ADDRESS "wicd_unit.exe"
```
jrvaldes, line 24: Does it matter the working directory in the Windows host where it will be invoked? I assume will default to the user profile.

pkg/winsvc/winsvc.go
```diff
@@ -0,0 +1,174 @@
...
+// The service will be executed by running exepath binary.
+// Use config c to specify service parameters.
+// Any args will be passed as command-line arguments when
+// the service is started; these arguments are distinct from
+// the arguments passed to Service.Start or via the "Start
+// parameters" field in the service's Properties dialog box.
+func (t testMgr) CreateService(name, exepath string, config mgr.Config, args ...string) (Service, error) {
+	// Throw an error if the service already exists
+	if _, ok := t.svcList.read(name); ok {
+		return nil, fmt.Errorf("service already exists")
```
jrvaldes, line 83:
```suggestion
                return nil, errors.New("service already exists")
```

pkg/winsvc/winsvc.go
```diff
@@ -0,0 +1,174 @@
...
+	Control(svc.Cmd) (svc.Status, error)
+	Query() (svc.Status, error)
+	UpdateConfig(mgr.Config) error
+}
+
+type FakeService struct {
+	name        string
+	config      mgr.Config
+	status      svc.Status
+	serviceList *fakeServiceList
```
saifshaikh48, line 124: This is the only thing confusing me, why does each FakeService contains a serviceList?

hack/wicd-unit.sh
sebsoto: It shouldn't matter at all. Since this is for CI purposes the home directory was used for simplicity.

pkg/winsvc/winsvc.go
sebsoto: They all need to share the same serviceList as the serviceList must be updated by a `FakeService.Delete()` call, across all `FakeService` structs and the `Mgr` struct. This is mocking out Windows syscalls, where there would be one service list for the whole OS.

saifshaikh48: Thanks for the explanation. So looks like each service's `serviceList` will point to the same object (done in `NewTestMgr()`'s loop), so no need to update every mocked service's `serviceList` individually on a `Delete()` call.

saifshaikh48 (commented):
/lgtm

hack/wicd-unit.sh
```diff
@@ -17,8 +15,11 @@ if [ -z "$KUBE_SSH_KEY_PATH" ]; then
...
 fi
+WICD_UNIT_EXE="wicd_unit.exe"
+# Build unit tests for Windows. Specifically targetting packages used by WICD.
+GOOS=windows GOFLAGS=-v go test -c ./pkg/winsvc/... -o $WICD_UNIT_EXE
 
-# Ensure the image used in the release job has ssh installed, as this will be used to remotely run tests against a
-# Windows VM. Running a dummy command to ensure connection to the VM is possible
-ssh -o StrictHostKeyChecking=no -i $KUBE_SSH_KEY_PATH $INSTANCE_USERNAME@$INSTANCE_ADDRESS "dir"
+# Run the unit tests against the Windows host
+scp -o StrictHostKeyChecking=no -i $KUBE_SSH_KEY_PATH wicd_unit.exe $INSTANCE_USERNAME@$INSTANCE_ADDRESS:wicd_unit.exe
```
jrvaldes, line 23: s/wicd_unit.exe/WICD_UNIT_EXE

pkg/winsvc/winsvc.go
```diff
@@ -0,0 +1,175 @@
...
+	"golang.org/x/sys/windows/svc/mgr"
+)
+
+type Mgr interface {
+	CreateService(string, string, mgr.Config, ...string) (Service, error)
+	ListServices() ([]string, error)
+	OpenService(string) (Service, error)
+}
+
+type fakeServiceList struct {
```
jrvaldes, line 20: An inline comment for this struct may be helpful.

pkg/winsvc/winsvc.go
```diff
@@ -0,0 +1,175 @@
...
+// write overwrites the given service to the svcs map
+func (l *fakeServiceList) write(name string, svc Service) {
+	l.m.Lock()
+	defer l.m.Unlock()
+	l.svcs[name] = svc
+}
+
+// read returns the entry with the given name, and a bool indicating if it exists or not
+func (l *fakeServiceList) read(name string) (Service, bool) {
+	l.m.Lock()
```
jrvaldes, line 34: Availability and consistency are good here, why the lock for a read operation?

pkg/winsvc/winsvc.go
```diff
@@ -0,0 +1,175 @@
...
+	defer l.m.Unlock()
+	service, exists := l.svcs[name]
+	return service, exists
+}
+
+// listServiceNames returns a slice of all service names
+func (l *fakeServiceList) listServiceNames() []string {
+	l.m.Lock()
+	defer l.m.Unlock()
+	var names []string
```
jrvaldes, line 44: As the size of the list is known, consider:
```
names := make([]string, len(l.svcs))
```

pkg/winsvc/winsvc.go
```diff
@@ -0,0 +1,175 @@
...
+		names = append(names, svcName)
+	}
+	return names
+}
+
+// remove deletes the entry with the given name, throwing an error if it doesn't exist
+func (l *fakeServiceList) remove(name string) error {
+	l.m.Lock()
+	defer l.m.Unlock()
+	_, exists := l.svcs[name]
```
jrvaldes, line 55: Why not using `read()`?

pkg/winsvc/winsvc.go
sebsoto: Writing and reading from a map at the same time will cause a panic, so the lock is necessary

pkg/winsvc/winsvc.go
sebsoto: read cannot be called as that would cause a deadlock due to both remove and read locking.

pkg/winsvc/winsvc.go
sebsoto: would prefer to leave it like this, as this is test code and it is a small optimization with the trade off of having to do all of this:
```
	names := make([]string, len(l.svcs))
	i := 0
	for svcName := range l.svcs {
		names[i] = svcName
		i++
	}
```

hack/wicd-unit.sh
```diff
@@ -17,8 +15,11 @@ if [ -z "$KUBE_SSH_KEY_PATH" ]; then
...
 fi
+WICD_UNIT_EXE="wicd_unit.exe"
+# Build unit tests for Windows. Specifically targetting packages used by WICD.
+GOOS=windows GOFLAGS=-v go test -c ./pkg/winsvc/... -o $WICD_UNIT_EXE
 
-# Ensure the image used in the release job has ssh installed, as this will be used to remotely run tests against a
-# Windows VM. Running a dummy command to ensure connection to the VM is possible
-ssh -o StrictHostKeyChecking=no -i $KUBE_SSH_KEY_PATH $INSTANCE_USERNAME@$INSTANCE_ADDRESS "dir"
+# Run the unit tests against the Windows host
+scp -o StrictHostKeyChecking=no -i $KUBE_SSH_KEY_PATH wicd_unit.exe $INSTANCE_USERNAME@$INSTANCE_ADDRESS:$WICD_UNIT_EXE
```
jrvaldes, line 23:
```suggestion
scp -o StrictHostKeyChecking=no -i $KUBE_SSH_KEY_PATH $WICD_UNIT_EXE $INSTANCE_USERNAME@$INSTANCE_ADDRESS:$WICD_UNIT_EXE
```

sebsoto:
/hold
ci blocked by https://github.com/openshift/installer/pull/5808

PR #1034, opened 2022-04-26
WINC-799: [controllers] Tighten node permissions

jrvaldes:
This commit adjusts the RBAC permission for the nodes resource by
explicitly listing the required verbs.

Ran:
```
make bundle
```

### DIFF
--- a/bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
@@ -200,8 +200,11 @@ spec:
           resources:
           - nodes
           verbs:
-          - '*'
+          - delete
+          - get
           - list
+          - patch
+          - watch
         - apiGroups:
           - ""
           resources:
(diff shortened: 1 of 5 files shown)


sebsoto:
lgtm assuming the existing tests pass

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
```diff
@@ -200,8 +200,12 @@ spec:
           resources:
           - nodes
           verbs:
-          - '*'
+          - create
```
aravindhp, line 203: Why do we need create?

jrvaldes: Good catch! We don't.

aravindhp (commented):
Thanks for doing this, @jrvaldes. LGTM.

s/Tight/Tighten in commit and PR header.

jrvaldes:
> Thanks for doing this, @jrvaldes. LGTM.
> 
> s/Tight/Tighten in commit and PR header.

Done

PR #1043, opened 2022-04-29
[wicd] Add cobra command framework

sebsoto:
This commit is adding the framework for the wicd command interface. This
change is adding the help text for wicd.exe, as well as the stub for the
`controller` sub command.
This is being added ahead of the implementation of the controller
subcommand, as it will allow for parellel work between different
subcommands.


sebsoto:
wicd.exe output now looks like this:

```
$ wicd.exe
The Windows Instance Config Daemon performs multiple functions related to mainting the expected state of a Windows Node.

Usage:
  wicd.exe [command]

Available Commands:
  completion  generate the autocompletion script for the specified shell
  controller  Manages local Windows Services
  help        Help about any command

Flags:
  -h, --help   help for wicd.exe

Use "wicd.exe [command] --help" for more information about a command.

```

cmd/daemon/main.go
```diff
@@ -17,9 +17,24 @@ limitations under the License.
...
+	rootCmd = &cobra.Command{
+		Use:   "wicd.exe",
+		Short: "Windows Instance Config Daemon",
+		Long: "The Windows Instance Config Daemon performs multiple functions related to maintaining the expected " +
+			"state of a Windows Node.",
+	}
+)
+
 func main() {
-	klog.Info("Daemon is running...")
```
selansen, line 24: Will there be any log message showing WICD is running?

sebsoto: @selansen yes the above help text will be printed when WICD is ran https://github.com/openshift/windows-machine-config-operator/pull/1043#issuecomment-1113515273

go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/pkg/sftp v1.11.0
 	github.com/prometheus-operator/prometheus-operator/pkg/client v0.45.0
+	github.com/spf13/cobra v1.2.1
```
jrvaldes, line 17: Can we use the a more recent version (i.e. [v1.4.0](https://github.com/spf13/cobra/releases/tag/v1.4.0))?

Makefile
```diff
@@ -91,7 +91,7 @@ build: fmt vet
 
 .PHONY: build-daemon
 build-daemon:
-	env GOOS=windows GOARCH=amd64 go build -o ${OUTPUT_DIR}/bin/windows-instance-config-daemon.exe cmd/daemon/main.go
+	env GOOS=windows GOARCH=amd64 go build -o ${OUTPUT_DIR}/bin/windows-instance-config-daemon.exe cmd/daemon/*
```
jrvaldes, line 94: Is the trailing  `/*` required?

go.mod
sebsoto: Good point, updated

Makefile
sebsoto: Yes it is

PR #1045, opened 2022-04-30
Bug 2070892: Skip node addresses without PTR record during instance de-configuration

jrvaldes:
This PR updates the de-configuration workflow to ignore all
node addresses that fail to resolve a valid reverse lookup record
while looking for the associated Windows instance.

### DIFF
--- a/controllers/configmap_controller.go
+++ b/controllers/configmap_controller.go
@@ -311,11 +311,7 @@ func (r *ConfigMapReconciler) deconfigureInstances(instances []*instance.Info, n
 		Namespace: r.watchNamespace}}
 	for _, node := range nodes.Items {
 		// Check for instances associated with this node
-		hasEntry, err := hasAssociatedInstance(node.Status.Addresses, instances)
-		if err != nil {
-			return errors.Wrapf(err, "unable to find instance associated with node %s", node.GetName())
-		}
-		if hasEntry {
+		if hasAssociatedInstance(node.Status.Addresses, instances) {
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


selansen (commented):
/lgtm

controllers/configmap_controller.go
```diff
@@ -330,9 +330,14 @@ func (r *ConfigMapReconciler) deconfigureInstances(instances []*instance.Info, n
...
 // hasAssociatedInstance returns true if any of the given addresses is associated with any instance in the given slice.
-// The instance's network address must be a valid IPv4 address or resolve to one.
+// The instance's network address must be a valid internal IPv4 address or resolve to one.
 func hasAssociatedInstance(nodeAddresses []core.NodeAddress, instances []*instance.Info) (bool, error) {
 	for _, nodeAddress := range nodeAddresses {
+		// check node's address type
+		if nodeAddress.Type == core.NodeExternalIP || nodeAddress.Type == core.NodeExternalDNS {
+			// skip external addresses
+			continue
+		}
```
sebsoto, line 340: I dont see why we should skip external addresses completely, it seems like we should instead not fail on reverse lookup failure on these lines below, and continue trying other addresses
```
		if parseAddr := net.ParseIP(nodeAddress.Address); parseAddr != nil {
			dnsAddresses, err := net.LookupAddr(nodeAddress.Address)
			if err != nil {
				return false, errors.Wrapf(err, "failed to lookup DNS for IP %s", nodeAddress.Address)
			}
			```

**Review by sebsoto — CHANGES_REQUESTED** [review]

File: controllers/configmap_controller.go (hunk 1 above)
Comment (jrvaldes) [reply controllers/configmap_controller.go]: Makes sense, adjusted in https://github.com/openshift/windows-machine-config-operator/pull/1045/commits/95d053bbbeb62c300887fb9828a454679aa540f5

File: controllers/configmap_controller.go [hunk 2]
```diff
@@ -343,7 +343,8 @@ func hasAssociatedInstance(nodeAddresses []core.NodeAddress, instances []*instan
 		if parseAddr := net.ParseIP(nodeAddress.Address); parseAddr != nil {
 			dnsAddresses, err := net.LookupAddr(nodeAddress.Address)
 			if err != nil {
-				return false, errors.Wrapf(err, "failed to lookup DNS for IP %s", nodeAddress.Address)
+				// skip node's address without reverse lookup records
+				continue
```
sebsoto, line 347: just realized that makes it so no return error is needed for this function signature

sebsoto (changes requested):

controllers/configmap_controller.go
jrvaldes: Oh! I missed that 😞 . Fixed in [923f7f9 ](https://github.com/openshift/windows-machine-config-operator/commit/923f7f99fa3712830457fa80cdc3b1bd823044aa)

saifshaikh48 (commented):
/lgtm

appreciated the unit tests :+1:

PR #1058, opened 2022-05-06
 [docs] clean up Docker specific debug information

selansen:
As containerd becomes the default runtime, Docker specific debug information should be removed from troubleshoot doc.This commit cleans up Docker runtime specific debug information.

### DIFF
--- a/docs/TROUBLESHOOTING.md
+++ b/docs/TROUBLESHOOTING.md
@@ -74,9 +74,9 @@ $ oc adm node-logs -l kubernetes.io/os=windows --path=/containerd/containerd.log
 ## How to collect Windows application event logs
 
 The Get-WinEvent shim on the kubelet logs endpoint can be used to collect application event logs from Windows machines.
-E.g. for the docker runtime service:
+E.g. getting logs for any given service:
 ```shell script
-$ oc adm node-logs -l kubernetes.io/os=windows --path=journal -u docker
+$ oc adm node-logs -l kubernetes.io/os=windows --path=journal -u <LOG_NAME>
 ```
 The same command is executed when collecting logs with `oc adm must-gather`.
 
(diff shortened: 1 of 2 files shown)


docs/TROUBLESHOOTING.md
```diff
@@ -74,9 +74,9 @@ $ oc adm node-logs -l kubernetes.io/os=windows --path=/containerd/containerd.log
 ## How to collect Windows application event logs
 
 The Get-WinEvent shim on the kubelet logs endpoint can be used to collect application event logs from Windows machines.
-E.g. for the docker runtime service:
+E.g. for the containerd runtime service:
 ```shell script
-$ oc adm node-logs -l kubernetes.io/os=windows --path=journal -u docker
+$ oc adm node-logs -l kubernetes.io/os=windows --path=/containerd/containerd.log
 ```
```
saifshaikh48, line 80: We can remove this, it is a repeat of the `## How to collect containerd runtime logs` section above

mansikulkarni96: I agree with @saifshaikh48 this whole section can be removed, no need for example here as we are already giving an example above this step both for containerd and kubelet logs. Moreover this is not collecting win-event logs now.

jrvaldes: This section is still useful and valid, we just need to find a good example that fit 
```
oc adm node-logs -l kubernetes.io/os=windows --path=journal -u <LOG_NAME>
```

selansen: I will use the above template line instead of containerd example

docs/TROUBLESHOOTING.md
```diff
@@ -74,9 +74,9 @@ $ oc adm node-logs -l kubernetes.io/os=windows --path=/containerd/containerd.log
 ## How to collect Windows application event logs
```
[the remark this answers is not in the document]
selansen: good catch. Thanks.

docs/machineset-aws.md
```diff
@@ -1,9 +1,10 @@
 # Creating an AWS Windows MachineSet
 
-_\<windows_container_ami\>_ should be replaced with the AMI ID of a Windows
-image with the Docker container run-time installed. You must use Windows Server
-2019 with a version 10.0.17763.1457 or earlier to work around Windows containers
-behind a Kubernetes load balancer becoming unreachable
+_\<windows_container_ami\>_ can be replaced with the AMI ID of a Windows
+image with the Docker container run-time installed or without it as the default
+runtime will be containerd and installing containerd will be taken care by WMCO.
```
sebsoto, line 5: In my opinion, if it doesnt matter if the image has or doesnt have docker its better just to drop mentioning docker at all

sebsoto (changes requested):

PR #1060, opened 2022-05-09
WINC-819: Enable required Windows feature for Containers

mansikulkarni96:
Windows Server 2022 onwards required Windows containers feature is not enabled by default. This PR adds
the ability to enable the Containers feature.
The additional step to do this from the golden image template has been reverted.
This work has been combined with the host name change method to minimize the number of times for VM reboot.

### DIFF
--- a/pkg/windows/windows.go
+++ b/pkg/windows/windows.go
@@ -95,6 +95,8 @@ const (
 	// ManagedTag indicates that the service being described is managed by OpenShift. This ensures that all services
 	// created as part of Node configuration can be searched for by checking their description for this string
 	ManagedTag = "OpenShift managed"
+	// containersFeatureName is the name of the Windows feature that is required to be enabled on the Windows instance.
+	containersFeatureName = "Containers"
 )
 
 var (
@@ -422,11 +424,8 @@ func (vm *windows) Configure() error {
 	if err := vm.EnsureRequiredServicesStopped(); err != nil {
 		return errors.Wrap(err, "unable to stop all services")
 	}
-	// Set the hostName of the Windows VM if needed
-	if vm.instance.NewHostname != "" {
-		if err := vm.ensureHostName(); err != nil {
-			return err
-		}
+	if err := vm.ensureHostNameAndContainersFeature(); err != nil {
+		return err
 	}
 	if err := vm.createDirectories(); err != nil {
 		return errors.Wrap(err, "error creating directories on Windows VM")
@@ -602,17 +601,41 @@ func (vm *windows) ConfigureKubeProxy(nodeName, hostSubnet string) error {
 
 // Interface helper methods
 
-// ensureHostName ensures hostname of the Windows VM matches the expected name
-func (vm *windows) ensureHostName() error {
-	hostNameChangedNeeded, err := vm.isHostNameChangeNeeded()
+// ensureHostNameAndContainersFeature ensures hostname of the Windows VM matches the expected name
+// and the required Windows feature is enabled.
+func (vm *windows) ensureHostNameAndContainersFeature() error {
+	rebootNeeded := false
+	// Set the hostName of the Windows VM if needed
+	if vm.instance.NewHostname != "" {
+		hostNameChangedNeeded, err := vm.isHostNameChangeNeeded()
+		if err != nil {
+			return err
+		}
+		if hostNameChangedNeeded {
+			if err := vm.changeHostName(); err != nil {
+				return err
+			}
+			rebootNeeded = true
+		}
+	}
+	isContainersFeatureEnabled, err := vm.isContainersFeatureEnabled()
 	if err != nil {
 		return err
 	}
-	if hostNameChangedNeeded {
-		if err := vm.changeHostName(); err != nil {
-			return err
+	if !isContainersFeatureEnabled {
+		if err := vm.enableContainersWindowsFeature(); err != nil {
+			return errors.Wrapf(err, "error enabling Windows Containers feature")
+		}
+		rebootNeeded = true
+	}
+	// Changing the host name or enabling the Containers feature requires a VM restart for
+	// the change to take effect.
+	if rebootNeeded {
+		if err := vm.rebootAndReinitialize(); err != nil {
+			return errors.Wrapf(err, "error restarting the Windows instance and reinitializing SSH connection")
 		}
 	}
+
 	return nil
 }
 
@@ -627,15 +650,11 @@ func (vm *windows) isHostNameChangeNeeded() (bool, error) {
 
 // changeHostName changes the hostName of the Windows VM to match the expected value
 func (vm *windows) changeHostName() error {
-	changeHostNameCommand := "Rename-Computer -NewName " + vm.instance.NewHostname + " -Force -Restart"
+	changeHostNameCommand := "Rename-Computer -NewName " + vm.instance.NewHostname + " -Force"
 	out, err := vm.Run(changeHostNameCommand, true)
 	if err != nil {
 		vm.log.Info("changing host name failed", "command", changeHostNameCommand, "output", out)
 		return errors.Wrap(err, "changing host name failed")
-	}
-	// Reinitialize the SSH connection given changing the host name requires a VM restart
-	if err := vm.Reinitialize(); err != nil {
-		return errors.Wrap(err, "error reinitializing VM after changing hostname")
 	}
 	return nil
 }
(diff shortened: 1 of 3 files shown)


pkg/internal/enable-containers.ps1
```diff
@@ -0,0 +1,8 @@
+$requiredWindowsFeature = "Containers"
+
+$f = Get-WindowsFeature -Name $requiredWindowsFeature
+if (-not $f.Installed) {
+     Write-Warning "Windows feature: '$requiredWindowsFeature' is not installed."
+     Write-Output "Installing required windows features and re-booting for the enablement to take effect"
```
saifshaikh48, line 6: `Windows feature`

pkg/windows/windows.go
```diff
@@ -1008,6 +1014,21 @@ func (vm *windows) removeHNSNetwork(networkName string) error {
 	return nil
 }
 
+// enableWindowsFeatures enables the required Windows Container features on the Windows host.
+func (vm *windows) enableWindowsFeatures() error {
```
saifshaikh48, line 1018: Consider renaming to `enableRequiredWindowsFeature`

saifshaikh48 (commented):
Nit throughout: there’s only 1 Windows feature in play here i.e. the Containers feature, no need to make the references plural (`features`)

pkg/windows/windows.go
```diff
@@ -1008,6 +1014,21 @@ func (vm *windows) removeHNSNetwork(networkName string) error {
...
+	out, err := vm.Run(enableContainersCmd, true)
+	if err != nil {
+		return errors.Wrapf(err, "failed to enable required Windows features with output: %s", out)
+	}
+
+	// Reinitialize the SSH connection given enabling Windows features requires a VM restart
+	if err := vm.Reinitialize(); err != nil {
```
selansen, line 1025: will this reinitialize get restart the node even though the container feature is already enabled?

mansikulkarni96: It will not restart VM here, that's done in PS script. This is reinitializing SSH connection. But I think that is not required as well if the feature is already installed.

pkg/internal/enable-containers.ps1
```diff
@@ -0,0 +1,8 @@
+$requiredWindowsFeature = "Containers"
```
jrvaldes, line 1: This is specific to `Containers`, so I suggest naming this as:
```
$ContainerFeatureName = "Containers"
```

On the other hand, you can make this extensible and have a collection with a list of the required features. (`requiredWindowsFeatures`).

pkg/windows/windows.go
```diff
@@ -1008,6 +1014,21 @@ func (vm *windows) removeHNSNetwork(networkName string) error {
 	return nil
 }
 
+// enableWindowsFeatures enables the required Windows Container features on the Windows host.
```
jrvaldes, line 1017:
```suggestion
// enableWindowsFeatures enables the required Windows features on the Windows host, for example Containers
```

pkg/windows/windows.go
jrvaldes: I would leave it plural (`...WindowsFeatures()`), so that it can be extended with other required features.

build/Dockerfile.ci
```diff
@@ -157,6 +158,7 @@ COPY --from=build /build/windows-machine-config-operator/pkg/internal/cni-conf-t
 WORKDIR /payload/powershell/
 COPY --from=build /build/windows-machine-config-operator/pkg/internal/wget-ignore-cert.ps1 .
 COPY --from=build /build/windows-machine-config-operator/pkg/internal/hns.psm1 .
+COPY --from=build /build/windows-machine-config-operator/pkg/internal/enable-containers.ps1 .
```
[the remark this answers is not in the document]
mansikulkarni96: yes, mentioned this in the PR message. Keeping it this way to be extensible in the future with WICD changes.
cc: @selansen

hack/machineset.sh
```diff
@@ -79,9 +79,9 @@ get_aws_ms() {
   local provider=$4
 
   # get the AMI id for the Windows VM
```
jrvaldes, line 81: Consider updating these [docs](https://github.com/openshift/windows-machine-config-operator/blob/256c3e5159bafb380db364f386b04fbc9e4fed9f/docs/machineset-aws.md#creating-an-aws-windows-machineset) too.

pkg/windows/windows.go
```diff
@@ -434,6 +437,9 @@ func (vm *windows) Configure() error {
 	if err := vm.transferFiles(); err != nil {
 		return errors.Wrap(err, "error transferring files to Windows VM")
 	}
+	if err := vm.enableWindowsFeatures(); err != nil {
```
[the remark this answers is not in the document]
mansikulkarni96: That's a good suggestion but I do not have access to powershell scripts before transferFiles stage.

hack/machineset.sh
mansikulkarni96: yes still in the works.

pkg/windows/windows.go
alinaryan: I'd keep it singular - it can be changed once we have more

mansikulkarni96: I lean towards keeping it singular as well in the short term.

hack/machineset.sh
```diff
@@ -185,7 +185,7 @@ get_vsphere_ms() {
 
   # set golden image template name
   # TODO: read from parameter
-  template="windows-golden-images/windows-server-2022-template"
+  template="windows-golden-images/windows-server-2022-template-containersDisabled"
```
sebsoto, line 188: this commit should be after the functionality commit

jrvaldes, line 188: I don't think we need this chance, I suggest using `windows-golden-images/windows-server-2022-template` as the latest template name, it's easier to update the template in vCenter.

pkg/windows/windows.go
```diff
@@ -602,17 +601,41 @@ func (vm *windows) ConfigureKubeProxy(nodeName, hostSubnet string) error {
...
+	// Set the hostName of the Windows VM if needed
+	if vm.instance.NewHostname != "" {
+		hostNameChangedNeeded, err := vm.isHostNameChangeNeeded()
+		if err != nil {
+			return err
+		}
+		if hostNameChangedNeeded {
+			if err := vm.changeHostName(); err != nil {
+				return err
+			}
```
sebsoto, line 617: could these two functions be combined and simplified into a single ensureHostName function?

pkg/windows/windows.go
```diff
@@ -1008,6 +1027,39 @@ func (vm *windows) removeHNSNetwork(networkName string) error {
...
+	command := "Install-WindowsFeature -Name " + containersFeatureName
+	out, err := vm.Run(command, true)
+	if err != nil {
+		return errors.Wrapf(err, "failed to enable required Windows feature: %s with output: %s",
+			containersFeatureName, out)
+	}
+	return nil
+}
+
+// isContainersFeatureEnabled tells if the required Containers Windows feature is enabled on the Windows instance
```
sebsoto, line 1041: tells if -> returns true if

pkg/windows/windows.go
```diff
@@ -1008,6 +1027,39 @@ func (vm *windows) removeHNSNetwork(networkName string) error {
...
+	command := "Get-WindowsOptionalFeature -FeatureName " + containersFeatureName + " -Online"
+	out, err := vm.Run(command, true)
+	if err != nil {
+		return false, errors.Wrapf(err, "failed to get Windows feature: %s", containersFeatureName)
+	}
+	return strings.Contains(out, "Enabled"), nil
+}
+
+// rebootAnReinitialize restarts the Windows instance and re-initializes the SSH connection for further configuration
+func (vm *windows) rebootAnReinitialize() error {
```
sebsoto, line 1052: rebootAnReinitialize -> rebootAndReinitialize

sebsoto (changes requested):
These changes are fine, but i'm now unsure how this is going to be handled by WICD since a restart is required before bootstrap. Not a blocker but something to think about

pkg/windows/windows.go
mansikulkarni96: @sebsoto  sure seems doable, but I wanted to keep them together to figure out if the reboot is needed. Else I would have to make enusreHostName() return a bool to signal a reboot which seems a bit counter-intutive to ensure host name has been changed.

hack/machineset.sh
mansikulkarni96: addressed this and rest of the minor comments.

aravindhp:
I meant to say revert [[vsphere_ci][docs] Revert enabling Containers feature](https://github.com/openshift/windows-machine-config-operator/pull/1060/commits/e43726a2325dc4d5be88438c6f4f4bd6de1d1a55) and not the whole PR.

PR #1067, opened 2022-05-16
Correct Windows service start/stop behavior

sebsoto:
This commit corrects an issue where trying to start/stop an already started/stopped
Windows service would error out. The ServiceController now checks that
the service is not already in the correct state before telling Windows
to start/stop the service. Additionally, the mock service API used in
the unit tests has been altered to enact the behavior that the actual
service API has.

### DIFF
--- a/pkg/daemon/winsvc/winsvc.go
+++ b/pkg/daemon/winsvc/winsvc.go
@@ -3,10 +3,10 @@
 package winsvc
 
 import (
-	"errors"
 	"fmt"
 	"sync"
 
+	"github.com/pkg/errors"
 	"golang.org/x/sys/windows/svc"
 	"golang.org/x/sys/windows/svc/mgr"
 )
@@ -156,6 +156,9 @@ func (f *FakeService) Close() error {
 }
 
 func (f *FakeService) Start(_ ...string) error {
+	if f.status.State == svc.Running {
+		return errors.New("service already running")
+	}
 	f.status.State = svc.Running
 	return nil
 }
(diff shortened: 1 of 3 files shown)


pkg/daemon/winsvc/winsvc.go
```diff
@@ -208,3 +211,23 @@ func NewTestMgr(existingServices map[string]*FakeService) *testMgr {
 	}
 	return testMgr
 }
+
+// EnsureServiceState ensures the service is in the given state
+func EnsureServiceState(service Service, state svc.State) error {
```
selansen, line 216: we are returning err from multiple places. Would it be nice to append some additional error messages so that we will get more details while debugging?

pkg/daemon/winsvc/winsvc_test.go
```diff
@@ -173,3 +173,44 @@ func TestDeleteService(t *testing.T) {
 		})
 	}
 }
+
+func TestStopService(t *testing.T) {
```
jrvaldes, line 177: Consider naming this test as `TestEnsureServiceState` and add full coverage for `EnsureServiceState` function.

pkg/daemon/winsvc/winsvc.go
alinaryan: +1

pkg/daemon/winsvc/winsvc_test.go
sebsoto: Added the TestEnsureServiceState function below
both are needed

saifshaikh48 (commented):
/lgtm

PR #1073, opened 2022-05-17
WINC-825: WICD reacts to Node events

sebsoto:
The controller-runtime manager is now being used to watch for Node
events relevant to the instance. When a relevant Node event occurs,
WICD will reconcile the Windows services specified within the cmData
Node annotation. In the future, instead of using this annotation, WICD
will use the actual contents of the appropriate services ConfigMap.

### DIFF
--- a/pkg/daemon/controller/controller.go
+++ b/pkg/daemon/controller/controller.go
@@ -29,12 +29,17 @@ import (
 	"golang.org/x/sys/windows/svc"
 	"golang.org/x/sys/windows/svc/mgr"
 	core "k8s.io/api/core/v1"
+	k8sapierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/util/jsonpath"
 	"k8s.io/klog/v2"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/winsvc"
 	"github.com/openshift/windows-machine-config-operator/pkg/nodeutil"
@@ -42,19 +47,23 @@ import (
 	"github.com/openshift/windows-machine-config-operator/pkg/windows"
 )
 
+const (
+	// desiredVersionAnnotation is a Node annotation, indicating the Service ConfigMap that should be used to configure it
+	desiredVersionAnnotation = "windowsmachineconfig.openshift.io/desired-version"
+	// wmcoNamespace defines the namespace service ConfigMaps are expected to be in
+	wmcoNamespace = "openshift-windows-machine-config-operator"
+)
+
 type ServiceController struct {
 	winsvc.Mgr
-	client   client.Client
-	ctx      context.Context
-	nodeName string
+	client         client.Client
+	ctx            context.Context
+	nodeName       string
+	watchNamespace string
 }
 
 // RunController is the entry point of WICD's controller functionality
 func RunController(kubeconfigPath string) error {
-	// TODO: This is currently a barebones controller, which pulls only the desired version's ConfigMap,
-	//       and reconciles the VM to match the ConfigMap's specs. This barebones implementation will not watch the
-	//       ConfigMap or the VM for state changes.
-
 	svcMgr, err := winsvc.NewMgr()
 	if err != nil {
 		return err
@@ -64,21 +73,22 @@ func RunController(kubeconfigPath string) error {
 		return errors.Wrap(err, "error using kubeconfig to build config")
 	}
 
-	// TODO: This is a client that reads directly from the server, not a cached client like used in WMCO. This should
-	//       be replaced with the cache client received by mgr.GetClient()
 	clientScheme := runtime.NewScheme()
 	err = clientgoscheme.AddToScheme(clientScheme)
 	if err != nil {
 		return err
 	}
-	clientset, err := client.New(config, client.Options{Scheme: clientScheme})
+	// This is a client that reads directly from the server, not a cached client. This is required to be used here, as
+	// the cached client, created by ctrl.NewManager() will not be functional until the manager is started.
+	directClient, err := client.New(config, client.Options{Scheme: clientScheme})
 	if err != nil {
 		return err
 	}
 
 	// use the client to find the name of the node associated with the VM this is running on
+	ctx := ctrl.SetupSignalHandler()
 	var nodes core.NodeList
-	err = clientset.List(context.TODO(), &nodes)
+	err = directClient.List(ctx, &nodes)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


pkg/daemon/controller/controller.go
```diff
@@ -64,58 +73,142 @@ func RunController(kubeconfigPath string) error {
...
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return strings.HasPrefix(e.ObjectNew.GetName(), servicescm.NamePrefix)
+		},
+		GenericFunc: func(e event.GenericEvent) bool {
+			return strings.HasPrefix(e.Object.GetName(), servicescm.NamePrefix)
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return strings.HasPrefix(e.Object.GetName(), servicescm.NamePrefix)
+		},
+	}
```
saifshaikh48, line 160: consider condensing with [predicate.NewPredicateFuncs()](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/predicate#NewPredicateFuncs)

mansikulkarni96 (commented):

pkg/daemon/controller/controller.go
```diff
@@ -64,58 +73,132 @@ func RunController(kubeconfigPath string) error {
...
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			// Only process update events if the desired version has changed
+			return sc.nodeName == e.ObjectNew.GetName() &&
+				e.ObjectOld.GetAnnotations()[desiredVersionAnnotation] != e.ObjectNew.GetAnnotations()[desiredVersionAnnotation]
+		},
+		GenericFunc: func(e event.GenericEvent) bool {
+			return sc.nodeName == e.Object.GetName()
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
```
mansikulkarni96, line 143: would the delete come into play with the cleanup command?

pkg/daemon/controller/controller.go
```diff
@@ -64,58 +73,132 @@ func RunController(kubeconfigPath string) error {
...
-	node, err := currentNode(&nodes, addrs)
+	node, err := findNodeByAddress(&nodes, addrs)
+	if err != nil {
+		return err
+	}
+
+	ctrlMgr, err := ctrl.NewManager(config, ctrl.Options{
+		Namespace: "openshift-windows-machine-config-operator",
+		Scheme:    clientScheme,
+	})
```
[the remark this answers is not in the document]
sebsoto: The function that calls this function handles that

pkg/daemon/controller/controller.go
```diff
@@ -64,58 +73,132 @@ func RunController(kubeconfigPath string) error {
...
+	if err := ctrlMgr.Start(ctx); err != nil {
+		return err
+	}
+	return nil
+}
+
+// NewServiceController returns a pointer to a ServiceController object
+func NewServiceController(ctx context.Context, client client.Client, mgr winsvc.Mgr, nodeName string) *ServiceController {
+	return &ServiceController{client: client, Mgr: mgr, ctx: ctx, nodeName: nodeName,
+		watchNamespace: "openshift-windows-machine-config-operator"}
```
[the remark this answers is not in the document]
sebsoto: Good point, added

pkg/daemon/controller/controller.go
sebsoto: I'm not sure at this moment, for now I'll change it to drop delete events as they dont have a purpose at the moment.

pkg/daemon/controller/controller.go
```diff
@@ -64,58 +73,132 @@ func RunController(kubeconfigPath string) error {
...
+			return ctrl.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		return ctrl.Result{}, err
+	}
+	desiredVersion, present := node.Annotations[desiredVersionAnnotation]
 	if !present {
-		return errors.New("node missing desired version annotation")
+		// node missing desired version annotation, don't requeue
+		return ctrl.Result{}, nil
```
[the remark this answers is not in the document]
sebsoto: For WICD, wait until it has the annotation in a future event.
Its up to WMCO to apply the annotation after the node bootstrapping process completes

Added logic to the predicate to drop events without the annotation.

pkg/daemon/controller/controller.go
```diff
@@ -29,32 +29,45 @@ import (
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/winsvc"
 	"github.com/openshift/windows-machine-config-operator/pkg/nodeutil"
 	"github.com/openshift/windows-machine-config-operator/pkg/servicescm"
 	"github.com/openshift/windows-machine-config-operator/pkg/windows"
 )
 
+var (
```
jrvaldes, line 54: s/var/const

Will the value of these vars change?

pkg/daemon/controller/controller.go
```diff
@@ -29,32 +29,45 @@ import (
...
+	// desiredVersionAnnotation is a Node annotation, indicating the Service ConfigMap that should be used to configure it
+	desiredVersionAnnotation = "windowsmachineconfig.openshift.io/desired-version"
+	// watchNamespace defines the namespace service ConfigMaps are expected to be in
+	watchNamespace = "openshift-machine-config-operator"
```
jrvaldes, line 58: This is the namespace of the machine-config-opereator, is that the correct one? If so, consider naming this as `namespaceMachineConfigOperator`.

pkg/daemon/controller/controller.go
```diff
@@ -64,58 +77,133 @@ func RunController(kubeconfigPath string) error {
...
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&core.Node{}, builder.WithPredicates(nodePredicate)).
+		Watches(&source.Kind{Type: &core.ConfigMap{}}, handler.EnqueueRequestsFromMapFunc(sc.mapToCurrentNode),
+			builder.WithPredicates(cmPredicate)).
+		Complete(sc)
+}
+
+// mapToCurrentNode maps all events to the node associated with this Windows instance
+func (sc *ServiceController) mapToCurrentNode(_ client.Object) []reconcile.Request {
+	return []reconcile.Request{{types.NamespacedName{Name: sc.nodeName}}}
```
jrvaldes, line 165: How do this works when there is more than one ConfigMap with prefix `servicescm.NamePrefix` ?

pkg/daemon/controller/controller.go
```diff
@@ -64,58 +77,133 @@ func RunController(kubeconfigPath string) error {
...
+
+// Reconcile fulfills the Reconciler interface
+func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (result ctrl.Result, reconcileErr error) {
+	klog.Infof("reconciling %s", req.NamespacedName)
+	var node core.Node
+	err := sc.client.Get(sc.ctx, req.NamespacedName, &node)
+	if err != nil {
+		if k8sapierrors.IsNotFound(err) {
+			// Request object not found, could have been deleted after reconcile request.
+			klog.Error("node not found")
```
jrvaldes, line 176: It may be useful to include the node name if the log message.

pkg/daemon/controller/controller.go
```diff
@@ -194,6 +283,7 @@ func (sc *ServiceController) reconcileService(service winsvc.Service, expected s
 	}
 
 	if updateRequired {
+		klog.Infof("updating %s service", expected.Name)
```
jrvaldes, line 286:
```suggestion
		klog.Infof("updating service %s ", expected.Name)
```

For consistency with above logs.

jrvaldes (commented):
Thanks for working on this @sebsoto, added a few recommendations. PTAL

pkg/daemon/controller/controller.go
sebsoto: Should have been the wmco namespace
changed variable name to be more explicit

pkg/daemon/controller/controller.go
sebsoto: This just makes it so anytime there is a service ConfigMap event that makes it through the predicate, the current VM's node is reconciled. There being multiple ConfigMaps should only happen during upgrades, and won't be a problem.

pkg/daemon/controller/controller.go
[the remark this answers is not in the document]
saifshaikh48: correct me if I'm wrong, but this check is now redundant as we already filter out nodes without `desiredVersionAnnotation` in the predicate function?

pkg/daemon/controller/controller_test.go
```diff
@@ -217,6 +222,148 @@ func TestReconcileService(t *testing.T) {
...
+	}
+	for _, test := range testIO {
+		t.Run(test.name, func(t *testing.T) {
+			desiredVersion := "testversion"
+			// This ConfigMap's name must match with the given Node object's desired-version annotation
+			cm, err := servicescm.GenerateWithData(servicescm.NamePrefix+desiredVersion,
+				"openshift-windows-machine-config-operator", &test.configMapServices, &[]servicescm.FileInfo{})
+			require.NoError(t, err)
+			// TODO: When the controller is able to watch ConfigMaps, add the generated CM to the cluster obj list,
+			//       instead of adding this annotation to the Node.
```
saifshaikh48, line 303: Isn't the generated CM already being added to the clusterObjs list below?

PR #1076, opened 2022-05-23
WINC-745: Clean up dockershim related flags

mansikulkarni96:
Update to https://github.com/openshift/windows-machine-config-bootstrapper/commit/897c494734c08abaaab7e7cf88ad594edf30fa34

Adapt to the WMCB submodule updates.

### DIFF
--- a/pkg/windows/windows.go
+++ b/pkg/windows/windows.go
@@ -186,8 +186,8 @@ type Windows interface {
 	Reinitialize() error
 	// Configure prepares the Windows VM for the bootstrapper and then runs it
 	Configure() error
-	// ConfigureCNI ensures that the CNI configuration in done on the node
-	ConfigureCNI(string) error
+	// EnsureCNIConfig ensures that the CNI config file is present on the node
+	EnsureCNIConfig(string) error
 	// ConfigureHybridOverlay ensures that the hybrid overlay is running on the node
 	ConfigureHybridOverlay(string) error
 	// ConfigureWindowsExporter ensures that the Windows metrics exporter is running on the node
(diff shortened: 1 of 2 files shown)


pkg/windows/windows.go
```diff
@@ -541,17 +541,7 @@ func (vm *windows) ConfigureCNI(configFile string) error {
...
-	// run the configure-cni command on the Windows VM
-	configureCNICmd := k8sDir + "wmcb.exe configure-cni --cni-dir=" + cniDir + " --cni-config=" + cniConfigDest
-
-	out, err := vm.Run(configureCNICmd, true)
-	if err != nil {
-		return errors.Wrap(err, "CNI configuration failed")
-	}
-
-	vm.log.Info("configured kubelet for CNI", "cmd", configureCNICmd, "output", out)
+	vm.log.Info("CNI config file copied", "remote dir", cniConfDir)
```
sebsoto, line 544: This info is already logged by EnsureFile()

pkg/windows/windows.go
```diff
@@ -532,7 +532,7 @@ func (vm *windows) ConfigureHybridOverlay(nodeName string) error {
 	return nil
 }
 
-func (vm *windows) ConfigureCNI(configFile string) error {
+func (vm *windows) EnsureCNI(configFile string) error {
```
sebsoto, line 535: This can be renamed to `EnsureCNIConfig` as there are other responsibilities such as copying CNI plugins that this function does not handle

sebsoto (changes requested):

PR #1081, opened 2022-05-26
WINC-746: Enable running WICD as a Windows service

sebsoto:
Adds the --windows-service flag to enable running as a Windows service.
Also adds the --log-dir flag so that logs can be accessible when running as a Windows service.

### DIFF
--- a/cmd/daemon/controller.go
+++ b/cmd/daemon/controller.go
@@ -19,6 +19,7 @@ limitations under the License.
 package main
 
 import (
+	"flag"
 	"os"
 
 	"github.com/spf13/cobra"
@@ -38,16 +39,26 @@ var (
 	}
 	kubeconfig     string
 	windowsService bool
+	logDir         string
 )
 
 func init() {
 	rootCmd.AddCommand(controllerCmd)
 	controllerCmd.PersistentFlags().StringVar(&kubeconfig, "kubeconfig", "", "Path to kubeconfig")
+	controllerCmd.PersistentFlags().StringVar(&logDir, "log-dir", "", "Directory to write logs to, "+
+		"if not provided, the command will log to stdout/stderr")
 	controllerCmd.PersistentFlags().BoolVar(&windowsService, "windows-service", false,
 		"Enables running as a Windows service")
 }
 
 func runControllerCmd(cmd *cobra.Command, args []string) {
+	if logDir != "" {
+		var fs flag.FlagSet
+		klog.InitFlags(&fs)
+		// When the logtostderr flag is set to true, which is the default, the log_dir arg is ignored
+		fs.Set("logtostderr", "false")
+		fs.Set("log_dir", logDir)
+	}
 	ctx := ctrl.SetupSignalHandler()
 	if windowsService {
 		if err := initService(ctx); err != nil {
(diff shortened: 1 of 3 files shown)


cmd/daemon/controller.go
```diff
@@ -35,17 +37,35 @@ var (
...
+	controllerCmd.PersistentFlags().BoolVar(&windowsService, "windows-service", false,
+		"Enables running as a Windows service")
 }
 
 func runControllerCmd(cmd *cobra.Command, args []string) {
+	if logDir != "" {
+		var fs flag.FlagSet
+		klog.InitFlags(&fs)
+		fs.Set("logtostderr", "false")
+		fs.Set("log_dir", logDir)
```
jrvaldes, line 58: Is there a default `log_dir` value, in case no `logDir` is provided? If so, you should mention it in the command's [description](https://github.com/openshift/windows-machine-config-operator/pull/1081/files#diff-dcdb946703bcd804e30b1506a5a49888d8945bef67c7f354d1a52559b90bc381R48).

cmd/daemon/service.go
```diff
@@ -0,0 +1,72 @@
...
+			break loop
+		case c := <-r:
+			switch c.Cmd {
+			case svc.Interrogate:
+				// The Interrogate command is always accepted by default
+				s <- c.CurrentStatus
+			case svc.Stop, svc.Shutdown:
+				klog.Info("Service stopping")
+				s <- svc.Status{State: svc.StopPending}
+				cancel()
```
jrvaldes, line 66: Is the context still valid after invoking the cancellation function (`cancel()`) ? . You may need to just `break` instead of `break loop`.

cmd/daemon/controller.go
saifshaikh48: +1

saifshaikh48 (commented):
/lgtm aside from the logDir default question

cmd/daemon/controller.go
```diff
@@ -38,16 +39,24 @@ var (
 	}
 	kubeconfig     string
 	windowsService bool
+	logDir         string
 )
 
 func init() {
 	rootCmd.AddCommand(controllerCmd)
 	controllerCmd.PersistentFlags().StringVar(&kubeconfig, "kubeconfig", "", "Path to kubeconfig")
+	controllerCmd.PersistentFlags().StringVar(&logDir, "log-dir", "", "Directory to write logs to")
```
selansen, line 48: Can we add some default value here?

sebsoto: Setting the default value to an empty string is standard as it defaults to log to stdout/stderr.

saifshaikh48 (commented):
/lgtm

cmd/daemon/service.go
sebsoto: Not sure I follow, but in any case we want to return from this function so the program can stop in the case of a `stop` request

cmd/daemon/service.go
```diff
@@ -0,0 +1,72 @@
+//go:build windows
+// +build windows
```
mansikulkarni96, line 2: The functions in this file have been taken from the windows/svc source code right? Should we add a reference for that in the file doc?

alinaryan: +1

sebsoto: Not exactly, I did look at what was being done there for some help with the waitgroups, but the code is mostly the same as the example given in the svc library https://github.com/golang/sys/blob/master/windows/svc/example/service.go

mansikulkarni96:
/lgtm

PR #1086, opened 2022-05-31
[servicescm] Service dependency cycle detection

saifshaikh48:
This PR adds cycle detection to the validation that occurs
on the services defined in `windows-services` ConfigMaps.
A dependency chain with cycles is considered invalid.

### DIFF
--- a/pkg/servicescm/servicescm.go
+++ b/pkg/servicescm/servicescm.go
@@ -247,6 +247,7 @@ func (f FileInfo) isPresentAndCorrect(files []FileInfo) bool {
 }
 
 // validateDependencies ensures that no bootstrap service depends on a non-bootstrap service
+// and ensures there is no cyclical dependency chain
 func validateDependencies(services []Service) error {
 	bootstrapServices := []Service{}
 	nonBootstrapServices := []Service{}
@@ -263,7 +264,8 @@ func validateDependencies(services []Service) error {
 			return errors.Errorf("bootstrap service %s cannot depend on non-bootstrap service", bootstrapSvc.Name)
 		}
 	}
-	return nil
+
+	return validateCycles(services)
 }
 
 // hasDependency checks if a service is dependent on any services in the given slice
@@ -275,6 +277,50 @@ func (s *Service) hasDependency(possibleDependencies []Service) bool {
 			}
 		}
 	}
+	return false
+}
+
+// validateCycles detects cycles between any of the given services by traversing the resulting dependency graph.
+// Wrapper for Service.hasCycle to handle disconnected graphs
+func validateCycles(services []Service) error {
+	// Convert list to map for fast lookup of service object by its name
+	servicesMap := make(map[string]*Service)
+	for _, svc := range services {
+		servicesMap[svc.Name] = &svc
+	}
+
+	// state is a map that keeps track of whether a service's dependency chain has been explored for cycles:
+	// 1. if a service does not have an entry in the map, it has not been processed yet
+	// 2. if a service has an entry in the map with value "true", it is currently being processed
+	// 3. if a service has an entry in the map with value "false", it has already been fully processed in the past
+	state := make(map[string]bool)
+	for _, svc := range services {
+		// Check if helper has already been called on this service to prevent duplicate calls
+		if _, seen := state[svc.Name]; !seen {
+			if svc.hasCycle(servicesMap, state) {
+				return errors.Errorf("invalid cyclical chain in %s service's dependencies", svc.Name)
+			}
+		}
+	}
+	return nil
+}
+
+// hasCycle uses depth-first traversal to check for cycles in the service dependency graph, using s as the source node
+func (s *Service) hasCycle(servicesMap map[string]*Service, state map[string]bool) bool {
+	// Mark this service as visited and in the current traversal path
+	state[s.Name] = true
+
+	for _, dependencyName := range s.Dependencies {
+		if inCurrentPath, seen := state[dependencyName]; seen && inCurrentPath {
+			// Cycle detected if service that is still being processed is seen again in the same dependency path
+			return true
+		}
+		if _, seen := state[dependencyName]; !seen {
+			return servicesMap[dependencyName].hasCycle(servicesMap, state)
+		}
+	}
+	// Backtracking step, remove this service from current traversal path by marking it as fully processed
+	state[s.Name] = false
 	return false
 }
 
--- a/pkg/servicescm/servicescm_test.go
+++ b/pkg/servicescm/servicescm_test.go
@@ -153,7 +153,7 @@ func TestValidateDependencies(t *testing.T) {
 				{
 					Name:         "test-controller-service-2",
 					Command:      "C:\\test-controller-service-2",
-					Dependencies: []string{"test-controller-service-2"},
+					Dependencies: []string{"test-controller-service"},
 					Bootstrap:    false,
 					Priority:     2,
 				},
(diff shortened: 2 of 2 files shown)


selansen:
Thanks for the review discussion. 

/approve

pkg/servicescm/servicescm.go
```diff
@@ -278,6 +283,48 @@ func (s *Service) hasDependency(possibleDependencies []Service) bool {
...
+	}
+
+	visited := make(map[string]struct{})
+	currentPath := make(map[string]struct{})
+	for _, svc := range services {
+		// Check if helper has already been called on this service to prevent duplicate calls
+		if _, present := visited[svc.Name]; !present {
+			if cycleDetected := svc.hasCycle(servicesMap, visited, currentPath); cycleDetected {
+				return true
+			}
```
jrvaldes, line 302:
```suggestion
			if svc.hasCycle(servicesMap, visited, currentPath) {
				return true
			}
```

pkg/servicescm/servicescm.go
```diff
@@ -278,6 +283,48 @@ func (s *Service) hasDependency(possibleDependencies []Service) bool {
 	return false
 }
 
+// hasCycle detects cycles between any of the given services by traversing the resulting dependency graph.
+// Wrapper for Service.hasCycle to handle disconnected graphs
```
jrvaldes, line 287: Can you add more color around the "disconnected graphs" case?

pkg/servicescm/servicescm_test.go
```diff
@@ -153,7 +153,7 @@ func TestValidateDependencies(t *testing.T) {
 				{
 					Name:         "test-controller-service-2",
 					Command:      "C:\\test-controller-service-2",
-					Dependencies: []string{"test-controller-service-2"},
+					Dependencies: []string{"test-controller-service"},
```
jrvaldes, line 156: Is this change related to the cyclical dependency test?

pkg/servicescm/servicescm_test.go
```diff
@@ -238,6 +238,60 @@ func TestValidateDependencies(t *testing.T) {
...
 		},
+		{
+			name: "Cyclical dependency structure",
+			input: []Service{
+				{
+					Name:         "new-bootstrap-service",
+					Command:      "C:\\new-service",
+					Dependencies: []string{"test-controller-service"},
+					Bootstrap:    false,
+					Priority:     0,
```
jrvaldes, line 249: It may be easier to read if you use the priority value to fill all the fields. For example:

```
					Name:         "service-0",
					Command:      "C:\\service-0",
					Dependencies: []string{"service-1"},
					Bootstrap:    false,
					Priority:     0,
```

pkg/servicescm/servicescm.go
```diff
@@ -278,6 +283,48 @@ func (s *Service) hasDependency(possibleDependencies []Service) bool {
...
+		if _, present := visited[svc.Name]; !present {
+			if cycleDetected := svc.hasCycle(servicesMap, visited, currentPath); cycleDetected {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// hasCycle uses depth-first traversal to check for cycles in the service dependency graph, using s as the source node
```
jrvaldes, line 308: Did you considered a [tortoise and hare](https://en.wikipedia.org/wiki/Cycle_detection#Floyd's_tortoise_and_hare)-like approach for finding the cycle?

pkg/servicescm/servicescm.go
```diff
@@ -278,6 +283,48 @@ func (s *Service) hasDependency(possibleDependencies []Service) bool {
...
+	visited[s.Name] = struct{}{}
+	currentPath[s.Name] = struct{}{}
+
+	for _, dependencyName := range s.Dependencies {
+		if _, present := currentPath[dependencyName]; present {
+			// Cycle detected if service already exists in current path
+			return true
+		}
+		if _, present := visited[dependencyName]; !present {
+			return servicesMap[dependencyName].hasCycle(servicesMap, visited, currentPath)
```
jrvaldes, line 320: In the best-case scenario (no cycles) `servicesMap` and `visited` have the same keys, consider using `servicesMap` as the primary lookup entity so that you reduce the extra space complexity by removing the `visited` set.

pkg/servicescm/servicescm_test.go
```diff
@@ -238,6 +238,60 @@ func TestValidateDependencies(t *testing.T) {
...
+			expectedErr: true,
+		},
+		{
+			name: "Disconnect cyclical dependency structure",
+			input: []Service{
+				{
+					Name:         "new-bootstrap-service",
+					Command:      "C:\\new-service",
+					Dependencies: []string{},
+					Bootstrap:    false,
```
jrvaldes, line 275: Based on the name (`new-bootstrap-service`), shouldn't this be `true`?

pkg/servicescm/servicescm_test.go
```diff
@@ -238,6 +238,60 @@ func TestValidateDependencies(t *testing.T) {
...
+					Command:      "C:\\test-controller-service-2",
+					Dependencies: []string{"new-bootstrap-service"},
+					Bootstrap:    false,
+					Priority:     2,
+				},
+			},
+			expectedErr: true,
+		},
+		{
+			name: "Disconnect cyclical dependency structure",
```
jrvaldes, line 269: Why disconnected? IMO, this test is equivalent to the first one (`Cyclical dependency structure`).

pkg/servicescm/servicescm_test.go
```diff
@@ -238,6 +238,60 @@ func TestValidateDependencies(t *testing.T) {
...
+				{
+					Name:         "test-controller-service-2",
+					Command:      "C:\\test-controller-service-2",
+					Dependencies: []string{"test-controller-service"},
+					Bootstrap:    false,
+					Priority:     2,
+				},
+			},
+			expectedErr: true,
+		},
```
jrvaldes, line 294: Can a service depend on itself? Is this case covered?

```
				{
					Name:         "test-service",
					Command:      "C:\\test-service",
					Dependencies: []string{"test-service"},
					Bootstrap:    false,
					Priority:     0,
				},
```

jrvaldes (commented):
Nice work @saifshaikh48, added a few comments. PTAL

pkg/servicescm/servicescm.go
```diff
@@ -278,6 +283,48 @@ func (s *Service) hasDependency(possibleDependencies []Service) bool {
 	return false
 }
 
+// hasCycle detects cycles between any of the given services by traversing the resulting dependency graph.
+// Wrapper for Service.hasCycle to handle disconnected graphs
+func hasCycle(services []Service) bool {
```
sebsoto, line 288: please rename this to something equivalent, would prefer to not have multiple functions with the same name if they are not due to an interface/generics

`validateCycles` perhaps? That way you can also return an error and state which service in particular has the cycle.

pkg/servicescm/servicescm.go
sebsoto: +1 on a solution that does not use recursion.
Recursion can feel elegant, but we should prefer simplicity when we can, especially since we don't need these functions to be performant.
In many cases recursion makes things harder to read and visualize, especially when anticipating edge cases. This gets increasingly worse as the requirements of our code evolves. Granted its early, but its taking me a long time to figure out whats going on here, compared to if this was done with some simple loops.

pkg/servicescm/servicescm_test.go
saifshaikh48: A service that depends on itself is invalid as it is considered a cycle. I'll add a unit test around this

pkg/servicescm/servicescm_test.go
saifshaikh48: I'll fix this, though it has no effect on the test case

pkg/servicescm/servicescm_test.go
saifshaikh48: Yes, this test case needed to be modified since a service cannot depend on itself.

pkg/servicescm/servicescm.go
saifshaikh48: Consider 3 services: A, B, and C
- A has no dependencies, B depends on C, C depends on B. 
- The resulting graph has 2 distinct, disconnected components: one is {A} and the other is {B,C}

As such, we cannot chose any node to call Service.hasCycles() on. If we simply did A.hasCycles(), it would return false despite the B<-->C cycle. So we need to call the method once on each subgraph.

pkg/servicescm/servicescm.go
saifshaikh48: I'm going to push back on this after exploring the iterative solution. The iterative solution is quite hard to follow itself and requires a stack (not hard to implement, just adds avoidable custom logic)

sebsoto: okay, thats fine with me, I'll remove the hold
thanks for exploring the other solution

selansen:
I did review the code with @saifshaikh48  to go through DFP graph loops. It gave me the opportunity to refresh my memory.
But I am +1 for making it simpler. 
@saifshaikh48 do you still want to merge this PR and rework a more straightforward solution later?
As this is not a blocker for WICD work, I suggest you keep this PR open and work on the alternate solution rather than pushing and then replacing it with an alternate solution.

pkg/servicescm/servicescm.go
saifshaikh48: After revisiting this again, I was able to find a way to consolidate the 2 maps using a variant of the [color-coded backtracking](https://www.geeksforgeeks.org/detect-cycle-direct-graph-using-colors/) algorithm. Thanks for the suggestion Jose.

saifshaikh48:
@selansen I did a fair bit of investigation here and tried to write up a few alternate solutions. A recursive DFS strategy is definitely the best suited/simplest for this problem.

jrvaldes:
LGTM

mansikulkarni96:
LGTM will wait for @sebsoto to give final lgtm

PR #1090, opened 2022-06-03
Wicd bootstrap command 2

saifshaikh48:
This PR adds WICD's `bootstrap` command, a one-shot operation 
that starts the primary OpenShift-managed Windows services.
This command results in the creation of a Node object for the instance it is ran on. 

A Node object is created as a result of the kubelet service running and the appropriate CSRs being approved. 
WICD therefore has to create the kubelet and all other services required for node object creation.


saifshaikh48:
/test wicd-unit-vsphere

fixed example service in tests

pkg/daemon/controller/controller.go
```diff
@@ -153,6 +153,20 @@ func (sc *ServiceController) SetupWithManager(mgr ctrl.Manager) error {
 		Complete(sc)
 }
 
+// Bootstrap starts all required Windows services on this instance. This will result in a node object being created
```
sebsoto, line 156: a node object won't be created until the csr's are approved

pkg/daemon/controller/controller.go
```diff
@@ -153,6 +153,20 @@ func (sc *ServiceController) SetupWithManager(mgr ctrl.Manager) error {
...
+func (sc *ServiceController) Bootstrap(services []servicescm.Service) error {
+	bootstrapServices := []servicescm.Service{}
+	for _, svc := range services {
+		if svc.Bootstrap {
+			bootstrapServices = append(bootstrapServices, svc)
+		} else {
+			// services are pre-sorted by priority, with all bootstrap services ordered towards the front of the slice
+			break
+		}
```
sebsoto, line 165: reduce indentation by handling the negative case first
the extra else is unnecessary 
```
if !svc.Bootstrap {
  break
}
bootstrapServices = append(bootstrapServices, svc)
```

cmd/daemon/bootstrap.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	svcMgr, err := winsvc.NewMgr()
+	if err != nil {
+		klog.Exitf(err.Error())
+	}
+	// TODO: for now, no client is needed for bootstrapping. This will not hold when WICD can get the service ConfigMap
+	serviceController := controller.NewServiceController(ctrl.SetupSignalHandler(), nil, svcMgr, "")
+
+	klog.Info("starting services to bootstrap node")
```
sebsoto, line 63: Moving this code into a function such as `RunBootstrap` in the controller package would be preferred as this is logic around parsing the services, and not about running the actual exe/flags.

pkg/daemon/controller/controller.go
sebsoto: Logically this should be done at the same time as
```
cmData, err := servicescm.DecodeAndUnmarshall(encodedCMData)
```

pkg/daemon/controller/controller.go
```diff
@@ -315,6 +329,10 @@ func (sc *ServiceController) expectedServiceCommand(expected servicescm.Service)
 // resolveNodeVariables returns a map, with the keys being each variable, and the value being the string to replace the
 // variable with
 func (sc *ServiceController) resolveNodeVariables(svc servicescm.Service) (map[string]string, error) {
+	if sc.nodeName == "" {
+		return nil, errors.New("instance has no associated node")
+	}
```
sebsoto, line 334: I feel this is unnecessary as this will be caught below in the Get(), and this is not going to cause a panic as it is not a pointer. I don't think its a good pattern to validate that struct fields (excluding pointers), are set before they are used in a method.

This should never occur, instead the service CM validation should ensure that bootstrap services don't rely on node variables, which is what you are doing in the other commit

pkg/daemon/controller/controller_test.go
```diff
@@ -222,6 +222,118 @@ func TestReconcileService(t *testing.T) {
...
+
+			createdServiceNameCmdPairs := make(map[string]string)
+			for name, createdService := range createdServices {
+				serviceStatus, _ := createdService.Query()
+				assert.Equal(t, svc.Running, serviceStatus.State)
+
+				config, err := createdService.Config()
+				require.NoError(t, err)
+				createdServiceNameCmdPairs[name] = config.BinaryPathName
+			}
```
sebsoto, line 330: This is repeated code w/ the Reconcile test
it probably should be made into a helper used by both of them

`getFakeServiceCmdNamePairs` or something

sebsoto (changes requested):
Good work, a few comments

saifshaikh48:
Manually tested the executable on a Windows instance:
```
.\windows-instance-config-daemon.exe bootstrap --services-data $base64_data
```
- I used data.MarshallAndEncode to generate encoded string for a services CM that had 2 services, one being a bootstrap service
```
{
	Name:         "containerd",
	Command:      "\"C:\\Program Files\\containerd\\containerd.exe\" --run-service",
	Dependencies: nil,
	Bootstrap:    true,
	Priority:     0,
},
```

- creates/starts only the bootstrap service with expected config as expected

pkg/daemon/controller/controller.go
```diff
@@ -153,6 +153,30 @@ func (sc *ServiceController) SetupWithManager(mgr ctrl.Manager) error {
...
+
+	bootstrapSvcs := []servicescm.Service{}
+	for _, svc := range cmData.Services {
+		if !svc.Bootstrap {
+			// services are pre-sorted by priority, with all bootstrap services ordered towards the front of the slice
+			break
+		}
+		bootstrapSvcs = append(bootstrapSvcs, svc)
+	}
+	return serviceController.reconcileServices(bootstrapSvcs)
```
saifshaikh48, line 177: Open question: since `RunBootstrap()` is majorly functions a wrapper for `reconcileServices()` do the existing reconcileService unit tests suffice?

pkg/daemon/controller/controller.go
```diff
@@ -153,6 +153,30 @@ func (sc *ServiceController) SetupWithManager(mgr ctrl.Manager) error {
 		Complete(sc)
 }
 
+// RunBootstrap starts all Windows services that are pre-requisites to create a node object as defined in the given data
+func RunBootstrap(encodedCMData string) error {
+	cmData, err := servicescm.DecodeAndUnmarshall(encodedCMData)
+	if err != nil {
+		return errors.Wrap(err, "unable to decode ConfigMap data")
+	}
```
sebsoto, line 161: please move this down to where it will be used

pkg/daemon/controller/controller_test.go
```diff
@@ -330,9 +330,12 @@ func TestReconcile(t *testing.T) {
 			createdServices, err := getAllFakeServices(winSvcMgr)
 			require.NoError(t, err)
 
-			// Specifically testing that the name/command is as expected
+			// Specifically testing that the service is running and the name/command is as expected
 			createdServiceNameCmdPairs := make(map[string]string)
 			for name, createdService := range createdServices {
+				serviceStatus, _ := createdService.Query()
+				assert.Equal(t, svc.Running, serviceStatus.State)
+
```
sebsoto, line 338: Good changes but should be in a separate commit

pkg/daemon/controller/controller.go
sebsoto: Yes if the 'wrapper' doesn't have any real logic.
The presence of this:
```
	cmData, err := servicescm.DecodeAndUnmarshall(encodedCMData)
	if err != nil {
		return errors.Wrap(err, "unable to decode ConfigMap data")
	}

	bootstrapSvcs := []servicescm.Service{}
	for _, svc := range cmData.Services {
		if !svc.Bootstrap {
			// services are pre-sorted by priority, with all bootstrap services ordered towards the front of the slice
			break
		}
		bootstrapSvcs = append(bootstrapSvcs, svc)
	}
```
makes it less of a wrapper, and i'm somewhat inclined to think this needs to be unit tested.

I'm back and forth on wondering if the signature of `reconcileServices` should be changed to accept a `bootstrap bool`, so that  that way there's less stuff going on in this wrapper.

sebsoto (changes requested):

pkg/daemon/controller/controller.go
saifshaikh48: What I'll do is make the new signature `RunBootstrap(svcMgr winsvc.Mgr, encodedCMData string)` to make it unit testable, and leave the bootstrap logic here. `reconcileServices` does enough on it's own as a single func

saifshaikh48:
/test wicd-unit-vsphere

fixed a context issue, passed on a local instance

cmd/daemon/bootstrap.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/openshift/windows-machine-config-operator/pkg/daemon/controller"
+	"github.com/openshift/windows-machine-config-operator/pkg/daemon/winsvc"
+)
+
+var (
+	bootstrapCmd = &cobra.Command{
+		Use:   "bootstrap",
+		Short: "Starts required Windows services to bootstrap a Node",
+		Long: "Starts all Windows services on an instance that are pre-requisites for Node object creation, " +
+			"configured according to the information given by the Windows services ConfigMap",
```
jrvaldes, line 34: s/Windows services ConfigMap/ windows-services ConfigMap

cmd/daemon/bootstrap.go
```diff
@@ -0,0 +1,58 @@
...
+		Run: runBootstrapCmd,
+	}
+	// encodedCMData holds encoded data from the Windows services ConfigMap
+	encodedCMData string
+)
+
+func init() {
+	rootCmd.AddCommand(bootstrapCmd)
+	bootstrapCmd.PersistentFlags().StringVar(&encodedCMData, "services-data", "", "Windows services ConfigMap data")
+	bootstrapCmd.MarkPersistentFlagRequired("services-data")
```
jrvaldes, line 44: Consider adding in the command's description that this flag is a required one.

pkg/daemon/controller/controller.go
```diff
@@ -153,6 +153,25 @@ func (sc *ServiceController) SetupWithManager(mgr ctrl.Manager) error {
 		Complete(sc)
 }
 
+// RunBootstrap starts all Windows services that are pre-requisites to create a node object as defined in the given data
+func RunBootstrap(svcMgr winsvc.Mgr, encodedCMData string) error {
```
alinaryan, line 157: runBootstrapCmd() and RunBootstrap() sound like they do the same thing based on their descriptions, but they don't. Could the description or naming be changed to differentiate?

jrvaldes (commented):
@saifshaikh48  thanks for working on this, mostly LGTM. Added minor suggestions.

cmd/daemon/bootstrap.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/openshift/windows-machine-config-operator/pkg/daemon/controller"
+	"github.com/openshift/windows-machine-config-operator/pkg/daemon/winsvc"
+)
+
+var (
+	bootstrapCmd = &cobra.Command{
+		Use:   "bootstrap",
+		Short: "Starts required Windows services to bootstrap a Node",
+		Long: "Starts all Windows services on an instance that are pre-requisites for Node object creation, " +
```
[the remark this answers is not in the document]
saifshaikh48: They're just descriptions of what the `bootstrap` command does to help a user:
```ruby
    // Short is the short description shown in the 'help' output.
    Short string

    // Long is the long message shown in the 'help <this-command>' output.
    Long string
```

pkg/daemon/controller/controller.go
saifshaikh48: `runBootstrapCmd()` is essentially a wrapper for `RunBootstrap()`, and the split is only really needed to allow unit testing

cmd/daemon/bootstrap.go
jrvaldes: Makes sense.

pkg/daemon/controller/controller_test.go
```diff
@@ -222,6 +222,109 @@ func TestReconcileService(t *testing.T) {
...
+			require.NoError(t, err)
+			encodedData, err := data.MarshallAndEncode()
+			require.NoError(t, err)
+
+			winSvcMgr := winsvc.NewTestMgr(make(map[string]*winsvc.FakeService))
+			err = RunBootstrap(winSvcMgr, encodedData)
+			if test.expectErr {
+				assert.Error(t, err)
+				return
+			}
```
sebsoto, line 319: I dont see any expected failure cases in the test list.
This, along with the `expectErr` in the test struct should be removed in this case.

sebsoto:
/approve
with the condition that this is [handled](https://github.com/openshift/windows-machine-config-operator/pull/1090#discussion_r909971285)

cmd/daemon/bootstrap.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+func init() {
+	rootCmd.AddCommand(bootstrapCmd)
+	bootstrapCmd.PersistentFlags().StringVar(&desiredVersion, "desired-version", "",
+		"Version of the services ConfigMap to use as the source of truth for service configuration")
+	bootstrapCmd.MarkPersistentFlagRequired("desired-version")
+}
+
+// runBootstrapCmd starts all Windows services that are pre-requisites for node object creation taken from encoded arg
```
saifshaikh48, line 52: Sorry, encoded arg doesn't exist anymore. I'll fix this comment after a round of reviews

cmd/daemon/bootstrap.go
```diff
@@ -0,0 +1,81 @@
...
+		klog.Exit(err.Error())
+	}
+
+	svcMgr, err := winsvc.NewMgr()
+	if err != nil {
+		klog.Exitf("could not create service manager: %s", err.Error())
+	}
+	sc := controller.NewServiceController(context.TODO(), directClient, svcMgr, "")
+
+	klog.Info("starting services to bootstrap node")
```
sebsoto, line 77: nit: can be condensed to `bootstrapping node`

pkg/daemon/controller/controller.go
```diff
@@ -67,13 +67,37 @@ type ServiceController struct {
...
+	}
+
+	bootstrapSvcs := []servicescm.Service{}
+	for _, svc := range cmData.Services {
+		if !svc.Bootstrap {
+			// services are pre-sorted by priority, with all bootstrap services ordered towards the front of the slice
+			break
+		}
+		bootstrapSvcs = append(bootstrapSvcs, svc)
+	}
```
sebsoto, line 90: can break this out into a new function for cleaner code
`func getBootstrapServices(data) []servicescm.Service{}`

saifshaikh48: I feel like getBootstrapServices() is better suited as a function on the `data` object (servicescm package). I'll add this in

pkg/daemon/config/config.go
```diff
@@ -0,0 +1,48 @@
...
+	"k8s.io/client-go/rest"
+	certutil "k8s.io/client-go/util/cert"
+	"k8s.io/klog/v2"
+)
+
+// FromServiceAccount uses service account credentials to create a client config able to authenticate with a cluster
+func FromServiceAccount(apiServerURL, caFile, tokenFile string) (*rest.Config, error) {
+	token, err := ioutil.ReadFile(tokenFile)
+	if err != nil {
+		klog.Errorf("error reading token file: %v", err)
```
mansikulkarni96, line 33: Knowing that this was a part of the previous version of this function
Is the logging required here if we can return the same error?

pkg/servicescm/servicescm.go
```diff
@@ -171,26 +170,16 @@ func Parse(dataFromCM map[string]string) (*data, error) {
...
-	return base64.StdEncoding.EncodeToString(marshalled), nil
-}
-
-// DecodeAndUnmarshall takes in a JSON marshalled base64 encoded string and returns the decoded data
-// TODO: Remove this when the WICD controller has permissions to watch ConfigMaps
-func DecodeAndUnmarshall(base64Encoded string) (*data, error) {
-	marshalledData, err := base64.StdEncoding.DecodeString(base64Encoded)
-	if err != nil {
-		return nil, errors.Wrap(err, "unable to decode ConfigMap data")
+func (cmData *data) GetBootstrapServices() []Service {
```
mansikulkarni96, line 173: missing doc

saifshaikh48:
@aravindhp could you take a look at this PR?

pkg/servicescm/servicescm_test.go
```diff
@@ -161,17 +155,37 @@ func TestValidateDependencies(t *testing.T) {
 			expectedErr: false,
 		},
 		{
-			name: "bootstrap depends on all non-bootstrap services",
+			name: "bootstrap service requires node variable in command",
```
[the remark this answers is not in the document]
saifshaikh48: The bootstrap service here requires a node variable substitution in it's command, which is invalid. Since we have `expectedErr: true` for this test case, the naming can stay IMO. Let me know if you feel strongly otherwise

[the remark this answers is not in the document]
aravindhp: The confusion is that the test name makes it seem it is valid to have a node variable in the bootstrap command. But I am not going to hold the PR due to this.

saifshaikh48:
/hold

for enhancement update https://github.com/openshift/enhancements/pull/1157

saifshaikh48:
/retest-required

Reported the ci-operator build issue to testplatform. Retesting as I see it has resolved itself for some other teams

PR #1092, opened 2022-06-06
WINC-833: [e2e] unblock aws CI failure

selansen:
This commit addresses CI failures on AWS platform.
the AMI which is created on May 25th is causing the issue, so
as a temp workaround we will be pointing to the older AMI image until
we figure out what is the real issue with the new AMI.

### DIFF
--- a/test/e2e/providers/aws/aws.go
+++ b/test/e2e/providers/aws/aws.go
@@ -120,6 +120,8 @@ func getLatestWindowsAMI(ec2Client *ec2.EC2, hasCustomVXLANPort bool) (string, e
 	windowsAMIOwner := "amazon"
 	windowsAMIFilterName := "name"
 	windowsAMIFilterValue := ""
+	winDateFilterName := "creation-date"
+	winDateFilterVal := "2022-05-11T*"
 	// This filter will grab all ami's that match the exact name. The '?' indicate any character will match.
 	// The ami's will have the name format: Windows_Server-2022-English-Full-ContainersLatest-2022.01.19
 	// so the question marks will match the date of creation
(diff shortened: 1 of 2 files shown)


test/e2e/providers/aws/aws.go
```diff
@@ -146,8 +145,9 @@ func getLatestWindowsAMI(ec2Client *ec2.EC2, hasCustomVXLANPort bool) (string, e
 	}
 
 	// Find the last created image
-	latestImage := describedImages.Images[0]
-	latestTime, err := time.Parse(time.RFC3339, *latestImage.CreationDate)
+	latestImage := describedImages.Images[2]
+	// WINC-833 temp fix. We don't need comp logic at present.
+	/*latestTime, err := time.Parse(time.RFC3339, *latestImage.CreationDate)
```
sebsoto, line 150: without this logic this will be picking a random image from the list wouldnt it?
potentially being the broken image?

selansen: As per the comment, it says `// Find the last created image`. It is stored in 0th index. So I assumed it is already sorted. Hence I am picking up from 2nd index.

sebsoto: My understanding is the list is not sorted, otherwise we wouldn't have needed the logic being commented out. 
`Find the last created image` would be referring to this entire code section

test/e2e/providers/aws/aws.go
```diff
@@ -133,9 +135,10 @@ func getLatestWindowsAMI(ec2Client *ec2.EC2, hasCustomVXLANPort bool) (string, e
 		windowsAMIFilterValue = "Windows_Server-2019-English-Full-ContainersLatest-????.??.??"
```
jrvaldes, line 135: As the date will be hard-coded, why not do it here (`????.??.??`)? And how different is this approach from hard-coding the specific AMI id?

selansen: @jrvaldes like I explained we can't hard-code as the  AMI is different for each region. 
Also, this is a temp fix we will be getting rid of it as soon as the new AMI works or we identify the root cause.  

@sebsoto you are 100% right. Once we make CI green, next task is to see how we can escalate AMI issue n fix it

selansen:
Azure failure is a known issue and the platform team is looking into it. 
Will do force merge to unblock networking team CI failures.

PR #1100, opened 2022-06-22
Bump the verbosity of the hybrid-overlay service in debug mode

jrvaldes:
This PR aims to adjust the hybrid-overlay log verbosity according
to the controller-runtime logging level.

### DIFF
--- a/pkg/windows/windows.go
+++ b/pkg/windows/windows.go
@@ -496,6 +496,13 @@ func (vm *windows) ConfigureHybridOverlay(nodeName string) error {
 	hybridOverlayServiceArgs := "--node " + nodeName + customVxlanPortArg + " --k8s-kubeconfig c:\\k\\kubeconfig " +
 		"--windows-service " + "--logfile " + hybridOverlayLogDir + "hybrid-overlay.log"
 
+	// check log level and increase hybrid-overlay verbosity if needed
+	if vm.log.V(1).Enabled() {
+		// append loglevel param using 5 for debug (default: 4)
+		// See https://github.com/openshift/ovn-kubernetes/blob/master/go-controller/pkg/config/config.go#L736
+		hybridOverlayServiceArgs = hybridOverlayServiceArgs + " --loglevel 5"
+	}
+
 	vm.log.Info("configure", "service", hybridOverlayServiceName, "args", hybridOverlayServiceArgs)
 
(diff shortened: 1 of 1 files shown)


pkg/windows/windows.go
```diff
@@ -496,6 +496,12 @@ func (vm *windows) ConfigureHybridOverlay(nodeName string) error {
 	hybridOverlayServiceArgs := "--node " + nodeName + customVxlanPortArg + " --k8s-kubeconfig c:\\k\\kubeconfig " +
 		"--windows-service " + "--logfile " + hybridOverlayLogDir + "hybrid-overlay.log"
 
+	// check log level and increase hybrid-overlay verbosity if needed
+	if ctrl.Log.V(1).Enabled() {
```
sebsoto, line 500: Please use windows.log for this check

pkg/windows/windows.go
```diff
@@ -496,6 +496,12 @@ func (vm *windows) ConfigureHybridOverlay(nodeName string) error {
...
+		// append loglevel param using 5 for debug (default: 4)
+		hybridOverlayServiceArgs = hybridOverlayServiceArgs + " --loglevel 5"
```
sebsoto, line 502: if the default is 4 wouldnt raising it to 5 make it less verbose?

sebsoto (changes requested):

pkg/windows/windows.go
jrvaldes: Will add a reference link to the [code](https://github.com/openshift/ovn-kubernetes/blob/master/go-controller/pkg/config/config.go#L736) for clarification.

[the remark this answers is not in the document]
jrvaldes: See [code](https://github.com/openshift/ovn-kubernetes/blob/master/go-controller/pkg/config/config.go#L736) for details.

pkg/windows/windows.go
jrvaldes: Addressed in [41fa41a ](https://github.com/openshift/windows-machine-config-operator/commit/41fa41ae3430f6f999fe1c15ddd576d204283dee)

jrvaldes:
/hold cancel 

https://github.com/openshift/windows-machine-config-operator/pull/1102 merged and CI is back to green.

Issue #1107, opened 2022-06-28
Bootstrap in vSphere Environment Fails if Windows Defender Feature is not Installed 


MattPOlson:
**Describe the bug**
In a vSphere environment if the Windows Defender Feature is not Installed in the golden image the operator fails to bootstrap the node. It tries to configure the windows defender exclusion for the containerd service by calling Add-MpPreference. Add-MpPreference is only available if the Windows Defender Feature is installed.

`1.656434951067332e+09 DEBUG wc 10.33.154.128 setting Windows defender exclusion failed {"command": "Add-MpPreference -ExclusionProcess C:\\k\\containerd\\containerd.exe", "output": "Add-MpPreference : The term 'Add-MpPreference' is not recognized as the name of a cmdlet, function, script file, or \r\noperable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try \r\nagain.\r\nAt line:1 char:1\r\n+ Add-MpPreference -ExclusionProcess C:\\k\\containerd\\containerd.exe\r\n+ ~~~~~~~~~~~~~~~~\r\n + CategoryInfo : ObjectNotFound: (Add-MpPreference:String) [], CommandNotFoundException\r\n + FullyQualifiedErrorId : CommandNotFoundException\r\n \r\n"}`

**To Reproduce**
1. Configure a golden image without the Windows Defender feature installed.
2. Create a machine set to use this image and scale it to 1
3. Monitor the logs of the operator pod and the error above will be thrown.

**Expected behavior**
If the Windows Defender Feature is not installed the operator should not try to set the exclusion.

mansikulkarni96:
Closing this issue as the PR linked above has been merged and the issue has been addressed.
/close

PR #1108, opened 2022-06-29
[windows] Check if Windows Defender is installed

saifshaikh48:
This PR fixes an issue where containerd configuration fails on instances
that do not have Windows Defender installed. WCMO now checks for Windows
Defender to be present before it creates an antivirus exclusion for containerd.

### DIFF
--- a/pkg/windows/windows.go
+++ b/pkg/windows/windows.go
@@ -446,14 +446,17 @@ func (vm *windows) Configure() error {
 // configureContainerd configures the Windows defender exclusion and starts the
 // Windows containerd service
 func (vm *windows) configureContainerd() error {
-	// set Windows defender exclusions for containerd
-	setExclusionCmd := "Add-MpPreference -ExclusionProcess " + containerdPath
-	out, err := vm.Run(setExclusionCmd, true)
+	// if Windows Defender is installed on the instance, create an exclusion for containerd
+	exclusionNeeded, err := vm.isWindowsDefenderEnabled()
 	if err != nil {
-		vm.log.V(1).Info("setting Windows defender exclusion failed", "command", setExclusionCmd,
-			"output", out)
-		return errors.Wrap(err, "setting Windows defender process exclusion failed")
+		return err
 	}
+	if exclusionNeeded {
+		if err := vm.createWindowsDefenderExclusion(containerdPath); err != nil {
+			return err
+		}
+	}
+
 	containerdServiceArgs := "--config " + containerdConfPath + " --log-file " + ContainerdLogDir + "containerd.log" +
 		" --log-level info" + " --run-service"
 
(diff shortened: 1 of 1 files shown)


mansikulkarni96:
Thanks for working on the quick fix @saifshaikh48 
/lgtm

mansikulkarni96:
fixes #1107

pkg/windows/windows.go
```diff
@@ -446,14 +446,17 @@ func (vm *windows) Configure() error {
 // configureContainerd configures the Windows defender exclusion and starts the
 // Windows containerd service
 func (vm *windows) configureContainerd() error {
-	// set Windows defender exclusions for containerd
-	setExclusionCmd := "Add-MpPreference -ExclusionProcess " + containerdPath
-	out, err := vm.Run(setExclusionCmd, true)
+	// if Windows Defender is installed on the instance, create an exlcusion for containerd
```
sebsoto, line 449: exlcusion -> exclusion

pkg/windows/windows.go
```diff
@@ -1037,6 +1040,27 @@ func (vm *windows) isContainersFeatureEnabled() (bool, error) {
...
+	}
+	return strings.TrimSpace(out) == "True", nil
+}
+
+// createWindowsDefenderExclusion sets a Windows Defender exclusion for the given file
+func (vm *windows) createWindowsDefenderExclusion(pathToFile string) error {
+	command := "Add-MpPreference -ExclusionProcess " + pathToFile
+	out, err := vm.Run(command, true)
+	if err != nil {
+		vm.log.V(1).Info("setting Windows defender exclusion failed", "command", command, "output", out)
```
sebsoto, line 1058: is this log necessary? Its duplicate information that will be present in the error being returned.
If the `out` is valuable info, wouldnt it make sense to add it to the error message?

sebsoto (changes requested):
Thanks for jumping on this quickly

pkg/windows/windows.go
saifshaikh48: You're right, I will fix this

saifshaikh48:
@jfrancoa @rrasouli 

Want to bring this PR to the attention of QE

jfrancoa:
> @jfrancoa @rrasouli
> 
> Want to bring this PR to the attention of QE

Should we create a new vSphere golden image without Windows Defender to verify this? Or could we use, for example the BYOH workflow (there we also have control on what the instance has enabled before WMCO adds it into the cluster). So, changing the user_data and making sure Windows Defender isn't enabled, should still allow us creating workloads on that BYOH node, right? However, if you consider we need to create a new vSphere golden image we can do it too.

rrasouli:
@saifshaikh48 What is the outcome of NOT having Windows Defender, ERROR, INFO failed test?

philipp1992:
do we need to install windows defender for the WMCO to work? we have a BYOH windows VM without defender and the WMCO throws the following error:

`{"level":"error","ts":1666266078.4728394,"msg":"Reconciler error","controller":"certificatesigningrequest","controllerGroup":"certificates.k8s.io","controllerKind":"CertificateSigningRequest","machine":{"name":"chmuw-default-windows-sq2nv","namespace":"openshift-machine-api"},"namespace":"openshift-machine-api","name":"chmuw-default-windows-sq2nv","reconcileID":"30aa1ad4-4645-44ce-b107-5dcd044fe0f1","error":"unable to configure instance 42245dad-2e8c-be76-eb07-aa192be0911e: configuring the Windows VM failed: error configuring containerd: setting Windows defender process exclusion failed with output: Add-MpPreference : Invalid class \r\nAt line:1 char:1\r\n+ Add-MpPreference -ExclusionProcess C:\\k\\containerd\\containerd.exe\r\n+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~\r\n    + CategoryInfo          : MetadataError: (MSFT_MpPreference:root\\Microsoft\\...FT_MpPreference) [Add-MpPreference], \r\n    CimException\r\n    + FullyQualifiedErrorId : HRESULT 0x80041010,Add-MpPreference\r..`

We are running OCP 4.11.8

We do NOT want to use Defender

sebsoto:
@philipp1992 Please see https://github.com/openshift/windows-machine-config-operator/issues/1240

saifshaikh48:
> What is the outcome of NOT having Windows Defender, ERROR, INFO failed test?

The outcome should be that the containerd service is properly configured/started and the Windows instance is successfully configured as a node without any errors.

> Should we create a new vSphere golden image without Windows Defender to verify this? Or could we use, for example the BYOH workflow (there we also have control on what the instance has enabled before WMCO adds it into the cluster). So, changing the user_data and making sure Windows Defender isn't enabled, should still allow us creating workloads on that BYOH node, right? However, if you consider we need to create a new vSphere golden image we can do it too.

Either method should work fine. The main thing here is to test have tests verifying functionality when Windows Defender is not enabled and when it is not

PR #1123, opened 2022-07-06
WINC-731: Introduce WICD ServiceAccount

sebsoto:
This PR adds a new WICD ServiceAccount with only the required permissions for WICD.
WICD will now use this SA instead of the WMCO SA.


bundle/manifests/windows-instance-config-daemon_rbac.authorization.k8s.io_v1_clusterrole.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  creationTimestamp: null
```
aravindhp, line 4: Is this auto-generated? If not it can be removed here and elsewhere.

aravindhp (approved):

/approve

PR #1136, opened 2022-07-18
[metrics] Remove old cleanup code

sebsoto:
This code was added to transition from the old metrics resources to the
new resources. This is no longer relevant now that it is not possible
to upgrade from a version with the old resource type.

This was added in commit f7d52ed9f2567ad8abd81d798ece5f3acfbfcbd3


aravindhp (approved):
Thanks for removing this, @sebsoto 

/approve

sebsoto:
its safe to cherry-pick this back to 4.10, as the original [PR](https://github.com/openshift/windows-machine-config-operator/pull/330) merged into 4.8.

PR #1138, opened 2022-07-18
Bug 2107261: Maintain endpoint subsets on restart

sebsoto:
WMCO deletes and recreates the windows-exporter endpoint upon each
operator start. This is resulting in an issue where the correct
endpoint subsets would be missing upon operator restart, stopping
metrics collection for Windows nodes.

This is being fixed by including the existing endpoint subsets in the
new endpoint created on restart.

A better approach would be to create a controller which maintains the
windows exporter endpoint. That is a larger change and should be done in
the future. This approach was taken as it is the smallest change which
can resolve this bug.

Fixes BZ#2107261

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -224,11 +224,20 @@ func (c *Config) Configure(ctx context.Context) error {
 	}
 	// In the case of an operator restart, a previous Endpoint object will be deleted and a new one will
 	// be created to ensure we have a correct spec.
-	err = c.CoreV1().Endpoints(c.namespace).Delete(ctx, WindowsMetricsResource, metav1.DeleteOptions{})
-	if err != nil && !apierrors.IsNotFound(err) {
-		return errors.Wrap(err, "error deleting existing metrics Endpoint")
+	var subsets []v1.EndpointSubset
+	existingEndpoint, err := c.CoreV1().Endpoints(c.namespace).Get(ctx, WindowsMetricsResource, metav1.GetOptions{})
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return errors.Wrapf(err, "error retrieving %s endpoint", WindowsMetricsResource)
+		}
+	} else {
+		subsets = existingEndpoint.Subsets
+		err = c.CoreV1().Endpoints(c.namespace).Delete(ctx, WindowsMetricsResource, metav1.DeleteOptions{})
+		if err != nil {
+			return errors.Wrapf(err, "error deleting %s endpoint", WindowsMetricsResource)
+		}
 	}
-	if err := c.createEndpoint(); err != nil {
+	if err := c.createEndpoint(subsets); err != nil {
 		return errors.Wrap(err, "error creating metrics Endpoint")
 	}
 	return nil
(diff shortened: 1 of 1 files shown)


pkg/metrics/metrics.go
```diff
@@ -224,11 +224,20 @@ func (c *Config) Configure(ctx context.Context) error {
...
-	if err != nil && !apierrors.IsNotFound(err) {
-		return errors.Wrap(err, "error deleting existing metrics Endpoint")
+	var subsets []v1.EndpointSubset
+	existingEndpoint, err := c.CoreV1().Endpoints(c.namespace).Get(ctx, WindowsMetricsResource, metav1.GetOptions{})
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return errors.Wrap(err, "error retrieving existing metrics endpoint")
+		}
+	} else {
+		subsets = existingEndpoint.Subsets
```
saifshaikh48, line 234: Can `existingEndpoint` be nil such that we would try to access `.Subsets` of a nil pointer obj? Looks like no based on my understanding of the client API, but just want another set of eyes to make sure.

saifshaikh48 (commented):
Thanks for fixing this

pkg/metrics/metrics.go
sebsoto: I dont see how, if Get() is returning no error, existingEndpoint shouldnt be nil

pkg/metrics/metrics.go
```diff
@@ -224,11 +224,20 @@ func (c *Config) Configure(ctx context.Context) error {
...
 	// In the case of an operator restart, a previous Endpoint object will be deleted and a new one will
 	// be created to ensure we have a correct spec.
-	err = c.CoreV1().Endpoints(c.namespace).Delete(ctx, WindowsMetricsResource, metav1.DeleteOptions{})
-	if err != nil && !apierrors.IsNotFound(err) {
-		return errors.Wrap(err, "error deleting existing metrics Endpoint")
+	var subsets []v1.EndpointSubset
+	existingEndpoint, err := c.CoreV1().Endpoints(c.namespace).Get(ctx, WindowsMetricsResource, metav1.GetOptions{})
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return errors.Wrap(err, "error retrieving existing metrics endpoint")
```
jrvaldes, line 231: It may be helpful to include the name of the resource that failed the lookup. i.e. `WindowsMetricsResource`

pkg/metrics/metrics.go
```diff
@@ -224,11 +224,20 @@ func (c *Config) Configure(ctx context.Context) error {
...
+		}
+	} else {
+		subsets = existingEndpoint.Subsets
+		err = c.CoreV1().Endpoints(c.namespace).Delete(ctx, WindowsMetricsResource, metav1.DeleteOptions{})
+		if err != nil && !apierrors.IsNotFound(err) {
```
jrvaldes, line 236: At this point, it's guaranteed the `WindowsMetricsResource` endpoint was present. I don't feel you should exclude the not found errors (`apierrors.IsNotFound(err)`)

jrvaldes (commented):
LGTM, just minor suggestions.

rrasouli:
cc @jfrancoa

saifshaikh48 (commented):
/lgtm

aravindhp (approved):
@sebsoto it is nice that you have mentioned the better approach for the fix in the commit message. It would be even better if you mentioned how you are fixing the issue currently without introducing a controller.

/approve

PR #1142, opened 2022-07-19
[servicescm] Handle unmanaged dependency services

saifshaikh48:
This PR prevents a panic during service validation if a service that is not defined in
the services ConfigMap is listed as a dependency. Cycle detection now skips over
the unmanaged service and continues as normal for any other dependencies.


jrvaldes (approved):
/lgtm

PR #1145, opened 2022-07-19
WINC-820: [docs] Remove Windows Server 20H2 support

saifshaikh48:
Microsoft will drop support for OS version Windows Server 20H2 on 2022-08-09.
In line with this, WMCO will no longer support this version for vSphere worker nodes.


saifshaikh48:
cc: @mburke5678

This likely has docs implications for all WMCO 3.x and above.

PR #1150, opened 2022-07-20
Move service CM definition out of servicecm package

sebsoto:
The servicescm should not be the package which declares the expected
contents of the Windows service ConfigMap, that is a more fitting
responsibility of the ConfigMap controller. The reasoning for this is
that the CM controller is the entity which is responsible for
maintaining the CM's expected state. The state is also highly dependent
on cluster configuration, such as the platform type, information which
is more fitting on the controller level, rather than where it currently
exists.

As part of this change, the servicescm.GenerateWithData() function is
being removed, and its functionality is being moved into
servicescm.Generate().

### DIFF
--- a/controllers/configmap_controller.go
+++ b/controllers/configmap_controller.go
@@ -70,6 +70,7 @@ const (
 // ConfigMapReconciler reconciles a ConfigMap object
 type ConfigMapReconciler struct {
 	instanceReconciler
+	servicesManifest *servicescm.Data
 }
 
 // NewConfigMapReconciler returns a pointer to a ConfigMapReconciler
@@ -84,6 +85,11 @@ func NewConfigMapReconciler(mgr manager.Manager, clusterConfig cluster.Config, w
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to initialize Prometheus configuration")
 	}
+	svcData, err := servicescm.NewData(&[]servicescm.Service{}, &[]servicescm.FileInfo{})
+	if err != nil {
+		return nil, err
+	}
+
 	return &ConfigMapReconciler{
 		instanceReconciler: instanceReconciler{
 			client:               mgr.GetClient(),
@@ -96,6 +102,7 @@ func NewConfigMapReconciler(mgr manager.Manager, clusterConfig cluster.Config, w
 			prometheusNodeConfig: pc,
 			platform:             clusterConfig.Platform(),
 		},
+		servicesManifest: svcData,
 	}, nil
 }
 
@@ -169,7 +176,7 @@ func (r *ConfigMapReconciler) reconcileServices(ctx context.Context, windowsServ
 
 	// If a ConfigMap with invalid values is found, WMCO will delete and recreate it with proper values
 	data, err := servicescm.Parse(windowsServices.Data)
-	if err != nil || data.ValidateRequiredContent() != nil {
+	if err != nil || data.ValidateExpectedContent(r.servicesManifest) != nil {
 		// Deleting will trigger an event for the configmap_controller, which will re-create a proper ConfigMap
 		if err = r.client.Delete(ctx, windowsServices); err != nil {
 			return err
@@ -403,7 +410,7 @@ func (r *ConfigMapReconciler) isValidConfigMap(o client.Object) bool {
 
 // createServicesConfigMap creates a valid ServicesConfigMap and returns it
 func (r *ConfigMapReconciler) createServicesConfigMap(ctx context.Context) (*core.ConfigMap, error) {
-	windowsServices, err := servicescm.Generate(servicescm.Name, r.watchNamespace)
+	windowsServices, err := servicescm.Generate(servicescm.Name, r.watchNamespace, r.servicesManifest)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


controllers/configmap_controller.go
```diff
@@ -67,9 +67,22 @@ const (
 	ConfigMapController = "configmap"
 )
 
+// serviceConfig defines the contents of the Windows service ConfigMap
+type serviceConfig struct {
```
saifshaikh48, line 71: Would the [`servicescm.data`](https://github.com/openshift/windows-machine-config-operator/blob/master/pkg/servicescm/servicescm.go#L91) struct be useful here? I feel like exporting it would allow you to remove this `serviceConfig` introduction.

newServiceConfig() would stay our source of truth for services CM data then return `*servicescm.Data`.

sebsoto: I think it would be, i'll make that change

test/e2e/validation_test.go
```diff
@@ -532,20 +534,21 @@ func (tc *testContext) testInvalidServicesCM(cmName string) error {
...
-	windowsServices, err := tc.waitForValidWindowsServicesConfigMap(cmName)
+	windowsServices, err := tc.waitForValidWindowsServicesConfigMap(cmName, expected)
 	if err != nil {
 		return errors.Wrapf(err, "error retrieving ConfigMap %s", cmName)
 	}
 	data, err := servicescm.Parse(windowsServices.Data)
-	if err != nil || data.ValidateRequiredContent() != nil {
+	if err != nil || data.ValidateExpectedContent(expected) != nil {
 		return errors.Wrapf(err, "error ensuring ConfigMap %s is re-created validly", cmName)
 	}
```
saifshaikh48, line 544: Isn't this already done within `waitForValidWindowsServicesConfigMap()` which is called just above? We can remove this IMO (and update the error msg on line 539)

pkg/servicescm/servicescm.go
```diff
@@ -185,29 +170,29 @@ func (cmData *data) GetBootstrapServices() []Service {
...
-func (cmData *data) ValidateRequiredContent() error {
+// ValidateExpectedContent ensures that the given slices are comprised of all the expected services/files and only these
+func (cmData *Data) ValidateExpectedContent(expected *Data) error {
 	// Validate services
-	if len(cmData.Services) != len(*requiredServices) {
+	if len(cmData.Services) != len(expected.Services) {
 		return errors.New("Unexpected number of services")
 	}
-	for _, requiredSvc := range *requiredServices {
+	for _, requiredSvc := range expected.Services {
```
mansikulkarni96, line 186: expectedSvc? instead of required
Same comment for files

controllers/configmap_controller.go
```diff
@@ -67,9 +67,16 @@ const (
 	ConfigMapController = "configmap"
 )
 
+// newServiceConfig returns a Data object which contains the expected Windows service configuration based on
```
mansikulkarni96, line 70: nit: Can you add 'Data' in single quotes?

test/e2e/validation_test.go
```diff
@@ -532,20 +534,17 @@ func (tc *testContext) testInvalidServicesCM(cmName string) error {
...
 	// Try to retrieve newly created ConfigMap and validate its contents
-	windowsServices, err := tc.waitForValidWindowsServicesConfigMap(cmName)
+	_, err = tc.waitForValidWindowsServicesConfigMap(cmName, expected)
 	if err != nil {
-		return errors.Wrapf(err, "error retrieving ConfigMap %s", cmName)
-	}
-	data, err := servicescm.Parse(windowsServices.Data)
-	if err != nil || data.ValidateRequiredContent() != nil {
-		return errors.Wrapf(err, "error ensuring ConfigMap %s is re-created validly", cmName)
+		return errors.Wrapf(err, "error waitng for valid ConfigMap %s", cmName)
```
saifshaikh48, line 539: typo in waiting :)

mansikulkarni96 (commented):
Motsly LGTM

saifshaikh48 (approved):
/lgtm

sebsoto:
@aravindhp PTAL

controllers/configmap_controller.go
```diff
@@ -67,9 +67,16 @@ const (
 	ConfigMapController = "configmap"
 )
 
+// newServiceConfig returns a 'Data' object which contains the expected Windows service configuration based on
+// the state of the cluster
+func newServiceConfig() (*servicescm.Data, error) {
```
aravindhp, line 72: I don't understand the  need for this wrapper. Also shouldn't it be `newServicesConfig()`?

controllers/configmap_controller.go
```diff
@@ -67,9 +67,16 @@ const (
...
+	return servicescm.NewData(&[]servicescm.Service{}, &[]servicescm.FileInfo{})
+}
+
 // ConfigMapReconciler reconciles a ConfigMap object
 type ConfigMapReconciler struct {
 	instanceReconciler
+	serviceConfig *servicescm.Data
```
aravindhp, line 79: Why is this called `serviceConfig` instead of `servicesData` or even better something like `servicesManifest`?

aravindhp (commented):
Just a couple of questions around naming. LGTM otherwise.

controllers/configmap_controller.go
sebsoto: I'm jumping the gun adding this function, removing it for now.
There will be a need for it as the logic for the service definitions grows, but its not necessary now.

aravindhp (approved):

/approve

PR #1152, opened 2022-07-22
WINC-848: Manage Windows exporter with WICD

sebsoto:
This PR adds Windows Exporter to the Windows service ConfigMap, as well as changes required to stop the service when needed.

### DIFF
--- a/controllers/configmap_controller.go
+++ b/controllers/configmap_controller.go
@@ -48,6 +48,7 @@ import (
 	"github.com/openshift/windows-machine-config-operator/pkg/secrets"
 	"github.com/openshift/windows-machine-config-operator/pkg/servicescm"
 	"github.com/openshift/windows-machine-config-operator/pkg/signer"
+	"github.com/openshift/windows-machine-config-operator/pkg/windows"
 	"github.com/openshift/windows-machine-config-operator/pkg/wiparser"
 	"github.com/openshift/windows-machine-config-operator/version"
 )
@@ -85,9 +86,11 @@ func NewConfigMapReconciler(mgr manager.Manager, clusterConfig cluster.Config, w
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to initialize Prometheus configuration")
 	}
-	svcData, err := servicescm.NewData(&[]servicescm.Service{}, &[]servicescm.FileInfo{})
+
+	// Get expected state of the Windows service ConfigMap
+	svcData, err := generateServicesManifest()
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "error generating expected Windows service state")
 	}
 
 	return &ConfigMapReconciler{
@@ -470,4 +473,20 @@ func (r *ConfigMapReconciler) EnsureServicesConfigMapExists() error {
 func (r *ConfigMapReconciler) isKubeAPIServerServingCAConfigMap(obj client.Object) bool {
 	return obj.GetNamespace() == certificates.KubeApiServerOperatorNamespace &&
 		obj.GetName() == certificates.KubeAPIServerServingCAConfigMapName
+}
+
+// generateServicesManifest returns the expected state of the Windows service configmap
+func generateServicesManifest() (*servicescm.Data, error) {
+	services := &[]servicescm.Service{{
+		Name:                         windows.WindowsExporterServiceName,
+		Command:                      windows.WindowsExporterServiceCommand,
+		NodeVariablesInCommand:       nil,
+		PowershellVariablesInCommand: nil,
+		Dependencies:                 nil,
+		Bootstrap:                    false,
+		Priority:                     1,
+	}}
+	// TODO: All payload filenames and checksums must be added here https://issues.redhat.com/browse/WINC-847
+	files := &[]servicescm.FileInfo{}
+	return servicescm.NewData(services, files)
 }
--- a/pkg/windows/windows.go
+++ b/pkg/windows/windows.go
@@ -77,13 +77,13 @@ const (
 	kubeProxyServiceName = "kube-proxy"
 	// kubeletServiceName is the name of the kubelet Windows service
 	kubeletServiceName = "kubelet"
-	// windowsExporterServiceName is the name of the windows_exporter Windows service
-	windowsExporterServiceName = "windows_exporter"
+	// WindowsExporterServiceName is the name of the windows_exporter Windows service
+	WindowsExporterServiceName = "windows_exporter"
 	// AzureCloudNodeManagerServiceName is the name of the azure cloud node manager service
 	AzureCloudNodeManagerServiceName = "cloud-node-manager"
-	// windowsExporterServiceArgs specifies metrics for the windows_exporter service to collect
+	// WindowsExporterServiceCommand specifies metrics for the windows_exporter service to collect
 	// and expose metrics at endpoint with default port :9182 and default URL path /metrics
-	windowsExporterServiceArgs = "--collectors.enabled " +
+	WindowsExporterServiceCommand = windowsExporterPath + " --collectors.enabled " +
 		"cpu,cs,logical_disk,net,os,service,system,textfile,container,memory,cpu_info"
 	// serviceQueryCmd is the Windows command used to query a service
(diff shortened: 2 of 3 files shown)


controllers/configmap_controller.go
```diff
@@ -85,7 +86,15 @@ func NewConfigMapReconciler(mgr manager.Manager, clusterConfig cluster.Config, w
...
-	svcData, err := servicescm.NewData(&[]servicescm.Service{}, &[]servicescm.FileInfo{})
+	svcData, err := servicescm.NewData(&[]servicescm.Service{{
+		Name:                         windows.WindowsExporterServiceName,
+		Command:                      windows.WindowsExporterServiceCommand,
+		NodeVariablesInCommand:       nil,
+		PowershellVariablesInCommand: nil,
+		Dependencies:                 nil,
+		Bootstrap:                    false,
+		Priority:                     1,
+	}}, &[]servicescm.FileInfo{})
```
saifshaikh48, line 97: Consider introducing 2 constants to hold the `managedServices` and `managedFiles`. Doing this in-line may get messy as this continues to grow

controllers/configmap_controller.go
```diff
@@ -85,7 +86,15 @@ func NewConfigMapReconciler(mgr manager.Manager, clusterConfig cluster.Config, w
...
 	}
-	svcData, err := servicescm.NewData(&[]servicescm.Service{}, &[]servicescm.FileInfo{})
+	svcData, err := servicescm.NewData(&[]servicescm.Service{{
+		Name:                         windows.WindowsExporterServiceName,
+		Command:                      windows.WindowsExporterServiceCommand,
+		NodeVariablesInCommand:       nil,
+		PowershellVariablesInCommand: nil,
+		Dependencies:                 nil,
+		Bootstrap:                    false,
+		Priority:                     1,
```
saifshaikh48, line 96: Could you explain how you settled on priority of 1? Just thinking ahead, will all bootstrap services share priority of 0?

alinaryan:
Can you add a commit message for the third commit?

controllers/configmap_controller.go
saifshaikh48: Fair enough, how about separate variables then? There will be services managed by WICD across all cluster configs.

sebsoto: Done. Additionally I expect this logic to be broken out into a separate function once the generation gets more complicated.

sebsoto:
@aravindhp PTAL when you get a chance

saifshaikh48 (commented):
/lgtm 

First service being moved to the configmap is a milestone in WICD work 🚀

pkg/windows/windows.go
```diff
@@ -379,6 +379,12 @@ func (vm *windows) Reinitialize() error {
 }
 
 func (vm *windows) EnsureRequiredServicesStopped() error {
+	// In this transitional period of migrating services to WICD's responsibility, WICD must be stopped first to ensure
+	// it does not start any of these stopped services
```
aravindhp, line 381: So will this go away after the transition? If yes, then the story where this should be updated to call this out and the Jira issue linked here.

test/e2e/validation_test.go
```diff
@@ -475,22 +475,24 @@ func testServicesConfigMap(t *testing.T) {
...
 
 	// Ensure the windows-services ConfigMap exists in the cluster
+	var expectedData *servicescm.Data
 	t.Run("Services ConfigMap existence", func(t *testing.T) {
-		_, err = tc.client.K8s.CoreV1().ConfigMaps(tc.namespace).Get(context.TODO(), servicesConfigMapName,
+		cm, err := tc.client.K8s.CoreV1().ConfigMaps(tc.namespace).Get(context.TODO(), servicesConfigMapName,
 			meta.GetOptions{})
-		assert.NoErrorf(t, err, "error ensuring ConfigMap %s exists", servicesConfigMapName)
+		require.NoErrorf(t, err, "error ensuring ConfigMap %s exists", servicesConfigMapName)
+		expectedData, err = servicescm.Parse(cm.Data)
```
aravindhp, line 485: So you are assuming that at this point the data in the ConfigMap is what is expected and correct? Ideally you must be generating the expected data here based on some WMCO constant rather than reusing the `Parse` method. What happens if there is a bug in the `Parse` method?

controllers/configmap_controller.go
aravindhp: With the pattern you are introducing, `NewConfigMapReconciler()` will grow very large as we keep adding service. I suggest encapsulating this with  a helper. Ideally we should be using a constant that this helper method uses to generate the data.

aravindhp (changes requested):

test/e2e/validation_test.go
sebsoto: The difficulty in validating the correct data is why I made this change. The ConfigMap contents needs to be generated using the cluster configuration. To validate the contents we'd need to either duplicate the logic of, or export the `generateServiceManifest` function from the `controllers` package. I'd rather not do either of those as that leads to the issue you present of there possibly being a bug in the `generateServiceManifest` function, and us reusing that function isn't doing anything valuable by confirming that `generateServiceManifest` generates the same content when invoked in two different places.

I don't see an issue with using `Parse` as that is how the consumer of the ConfigMap, WICD, is going to interact with it.

In order to get some amount of validation here I've expanded the test to check that the required services exist within the ConfigMap.

sebsoto:
@aravindhp Updated to address your comments PTAL

test/e2e/validation_test.go
```diff
@@ -433,39 +433,48 @@ func (tc *testContext) getWinServices(addr string) (map[string]winService, error
...
 	require.NoError(t, err)
-
 	for _, node := range gc.allNodes() {
 		t.Run(node.GetName(), func(t *testing.T) {
 			addr, err := controllers.GetAddress(node.Status.Addresses)
 			require.NoError(t, err, "unable to get node address")
 			svcs, err := tc.getWinServices(addr)
 			require.NoError(t, err, "error getting service map")
-			for _, svcName := range windows.RequiredServices {
+			for svcName, shouldBeRunning := range expectedSvcs {
```
aravindhp, line 444: Is the Azure Cloud Manager service always present on the instance even on non-Azure clouds? Is it that it will just not be running?

controllers/configmap_controller.go
```diff
@@ -471,3 +474,18 @@ func (r *ConfigMapReconciler) isKubeAPIServerServingCAConfigMap(obj client.Objec
 	return obj.GetNamespace() == certificates.KubeApiServerOperatorNamespace &&
 		obj.GetName() == certificates.KubeAPIServerServingCAConfigMapName
 }
+
+// generateServiceManifest returns the expected state of the Windows service configmap
+func generateServiceManifest() (*servicescm.Data, error) {
```
aravindhp, line 479: `generateServicesManifest()` will be more accurate.

controllers/configmap_controller.go
```diff
@@ -471,3 +474,18 @@ func (r *ConfigMapReconciler) isKubeAPIServerServingCAConfigMap(obj client.Objec
...
+	services := &[]servicescm.Service{{
+		Name:                         windows.WindowsExporterServiceName,
+		Command:                      windows.WindowsExporterServiceCommand,
+		NodeVariablesInCommand:       nil,
+		PowershellVariablesInCommand: nil,
+		Dependencies:                 nil,
+		Bootstrap:                    false,
+		Priority:                     1,
+	}}
+	files := &[]servicescm.FileInfo{}
```
aravindhp, line 489: Sorry didn't catch this last time. Shouldn't the Windows exporter binary be added here?

aravindhp (commented):
Mostly LGTM.

test/e2e/validation_test.go
sebsoto: That service will only run on Azure instances

controllers/configmap_controller.go
sebsoto: All payload files can be added, adding a TODO with a link to the appropriate story.

saifshaikh48 (commented):
/lgtm

aravindhp (approved):
/approve

PR #1153, opened 2022-07-25
[WIP] Investigate AWS WS2022 reconfigure test failures

saifshaikh48:
I want to see if we can isolate the reconfiguration tests as the issue creator, or if the order of the tests is contributing to the EC2 metadata service network issues.

Current test flow is upgrade test suite (replaces the version annotation) then reconfigure test suite (deletes the version annotation) --> results in error

With this PR I want to run 2 experiments:
- skip the upgrade tests and see if the reconfigure suite still fails
- run the upgrade test suite 2x in a row and see if it still passes

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,8 +4,8 @@
 	branch = release-4.11
 [submodule "ovn-kubernetes"]
 	path = ovn-kubernetes
-	url = https://github.com/openshift/ovn-kubernetes
-	branch = release-4.11
+	url = https://github.com/jrvaldes/ovn-kubernetes
+	branch = duplicate-link-local-routes
 [submodule "containernetworking-plugins"]
 	path = containernetworking-plugins
 	url = https://github.com/openshift/containernetworking-plugins
--- a/controllers/controllers.go
+++ b/controllers/controllers.go
@@ -79,7 +79,7 @@ func (r *instanceReconciler) ensureInstanceIsUpToDate(instanceInfo *instance.Inf
 
 	// Check if the instance was configured by a previous version of WMCO and must be deconfigured before being
 	// configured again.
-	if instanceInfo.UpgradeRequired() {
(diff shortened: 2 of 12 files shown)


saifshaikh48:
Skipping upgrade tests and only running reconfiguration tests results in the error loop seen before
```
\"logger\":\"wmcb\",\"msg\":\"could not run bootstrapper\",\"error\":\"failed to initialize kubelet: could not parse ignition file: cannot generate initial kubelet args: unable to retrieve the hostname from the EC2 instance: operation error ec2imds: GetMetadata, exceeded maximum number of attempts, 3, request send failed, Get \\\"http://169.254.169.254/latest/meta-data/local-hostname\\\": dial tcp 169.254.169.254:80: connectex: A socket operation was attempted to an unreachable network.\"
```

saifshaikh48:
/test aws-e2e-upgrade

Now ensuring upgrade test suite 2x in a row still passes (to try to rule out instance reboots causing network issues)

saifshaikh48:
Running upgrade test suite 2x in a row passes without error.

saifshaikh48:
/test aws-e2e-upgrade

The upgrade test suite replaces the version annotation to kick off instance reconfiguration. The reconfigure test suite deletes the version annotation altogether to kick off instance reconfiguration. Only the latter results in an error.

In this test run, I made the reconfigure test logic to mimic the upgrade test, replacing the version annotation instead of removing.

saifshaikh48:
/test aws-e2e-upgrade

Pushed a fix to properly detect and fix an out-of-sync instance (one that has previously been configured but requires deconfiguration before it can be re-configured).

saifshaikh48:
/test aws-e2e-upgrade

looks like an unrelated failure. trying this again

saifshaikh48:
/test aws-e2e-upgrade

BYOH node properly configured but Machine hit cloud provider issues
`unable to get instance profile Throttling: Rate exceeded status code: 400`

PR #1154, opened 2022-07-26
WINC-718: WICD cleanup command

saifshaikh48:
WMCO will invoke WICD `cleanup` to remove an instance from the cluster. 
This PR covers the initial implementation of said command to remove services.

When the WICD cleanup command is executed, all services managed by us 
(as defined in the expected services ConfigMap) will be stopped and deleted. 

It will then delete the instance's Node object by default. 
When ran with the optional `--preserveNode` arg, the Node object won't be deleted.

### DIFF
--- a/cmd/daemon/cleanup.go
+++ b/cmd/daemon/cleanup.go
@@ -1,0 +1,58 @@
+//go:build windows
+
+/*
+Copyright 2022.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"github.com/spf13/cobra"
+	"k8s.io/klog/v2"
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	"github.com/openshift/windows-machine-config-operator/pkg/daemon/cleanup"
+	"github.com/openshift/windows-machine-config-operator/pkg/daemon/config"
+)
+
+var (
+	cleanupCmd = &cobra.Command{
+		Use:   "cleanup",
+		Short: "Cleans up WMCO-managed Windows services",
+		Long: "Stops and removes all Windows services as part of Node deconfiguration, " +
+			"according to information given by Windows Service ConfigMaps present within the cluster",
+		Run: runCleanupCmd,
+	}
+	// preserveNode is an optional flag that instructs WICD to deconfigure an instance without deleting the Node object.
+	// This is useful in the node upgrade scenario.
+	preserveNode bool
+)
+
+func init() {
+	rootCmd.AddCommand(cleanupCmd)
+	cleanupCmd.PersistentFlags().BoolVar(&preserveNode, "preserveNode", false,
+		"If set to true, preserves the Node associated with this instance. Defaults to false.")
+}
+
+func runCleanupCmd(cmd *cobra.Command, args []string) {
+	cfg, err := config.FromServiceAccount(apiServerURL, saCA, saToken)
+	if err != nil {
+		klog.Exitf("error using service account to build config: %s", err.Error())
+	}
+	ctx := ctrl.SetupSignalHandler()
+	if err := cleanup.Deconfigure(cfg, ctx, preserveNode); err != nil {
+		klog.Exitf(err.Error())
+	}
+}
--- a/pkg/daemon/cleanup/cleanup.go
+++ b/pkg/daemon/cleanup/cleanup.go
@@ -1,0 +1,111 @@
+//go:build windows
+
+/*
+Copyright 2022.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cleanup
+
+import (
+	"context"
+
+	"github.com/pkg/errors"
+	core "k8s.io/api/core/v1"
+	"k8s.io/client-go/rest"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/openshift/windows-machine-config-operator/pkg/daemon/controller"
+	"github.com/openshift/windows-machine-config-operator/pkg/daemon/manager"
+	"github.com/openshift/windows-machine-config-operator/pkg/nodeconfig"
+	"github.com/openshift/windows-machine-config-operator/pkg/servicescm"
+)
+
+// Deconfigure removes an instance from the cluster. If preserveNode is true, the node object is not deleted.
+// If we are able to get the services ConfigMap tied to the desired version, all services defined in it are cleaned up.
+// TODO: Otherwise, perform cleanup based on a combination of the OpenShift managed tag and latest services ConfigMap.
+func Deconfigure(cfg *rest.Config, ctx context.Context, preserveNode bool) error {
+	// Cannot use a cached client as no manager will be started to populate cache
+	directClient, err := controller.NewDirectClient(cfg)
+	if err != nil {
+		return errors.Wrap(err, "could not create authenticated client from service account")
+	}
+	addrs, err := controller.LocalInterfaceAddresses()
+	if err != nil {
+		return err
+	}
+	var node *core.Node
+	var cmData *servicescm.Data
+	err = func() error {
+		node, err = controller.GetAssociatedNode(directClient, addrs)
+		if err != nil {
+			return err
+		}
+		desiredVersion, present := node.Annotations[nodeconfig.DesiredVersionAnnotation]
+		if !present {
+			return errors.Wrapf(err, "node missing desired version annotation")
+		}
+		// Fetch the CM of the desired version
+		cm := &core.ConfigMap{}
+		err = directClient.Get(ctx,
+			client.ObjectKey{Namespace: controller.WMCONamespace, Name: servicescm.NamePrefix + desiredVersion}, cm)
+		if err != nil {
+			return err
+		}
+		cmData, err = servicescm.Parse(cm.Data)
+		return err
+	}()
+	if err != nil {
+		// TODO: best effort cleanup of all OpenShift managed services and those in the latest services ConfigMap
+		// https://issues.redhat.com/browse/WINC-733
+		return err
+	}
+
+	svcMgr, err := manager.New()
+	if err != nil {
+		klog.Exitf("could not create service manager: %s", err.Error())
+	}
+	if err = removeServices(svcMgr, cmData.Services); err != nil {
+		return err
+	}
+
+	// Delete node if needed
+	if preserveNode || node == nil {
+		return nil
+	}
+	klog.Infof("deleting node %s", node.GetName())
+	return directClient.Delete(ctx, node)
+}
+
+// removeServices uses the given manager to remove all the given Windows services from this instance.
+func removeServices(svcMgr manager.Manager, services []servicescm.Service) error {
+	// Build up log message and failures
+	servicesRemoved := []string{}
+	failedRemovals := []error{}
+	// The services are ordered by increasing priority already, so stop them in reverse order to avoid dependency issues
+	for i := len(services) - 1; i >= 0; i-- {
+		service := services[i]
+		if err := svcMgr.DeleteService(service.Name); err != nil {
+			failedRemovals = append(failedRemovals, err)
+		} else {
+			servicesRemoved = append(servicesRemoved, service.Name)
+		}
+	}
+	klog.Infof("removed services: %q", servicesRemoved)
+	if len(failedRemovals) > 0 {
+		return errors.Errorf("%#v", failedRemovals)
+	}
+	return nil
+}
--- a/pkg/daemon/cleanup/cleanup_test.go
+++ b/pkg/daemon/cleanup/cleanup_test.go
@@ -1,0 +1,109 @@
+//go:build windows
+
+package cleanup
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/sys/windows/svc"
+	"golang.org/x/sys/windows/svc/mgr"
+
+	"github.com/openshift/windows-machine-config-operator/pkg/daemon/fake"
+	"github.com/openshift/windows-machine-config-operator/pkg/servicescm"
+	"github.com/openshift/windows-machine-config-operator/pkg/windows"
+)
+
+func TestRemoveServices(t *testing.T) {
+	testIO := []struct {
+		name              string
+		existingServices  map[string]*fake.FakeService
+		configMapServices []servicescm.Service
+		expectedServices  map[string]struct{}
+	}{
+		{
+			name:              "No services",
+			existingServices:  map[string]*fake.FakeService{},
+			configMapServices: []servicescm.Service{},
+			expectedServices:  map[string]struct{}{},
+		},
+		{
+			name:              "Upgrade scenario with no services",
+			existingServices:  map[string]*fake.FakeService{},
+			configMapServices: []servicescm.Service{},
+			expectedServices:  map[string]struct{}{},
+		},
+		{
+			name:             "ConfigMap managed service doesn't exist on node",
+			existingServices: map[string]*fake.FakeService{},
+			configMapServices: []servicescm.Service{
+				{Name: "test1", Dependencies: nil, Priority: 0},
+			},
+			expectedServices: map[string]struct{}{},
+		},
+		{
+			name: "Single service",
+			existingServices: map[string]*fake.FakeService{
+				"test1": newTestService("test1", []string{}),
+			},
+			configMapServices: []servicescm.Service{
+				{Name: "test1", Dependencies: nil, Priority: 0},
+			},
+			expectedServices: map[string]struct{}{},
+		},
+		{
+			name: "Multiple services",
+			existingServices: map[string]*fake.FakeService{
+				"test1": newTestService("test1", []string{}),
+				"test2": newTestService("test2", []string{}),
+				"test3": newTestService("test3", []string{}),
+			},
+			configMapServices: []servicescm.Service{
+				{Name: "test1", Dependencies: nil, Priority: 0},
+				{Name: "test2", Dependencies: nil, Priority: 1},
+			},
+			expectedServices: map[string]struct{}{"test3": {}},
+		},
+		{
+			name: "Multiple services with dependencies",
+			existingServices: map[string]*fake.FakeService{
+				"test1": newTestService("test1", []string{}),
+				"test2": newTestService("test2", []string{}),
+				"test3": newTestService("test3", []string{}),
+				"test4": newTestService("test4", []string{"test3"}),
+				"test5": newTestService("test5", []string{"test1", "test3"}),
+			},
+			configMapServices: []servicescm.Service{
+				{Name: "test1", Dependencies: nil, Priority: 0},
+				{Name: "test3", Dependencies: nil, Priority: 0},
+				{Name: "test4", Dependencies: []string{"test3"}, Priority: 1},
+				{Name: "test5", Dependencies: []string{"test1", "test3"}, Priority: 2},
+			},
+			expectedServices: map[string]struct{}{"test2": {}},
+		},
+	}
+
+	for _, test := range testIO {
+		t.Run(test.name, func(t *testing.T) {
+			winSvcMgr := fake.NewTestMgr(test.existingServices)
+			err := removeServices(winSvcMgr, test.configMapServices)
+			require.NoError(t, err)
+			allServices, err := winSvcMgr.GetServices()
+			require.NoError(t, err)
+
+			assert.Equal(t, test.expectedServices, allServices)
+		})
+	}
+}
+
+func newTestService(name string, dependencies []string) *fake.FakeService {
+	return fake.NewFakeService(
+		name,
+		mgr.Config{
+			Description:  windows.ManagedTag + name,
+			Dependencies: dependencies,
+		},
+		svc.Status{State: svc.Running},
+	)
+}
--- a/pkg/daemon/controller/controller.go
+++ b/pkg/daemon/controller/controller.go
@@ -55,8 +55,8 @@ import (
 )
 
 const (
-	// wmcoNamespace defines the namespace service ConfigMaps are expected to be in
-	wmcoNamespace = "openshift-windows-machine-config-operator"
+	// WMCONamespace defines the namespace service ConfigMaps are expected to be in
+	WMCONamespace = "openshift-windows-machine-config-operator"
 )
 
 type ServiceController struct {
@@ -100,7 +100,7 @@ func RunController(ctx context.Context, apiServerURL, saCA, saToken string) erro
 		return err
 	}
 
-	addrs, err := localInterfaceAddresses()
+	addrs, err := LocalInterfaceAddresses()
 	if err != nil {
 		return err
 	}
@@ -110,7 +110,7 @@ func RunController(ctx context.Context, apiServerURL, saCA, saToken string) erro
 	}
 
 	ctrlMgr, err := ctrl.NewManager(cfg, ctrl.Options{
-		Namespace: wmcoNamespace,
+		Namespace: WMCONamespace,
 		Scheme:    directClient.Scheme(),
 	})
 	if err != nil {
@@ -130,7 +130,7 @@ func RunController(ctx context.Context, apiServerURL, saCA, saToken string) erro
 // NewServiceController returns a pointer to a ServiceController object
 func NewServiceController(ctx context.Context, client client.Client, mgr manager.Manager, nodeName string) *ServiceController {
 	return &ServiceController{client: client, Manager: mgr, ctx: ctx, nodeName: nodeName,
-		watchNamespace: wmcoNamespace}
+		watchNamespace: WMCONamespace}
 }
 
 // SetupWithManager sets up the controller with the Manager.
@@ -380,8 +380,8 @@ func GetAssociatedNode(c client.Client, addrs []net.Addr) (*core.Node, error) {
 	return node, nil
 }
 
-// localInterfaceAddresses returns a slice of all addresses associated with local network interfaces
-func localInterfaceAddresses() ([]net.Addr, error) {
+// LocalInterfaceAddresses returns a slice of all addresses associated with local network interfaces
+func LocalInterfaceAddresses() ([]net.Addr, error) {
 	var addresses []net.Addr
 	netIfs, err := net.Interfaces()
 	if err != nil {
--- a/pkg/daemon/controller/controller.go
+++ b/pkg/daemon/controller/controller.go
@@ -210,26 +210,10 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
 	return ctrl.Result{}, nil
 }
 
-// getExistingServices returns a map with the keys being all Windows services currently present on the VM
-func (sc *ServiceController) getExistingServices() (map[string]struct{}, error) {
-	// The most reliable way to determine if a service exists or not is to do a 'list' API call. It is possible to
-	// remove this call, and parse the error messages of a service 'open' API call, but I find that relying on human
-	// readable errors could cause issues when providing compatibility across different versions of Windows.
-	svcList, err := sc.ListServices()
-	if err != nil {
-		return nil, err
-	}
-	svcs := make(map[string]struct{})
-	for _, service := range svcList {
-		svcs[service] = struct{}{}
-	}
-	return svcs, nil
-}
-
 // reconcileServices ensures that all the services passed in via the services slice are created, configured properly
 // and started
 func (sc *ServiceController) reconcileServices(services []servicescm.Service) error {
-	existingSvcs, err := sc.getExistingServices()
+	existingSvcs, err := sc.GetServices()
 	if err != nil {
 		return errors.Wrap(err, "could not determine existing Windows services")
 	}
--- a/pkg/daemon/controller/controller.go
+++ b/pkg/daemon/controller/controller.go
@@ -33,6 +33,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
 	"k8s.io/client-go/util/jsonpath"
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -92,37 +93,25 @@ func RunController(ctx context.Context, apiServerURL, saCA, saToken string) erro
 		klog.Error(err)
 		return errors.Wrap(err, "error using service account to build config")
 	}
-
-	clientScheme := runtime.NewScheme()
-	err = clientgoscheme.AddToScheme(clientScheme)
-	if err != nil {
-		return err
-	}
 	// This is a client that reads directly from the server, not a cached client. This is required to be used here, as
 	// the cached client, created by ctrl.NewManager() will not be functional until the manager is started.
-	directClient, err := client.New(cfg, client.Options{Scheme: clientScheme})
+	directClient, err := NewDirectClient(cfg)
 	if err != nil {
 		return err
 	}
 
-	// use the client to find the name of the node associated with the VM this is running on
-	var nodes core.NodeList
-	err = directClient.List(ctx, &nodes)
-	if err != nil {
-		return err
-	}
 	addrs, err := localInterfaceAddresses()
 	if err != nil {
 		return err
 	}
-	node, err := findNodeByAddress(&nodes, addrs)
+	node, err := GetAssociatedNode(directClient, addrs)
 	if err != nil {
-		return err
+		errors.Wrap(err, "could not find node object associated with this instance")
 	}
 
 	ctrlMgr, err := ctrl.NewManager(cfg, ctrl.Options{
 		Namespace: wmcoNamespace,
-		Scheme:    clientScheme,
+		Scheme:    directClient.Scheme(),
 	})
 	if err != nil {
 		return errors.Wrap(err, "unable to start manager")
@@ -376,6 +365,35 @@ func (sc *ServiceController) resolveNodeVariables(svc servicescm.Service) (map[s
 func (sc *ServiceController) resolvePowershellVariables(svc servicescm.Service) (map[string]string, error) {
 	// TODO: Implement this function
 	return make(map[string]string), nil
+}
+
+// NewDirectClient creates and returns an authenticated client that reads directly from the API server.
+// It also returns the config and scheme used to created the client.
+func NewDirectClient(cfg *rest.Config) (client.Client, error) {
+	clientScheme := runtime.NewScheme()
+	err := clientgoscheme.AddToScheme(clientScheme)
+	if err = clientgoscheme.AddToScheme(clientScheme); err != nil {
+		return nil, err
+	}
+
+	directClient, err := client.New(cfg, client.Options{Scheme: clientScheme})
+	if err != nil {
+		return nil, err
+	}
+	return directClient, nil
+}
+
+// GetAssociatedNode uses the given client to find the name of the node associated with the VM this is running on
+func GetAssociatedNode(c client.Client, addrs []net.Addr) (*core.Node, error) {
+	var nodes core.NodeList
+	if err := c.List(context.TODO(), &nodes); err != nil {
+		return nil, err
+	}
+	node, err := findNodeByAddress(&nodes, addrs)
+	if err != nil {
+		return nil, err
+	}
+	return node, nil
 }
 
 // localInterfaceAddresses returns a slice of all addresses associated with local network interfaces
--- a/pkg/daemon/fake/manager.go
+++ b/pkg/daemon/fake/manager.go
@@ -93,8 +93,13 @@ func (t *testMgr) CreateService(name, exepath string, config mgr.Config, args ..
 	return &service, nil
 }
 
-func (t *testMgr) ListServices() ([]string, error) {
-	return t.svcList.listServiceNames(), nil
+func (t *testMgr) GetServices() (map[string]struct{}, error) {
+	svcsList := t.svcList.listServiceNames()
+	svcsMap := make(map[string]struct{})
+	for _, svc := range svcsList {
+		svcsMap[svc] = struct{}{}
+	}
+	return svcsMap, nil
 }
 
 func (t *testMgr) OpenService(name string) (winsvc.Service, error) {
--- a/pkg/daemon/fake/manager.go
+++ b/pkg/daemon/fake/manager.go
@@ -111,10 +111,14 @@ func (t *testMgr) OpenService(name string) (winsvc.Service, error) {
 }
 
 func (t *testMgr) DeleteService(name string) error {
-	_, exists := t.svcList.read(name)
+	winSvc, exists := t.svcList.read(name)
 	if !exists {
-		// This is to mimic the behavior of the Windows OS. Trying to delete a nonexistant service results in an error.
-		return fmt.Errorf("service %s does not exist", name)
+		// Nothing to do if it already does not exist
+		return nil
+	}
+	// Ensure service is stopped before deleting
+	if err := winsvc.EnsureServiceState(winSvc, svc.Stopped); err != nil {
+		return errors.Wrapf(err, "failed to stop service %q", name)
 	}
 	return t.svcList.remove(name)
 }
--- a/pkg/daemon/manager/manager.go
+++ b/pkg/daemon/manager/manager.go
@@ -11,8 +11,9 @@ import (
 type Manager interface {
 	// CreateService creates a Windows service with the given configuration parameters
 	CreateService(string, string, mgr.Config, ...string) (winsvc.Service, error)
-	// ListServices enumerates all the Windows services that exist on an instance
-	ListServices() ([]string, error)
+	// GetServices returns a map of all the Windows services that exist on an instance.
+	// The keys are service names and values are empty structs, used as 0 byte placeholders.
+	GetServices() (map[string]struct{}, error)
 	// OpenService gets the Windows service of the given name if it exists, by which it can be queried or controlled
 	OpenService(string) (winsvc.Service, error)
 	// DeleteService marks a Windows service of the given name for deletion, or an error if it does not exist
@@ -31,14 +32,28 @@ func (m *manager) CreateService(name, exepath string, config mgr.Config, args ..
 	service, err := underlyingMgr.CreateService(name, exepath, config, args...)
 	return winsvc.Service(service), err
 }
-func (m *manager) ListServices() ([]string, error) {
-	underlyingMgr := (*mgr.Mgr)(m)
-	return underlyingMgr.ListServices()
+
+func (m *manager) GetServices() (map[string]struct{}, error) {
+	// The most reliable way to determine if a service exists or not is to do a 'list' API call. It is possible to
+	// remove this call, and parse the error messages of a service 'open' API call, but I find that relying on human
+	// readable errors could cause issues when providing compatibility across different versions of Windows.
+	manager := (*mgr.Mgr)(m)
+	svcList, err := manager.ListServices()
+	if err != nil {
+		return nil, err
+	}
+	svcs := make(map[string]struct{})
+	for _, service := range svcList {
+		svcs[service] = struct{}{}
+	}
+	return svcs, nil
 }
+
 func (m *manager) OpenService(name string) (winsvc.Service, error) {
 	underlyingMgr := (*mgr.Mgr)(m)
 	return underlyingMgr.OpenService(name)
 }
+
 func (m *manager) DeleteService(name string) error {
 	underlyingMgr := (*mgr.Mgr)(m)
 	winSvc, err := underlyingMgr.OpenService(name)
--- a/pkg/daemon/manager/manager.go
+++ b/pkg/daemon/manager/manager.go
@@ -3,6 +3,8 @@
 package manager
 
 import (
+	"github.com/pkg/errors"
+	"golang.org/x/sys/windows/svc"
 	"golang.org/x/sys/windows/svc/mgr"
 
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/winsvc"
@@ -16,7 +18,7 @@ type Manager interface {
 	GetServices() (map[string]struct{}, error)
 	// OpenService gets the Windows service of the given name if it exists, by which it can be queried or controlled
 	OpenService(string) (winsvc.Service, error)
-	// DeleteService marks a Windows service of the given name for deletion, or an error if it does not exist
+	// DeleteService marks a Windows service of the given name for deletion. No-op if the service already doesn't exist
 	DeleteService(string) error
 }
 
@@ -55,12 +57,29 @@ func (m *manager) OpenService(name string) (winsvc.Service, error) {
 }
 
 func (m *manager) DeleteService(name string) error {
-	underlyingMgr := (*mgr.Mgr)(m)
-	winSvc, err := underlyingMgr.OpenService(name)
+	existingSvcs, err := m.GetServices()
 	if err != nil {
 		return err
 	}
-	return winSvc.Delete()
+	// Nothing to do if it already does not exist
+	if _, present := existingSvcs[name]; !present {
+		return nil
+	}
+
+	manager := (*mgr.Mgr)(m)
+	service, err := manager.OpenService(name)
+	if err != nil {
+		return errors.Wrapf(err, "failed to open service %q", name)
+	}
+	defer service.Close()
+	// Ensure service is stopped before deleting
+	if err = winsvc.EnsureServiceState(service, svc.Stopped); err != nil {
+		return errors.Wrapf(err, "failed to stop service %q", name)
+	}
+	if err = service.Delete(); err != nil {
+		return errors.Wrapf(err, "failed to delete service %q", name)
+	}
+	return nil
 }
 
 func New() (Manager, error) {
--- a/pkg/servicescm/servicescm.go
+++ b/pkg/servicescm/servicescm.go
@@ -1,13 +1,16 @@
 package servicescm
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"reflect"
 	"sort"
+	"strings"
 
 	core "k8s.io/api/core/v1"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/openshift/windows-machine-config-operator/version"
 	"github.com/pkg/errors"
@@ -97,6 +100,21 @@ func NewData(services *[]Service, files *[]FileInfo) (*Data, error) {
 		return nil, errors.Wrap(err, "unable to create services ConfigMap data object")
 	}
 	return cmData, nil
+}
+
+// List returns a list of all windows-services ConfigMaps in the given namespace
+func List(c client.Client, ctx context.Context, namespace string) ([]core.ConfigMap, error) {
+	watchNamespaceCMs := &core.ConfigMapList{}
+	if err := c.List(ctx, watchNamespaceCMs, &client.ListOptions{Namespace: namespace}); err != nil {
+		return nil, err
+	}
+	servicesConfigMaps := []core.ConfigMap{}
+	for _, cm := range watchNamespaceCMs.Items {
+		if strings.HasPrefix(cm.Name, NamePrefix) {
+			servicesConfigMaps = append(servicesConfigMaps, cm)
+		}
+	}
+	return servicesConfigMaps, nil
 }
 
 // Generate creates an immutable service ConfigMap which provides WICD with the specifications
--- a/cmd/daemon/bootstrap.go
+++ b/cmd/daemon/bootstrap.go
@@ -22,10 +22,7 @@ import (
 	"context"
 
 	"github.com/spf13/cobra"
-	"k8s.io/apimachinery/pkg/runtime"
-	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/klog/v2"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/config"
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/controller"
@@ -58,15 +55,10 @@ func runBootstrapCmd(cmd *cobra.Command, args []string) {
 		klog.Exitf("error using service account to build config: %s", err.Error())
 	}
 
-	clientScheme := runtime.NewScheme()
-	err = clientgoscheme.AddToScheme(clientScheme)
-	if err = clientgoscheme.AddToScheme(clientScheme); err != nil {
-		klog.Exit(err.Error())
-	}
-	// Client reads directly from the server. Cannot use a cached client as no manager will be started to populate cache
-	directClient, err := client.New(cfg, client.Options{Scheme: clientScheme})
+	// Cannot use a cached client as no manager will be started to populate cache
+	directClient, err := controller.NewDirectClient(cfg)
 	if err != nil {
-		klog.Exit(err.Error())
+		klog.Exitf("could not create authenticated client from service account: %s", err.Error())
 	}
 
 	svcMgr, err := manager.New()
@@ -75,7 +67,7 @@ func runBootstrapCmd(cmd *cobra.Command, args []string) {
 	}
 	sc := controller.NewServiceController(context.TODO(), directClient, svcMgr, "")
 
-	klog.Info("bootstrapping node")
+	klog.Info("bootstrapping Windows instance")
 	if err := sc.Bootstrap(desiredVersion); err != nil {
 		klog.Exit(err.Error())
 	}
--- a/controllers/configmap_controller.go
+++ b/controllers/configmap_controller.go
@@ -200,7 +200,7 @@ func (r *ConfigMapReconciler) removeOutdatedServicesConfigMaps(ctx context.Conte
 	}
 	versionAnnotations := getVersionAnnotations(nodes.Items)
 
-	servicesConfigMaps, err := r.listServiceConfigMaps(ctx)
+	servicesConfigMaps, err := servicescm.List(r.client, ctx, r.watchNamespace)
 	if err != nil {
 		return errors.Wrapf(err, "unable to retrieve list of services ConfigMaps")
 	}
@@ -231,21 +231,6 @@ func isTiedToRelevantVersion(v string, versions map[string]struct{}) bool {
 		}
 	}
 	return false
-}
-
-// listServiceConfigMaps returns a list of all windows-services ConfigMaps in the operator namespace
-func (r *ConfigMapReconciler) listServiceConfigMaps(ctx context.Context) ([]core.ConfigMap, error) {
-	watchNamespaceCMs := &core.ConfigMapList{}
-	if err := r.client.List(ctx, watchNamespaceCMs, &client.ListOptions{Namespace: r.watchNamespace}); err != nil {
-		return nil, err
-	}
-	servicesConfigMaps := []core.ConfigMap{}
-	for _, cm := range watchNamespaceCMs.Items {
-		if strings.HasPrefix(cm.Name, servicescm.NamePrefix) {
-			servicesConfigMaps = append(servicesConfigMaps, cm)
-		}
-	}
-	return servicesConfigMaps, nil
 }
 
 // reconcileNodes corrects the discrepancy between the "expected" instances, and the "actual" Node list
--- a/pkg/daemon/controller/controller_test.go
+++ b/pkg/daemon/controller/controller_test.go
@@ -448,12 +448,12 @@ func testServicesCreatedAsExpected(t *testing.T, createdServices map[string]fake
 // getAllFakeServices accepts a mocked Windows service manager, and returns a map of copies of all existing Windows
 // services
 func getAllFakeServices(svcMgr manager.Manager) (map[string]fake.FakeService, error) {
-	svcs, err := svcMgr.ListServices()
+	svcs, err := svcMgr.GetServices()
 	if err != nil {
 		return nil, err
 	}
 	fakeServices := make(map[string]fake.FakeService)
-	for _, winServiceName := range svcs {
+	for winServiceName := range svcs {
 		winService, err := svcMgr.OpenService(winServiceName)
 		if err != nil {
 			return nil, err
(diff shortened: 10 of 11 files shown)


saifshaikh48:
/test wicd-unit-vsphere

passed locally

pkg/daemon/controller/controller.go
```diff
@@ -203,6 +203,130 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
 	return ctrl.Result{}, nil
 }
 
+// Cleanup removes an instance from the cluster. If the upgradeInPlace flag is true, the node object is not deleted.
+// TODO: Remove/ensure files does not exist and also remove HNS networks, like WMCO does through windows.Deconfigure()
```
saifshaikh48, line 210: I'll remove this TODO after a round of reviews. As per the enhancement, this will remain a responsibility of WMCO.

pkg/daemon/controller/controller.go
```diff
@@ -377,6 +360,44 @@ func (sc *ServiceController) resolvePowershellVariables(svc servicescm.Service)
 	return make(map[string]string), nil
 }
 
+// NewDirectClient creates and returns an authenticated client that reads directly from the API server.
+// It also returns the config and shceme used to created the client.
```
sebsoto, line 552: shceme -> scheme

pkg/daemon/controller/controller.go
```diff
@@ -377,6 +360,44 @@ func (sc *ServiceController) resolvePowershellVariables(svc servicescm.Service)
...
+
+	directClient, err := client.New(cfg, client.Options{Scheme: clientScheme})
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	return directClient, cfg, clientScheme, nil
+}
+
+// GetUnderlyingNode uses the given client to find the name of the node associated with the VM this is running on
+func GetUnderlyingNode(c client.Client) (*core.Node, error) {
```
sebsoto, line 573: Consider renaming this `GetInstanceNode`/ `GetAssociatedNode`, or something similar. I'd argue `Overlying` would be more correct than `Underlying`, as the node is an abstraction layer on top of the instance :wink: .

sebsoto (changes requested):

pkg/daemon/controller/controller.go
```diff
@@ -203,6 +203,130 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
 	return ctrl.Result{}, nil
 }
 
+// Cleanup removes an instance from the cluster. If the upgradeInPlace flag is true, the node object is not deleted.
+// TODO: Remove/ensure files does not exist and also remove HNS networks, like WMCO does through windows.Deconfigure()
+func (sc *ServiceController) Cleanup(upgradeInPlace bool) error {
```
sebsoto, line 211: I'd prefer unit tests that cover the functionality you are introducing be added within the same commit

pkg/daemon/controller/controller.go
```diff
@@ -203,6 +203,130 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
...
+// removeDependentServices stops and deletes all services that depend on the given service
+func (sc *ServiceController) removeDependentServices(service string, existingSvcs map[string]struct{}) error {
+	for existingSvc := range existingSvcs {
+		existingWinSvcObj, err := sc.OpenService(existingSvc)
+		if err != nil {
+			return errors.Wrapf(err, "error opening %s service handle", existingSvc)
+		}
+		existingWinSvcConfig, err := existingWinSvcObj.Config()
+		if err != nil {
+			return errors.Wrapf(err, "error getting %s service config", existingSvc)
```
jrvaldes, line 372: Following the [above](https://github.com/openshift/windows-machine-config-operator/pull/1154/commits/93f50d7c64b18ea38f73d5c8dce98bd216c147cb#diff-e847d452a6040c690f8f6011bd0ecf612cc53aa12ce8ff997b29bb85870328f6R293) logging pattern, consider pushing the parameters to the end of the log message. For example:

```go
return errors.Wrapf(err, "error getting configuration for service %s", existingSvc)
```

pkg/daemon/controller/controller.go
```diff
@@ -203,6 +203,130 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
...
+}
+
+// ensureServiceIsRemoved deletes the given service and all services that depend on it, even those unmanaged by WICD.
+func (sc *ServiceController) ensureServiceIsRemoved(service string) error {
+	existingSvcs, err := sc.getExistingServices()
+	if err != nil {
+		return errors.Wrap(err, "could not determine existing Windows services")
+	}
+	// Nothing to do if it already does not exist
+	if _, present := existingSvcs[service]; !present {
```
jrvaldes, line 316: Is this a genuinely breaking statement? No other checks are required if the `service` is not in the list of existing services. Correct?

jrvaldes (changes requested):
Nice work @saifshaikh48, added a few comments. PTAL

pkg/daemon/controller/controller.go
jrvaldes: +1

pkg/daemon/controller/controller.go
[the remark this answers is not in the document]
jrvaldes: This is a [duplicate](https://github.com/openshift/windows-machine-config-operator/pull/1154#discussion_r934884245). Looks like @sebsoto  read my mind :)

cmd/daemon/cleanup.go
```diff
@@ -0,0 +1,70 @@
...
+var (
+	cleanupCmd = &cobra.Command{
+		Use:   "cleanup",
+		Short: "Cleans up WMCO-managed Windows services and files",
+		Long: "Stops and removes all Windows services and files as part of Node deconfiguration, " +
+			"according to information given by Windows Service ConfigMaps present within the cluster",
+		Run: runCleanupCmd,
+	}
+	// an optional flag that, when supplied, instructs WICD to deconfigure an instance without removing the Node object
+	upgradeInPlace bool
```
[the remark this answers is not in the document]
sebsoto: +1 for keepNode or something along those lines
IMO in this case its better to name according to function and not potential use case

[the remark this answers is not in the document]
sebsoto: preserveNode might be better. This also ties in with the flag `--upgrade` flag, its a bit misleading as this isnt actually upgrading the VM,

go.mod
```diff
@@ -20,9 +20,9 @@ require (
 	golang.org/x/crypto v0.0.0-20220214200702-86341886e292
 	golang.org/x/mod v0.6.0-dev.0.20220106191415-9b9b3d81d5e3
 	golang.org/x/sys v0.0.0-20220209214540-3681064d5158
-	k8s.io/api v0.24.1
-	k8s.io/apimachinery v0.24.1
-	k8s.io/client-go v0.24.1
+	k8s.io/api v0.24.3
```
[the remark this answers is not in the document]
saifshaikh48: TBH I was a bit confused when my editor was giving me errors. But after vendoring the errors disappear so looks like this is needed.

cmd/daemon/cleanup.go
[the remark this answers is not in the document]
saifshaikh48: I'll go with preserveNode, thanks for the suggestions

cmd/daemon/cleanup.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+func init() {
+	rootCmd.AddCommand(cleanupCmd)
+	cleanupCmd.PersistentFlags().BoolVar(&upgradeInPlace, "upgrade", true, "Node version upgrade flag")
```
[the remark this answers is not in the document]
saifshaikh48: On second look, it should be `false` according to the enhancement. Thanks for drawing attention to this. I'll make the change and add it to the default behavior to the description.

pkg/daemon/controller/controller.go
```diff
@@ -203,6 +203,130 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
 	return ctrl.Result{}, nil
 }
 
+// Cleanup removes an instance from the cluster. If the upgradeInPlace flag is true, the node object is not deleted.
+// TODO: Remove/ensure files does not exist and also remove HNS networks, like WMCO does through windows.Deconfigure()
+func (sc *ServiceController) Cleanup(upgradeInPlace bool) error {
+	klog.Infof("cleaning up %s, node upgrade=%t", sc.nodeName, upgradeInPlace)
```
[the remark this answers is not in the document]
saifshaikh48: Since we log before invoking `Cleanup` and if we delete the node object, I'll remove this log altogether. This was more a debug log anyway.

pkg/daemon/controller/controller.go
saifshaikh48: Yep, that's correct. If a service does not exist, there's no way to clean it up further.

jrvaldes: Makes sense.

go.mod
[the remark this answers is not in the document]
jrvaldes: Sounds good, I was wondering if we could move it into a separate PR to reduce the number of changes here.

pkg/daemon/controller/controller.go
```diff
@@ -220,6 +203,191 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
...
+		}
+	}
+
+	if !preserveNode {
+		if err = sc.client.Delete(sc.ctx, &node); err != nil {
+			return errors.Wrapf(err, "error deleting node %s", node.GetName())
+		}
+		klog.Infof("deleted node %s", node.GetName())
+	}
+	return nil
```
jrvaldes, line 302: I suggest, as possible, keeping the default execution flow leaning towards the positive side while branching edge cases and exceptions. For example, this code snippet could be refactored as:

```go
	if preserveNode {
		// nothing to do
		return nil
	}

	klog.Infof("deleting node %s", node.GetName())
	return sc.client.Delete(sc.ctx, &node)
```

It may be easy to read/maintain.

pkg/daemon/controller/controller.go
```diff
@@ -202,6 +202,128 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
...
+	if !present {
+		return errors.Errorf("node missing desired version annotation")
+	}
+	// Fetch the CM of the desired version
+	cm := &core.ConfigMap{}
+	err = sc.client.Get(sc.ctx,
+		client.ObjectKey{Namespace: sc.watchNamespace, Name: servicescm.NamePrefix + desiredVersion}, cm)
+	if err != nil {
+		return err
+	}
```
sebsoto, line 232: This is done in multiple places, you can move this to its own function and return a custom error to identify that the annotation is missing for the logic in the next commit

pkg/daemon/controller/controller.go
```diff
@@ -377,6 +359,44 @@ func (sc *ServiceController) resolvePowershellVariables(svc servicescm.Service)
...
+// GetAssociatedNode uses the given client to find the name of the node associated with the VM this is running on
+func GetAssociatedNode(c client.Client) (*core.Node, error) {
+	var nodes core.NodeList
+	if err := c.List(context.TODO(), &nodes); err != nil {
+		return nil, err
+	}
+	addrs, err := localInterfaceAddresses()
+	if err != nil {
+		return nil, err
+	}
```
sebsoto, line 578: This section prevents the functions using this from being unit tested, I'd prefer that this section be moved out of the function, and instead `[]net.Addrs`  be passed in.

pkg/daemon/controller/controller.go
```diff
@@ -202,6 +202,128 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
...
+		client.ObjectKey{Namespace: sc.watchNamespace, Name: servicescm.NamePrefix + desiredVersion}, cm)
+	if err != nil {
+		return err
+	}
+	return sc.removeServices(cm, node, preserveNode)
+}
+
+// removeServices removes all the services defined in the given ConfigMap from this instance.
+// The given Node object is deleted if the given flag is false
+func (sc *ServiceController) removeServices(cm *core.ConfigMap, node core.Node, preserveNode bool) error {
```
sebsoto, line 284: Theres no need to pass in the full CM here, you can parse the CM in the same place you `Get()` it. And then pass in the data struct to this function.

pkg/daemon/controller/controller.go
```diff
@@ -202,6 +202,128 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
...
+			return errors.Wrapf(err, "failed to delete service %s", name)
+		}
+		klog.Infof("removed service %s", name)
+		return nil
+	}
+
+	// Ensure no dependent services are running before stopping this one. This is needed to safely stop Windows services
+	if err := sc.removeDependentServices(name, existingSvcs); err != nil {
+		return err
+	}
```
sebsoto, line 347: I'm not sure why this needs to be done. Priority gives us the order in which services depend on each other. You should just be going through the services in reverse order to stop them.

sebsoto (changes requested):

pkg/daemon/controller/controller.go
sebsoto: I guess this would be required if we dont have the ConfigMap to use. Please limit the use of this function to that case.

pkg/daemon/controller/controller.go
```diff
@@ -377,6 +359,44 @@ func (sc *ServiceController) resolvePowershellVariables(svc servicescm.Service)
 	return make(map[string]string), nil
 }
 
+// NewDirectClient creates and returns an authenticated client that reads directly from the API server.
+// It also returns the config and scheme used to created the client.
+func NewDirectClient(apiServerURL, saCA, saToken string) (client.Client, *rest.Config, *runtime.Scheme, error) {
```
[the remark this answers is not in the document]
aravindhp: I agree with @sebsoto. the function `NewDirectClient()` should only return the direct client and an error.

pkg/daemon/controller/controller.go
aravindhp: Should we log the errors and make a best effort to clean up the node? Why should we not try to clean up the instances just because the node object is missing?

pkg/daemon/controller/controller.go
```diff
@@ -202,6 +202,128 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
...
+// The given Node object is deleted if the given flag is false
+func (sc *ServiceController) removeServices(cm *core.ConfigMap, node core.Node, preserveNode bool) error {
+	klog.Infof("removing services defined in ConfigMap %s", cm.Name)
+	cmData, err := servicescm.Parse(cm.Data)
+	if err != nil {
+		return err
+	}
+	for _, service := range cmData.Services {
+		if err := sc.ensureServiceIsRemoved(service.Name); err != nil {
+			return err
```
aravindhp, line 292: We should collect the errors and not fail on the first service we were not able to remove.

pkg/servicescm/servicescm.go
```diff
@@ -99,6 +102,21 @@ func NewData(services *[]Service, files *[]FileInfo) (*Data, error) {
 	return cmData, nil
 }
 
+// ListServiceConfigMaps returns a list of all windows-services ConfigMaps in the given namespace
+func ListServiceConfigMaps(c client.Client, ctx context.Context, namespace string) ([]core.ConfigMap, error) {
```
aravindhp, line 106: This can just be `List` as the call site will read `servicevm.List()` and it being a list implies plural.

pkg/daemon/controller/controller.go
```diff
@@ -202,6 +202,128 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
 	return ctrl.Result{}, nil
 }
 
+// Cleanup removes an instance from the cluster. If the parameter flag is true, the node object is not deleted.
+func (sc *ServiceController) Cleanup(preserveNode bool) error {
```
aravindhp, line 209: Why is `Cleanup` of services a function of the controller? Shouldn't the `Mgr` be the one to handle that?

pkg/daemon/controller/controller.go
aravindhp: Same question here:  Shouldn't the Mgr be the one to handle this?

aravindhp (changes requested):
Thanks for working on this, @saifshaikh48. Lets pause on this for a bit and rethink the object design.

cmd/daemon/cleanup.go
```diff
@@ -0,0 +1,71 @@
...
+		Run: runCleanupCmd,
+	}
+	// preserveNode is an optional flag that instructs WICD to deconfigure an instance without deleting the Node object
+	preserveNode bool
+)
+
+func init() {
+	rootCmd.AddCommand(cleanupCmd)
+	cleanupCmd.PersistentFlags().BoolVar(&preserveNode, "preserveNode", false,
+		"Flag to keep an instance's associated node in the cluster. Defaults to false, which deletes the node object.")
```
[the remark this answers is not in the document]
saifshaikh48: The option will be used in the BYOH upgrade case. Deconfigure without deleting node object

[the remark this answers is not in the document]
aravindhp: Please add a comment mentioning that.

pkg/daemon/controller/controller.go
saifshaikh48: Good point. The best effort cleanup was originally only if the desired version annotation or services ConfigMap tied to it was missing. I'll extend that to occur if the node object is missing.

pkg/daemon/controller/controller.go
saifshaikh48: I don't quite understand how the unit tests were running (and passing) without making this change. I made it anyway

pkg/daemon/controller/controller.go
saifshaikh48: Thanks for the suggestions, took them into consideration with the latest changes.

pkg/daemon/cleanup/cleanup.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/openshift/windows-machine-config-operator/pkg/daemon/controller"
+	"github.com/openshift/windows-machine-config-operator/pkg/daemon/manager"
+	"github.com/openshift/windows-machine-config-operator/pkg/daemon/winsvc"
+	"github.com/openshift/windows-machine-config-operator/pkg/nodeconfig"
+	"github.com/openshift/windows-machine-config-operator/pkg/servicescm"
+)
+
+// Execute removes an instance from the cluster. If the parameter flag is true, the node object is not deleted.
+// If we are able to get the services ConfigMap tied to the desired version, all services defined in it are cleaned up.
+// TODO: Best effort cleanup if node object, desired version annotation, or services ConfigMap are not attainable
```
saifshaikh48: The best effort cleanup is implemented in the last commit. I split them up due to the amount of code being introduced.

pkg/daemon/fake/manager.go
```diff
@@ -123,6 +126,89 @@ func (t *testMgr) GetExistingServices() (map[string]struct{}, error) {
 	return svcsMap, nil
 }
 
+func (t *testMgr) EnsureServiceIsRemoved(name string, managedOnly bool) error {
```
saifshaikh48, line 129: These methods on the fake manager are required for testing purposes. They are exact copies of those upon the real manager - if anyone has suggestions on how to avoid this, please let me know.

alinaryan (commented):
Good work, just a couple comments

pkg/daemon/cleanup/cleanup_test.go
```diff
@@ -0,0 +1,322 @@
...
+				_, present := allServices[svc]
+				assert.Truef(t, present, "service %s not present as expected", svc)
+			}
+			// Depending on if the test is an upgrade cleanup case, check if the node object has been deleted or not
+			nodes := &core.NodeList{}
+			require.NoError(t, fakeClient.List(context.TODO(), nodes))
+			if test.preserveNode {
+				assert.NotZero(t, len(nodes.Items))
+			} else {
+				assert.Zero(t, len(nodes.Items))
```
sebsoto, line 517: I suggest using `assert.Empty()` and `NotEmpty()`. The `len()` calls are unecessary

pkg/daemon/cleanup/cleanup.go
```diff
@@ -33,30 +35,41 @@ import (
...
 // If we are able to get the services ConfigMap tied to the desired version, all services defined in it are cleaned up.
-// TODO: Best effort cleanup if node object, desired version annotation, or services ConfigMap are not attainable
+// Otherwise, perform cleanup based on a combination of the OpenShift managed tag and the latest services ConfigMap.
 func Execute(c client.Client, addrs []net.Addr, svcMgr manager.Manager, preserveNode bool) error {
 	ctx := context.TODO()
 	node, err := controller.GetAssociatedNode(c, addrs)
 	if err != nil {
-		klog.Exitf("could not node object associated with this instance: %s", err.Error())
+		// If the node object is not attainable use a best effort cleanup stategy
+		klog.Warning(err.Error())
```
sebsoto, line 49: Since this is a warning, it would be good to add that the ConfigMap based cleanup will be skipped.

pkg/daemon/cleanup/cleanup.go
```diff
@@ -33,30 +35,41 @@ import (
...
+		return outOfSyncCleanup(c, ctx, svcMgr, node, preserveNode)
 	}
 	klog.Infof("cleaning up managed services on node %s", node.Name)
 
 	desiredVersion, present := node.Annotations[nodeconfig.DesiredVersionAnnotation]
 	if !present {
-		return errors.Errorf("node missing desired version annotation")
+		// If the desired version annotation is not present, use a best effort cleanup stategy
+		klog.Warning("node missing desired version annotation")
```
sebsoto, line 57: Same as above

pkg/daemon/cleanup/cleanup.go
```diff
@@ -33,30 +35,41 @@ import (
...
+		return outOfSyncCleanup(c, ctx, svcMgr, node, preserveNode)
 	}
 	// Fetch the CM of the desired version
 	cm := &core.ConfigMap{}
 	err = c.Get(ctx,
 		client.ObjectKey{Namespace: controller.WmcoNamespace, Name: servicescm.NamePrefix + desiredVersion}, cm)
 	if err != nil {
+		if k8sapierrors.IsNotFound(err) {
+			klog.Warningf("unable to find desired version ConfigMap %s", servicescm.NamePrefix+desiredVersion)
```
sebsoto, line 66: Same as above

pkg/daemon/cleanup/cleanup.go
```diff
@@ -33,30 +35,41 @@ import (
...
+			// If the ConfigMap tied to the desired version annotation is not present, use a best effort cleanup stategy
+			return outOfSyncCleanup(c, ctx, svcMgr, node, preserveNode)
+		}
 		return err
 	}
+
 	return cleanCMServices(c, ctx, svcMgr, cm, node, preserveNode)
```
sebsoto, line 73: Theres a lot of repeating syntax here with error checks + log warning + doing the outOfSyncCleanup.
This can be fixed by using an anonymous func, or a named func depending on what works best.
Take a look at the nodeconfig code to what [we're doing there](https://github.com/openshift/windows-machine-config-operator/blob/3492116e5071d2184226b67d47ce04040beac30d/pkg/nodeconfig/nodeconfig.go#L204).

This function can be refactored like this:
```
	cm := &core.ConfigMap{}
	err := func()error {
		node, err := controller.GetAssociatedNode(c, addrs)
		if err != nil {
			// If the node object is not attainable use a best effort cleanup stategy
			return err
		}

		desiredVersion, present := node.Annotations[nodeconfig.DesiredVersionAnnotation]
		if !present {
			// If the desired version annotation is not present, use a best effort cleanup stategy
			return errors.Wrap("node missing desired version annotation",err)
		}
		// Fetch the CM of the desired version
		err = c.Get(ctx,
			client.ObjectKey{Namespace: controller.WmcoNamespace, Name: servicescm.NamePrefix + desiredVersion}, cm)
		if err != nil {
			return errors.Wrap()
		}
		return nil
	}()
	if err == nil {
		// no issue getting cm
		// return with CM cleanup
	} 
	klog.Warningf(err)
	// return with out of sync cleanup
```

pkg/daemon/manager/manager.go
```diff
@@ -19,6 +25,9 @@ type Manager interface {
 	DeleteService(string) error
 	// GetExistingServices is a wrapper of ListServices, returning a map of all services that exist on an instance
 	GetExistingServices() (map[string]struct{}, error)
+	// EnsureServiceIsRemoved deletes the given service and all services that depend on it, even those unmanaged by WICD
+	// If the managedOnly flag is true, only services with the managed tag will be removed.
+	EnsureServiceIsRemoved(string, bool) error
```
sebsoto, line 30: I'm not sold on both `DeleteService` and this method being necessary.
Do you have any reason for them to not be combined?

sebsoto (changes requested):
Great work @saifshaikh48 
Looking forward to getting this merged

pkg/daemon/manager/manager.go
saifshaikh48: I wanted to keep this separate as it also deletes dependencies. It felt right to provide a less expensive option by exposing the `DeleteService` method instead of just making it a helper

sebsoto: Thanks that reasoning makes sense.

In that case two methods can make sense. But I see no reason for `DeleteService` to not 'ensure' that the service is deleted, and not throw an error if it doesnt exist.

I don't see a usecase for WICD where failure upon the service not existing is useful. Exposing the real API call's behavior isn't helpful for us, and our code can be simplified if both interface methods maintain the same behavior in the case that the service does not exist. This change would require a rename of both methods for clarity.

sebsoto: I'm thinking through this more, and I think what might be complicating things here is adding the concept of a managed service to the `Manager` interface. Bringing that concept out of this package is something that should be considered.

Additionally, why does the logic around the manged services need to exist? What services not managed by WICD are going to depend on services managed by WICD? I don't see this as a case we should be willing to support as it will prevent WICD from performing its required functionality.

pkg/daemon/cleanup/cleanup.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/openshift/windows-machine-config-operator/pkg/daemon/manager"
+	"github.com/openshift/windows-machine-config-operator/pkg/daemon/winsvc"
+	"github.com/openshift/windows-machine-config-operator/pkg/nodeconfig"
+	"github.com/openshift/windows-machine-config-operator/pkg/servicescm"
+)
+
+// Execute removes an instance from the cluster. If the parameter flag is true, the node object is not deleted.
+// If we are able to get the services ConfigMap tied to the desired version, all services defined in it are cleaned up.
+// TODO: Best effort cleanup if node object, desired version annotation, or services ConfigMap are not attainable
+func Execute(c client.Client, addrs []net.Addr, svcMgr manager.Manager, preserveNode bool) error {
```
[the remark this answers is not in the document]
saifshaikh48: Haha yes trying to avoid the cleanup stutter. How does cleanup.Deconfigure() sound?

pkg/daemon/cleanup/cleanup.go
```diff
@@ -0,0 +1,109 @@
...
+	for i := len(cmData.Services) - 1; i >= 0; i-- {
+		service := cmData.Services[i]
+		if _, ok := existingSvcs[service.Name]; !ok {
+			// Nothing to do if it already does not exist
+			continue
+		}
+		// Open the service to get a service handle
+		winSvcObj, err := svcMgr.OpenService(service.Name)
+		if err != nil {
+			failedRemovals = append(failedRemovals, err.Error())
```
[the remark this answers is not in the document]
saifshaikh48: I'll append the error message and the service name on failure

pkg/daemon/manager/manager.go
saifshaikh48: > What services not managed by WICD are going to depend on services managed by WICD?

One use case I can think of -- it's needed for the intermediate time period where some services are WICD managed and some are not. For example an unmanaged hybrid-overlay may depend on a WICD-managed kubelet service.

aravindhp (changes requested):
Thanks for working on this @saifshaikh48. Please address my comments.

pkg/daemon/cleanup/cleanup.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+		// Ensure service is stopped before deleting
+		if err := winsvc.EnsureServiceState(winSvcObj, svc.Stopped); err != nil {
+			failedRemovals = append(failedRemovals, err.Error())
+		}
+		klog.Infof("removing service %s", service.Name)
+		if err := svcMgr.DeleteService(service.Name); err != nil {
```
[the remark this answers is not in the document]
saifshaikh48: I assume you meant  "Can you _delete_ a service that you were not able to open/stop?" -- the answer is no. 
Note that the happy path (where we can get the desired services CM for the node) is not best effort. It should always be able to find, stop, and delete the services listed in the ConfigMap, otherwise that is an error worthy event.

pkg/daemon/cleanup/cleanup.go
```diff
@@ -0,0 +1,109 @@
...
+	cmData, err := servicescm.Parse(cm.Data)
+	if err != nil {
+		return err
+	}
+	existingSvcs, err := svcMgr.GetExistingServices()
+	if err != nil {
+		return err
+	}
+
+	klog.Infof("removing services defined in ConfigMap %s", cm.Name)
```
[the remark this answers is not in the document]
saifshaikh48: They are printed a few lines below, one by one as they are deleted.

pkg/daemon/cleanup/cleanup_test.go
```diff
@@ -0,0 +1,322 @@
...
+		},
+		{
+			name:                      "Upgrade scenario with no services",
+			existingServices:          map[string]*fake.FakeService{},
+			expectedRemovedServices:   []string{},
+			expectedRemainingServices: []string{},
+			preserveNode:              true,
+			configMapServices:         []servicescm.Service{},
+		},
+		{
```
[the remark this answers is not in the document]
saifshaikh48: Tried to add one below: `Defined service doesn't exist on node`

Ensures that trying to delete a service that is not there/can't be opened doesn't cause panic or halt cleanup.

pkg/daemon/cleanup/cleanup.go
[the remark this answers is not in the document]
aravindhp: >Can you delete a service that you were not able to open/stop?" -- the answer is no.

Then you should be logging and continuing not moving on to deleting on error when opening or stopping.

>Note that the happy path (where we can get the desired services CM for the node) is not best effort. It should always be able to find, stop, and delete the services listed in the ConfigMap, otherwise that is an error worthy event.

I don't agree with that. It always should be best effort, while you collect and display all the errors you hit. Stopping at the first error should not be the option.

pkg/daemon/cleanup/cleanup.go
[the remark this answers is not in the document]
aravindhp: I would rather collect all of them and have one message.

pkg/daemon/manager/manager.go
```diff
@@ -17,6 +17,8 @@ type Manager interface {
 	OpenService(string) (winsvc.Service, error)
 	// DeleteService marks a Windows service of the given name for deletion, or an error if it does not exist
 	DeleteService(string) error
+	// GetExistingServices is a wrapper of ListServices, returning a map of all services that exist on an instance
```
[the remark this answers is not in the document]
saifshaikh48: I decided to take the suggestion above. Just one function `GetServices()` that returns a map

pkg/daemon/manager/manager.go
```diff
@@ -75,6 +78,72 @@ func (m *manager) DeleteService(name string) error {
...
+	if err != nil {
+		return errors.Wrapf(err, "unable to open %s service handler", name)
+	}
+	defer winSvcObj.Close()
+
+	status, err := winSvcObj.Query()
+	if err != nil {
+		return errors.Wrapf(err, "unable to query %s service status", name)
+	}
+	if status.State != svc.Stopped {
```
aravindhp, line 101: Is there some other state that will prevent deletion?

pkg/daemon/manager/manager.go
```diff
@@ -53,7 +56,7 @@ func (m *manager) OpenService(name string) (winsvc.Service, error) {
 }
 
 func (m *manager) DeleteService(name string) error {
-	existingSvcs, err := m.GetServicesMap()
+	existingSvcs, err := m.GetServices()
```
aravindhp, line 59: This should be in the commit where you introduce `GetServices()`

aravindhp (changes requested):

pkg/daemon/manager/manager.go
```diff
@@ -51,11 +53,25 @@ func (m *manager) OpenService(name string) (winsvc.Service, error) {
...
+	if _, present := existingSvcs[name]; !present {
+		return nil
+	}
+
 	underlyingMgr := (*mgr.Mgr)(m)
 	winSvc, err := underlyingMgr.OpenService(name)
 	if err != nil {
 		return err
 	}
+	defer winSvc.Close()
```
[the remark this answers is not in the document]
saifshaikh48: From looking online it seems like this is the standard pattern
- open service
- defer close
- do something with service
- delete service
so closing the service obj (which is some OS level resource cleanup) happens after it is marked for deletion from the Windows service manager database on the VM

pkg/daemon/manager/manager.go
```diff
@@ -51,11 +53,25 @@ func (m *manager) OpenService(name string) (winsvc.Service, error) {
...
+	// Ensure service is stopped before deleting
+	if err := winsvc.EnsureServiceState(winSvc, svc.Stopped); err != nil {
```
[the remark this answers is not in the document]
saifshaikh48: `svc` also exists already :)
I'll the change it to `winSvcObj` as that's what is used elsewhere in the codebase (WICD controller)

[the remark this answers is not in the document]
aravindhp: I suggest staying away from `*Obj`. That's similar to calling a variable a `list` 😄 How about `service`?

pkg/daemon/cleanup/cleanup.go
```diff
@@ -0,0 +1,173 @@
...
+		return err
+	}
+	// Build up log message and failures
+	servicesRemoved := []string{}
+	var failedRemovals error
+	// The services are ordered by increasing priority already, so stop them in reverse order to avoid dependency issues
+	for i := len(cmData.Services) - 1; i >= 0; i-- {
+		service := cmData.Services[i]
+		if err := svcMgr.DeleteService(service.Name); err != nil {
+			failedRemovals = errors.Errorf("unable to delete %s (%v); %s", service.Name, err, failedRemovals)
```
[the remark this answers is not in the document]
sebsoto: This is going to cause the logs to be a mess by jamming all the errors into a single line.
That question also makes me wonder how the WMCO logs are going to look when running this.
When running this on a terminal most of the logging being added here is great.
But we do not want to overload WMCO error messages and make them impossible to read.

pkg/daemon/cleanup/cleanup.go
```diff
@@ -0,0 +1,173 @@
...
+		return failedRemovals
+	}
+	klog.Infof("deleted services %v from ConfigMap %s", servicesRemoved, cm.Name)
+	return nil
+}
+
+// outOfSyncCleanup removes all services on the instance with the Openshift managed tag, as well as any services defined
+// in the latest retreivable version of the services ConfigMap
+func outOfSyncCleanup(c client.Client, ctx context.Context, svcMgr manager.Manager) error {
+	klog.Infof("cleaning up all %s Windows services", windows.ManagedTag)
```
sebsoto, line 107: nit: Moving this logging up to the Deconfigure function is my preference.
Deconfigure is making the choice for this to happen, and that allows for the logging to not repeat in the chance this function needs to be called in multiple places.

pkg/daemon/cleanup/cleanup.go
```diff
@@ -0,0 +1,173 @@
...
+	allServices, err := svcMgr.GetServices()
+	if err != nil {
+		return errors.Wrap(err, "could not determine existing Windows services")
+	}
+	// Build up any failures
+	var failedRemovals error
+	for service := range allServices {
+		// Re-query existing services to avoid processing one that has already been removed
+		currServices, err := svcMgr.GetServices()
```
sebsoto, line 116: its not clear what `currServices` means, `currentServices`? I dont think the brevity is worth the loss of clarity

pkg/daemon/cleanup/cleanup.go
```diff
@@ -0,0 +1,173 @@
...
+		if !strings.HasPrefix(config.Description, windows.ManagedTag) {
+			continue
+		}
+		if err = svcMgr.DeleteServiceAndDependents(service); err != nil {
+			failedRemovals = errors.Errorf("unable to delete %s (%v); %s", service, err, failedRemovals)
+		}
+	}
+	if failedRemovals != nil {
+		// Since this is a best effort cleanup, warn but do not return error on failure
+		klog.Warning(failedRemovals)
```
sebsoto, line 146: Should this really not cause an error?
Its best effort in the sense that we should make the best effort to find any remaining OpenShfit managed services, but we still need to clean up the services.

pkg/daemon/cleanup/cleanup.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	// Try to get the latest services ConfigMap. If it exists, the services in it will be deconfigured.
+	// This will enable clean up instances even if service descriptions were somehow overwritten.
+	cm, err := getLatestServicesCM(c, ctx)
+	if err != nil {
+		klog.Warningf("unable to find latest services ConfigMap: %s", err.Error())
+		return nil
+	}
+	return removeCMServices(c, ctx, svcMgr, cm)
```
sebsoto, line 156: Reversing the order of this cleanup and the rest of the function makes more sense to me.
This is giving us a list of services that we know we installed. We should be handling that before combing through the rest of the services.

pkg/daemon/fake/manager.go
```diff
@@ -123,6 +124,77 @@ func (t *testMgr) DeleteService(name string) error {
...
+			}
+			// Found an existing service that has the root one as a dependency, process it if it's not already stopped
+			status, err := existingWinSvcObj.Query()
+			if err != nil {
+				return errors.Wrapf(err, "error querying %s service status", service)
+			}
+			if status.State == svc.Stopped {
+				continue
+			}
+			klog.Infof("processing service %s as it depends on %s", service, rootService)
```
sebsoto, line 189: Is this logging really necessary?
This is only going to make unit tests noisy.
This function can't be used in a real scenario

We should be getting enough feedback from asserts and error messages

pkg/daemon/fake/manager.go
```diff
@@ -123,6 +124,77 @@ func (t *testMgr) DeleteService(name string) error {
...
+	if err != nil {
+		return errors.Wrapf(err, "unable to query %s service status", name)
+	}
+	// Ensure no dependent services are running before we can safely stop this one
+	if status.State != svc.Stopped {
+		if err := t.removeDependentServices(name, existingSvcs); err != nil {
+			return err
+		}
+	}
+	klog.Infof("removing service %s", name)
```
sebsoto, line 153: Is this logging really necessary?
This is only going to make unit tests noisy.
This function can't be used in a real scenario

We should be getting enough feedback from asserts and error messages

pkg/daemon/manager/manager.go
```diff
@@ -75,6 +78,72 @@ func (m *manager) DeleteService(name string) error {
...
+	if err != nil {
+		return errors.Wrapf(err, "unable to query %s service status", name)
+	}
+	if status.State != svc.Stopped {
+		// Ensure no dependent services are running before we can safely stop this one
+		if err := m.removeDependentServices(name, existingSvcs); err != nil {
+			return err
+		}
+	}
+	klog.Infof("removing service %s", name)
```
sebsoto, line 107: Instead of logging here, I recommend returning a []string of services that have been removed. This will solve the issue presented [here](https://github.com/openshift/windows-machine-config-operator/pull/1154/commits/a5052500d56f0cd3ae07a9295bb77747291926a2#r947221980)

sebsoto: This slice can be appended to as you recurse

pkg/daemon/manager/manager.go
```diff
@@ -75,6 +78,72 @@ func (m *manager) DeleteService(name string) error {
...
+			// Found an existing service that has this one as a dependency, process it if it's not already stopped
+			status, err := existingWinSvcObj.Query()
+			if err != nil {
+				return errors.Wrapf(err, "error querying %s service status", existingSvc)
+			}
+			if status.State == svc.Stopped {
+				continue
+			}
+			klog.Infof("processing service %s as it depends on %s", existingSvc, service)
+			if err := m.DeleteServiceAndDependents(existingSvc); err != nil {
```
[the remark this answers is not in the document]
sebsoto: +1

pkg/daemon/manager/manager.go
sebsoto: Logging a concise list of the removed services, instead of having individual log lines for each service is going to help the WMCO error logs to not be a mess.

pkg/daemon/cleanup/cleanup.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+// getLatestServicesCM returns the most recently created services ConfigMap in the cluster or an error if none exists.
+func getLatestServicesCM(c client.Client, ctx context.Context) (*core.ConfigMap, error) {
```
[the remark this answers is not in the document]
sebsoto: +1

pkg/daemon/cleanup/cleanup.go
```diff
@@ -0,0 +1,173 @@
...
+	var failedRemovals error
+	for service := range allServices {
+		// Re-query existing services to avoid processing one that has already been removed
+		currServices, err := svcMgr.GetServices()
+		if err != nil {
+			return err
+		}
+		if _, ok := currServices[service]; !ok {
+			continue
+		}
```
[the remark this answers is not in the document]
sebsoto: +1, see my comment on DeleteServiceAndDependents for more feedback on this

sebsoto (changes requested):

pkg/daemon/cleanup/cleanup.go
```diff
@@ -0,0 +1,173 @@
...
+		// Happy path, no issues getting the desired services ConfigMap
+		if err = removeCMServices(c, ctx, svcMgr, cm); err != nil {
+			return err
+		}
+	} else {
+		klog.Warningf(err.Error())
+		if err = outOfSyncCleanup(c, ctx, svcMgr); err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
saifshaikh48: > Should we delete it even if we haven't cleaned up the services?
I lean towards no, do not delete the node if there was an error. @sebsoto could you also weigh in?

> Should we cordon the node before we start the deconfiguration process?
IMO this should happen from the WMCO side before WICD cleanup is invoked.

pkg/daemon/manager/manager.go
saifshaikh48: Not that I could tell from my reading

pkg/daemon/cleanup/cleanup.go
saifshaikh48: I disagree on this, outOfSyncCleanup is the one filtering for the managed tag. Besides, cleaning services witht the managed tag is only half of this functions responsibility

pkg/daemon/cleanup/cleanup_test.go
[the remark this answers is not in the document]
saifshaikh48: I'm not sure how we would simulate that

pkg/daemon/cleanup/cleanup.go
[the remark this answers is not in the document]
saifshaikh48: > not moving on to deleting on error when opening or stopping
> Stopping at the first error should not be the option.

Agreed on both

pkg/daemon/cleanup/cleanup.go
```diff
@@ -0,0 +1,109 @@
...
+		return err
+	}
+	existingSvcs, err := svcMgr.GetExistingServices()
+	if err != nil {
+		return err
+	}
+
+	klog.Infof("removing services defined in ConfigMap %s", cm.Name)
+	// Build up any failures
+	failedRemovals := []string{}
```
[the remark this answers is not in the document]
saifshaikh48: How would we return/display them in that case? I'm struggling with balancing verbosity and accuracy of the error messages, but think the current approach is the best of both.

pkg/daemon/cleanup/cleanup.go
```diff
@@ -0,0 +1,173 @@
...
+	"github.com/openshift/windows-machine-config-operator/pkg/nodeconfig"
+	"github.com/openshift/windows-machine-config-operator/pkg/servicescm"
+	"github.com/openshift/windows-machine-config-operator/pkg/windows"
+)
+
+// Deconfigure removes an instance from the cluster. If the parameter flag is true, the node object is not deleted.
+// If we are able to get the services ConfigMap tied to the desired version, all services defined in it are cleaned up.
+// Otherwise, perform cleanup based on a combination of the OpenShift managed tag and the latest services ConfigMap.
+func Deconfigure(c client.Client, addrs []net.Addr, svcMgr manager.Manager, preserveNode bool) error {
+	var err error
```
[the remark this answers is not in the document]
saifshaikh48: scoping issues within the anonymous func prevent this

pkg/daemon/cleanup/cleanup.go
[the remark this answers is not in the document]
aravindhp: I am seeing many examples of dealing with errors. There is [hashicorp/go-multierror](https://pkg.go.dev/github.com/hashicorp/go-multierror). Not saying we should use this but using a list of errors and looping through them at the call site might be easier to read rather than a huge concatenated list.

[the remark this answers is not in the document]
saifshaikh48: Even if we use the multierror library, when WICD errors are printed in the WMCO log, the list of errors will be printed as one giant list. 

Please let me know if the latest version is satisfactory - collect the service names on failed deletion and print them together at the end.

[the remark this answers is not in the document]
aravindhp: Please reply with an example WMCO log output with the error as you have implemented.

cmd/daemon/cleanup.go
```diff
@@ -0,0 +1,71 @@
...
+		klog.Exitf("could not create authenticated client from service account: %s", err.Error())
+	}
+	addrs, err := controller.LocalInterfaceAddresses()
+	if err != nil {
+		klog.Exitf(err.Error())
+	}
+	svcMgr, err := manager.New()
+	if err != nil {
+		klog.Exitf("could not create service manager: %s", err.Error())
+	}
```
aravindhp, line 67: Why is this being done here instead of inside `Deconfigure()`? In fact `svcMgr` is not even needed until you hit `removeCMServices`. IOW, please declare these closer to the scope where they are being used.

pkg/daemon/cleanup/cleanup.go
```diff
@@ -0,0 +1,97 @@
...
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/openshift/windows-machine-config-operator/pkg/daemon/controller"
+	"github.com/openshift/windows-machine-config-operator/pkg/daemon/manager"
+	"github.com/openshift/windows-machine-config-operator/pkg/nodeconfig"
+	"github.com/openshift/windows-machine-config-operator/pkg/servicescm"
+)
+
+// Deconfigure removes an instance from the cluster. If the parameter flag is true, the node object is not deleted.
```
sebsoto, line 36: `If the parameter flag` ->  `If preserveNode`

pkg/daemon/cleanup/cleanup.go
```diff
@@ -0,0 +1,97 @@
...
+	// Delete node if needed
+	if preserveNode || node == nil {
+		return nil
+	}
+	klog.Infof("deleting node %s", node.GetName())
+	return c.Delete(ctx, node)
+}
+
+// removeCMServices removes all the services defined in the given ConfigMap from this instance.
+func removeCMServices(c client.Client, ctx context.Context, svcMgr manager.Manager, cm *core.ConfigMap) error {
```
aravindhp, line 75: Why do you need `c` and `ctx`?

pkg/daemon/cleanup/cleanup.go
```diff
@@ -0,0 +1,97 @@
...
+	cmData, err := servicescm.Parse(cm.Data)
```
aravindhp, line 76: If `cm.Data` is what the function needs, then we should only be passing that in.

pkg/daemon/cleanup/cleanup_test.go
```diff
@@ -0,0 +1,355 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	clientfake "sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	"github.com/openshift/windows-machine-config-operator/pkg/daemon/fake"
+	"github.com/openshift/windows-machine-config-operator/pkg/nodeconfig"
+	"github.com/openshift/windows-machine-config-operator/pkg/servicescm"
+	"github.com/openshift/windows-machine-config-operator/pkg/windows"
+)
+
+func TestExecute(t *testing.T) {
```
sebsoto, line 25: TestDeconfigure

aravindhp (changes requested):
@saifshaikh48 thanks for reducing the scope of this PR. It made it easier to review. Downside is that I found issues that I missed before 😄

pkg/daemon/cleanup/cleanup_test.go
```diff
@@ -0,0 +1,355 @@
...
+			expectedRemovedServices:   []string{},
+			expectedRemainingServices: []string{},
+			preserveNode:              true,
+			servicesCMExists:          true,
+			configMapServices:         []servicescm.Service{},
+		},
+		{
+			name:                      "Defined service doesn't exist on node",
+			existingServices:          map[string]*fake.FakeService{},
+			expectedRemovedServices:   []string{"test1"},
```
sebsoto, line 57: If it doesn't exist how would it be removed?

pkg/daemon/cleanup/cleanup_test.go
```diff
@@ -0,0 +1,355 @@
...
+				"test5-unmanaged": fake.NewFakeService(
+					"test5-unmanaged",
+					mgr.Config{
+						BinaryPathName: "test5-unmanaged",
+						Dependencies:   []string{},
+					},
+					svc.Status{State: svc.Running},
+				),
+			},
+			expectedRemovedServices:   []string{"test1", "test5-umnanaged", "test3", "test4"},
```
sebsoto, line 205: umnanaged -> unmanaged

sebsoto: Not a big deal, but seems like these unmanaged services dont have a point yet

pkg/daemon/cleanup/cleanup_test.go
```diff
@@ -0,0 +1,355 @@
...
+						Description:    windows.ManagedTag + " test1",
+					},
+					svc.Status{State: svc.Running},
+				),
+				"test2": fake.NewFakeService(
+					"test2",
+					mgr.Config{
+						BinaryPathName: "test2 arg1 arg2",
+						Dependencies:   []string{},
+						Description:    windows.ManagedTag + " test2",
```
sebsoto, line 123: BinaryPathName doesn't really serve a purpose here right? It could be any value?
Might be helpful to add a helper function to reduce the bloat of these test cases, and making adding future cases easier
```
func newConfig(name string, managed bool, dependencies []string) mgr.Config {
  description: ""
  if managed {
    description = windows.Managed + " name"
  }
  return mgr.Config {
    BinaryPathName: "name",
    Description: description
    Dependencies: dependencies
  }
}
```

pkg/daemon/cleanup/cleanup_test.go
```diff
@@ -0,0 +1,355 @@
...
+			require.NoError(t, err)
+
+			for _, svc := range test.expectedRemovedServices {
+				_, present := allServices[svc]
+				assert.Falsef(t, present, "service %s not removed as expected", svc)
+			}
+			for _, svc := range test.expectedRemainingServices {
+				_, present := allServices[svc]
+				assert.Truef(t, present, "service %s not present as expected", svc)
+			}
```
sebsoto, line 344: Instead of doing this it makes more sense to check the final state of the services.
Instead of having two slices `expectedRemovedServices` and `expectedRemainingServices`, you just have
`expectedServices` and assert that the service list is equal to it

sebsoto (changes requested):

pkg/daemon/cleanup/cleanup.go
saifshaikh48: good catch, forgot to change after moving the node deletion outside this func

pkg/daemon/controller/controller.go
```diff
@@ -92,37 +93,25 @@ func RunController(ctx context.Context, apiServerURL, saCA, saToken string) erro
...
-	}
 	addrs, err := localInterfaceAddresses()
 	if err != nil {
 		return err
 	}
-	node, err := findNodeByAddress(&nodes, addrs)
+	node, err := GetAssociatedNode(directClient, addrs)
 	if err != nil {
-		return err
+		errors.Wrap(err, "could not node object associated with this instance")
```
mansikulkarni96, line 109: could not get* node

cmd/daemon/cleanup.go
saifshaikh48: This is done to ensure cleanup.Deconfigure is unit testable. Otherwise the client, network addrs, and manager declarations could have been moved into Deconfigure for sure. I don't see a way around this though.

pkg/daemon/cleanup/cleanup_test.go
saifshaikh48: Added a func to return a fake service altogether

pkg/daemon/manager/manager.go
```diff
@@ -55,12 +57,29 @@ func (m *manager) OpenService(name string) (winsvc.Service, error) {
...
 	if err != nil {
 		return err
 	}
-	return winSvc.Delete()
+	// Nothing to do if it already does not exist
+	if _, present := existingSvcs[name]; !present {
+		return nil
+	}
+
+	underlyingMgr := (*mgr.Mgr)(m)
```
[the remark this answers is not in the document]
saifshaikh48: handled in next commit where DeleteService() is edited.

cmd/daemon/cleanup.go
aravindhp: This is pointing at the fact that the code needs to be structured differently. I suggest thinking this through from an oops perspective. Also what are you getting by unit testing `Deconfigure()` vs `removeCMServices()`?

pkg/daemon/manager/manager.go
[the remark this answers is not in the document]
aravindhp: Why? I am seeing this pattern  of introducing code in one commit and changing it in another commit within the same PR. We don't need this sort of code churn that makes `git blame` a bit hard to parse.

pkg/daemon/cleanup/cleanup.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/openshift/windows-machine-config-operator/pkg/nodeconfig"
+	"github.com/openshift/windows-machine-config-operator/pkg/servicescm"
+)
+
+// Deconfigure removes an instance from the cluster. If preserveNode is true, the node object is not deleted.
+// If we are able to get the services ConfigMap tied to the desired version, all services defined in it are cleaned up.
+// TODO: Otherwise, perform cleanup based on a combination of the OpenShift managed tag and latest services ConfigMap.
+func Deconfigure(c client.Client, netInterfaceAddresses []net.Addr, svcMgr manager.Manager, preserveNode bool) error {
+	var node *core.Node
+	ctx := context.TODO()
```
sebsoto, line 41: Lets use a real context instead of TODO 
example: https://github.com/openshift/windows-machine-config-operator/blob/b81eca363e5280180e3e681277c05d40041e3973/cmd/daemon/controller.go#L60

pkg/daemon/cleanup/cleanup.go
```diff
@@ -0,0 +1,99 @@
...
+		service := services[i]
+		if err := svcMgr.DeleteService(service.Name); err != nil {
+			failedRemovals = append(failedRemovals, err)
+		} else {
+			servicesRemoved = append(servicesRemoved, service.Name)
+		}
+	}
+	klog.Infof("removed services: %q", servicesRemoved)
+	if len(failedRemovals) > 0 {
+		return errors.Errorf("%#v", failedRemovals)
```
sebsoto, line 96: What is the output difference with `%v` and `%#v`?

pkg/daemon/cleanup/cleanup_test.go
```diff
@@ -0,0 +1,189 @@
...
+			},
+			expectedServices: map[string]struct{}{},
+			preserveNode:     false,
+			servicesCMExists: true,
+			configMapServices: []servicescm.Service{
+				{Name: "test1", Dependencies: []string{"test2", "test3-unmanaged"}, Priority: 3},
+				{Name: "test2", Dependencies: []string{}, Priority: 2},
+				{Name: "test3-unmanaged", Dependencies: nil, Priority: 0},
+			},
+		},
```
sebsoto, line 122: Whats the functional difference between these two test cases?
I don't understand why both are needed. 
I cant find any logic that deal with dependencies and 'unmanaged' services

cmd/daemon/cleanup.go
sebsoto: Looks like this still needs to be addressed

sebsoto (changes requested):

pkg/daemon/cleanup/cleanup_test.go
saifshaikh48: It is a vestige from the out of sync cleanup logic. I'll remove the managed vs unmanaged concept from the test suite.

pkg/daemon/cleanup/cleanup.go
saifshaikh48: Benefit of comma separating the errors:
`%#v`
```
[]error{failed to delete service "test1": <something Windows OS returns>, failed to stop service "test2": <something Windows OS returns>, failed to open service "test3": <something Windows OS returns>}
```

`%v`
```
 [failed to delete service "test1": <something Windows OS returns> failed to stop service "test2": <something Windows OS returns> failed to open service "test3": <something Windows OS returns>]
```

pkg/daemon/manager/manager.go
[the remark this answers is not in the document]
saifshaikh48: DeleteService is not introduced or modified in this commit. It's only touched once, in the commit after this one. I think perhaps this comment is misplaced.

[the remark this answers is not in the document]
aravindhp: You are touching `underlyingMgr` here which is why I felt it was strange you are fixing it in the next commit.

cmd/daemon/cleanup.go
saifshaikh48: By testing Deconfigure, what we gain is making sure the node object is (not) deleted based on preserveNode. There were more benefits when we were handling the out-of-sync cleanup but since that is not in scope anymore, I'll refactor.
* bring client, network addrs, and manager declarations into `Deconfigure()`
* unit test removeCMServices only

pkg/daemon/manager/manager.go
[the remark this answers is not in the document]
saifshaikh48: Both commits touch different functions, each of which have `underlyingMgr`. This one refactors ListServices (replaces it with GetServices) and the next one refactors DeleteService. I'm not seeing any extra code churn

pkg/daemon/cleanup/cleanup.go
```diff
@@ -0,0 +1,111 @@
...
+		if err != nil {
+			return err
+		}
+		desiredVersion, present := node.Annotations[nodeconfig.DesiredVersionAnnotation]
+		if !present {
+			return errors.Wrapf(err, "node missing desired version annotation")
+		}
+		return directClient.Get(ctx,
+			client.ObjectKey{Namespace: controller.WMCONamespace, Name: servicescm.NamePrefix + desiredVersion}, cm)
+	}()
```
sebsoto, line 64: The logic deciding if a full (with correct configmap) cleanup should occur is no longer being unit tested, it would be nice if it was considering we have the capabilities to.

moving this logic into a new function `func normalCleanupPossible (bool,error)` would
1) make the code more readable
2) Enable us to make the distiniction between an actual error with interacting with the API, versus the required node annotation/config being missing, and handle that appropriately.
3) Unit test this logic

sebsoto (changes requested):

pkg/daemon/manager/manager.go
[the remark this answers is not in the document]
aravindhp: Then its fine. Your comment `handled in next commit where DeleteService() is edited.` threw me off.

pkg/daemon/cleanup/cleanup.go
sebsoto: feel free to change the name, this is just an example

sebsoto: @aravindhp any thoughts on this?
after discussing with @saifshaikh48 this has proven more difficult than i originally thought
specifically having to consider all the cases of what should be an error that causes program failure and what should cause the best effort cleanup

sebsoto: I've suggested at the least doing
```
	addrs, err := controller.LocalInterfaceAddresses()
	if err != nil {
		return err
	}
	var node *core.Node
	var cmData *servicescm.Data
	err := func() error {
		node, err = controller.GetAssociatedNode(directClient, addrs)
		if err != nil {
			return err
		}
		desiredVersion, present := node.Annotations[nodeconfig.DesiredVersionAnnotation]
		if !present {
			return errors.Wrapf(err, "node missing desired version annotation")
		}
		cm := &core.ConfigMap{}
		err = directClient.Get(ctx,
			client.ObjectKey{Namespace: controller.WMCONamespace, Name: servicescm.NamePrefix + desiredVersion}, cm)
		if err != nil {
			return err
		}
		cmData, err = servicescm.Parse(cm.Data)
		if err != nil {
			return err
		}
	}()
```
But I feel like we are missing out on the case that a `Node list`, or `ConfigMap Get` fails due to whatever reason, and WMCO re-reconciling the cleanup would fix things, but the cleanup chooses instead to resort to best effort cleanup.

This may not be important enough to hold up this PR but it might be good to have a TODO?
Or am I overthinking things here?

aravindhp: We have a similar problem if getting one of the services fail. But what I don't understand is why resorting to best effort clean is bad. As long as we don't delete the node object on intermittent failure, we should be fine with WMCO re-reconciling on failures. Or am I missing something?

sebsoto: I'm probably overthinking things then
the recently pushed changes are fine with me

saifshaikh48: Yep, there should be no issue if WMCO re-reconciles on failure. 

I've incorporated the above suggested code.

cmd/daemon/cleanup.go
```diff
@@ -0,0 +1,58 @@
...
+	cleanupCmd.PersistentFlags().BoolVar(&preserveNode, "preserveNode", false,
+		"If set to true, preserves the Node associated with this instance. Defaults to false.")
+}
+
+func runCleanupCmd(cmd *cobra.Command, args []string) {
+	cfg, err := config.FromServiceAccount(apiServerURL, saCA, saToken)
+	if err != nil {
+		klog.Exitf("error using service account to build config: %s", err.Error())
+	}
+	ctx := ctrl.SetupSignalHandler()
```
aravindhp, line 54: This sets up SIGTERM and SIGINT handlers. Shouldn't we be doing this for all commands?

aravindhp (approved):
Nice work, @saifshaikh48

PR #1151, opened 2022-07-21
WINC-823: Test generated community manifests in WMCO e2e

alinaryan:
Adds the 'community' target to our Makefile that calls the generate.sh
script in order to generate a community bundle in the artifact dir to be used in
our CI e2e test for community branch PRs.

Adjusts generate.sh to ensure it aligns with the
aws-e2e-operator platform test in CI.

PR #1155, opened 2022-07-28
WINC-823: Test generated community manifests in WMCO e2e Followup

alinaryan:
Adds a check for the COMMUNITY env var. If present(true), the manifest
location is set to the artifact dir. Otherwise, it is set to the local bundle/ location.

Moves ARTIFACT_DIR check up so that nounset is not triggered. 

Adjusts generate.sh to copy over wmco-package.yaml
to the community bundle as it's needed for e2e testing.

Depends on: https://github.com/openshift/windows-machine-config-operator/pull/1151 and https://github.com/openshift/release/pull/30711

### DIFF
--- a/hack/common.sh
+++ b/hack/common.sh
@@ -1,7 +1,13 @@
 #!/bin/bash
 
+# Default is false unless test is running in COMMUNITY branch
+COMMUNITY=${COMMUNITY:-false}
+
 # Location of the manifests file
 MANIFEST_LOC=bundle/
+if [ "$COMMUNITY" = "true" ]; then
+  MANIFEST_LOC="$ARTIFACT_DIR"
+fi
 
 # define namespace
 declare -r WMCO_DEPLOY_NAMESPACE=openshift-windows-machine-config-operator
@@ -52,7 +58,7 @@ OSDK_WMCO_management() {
 
   if [[ "$COMMAND" = "run" ]]; then
     local version=$(get_packagemanifests_version)
-    $OSDK_PATH run packagemanifests bundle \
+    $OSDK_PATH run packagemanifests $MANIFEST_LOC \
       --namespace $WMCO_DEPLOY_NAMESPACE \
       --install-mode=OwnNamespace \
       --version $version \
--- a/hack/community/generate.sh
+++ b/hack/community/generate.sh
@@ -61,7 +61,7 @@ generate_manifests() {
   local OUTPUT_DIR=$3
 
   echo "Update operator manifests"
-  cp -r "${BUNDLE_DIR}/manifests" "${BUNDLE_DIR}/metadata" "${OUTPUT_DIR}"
+  cp -r "${BUNDLE_DIR}/manifests" "${BUNDLE_DIR}/metadata" "${BUNDLE_DIR}/windows-machine-config-operator.package.yaml" "${OUTPUT_DIR}"
   local CO_CSV="${OUTPUT_DIR}/manifests/windows-machine-config-operator.clusterserviceversion.yaml"
   local CO_ANNOTATIONS="${OUTPUT_DIR}/metadata/annotations.yaml"
 
--- a/hack/run-ci-e2e-test.sh
+++ b/hack/run-ci-e2e-test.sh
@@ -3,6 +3,14 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+# If ARTIFACT_DIR is not set, create a temp directory for artifacts
+ARTIFACT_DIR=${ARTIFACT_DIR:-}
+if [ -z "$ARTIFACT_DIR" ]; then
+  ARTIFACT_DIR=`mktemp -d`
+  echo "ARTIFACT_DIR is not set. Artifacts will be stored in: $ARTIFACT_DIR"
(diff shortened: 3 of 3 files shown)


hack/common.sh
```diff
@@ -1,7 +1,11 @@
 #!/bin/bash
 
 # Location of the manifests file
-MANIFEST_LOC=bundle/
+if [ "$COMMUNITY" = "true" ]; then
```
sebsoto, line 4: CI is failing because this is triggering `set -o nounset`
Try doing `if [ -n "$COMMUNITY"  && "$COMMUNITY" = "true" ]; then`

alinaryan: Thanks, I was just about to ask about this

hack/common.sh
```diff
@@ -67,7 +71,7 @@ OSDK_WMCO_management() {
 
 build_WMCO() {
   local OSDK=$1
-  
+
```
sebsoto, line 74: unrelated change

sebsoto (changes requested):

hack/common.sh
```diff
@@ -1,7 +1,11 @@
 #!/bin/bash
 
-# Location of the manifests file
-MANIFEST_LOC=bundle/
+# Set location of the manifests file
+if [ -z ${COMMUNITY:-} ]; then
+  MANIFEST_LOC=bundle/
+else
+  MANIFEST_LOC="$ARTIFACT_DIR"
+fi
```
jrvaldes, line 8: I would keep a default value and update as necessary. That is:
```shell
# set default location of the manifests file
MANIFEST_LOC="$ARTIFACT_DIR"
# update for commutiy 
if [ -z ${COMMUNITY:-} ]; then
  MANIFEST_LOC=bundle/ 
fi
```

jrvaldes (commented):
LGTM, just minor comments.

hack/common.sh
alinaryan: @jrvaldes Thanks, but I might keep what I have. ARTIFACT_DIR is not the default location. That if statement checks if COMMUNITY is unset and keeps the default bundle/ location without triggering `set -o nounset`

saifshaikh48, line 8: +1 to Jose's code structure suggestion

note that `-z` checks if a string is null (that is, has zero length) whereas `-n` checks that a string is _not_ null

```suggestion
# set default location of the manifests file
MANIFEST_LOC="$ARTIFACT_DIR"
# update for community 
if [ -n ${COMMUNITY:-} && "$COMMUNITY" = "true" ]; then
  MANIFEST_LOC=bundle/ 
fi
```

sebsoto: How is `-n ${COMMUNITY:-}` ever going to evaluate to false?
`${COMMUNITY:-}` by definition cannot be unset by

saifshaikh48: @sebsoto `-n ${COMMUNITY:-}` would evaluate to false if `COMMUNITY=""` (empty string) or the variable is not set (tested this by running `unset COMMUNITY` before the conditional). Since we default to `COMMUNITY="false"` in CI config it won't ever be false but it the check doesn't hurt imo.

sebsoto: Got it. Needs quotes around `${COMMUNITY:-}` though.
Instead of putting it into the conditional,  i suggest doing `COMMUNITY=${COMMUNITY:-false}

sebsoto: so overall the check looks like this: 
```
MANIFEST_LOC=bundle/
COMMUNITY=${COMMUNITY:-false}
if [ "$COMMUNITY" = "true" ]; then
  MANIFEST_LOC="$ARTIFACT_DIR"
fi
```

hack/common.sh
```diff
@@ -2,6 +2,10 @@
 
 # Location of the manifests file
 MANIFEST_LOC=bundle/
+COMMUNITY=${COMMUNITY:-false}
```
jrvaldes, line 5: I suggest moving this to the top of the file and adding a comment that explains why the default value is set to `false`. Then, handle the MANIFEST_LOC logic.

jrvaldes (approved):
/lgtm

jrvaldes:
@alinaryan   any backports?

jrvaldes:
> @alinaryan: #1155 failed to apply on top of branch "release-4.11":
> 
> ```
> Applying: check for COMMUNITY env var
> Using index info to reconstruct a base tree...
> M	hack/community/generate.sh
> M	hack/run-ci-e2e-test.sh
> Falling back to patching base and 3-way merge...
> Auto-merging hack/run-ci-e2e-test.sh
> Auto-merging hack/community/generate.sh
> CONFLICT (content): Merge conflict in hack/community/generate.sh
> error: Failed to merge in the changes.
> hint: Use 'git am --show-current-patch=diff' to see the failed patch
> Patch failed at 0001 check for COMMUNITY env var
> When you have resolved this problem, run "git am --continue".
> If you prefer to skip this patch, run "git am --skip" instead.
> To restore the original branch and stop patching, run "git am --abort".
> ```

Depends on:
- https://github.com/openshift/windows-machine-config-operator/pull/1239

PR #1160, opened 2022-08-05
WINC-740: Configure kube-proxy with WICD

sebsoto:
* Adds functionality for WICD to run PowerShell scripts to resolve service variables 
* Moves network configuration into a PowerShell script, a requirement for it to be done by WICD
* Moves kube-proxy configuration into WICD


pkg/nodeconfig/payload/payload.go
```diff
@@ -82,3 +84,39 @@ func NewFileInfo(path string) (*FileInfo, error) {
 		SHA256: fmt.Sprintf("%x", sha256.Sum256(contents)),
 	}, nil
 }
+
+// Validate ensures all required payload files exist
+func Validate() error {
```
jrvaldes, line 92: Do you anticipate more validations to be included here? Otherwise, I don't feel the `Validate()` function is needed just to hold the collection of files. You can call directly `checkIfFilesExist`.

cmd/operator/main_test.go
```diff
@@ -1,25 +1 @@
 package main
```
jrvaldes, line 1: Do we need to keep the empty file?

jrvaldes (commented):
Mostly, LGTM. Just minor comments.

Great work @sebsoto

pkg/daemon/controller/controller.go
```diff
@@ -65,6 +66,7 @@ type ServiceController struct {
 	ctx            context.Context
 	nodeName       string
 	watchNamespace string
+	psCmdRunner    powershell.CommandRunner
```
aravindhp, line 69: Why do you need `psCmdRunner` to be part of the struct?

pkg/daemon/controller/controller.go
```diff
@@ -128,8 +130,12 @@ func RunController(ctx context.Context, apiServerURL, saCA, saToken string) erro
 }
 
 // NewServiceController returns a pointer to a ServiceController object
-func NewServiceController(ctx context.Context, client client.Client, mgr manager.Manager, nodeName string) *ServiceController {
-	return &ServiceController{client: client, Manager: mgr, ctx: ctx, nodeName: nodeName,
+func NewServiceController(ctx context.Context, client client.Client, mgr manager.Manager,
+	cmdRunner powershell.CommandRunner, nodeName string) *ServiceController {
+	if cmdRunner == nil {
+		cmdRunner = powershell.NewCommandRunner()
+	}
```
aravindhp, line 137: This is very puzzling. I don't see why `cmdRunner` needs to be passed in.

pkg/daemon/controller/controller.go
```diff
@@ -344,11 +350,18 @@ func (sc *ServiceController) resolveNodeVariables(svc servicescm.Service) (map[s
...
 }
 
-// resolvePowershellVariables returns a map, with the keys being each variable, and the value being the string to replace the
-// variable with
+// resolvePowershellVariables returns a map, with the keys being each variable, and the value being the string to
+// replace the variable with
 func (sc *ServiceController) resolvePowershellVariables(svc servicescm.Service) (map[string]string, error) {
-	// TODO: Implement this function
-	return make(map[string]string), nil
+	vars := make(map[string]string)
```
aravindhp, line 356: Can't you just instantiate a runner here?

build/Dockerfile.ci
```diff
@@ -151,16 +151,20 @@ WORKDIR /payload/kube-node/
...
+# Copy CNI plugin binaries
 WORKDIR /payload/cni/
 COPY --from=build /build/windows-machine-config-operator/containernetworking-plugins/bin/host-local.exe .
 COPY --from=build /build/windows-machine-config-operator/containernetworking-plugins/bin/win-bridge.exe .
 COPY --from=build /build/windows-machine-config-operator/containernetworking-plugins/bin/win-overlay.exe .
-COPY --from=build /build/windows-machine-config-operator/pkg/internal/cni-conf-template.json .
+
+# Create directory for generated files with open permissions
+RUN mkdir /payload/generated
+RUN chmod 0777 /payload/generated
```
aravindhp, line 162: Why is this needed?

pkg/windows/windows.go
```diff
@@ -547,18 +535,18 @@ func (vm *windows) ConfigureAzureCloudNodeManager(nodeName string) error {
 }
 
 func (vm *windows) ConfigureKubeProxy(nodeName, hostSubnet string) error {
-	endpointIP, err := vm.createHNSEndpoint()
+	endpointIP, err := vm.Run(NetworkConfScriptPath, true)
```
aravindhp: Why are you combining CNI configuration and kube proxy configuration?

pkg/internal/network-conf-template.ps1
```diff
@@ -0,0 +1,81 @@
+# This script ensures the contents of the CNI config file is correct, and returns the HNS endpoint IP
```
aravindhp, line 1: Shouldn't there be unit test for testing the generation from the teamplate?

pkg/services/services.go
```diff
@@ -56,3 +59,32 @@ func hybridOverlayConfiguration(vxlanPort string, debug bool) servicescm.Service
 		Priority:                     1,
 	}
 }
+
+func kubeProxyConfiguration() servicescm.Service {
```
aravindhp, line 63: If kube-proxy has debug levels, then this should follow the same treatment that `hybridOverlayConfiguration()` got i.e. take `debug` as input,

aravindhp (changes requested):

pkg/nodeconfig/payload/payload.go
```diff
@@ -82,3 +86,16 @@ func NewFileInfo(path string) (*FileInfo, error) {
...
+// GenerateNetworkConfScript generates the .ps1 file responsible for CNI configuration generation
+func GenerateNetworkConfScript(clusterCIDR, hnsNetworkName, hnsPSModulePath, cniConfigPath string) error {
+	out, err := ioutil.ReadFile(NetworkConfTemplate)
+	if err != nil {
+		return err
+	}
+	cniConfScript := strings.ReplaceAll(string(out), "HNS_NETWORK", hnsNetworkName)
+	cniConfScript = strings.ReplaceAll(cniConfScript, "SERVICE_NETWORK_CIDR", clusterCIDR)
+	cniConfScript = strings.ReplaceAll(cniConfScript, "HNS_MODULE_PATH", hnsPSModulePath)
+	cniConfScript = strings.ReplaceAll(cniConfScript, "CNI_CONFIG_PATH", cniConfigPath)
```
saifshaikh48, line 99: consider the following
```
    cniConfScript := string(out)
    replacements := map[string]string{"HNS_NETWORK" : hnsNetworkName, "SERVICE_NETWORK_CIDR" : clusterCIDR, "HNS_MODULE_PATH" : hnsPSModulePath, "CNI_CONFIG_PATH" : cniConfigPath}
    for key, val := range replacements {
        cniConfScript = strings.ReplaceAll(cniConfScript, key, val)
    }
```

pkg/windows/windows.go
```diff
@@ -534,28 +534,11 @@ func (vm *windows) ConfigureAzureCloudNodeManager(nodeName string) error {
...
-		[]string{HybridOverlayServiceName})
-	if err != nil {
-		return errors.Wrapf(err, "error creating %s service object", kubeProxyServiceName)
-	}
-
-	if err := vm.ensureServiceIsRunning(kubeProxyService); err != nil {
-		return errors.Wrapf(err, "error ensuring %s Windows service has started running", kubeProxyServiceName)
+func (vm *windows) ConfigureKubeProxy() error {
+	vm.log.V(1).Info("waiting for kube-proxy to start running")
+	if err := vm.waitForServiceToRun(KubeProxyServiceName); err != nil {
```
saifshaikh48, line 539: Why is this wait needed? Were we doing this in the past?

pkg/windows/windows.go
```diff
@@ -534,28 +534,11 @@ func (vm *windows) ConfigureAzureCloudNodeManager(nodeName string) error {
...
-		"--network-name=" + OVNKubeOverlayNetwork + " --source-vip=" + strings.TrimSpace(endpointIP) +
-		" --enable-dsr=false"
-
-	kubeProxyService, err := newService(KubeProxyPath, kubeProxyServiceName, kubeProxyServiceArgs,
-		[]string{HybridOverlayServiceName})
-	if err != nil {
-		return errors.Wrapf(err, "error creating %s service object", kubeProxyServiceName)
-	}
-
-	if err := vm.ensureServiceIsRunning(kubeProxyService); err != nil {
```
saifshaikh48, line 567: Will the kube-proxy service still have the "OpenShift managed" description if we don't call  `ensureServiceIsRunning` anymore?

sebsoto: Yes, [WICD does this as it is starting the service](https://github.com/openshift/windows-machine-config-operator/blob/0197a69eb52621ad57adea805cc63abaaf94e9d5/pkg/daemon/controller/controller.go#L262)

pkg/windows/windows.go
sebsoto: That's explained in the commit message

pkg/windows/windows.go
sebsoto: @aravindhp This is happening because I wanted to limit the amount of churn that would occur between commits 2 and 3. There is no real "kube proxy configuration" that takes place, besides starting the kube-proxy service. The powershell commands that are run are still setting up the CNI by creating a usable endpoint for the HNS network.
Should I squash those commits together?

pkg/windows/windows.go
aravindhp: I think you should call it out in a comment here too. Is there a better way to figure this out? Maybe look at the node object?

pkg/daemon/controller/controller.go
sebsoto: The main reason is that  `CommandRunner` provides a method of interacting with an external program. In order to enable useful unit tests, this interface must be made part of the definition of the `ServiceController`. This makes sense to me in the same way that a logger or any other interface which interacts with systems outside of would be part of the struct.

pkg/daemon/controller/controller.go
sebsoto: As an alternative I could introduce a `ControllerOptions` struct to be passed in, this could clean things up a bit.

sebsoto: `ControllerOptions` would reduce the amount of arguments to the NewController function allow the caller to specify an alternate CommandRunner, and potentially other inputs such as the manager used.

pkg/daemon/controller/controller.go
aravindhp: I don't buy this argument. Passing in `nil` values is just not a good pattern. Saying this is needed for unit is not good enough. It might be worth re-examining the object design here.

pkg/daemon/controller/controller.go
aravindhp: I don't know. I would have to see the implementation as I am not sure what you mean by `ControllerOptions`.

pkg/windows/windows.go
sebsoto: Commented. I don't know of a better way. This will be removed ASAP, added a TODO with a ticket.

pkg/internal/network-conf-template.ps1
sebsoto: Added. Not sure of the value added when compared to the extra file required to do this, but I am fine with keeping it. PTAL and let me know what you think.

build/Dockerfile.ci
sebsoto: Commented

aravindhp (changes requested):

saifshaikh48 (commented):
LGTM aside from the network-config script testing. I am fine with the CNI configuration and kube-proxy starting being separated at a later time.

cmd/daemon/bootstrap.go
```diff
@@ -50,23 +49,15 @@ func init() {
...
 func runBootstrapCmd(cmd *cobra.Command, args []string) {
+	// This command will not run in a pod, authenticate using the provided Service Account creds instead
 	cfg, err := config.FromServiceAccount(apiServerURL, saCA, saToken)
 	if err != nil {
 		klog.Exitf("error using service account to build config: %s", err.Error())
 	}
-
-	// Cannot use a cached client as no manager will be started to populate cache
-	directClient, err := controller.NewDirectClient(cfg)
+	sc, err := controller.NewServiceController(context.TODO(), "", controller.Options{Config: cfg})
```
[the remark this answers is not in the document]
aravindhp: That's a good point.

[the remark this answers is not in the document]
sebsoto: @saifshaikh48 I'm not sure what you mean. I'm not introducing that behavior here. `Bootstrap` is a method of the service controller struct. 
see code in current master:
https://github.com/openshift/windows-machine-config-operator/blob/82a718c41c4e202f304f04c940b496d8e2a34467/cmd/daemon/bootstrap.go#L71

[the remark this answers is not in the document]
saifshaikh48: Sorry, I wasn't clear. My comment is a general comment -- ideally we shouldn't have `Bootstrap` be a method on a controller object. Just something to keep in mind as future work.

saifshaikh48 (commented):
just one comment, LGTM otherwise

alinaryan (commented):
Looking good so far. Just curious, why isn’t WICD able to run PS commands like WMCO does? Also, how are you testing?

pkg/nodeconfig/payload/payload.go
```diff
@@ -46,8 +48,8 @@ const (
 	// WinOverlayCNIPlugin is the path of the win-overlay CNI Plugin binary. The container image should already have
 	// this binary mounted
 	WinOverlayCNIPlugin = payloadDirectory + cniDirectory + "win-overlay.exe"
-	// CNIConfigTemplatePath is the path for CNI config template
-	CNIConfigTemplatePath = payloadDirectory + cniDirectory + "cni-conf-template.json"
+	// NetworkConfigurationScript is the path for generated Network configuration Script
+	NetworkConfigurationScript = payloadDirectory + "/generated/network-conf.ps1"
```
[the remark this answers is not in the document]
sebsoto: Added the generated directory, as the diagram is describing how the container file is building the payload, not WMCO

mansikulkarni96:
LGTM

alinaryan:
LGTM

pkg/nodeconfig/payload/payload.go
```diff
@@ -63,6 +65,89 @@ const (
 	// AzureCloudNodeManagerPath contains the path of the azure cloud node manager binary. The container image should
 	// already have this binary mounted
 	AzureCloudNodeManagerPath = payloadDirectory + AzureCloudNodeManager
+	// networkConfTemplate is the template used to generate the network configuration script
+	networkConfTemplate = `# This script ensures the contents of the CNI config file is correct, and returns the HNS endpoint IP
```
aravindhp, line 69: It is fine to leave this as a file. I had only asked for the in-lining in the test case. Sorry if that was not clear.

pkg/windows/windows.go
aravindhp: Did you create a follow up story to separate this out i.e. introduce allowing WICD to run PowerShell scripts.

aravindhp (approved):
Nice work, @sebsoto

pkg/nodeconfig/payload/payload.go
aravindhp: That's fine with me as long as it is easy to maintain.

cmd/daemon/bootstrap.go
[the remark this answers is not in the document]
sebsoto: Sure. https://issues.redhat.com/browse/WINC-886

pkg/windows/windows.go
sebsoto: Yes, it was added as a TODO in the code. https://issues.redhat.com/browse/WINC-882

PR #1161, opened 2022-08-05
[community-4.10] Remove subscription requirement

alinaryan:
Removes the subscription requirement from the description as it
is not mandatory for community WMCO. 

Corrects doc links.

### DIFF
--- a/bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
@@ -21,126 +21,31 @@ spec:
   apiservicedefinitions: {}
   customresourcedefinitions: {}
   description: |-
+    ## WARNING
+    Community distribution of the Windows Machine Config Operator.
+    This is a preview build, and is not meant for production workloads.
+    Issues with this distribution of WMCO can be opened against the [WMCO repository](https://github.com/openshift/windows-machine-config-operator).
+    Please read through the [troubleshooting doc](https://github.com/openshift/windows-machine-config-operator/blob/community-4.10/docs/TROUBLESHOOTING.md)
+    before opening an issue.
+    Please ensure that when installing this operator the starting CSV you subscribe to is supported on the
+    version of OKD/OCP you are using. This CSV is meant for OKD/OCP 4.10.
+
     ### Introduction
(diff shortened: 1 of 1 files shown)


bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
```diff
@@ -21,126 +21,31 @@ spec:
...
-    - [AWS](https://github.com/openshift/windows-machine-config-operator/blob/master/docs/machineset-aws.md)
-    - [Azure](https://github.com/openshift/windows-machine-config-operator/blob/master/docs/machineset-azure.md)
+    Please see the usage section of [README.md](https://github.com/openshift/windows-machine-config-operator/blob/community-4.10/README.md#usage).
 
     ### Limitations
     #### DeploymentConfigs
     Windows Nodes do not support workloads created via DeploymentConfigs. Please use a normal Deployment, or other method to
     deploy workloads.
-
-    ### Reporting issues
```
saifshaikh48, line 140: Should we point people to our github instead?

saifshaikh48 (commented):
lgtm aside from one question

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
alinaryan: It's mentioned on [line 27](https://github.com/openshift/windows-machine-config-operator/pull/1161/commits/e4b4c08a71ae3092e7ea4468444a5348db8cfdb8#diff-53f141ca70f6ce0a959fa80b57ab2ac4e1e8f29689213aebce7adf958f5c71caR27)

sebsoto:
I'm confused as to why this change needs to be made. I thought the community CSV description lives in the [description.md](https://github.com/openshift/windows-machine-config-operator/blob/master/hack/community/csv/description.md) file you added?

Making the changes present in this PR, specifically to the community branch, means we'll have to do this every time a community branch is created. Thats one of the problems thats solved by the generate.sh script and the above linked file.

alinaryan:
@sebsoto It does currently, but while I was editing generate.sh I was wondering if it's ok that our community branches have separate csv descriptions than the community release ones. I see your point in it being a problem to update each time a community branch is created so I'm ok with closing it if you don't think it's necessary

PR #1166, opened 2022-08-05
[wicd] Restructure Windows service interaction

saifshaikh48:
This PR makes the Windows service manager interface own service deletion,
instead of deleting a service through the service object itself. As the manager
is in charge of service creation, it should also be in charge of deletion.
    
It also breaks the winsvc pkg into separate components for Windows
services (winsvc) and the overlying service manager interface (manager),
as well as a separate fake package to test both.


pkg/daemon/fake/fake_manager_test.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+func TestDeleteService(t *testing.T) {
+	testIO := []struct {
+		name         string
+		svcName      string
+		existingSvcs map[string]*FakeService
+	}{
+		{
+			name:         "service exists",
```
aravindhp, line 149: What about deleting an non-existent service?

pkg/daemon/manager/manager.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"golang.org/x/sys/windows/svc/mgr"
+
+	"github.com/openshift/windows-machine-config-operator/pkg/daemon/winsvc"
+)
+
+type Mgr interface {
+	CreateService(string, string, mgr.Config, ...string) (winsvc.Service, error)
+	ListServices() ([]string, error)
+	OpenService(string) (winsvc.Service, error)
```
aravindhp, line 14: Please add docs.

pkg/daemon/fake/fake_manager.go
```diff
@@ -105,6 +105,15 @@ func (t *testMgr) OpenService(name string) (winsvc.Service, error) {
 	return service, nil
 }
 
+func (t *testMgr) DeleteService(name string) error {
+	_, exists := t.svcList.read(name)
+	if !exists {
+		// This is to mimic the behavior of the Windows OS. Trying to delete a nonexistant service results in an error.
+		return fmt.Errorf("service %s already does not exist", name)
```
aravindhp, line 112: I suggest removing `already`,

aravindhp (commented):
Nice work @saifshaikh48. I am good with this. Had couple of minor comments. I will leave the approval to @sebsoto.

pkg/daemon/fake/fake_manager_test.go
saifshaikh48: Added in 2nd commit

pkg/daemon/fake/fake_manager.go
```diff
@@ -0,0 +1,117 @@
+//go:build windows
```
sebsoto, line 1: To me it seems like the naming scheme for these files should be:
pkg/daemon/fake/fake_manager.go -> pkg/daemon/fake/manager.go

sebsoto: Applies to all fake package files

pkg/daemon/manager/manager.go
```diff
@@ -0,0 +1,48 @@
...
+
+package manager
+
+import (
+	"golang.org/x/sys/windows/svc/mgr"
+
+	"github.com/openshift/windows-machine-config-operator/pkg/daemon/winsvc"
+)
+
+type Mgr interface {
```
sebsoto, line 11: Consider renaming this to `Manager`, that fits within go naming conventions a little better

pkg/daemon/manager/manager.go
```diff
@@ -0,0 +1,48 @@
...
+func (m *manager) ListServices() ([]string, error) {
+	underlyingMgr := (*mgr.Mgr)(m)
+	return underlyingMgr.ListServices()
+}
+func (m *manager) OpenService(name string) (winsvc.Service, error) {
+	underlyingMgr := (*mgr.Mgr)(m)
+	return underlyingMgr.OpenService(name)
+}
+
+func NewMgr() (Mgr, error) {
```
sebsoto, line 51: This can be renamed to `New()` if the above `Manager` change is made

pkg/daemon/manager/manager.go
```diff
@@ -37,6 +39,14 @@ func (m *manager) OpenService(name string) (winsvc.Service, error) {
...
 	return underlyingMgr.OpenService(name)
 }
+func (m *manager) DeleteService(name string) error {
+	underlyingMgr := (*mgr.Mgr)(m)
+	winSvc, err := underlyingMgr.OpenService(name)
+	if err != nil {
+		return err
+	}
+	return winSvc.Delete()
+}
```
sebsoto, line 49: Please write a test case for manually testing this to ensure:
1) This doesn't error out if another service handle is floating around due to `DeleteService(service)` being called within the same function as `OpenService(service)` 
2) The existence of other open handles for that service does not cause the service to be stuck in waiting for deletion.

sebsoto (changes requested):
Thanks @saifshaikh48 mostly good, but PTAL at my comments

pkg/daemon/manager/manager.go
saifshaikh48: Sure, I added this within the `manager.DeleteService()` [unit tests](https://github.com/openshift/windows-machine-config-operator/pull/1166/commits/8dc05cb9404101921b14e57ba677b21408db22e5#diff-770772d9b994c813411fdf9f00b851ef0e3a7b13e05398d09fce6a9d120a95b4R165-R171)

sebsoto: I meant this needs to be tested with the Windows service manager, not with our fake manager. I don't know the behavior of the real manager in this scenario.

saifshaikh48: I can confirm that there's no issue opening a service multiple times in a tow when using the real manager. Ran on Windows machine without error and printed expected msg:
```golang
func runTest() {
	svcMgr, err := manager.New()
	if err != nil {
	}

	_, err = svcMgr.CreateService("testsvc", "testsvc", mgr.Config{
		Description: "testsvc",
	})
	if err != nil {
		klog.Exitf("could not create: %s", err.Error())
	}
	_, err = svcMgr.OpenService("testsvc")
	if err != nil {
		klog.Exitf("could not open: %s", err.Error())
	}
	klog.Info("open1")
        _, err = svcMgr.OpenService("testsvc")
	if err != nil {
		klog.Exitf("could not open: %s", err.Error())
	}
	klog.Info("open2")

	err = svcMgr.DeleteService("testsvc") // <--------- has additional underlying OpenService call
	if err != nil {
		klog.Exitf("could not delete: %s", err.Error())
	}
	klog.Info("end")
}
```
Additionally the `testsvc` doesn't appear in the output of `Get-Service` so successfully deleted

sebsoto: Thanks for doing that

mansikulkarni96:
LGTM

PR #1174, opened 2022-08-11
[docs] Limitation to access secure registry images

mansikulkarni96:


mansikulkarni96:
/ cc: @mburke5678

README.md
```diff
@@ -228,6 +228,9 @@ deploy workloads.
 ### Cluster-wide proxy 
 WMCO does not support adding Windows workloads using a [cluster-wide proxy](https://docs.openshift.com/container-platform/latest/networking/enable-cluster-wide-proxy.html) config for the OpenShift Container Platform. WMCO will not be able to automatically route proxy connections for Windows workloads.
 
+### Accessing secure registries
+WMCO does not support pulling images from secure private registries on the Windows nodes. It is recommended to use images from public registries.
```
sebsoto, line 232: Alternatively the images can pre-pulled onto the VM image that is used

sebsoto: Or pulled before adding a BYOH instance to the cluster.

aravindhp: I would leave out the automation bit. That is up to the user.

README.md
```diff
@@ -228,6 +228,9 @@ deploy workloads.
 ### Cluster-wide proxy 
 WMCO does not support adding Windows workloads using a [cluster-wide proxy](https://docs.openshift.com/container-platform/latest/networking/enable-cluster-wide-proxy.html) config for the OpenShift Container Platform. WMCO will not be able to automatically route proxy connections for Windows workloads.
 
+### Accessing secure registries
+WMCO does not support pulling images from secure private registries on the Windows nodes. It is recommended to use images from public registries or pre-pull the images in the VM golden image.
```
saifshaikh48, line 232: I'd drop the word golden since this is the general readme, not vsphere specific

saifshaikh48: also nit: line length

jrvaldes: +1

README.md
```diff
@@ -228,6 +228,10 @@ deploy workloads.
 ### Cluster-wide proxy 
 WMCO does not support adding Windows workloads using a [cluster-wide proxy](https://docs.openshift.com/container-platform/latest/networking/enable-cluster-wide-proxy.html) config for the OpenShift Container Platform. WMCO will not be able to automatically route proxy connections for Windows workloads.
 
+### Accessing secure registries
+WMCO does not support pulling images from secure private registries on the Windows nodes.
```
sebsoto, line 232: This sentence could be restructured like so:
`Windows nodes managed by WMCO do not support pulling container images from secure private registries.`

I'd do this as the subject is the `Windows nodes managed by WMCO`, as WMCO itself is not doing the pulling.
I also replaced `images` with `container images`, to be explicit in differentiating it from the VM image referred to in the next sentence.

sebsoto (changes requested):
Sorry, one more comment :sweat_smile:

README.md
mansikulkarni96: Makes sense, thanks for the suggested wording. Made the changes.

README.md
mburke5678: @mansikulkarni96 This is related to containerd, for 6.0.0+ only, correct?

mansikulkarni96: @mburke5678 I believe this was always the case, even with Docker. It's only with a feature request that came in with containerd that we decided to document it.

PR #1175, opened 2022-08-11
[wicd] Make manager in charge of closing services

saifshaikh48:
This PR makes the Windows service manager interface own the closing of
Windows services, instead of the service itself. As the manager
is in charge of opening services, it should also be in charge of closing.

Follow-up to https://github.com/openshift/windows-machine-config-operator/pull/1166

### DIFF
--- a/pkg/daemon/manager/manager.go
+++ b/pkg/daemon/manager/manager.go
@@ -15,6 +15,8 @@ type Manager interface {
 	ListServices() ([]string, error)
 	// OpenService gets the Windows service of the given name if it exists, by which it can be queried or controlled
 	OpenService(string) (winsvc.Service, error)
+	// CloseService closes access to the handle of the given service
+	CloseService(string) error
 	// DeleteService marks a Windows service of the given name for deletion, or an error if it does not exist
 	DeleteService(string) error
 }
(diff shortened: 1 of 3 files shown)


pkg/daemon/manager/manager.go
```diff
@@ -39,6 +41,14 @@ func (m *manager) OpenService(name string) (winsvc.Service, error) {
 	underlyingMgr := (*mgr.Mgr)(m)
 	return underlyingMgr.OpenService(name)
 }
+func (m *manager) CloseService(name string) error {
```
aravindhp, line 44: PS: No need to bunch these functions together. Feel free to add new lines.

aravindhp (approved):
Thanks for doing this, @saifshaikh48

pkg/daemon/manager/manager.go
saifshaikh48: A some whitespace :)

pkg/daemon/manager/manager.go
```diff
@@ -31,14 +33,26 @@ func (m *manager) CreateService(name, exepath string, config mgr.Config, args ..
...
 }
+
+func (m *manager) CloseService(name string) error {
+	underlyingMgr := (*mgr.Mgr)(m)
+	winSvc, err := underlyingMgr.OpenService(name)
+	if err != nil {
+		return err
+	}
+	return winSvc.Close()
+}
```
sebsoto, line 54: This implementation doesn't make sense. you're opening a handle which won't be used anywhere only to close the specific handle.

In general, I also disagree with moving this close method to the manager at all.
In go handles are closed via a method on the handle itself.
See examples in the `os` and `io` standard libraries:
* [Pipes](https://pkg.go.dev/io#PipeReader.Close)
* [Files](https://pkg.go.dev/os#File.Close)

aravindhp: Good point @sebsoto. I agree with you. I obviously had forgotten how file and pipe handles closing work.

@saifshaikh48 sorry for not thinking this through and catching it sooner.

aravindhp (commented):
/approve cancel

I suggest we close this PR and leave it as is.

saifshaikh48:
/close

based on above feedback

PR #1189, opened 2022-08-17
[winsvc] Wait for service state changes

saifshaikh48:
This PR adds a wait to EnsureServiceState, ensuring that the Windows
service actually has the desired state when the function returns. This makes
any operations that occur after calling this function safe (for example, 
deleting after changing state to stopped).

### DIFF
--- a/pkg/daemon/winsvc/winsvc.go
+++ b/pkg/daemon/winsvc/winsvc.go
@@ -6,6 +6,9 @@ import (
 	"github.com/pkg/errors"
 	"golang.org/x/sys/windows/svc"
 	"golang.org/x/sys/windows/svc/mgr"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	"github.com/openshift/windows-machine-config-operator/pkg/retry"
 )
 
 type Service interface {
@@ -28,11 +31,27 @@ func EnsureServiceState(service Service, state svc.State) error {
 	}
 	switch state {
 	case svc.Running:
-		return service.Start()
+		err = service.Start()
+		if err != nil {
+			return err
+		}
 	case svc.Stopped:
 		_, err = service.Control(svc.Stop)
-		return err
+		if err != nil {
+			return err
+		}
 	default:
 		return errors.New("unexpected state request")
 	}
+	// Wait for the state change to actually take place
+	return wait.PollImmediate(retry.WindowsAPIInterval, retry.ResourceChangeTimeout, func() (bool, error) {
+		status, err := service.Query()
+		if err != nil {
+			return false, errors.Wrap(err, "error querying service state")
+		}
+		if status.State == state {
+			return true, nil
+		}
+		return false, nil
+	})
 }
--- a/pkg/retry/retry.go
+++ b/pkg/retry/retry.go
@@ -5,6 +5,8 @@ import "time"
 const (
 	// Count is the number of times we will retry an API call
 	Count = 20
+	// WindowsAPIInterval is the wait time between calls to the Windows OS API in case of failure
+	WindowsAPIInterval = 5 * time.Second
 	// Interval is the wait time between API calls on a failure
 	Interval = 15 * time.Second
 	// Timeout is the total time we will wait for an event to occur.


pkg/daemon/winsvc/winsvc.go
```diff
@@ -28,11 +30,27 @@ func EnsureServiceState(service Service, state svc.State) error {
...
 		_, err = service.Control(svc.Stop)
-		return err
+		if err != nil {
+			return err
+		}
 	default:
 		return errors.New("unexpected state request")
 	}
+	// wait for the state change to actually take place
+	return wait.PollImmediate(retry.Interval, retry.ResourceChangeTimeout, func() (bool, error) {
```
aravindhp, line 46: Nit: you should update the doc for `retry.ResourceChangeTimeout`.

aravindhp (approved):

pkg/daemon/winsvc/winsvc.go
saifshaikh48: I think it's still applicable: 

ResourceChangeTimeout is the total time waited for a change (create/**update**/delete) to take place

aravindhp: I guess `Resource` is generic enough.

sebsoto, line 46: These intervals and timeouts set for dealing with kubernetes resources. I'd rather not reuse them here. We shouldn't be waiting more than 30 seconds for this to occur, and we should be rechecking much more frequently.

sebsoto: As it is, 10 minutes is a very long time to wait for a program to stop.

saifshaikh48: Currently in WMCO, we try every 15 seconds for 5 minutes for services to start, and every 15 seconds for 10 minutes each (wow) for stopping and deleting. Before we reduce this to say, trying every 5 seconds for 30 seconds max, do we remember why we currently have such a long timeout?

aravindhp: I don't understand how we are waiting for 10 minutes. Aren't we waiting for 2 minutes and checking every 15 seconds?

pkg/daemon/winsvc/winsvc.go
```diff
@@ -28,11 +30,27 @@ func EnsureServiceState(service Service, state svc.State) error {
...
+		status, err := service.Query()
+		if err != nil {
+			return false, errors.Wrap(err, "error querying service state")
```
aravindhp, line 49: Please add the service name here unless you know that the original err will have that info.

pkg/daemon/winsvc/winsvc.go
aravindhp: My question above is regarding the code under review.

sebsoto: Ah you're right, theres `Timeout` and `ResourceChangeTimeout`.
It will be a 2 min wait. Thats definitely not as bad, but still a bit long for a program to respond to a stop request.

>Before we reduce this to say, trying every 5 seconds for 30 seconds max, do we remember why we currently have such a long timeout?

I'm not suggesting changing any WMCO timeouts, the existing WMCO timeouts don't have much relevance here. If we want to do that we can and probably should, but outside this PR. I'm strictly talking about the code being introduced.
 
As for the interval, 5 seconds sounds good to me. The time spent waiting adds up when we need to do it for multiple services.

aravindhp: I think 2 minutes will be an outlier and it might cut down on us prematurely timing out. WDYT?

aravindhp: Or maybe wait at least a minute? We can introduce new constants specifically for services if needed.

saifshaikh48: ^ @openshift/openshift-team-windows-containers

pkg/daemon/winsvc/winsvc.go
```diff
@@ -3,9 +3,11 @@
 package winsvc
 
 import (
+	"github.com/openshift/windows-machine-config-operator/pkg/retry"
 	"github.com/pkg/errors"
 	"golang.org/x/sys/windows/svc"
 	"golang.org/x/sys/windows/svc/mgr"
+	"k8s.io/apimachinery/pkg/util/wait"
```
aravindhp, line 10: Don't we bunch WMCO packages vs k8s vs internal using a line separator?

aravindhp (changes requested):

pkg/daemon/winsvc/winsvc.go
mansikulkarni96: I think 2 mins should be enough wait, we can increase the timeout if we see instances of the test prematurely timing out.

pkg/daemon/winsvc/winsvc.go
saifshaikh48: Getting the name would require another API call to service.Config(). Since this is a function on the service obj itself rather than from the manager, I think it's fine to leave the service name appending to call site error handling.

pkg/daemon/winsvc/winsvc.go
saifshaikh48: Thanks folks, I'll update with retrying every 5 seconds until a 2 min total timeout

saifshaikh48:
@aravindhp PTAL when you get a chance. Awaiting your approval

aravindhp (approved):

PR #1190, opened 2022-08-17
Update issue templates

sebsoto:
* Create an issue form for RFEs
* Update the bug template to use an issue form


.github/ISSUE_TEMPLATE/bug.yaml
```diff
@@ -0,0 +1,95 @@
...
+  - type: dropdown
+    id: platform
+    attributes:
+      label: Platform
+      description: What platform is your cluster running on?
+      options:
+        - AWS
+        - Azure
+        - Bare metal
+        - GCP
```
jrvaldes, line 49: GCP?

.github/ISSUE_TEMPLATE/bug.yaml
```diff
@@ -0,0 +1,95 @@
...
+    validations:
+      required: false
+  - type: markdown
+    attributes:
+      value: |
+        Please attach any logs and relevant screenshots in a comment.
+        At minimum, please include the output of the following
+        * oc get network.operator cluster -o yaml
+        * oc logs deployment/windows-machine-config-operator -n openshift-windows-machine-config-operator
+        * If relevant: oc get machineset <windows_machineSet_name> -n openshift-machine-api -o yaml
```
aravindhp, line 94: Shouldn't we also ask for the 
- windows-instances ConfigMap?
- Cluster config that will tell us if proxy is being used?

.github/ISSUE_TEMPLATE/bug.yaml
```diff
@@ -0,0 +1,95 @@
...
+        - OKD
+        - OCP
+    validations:
+      required: true
+  - type: input
+    id: cluster-version
+    attributes:
+      label: Cluster Version
+      description: What is the version of your cluster?
+      placeholder: 4.12
```
aravindhp, line 37: 4.11 will be better given 4.12 will be released towards the end of the year.

.github/ISSUE_TEMPLATE/bug.yaml
```diff
@@ -0,0 +1,95 @@
...
+        - Bare metal
+        - GCP
+        - Platform agnostic (type=none)
+        - vSphere
+    validations:
+      required: true
+  - type: input
+    id: wmco-version
+    attributes:
+      label: WMCO Version
```
aravindhp, line 57: Please add Windows version with the list that we support

aravindhp (changes requested):
Thanks for working on this, @sebsoto. Just a few comments.

aravindhp:
We should add `.github` to the list of directories for which tests should not run.

.github/ISSUE_TEMPLATE/bug.yaml
sebsoto: Yep, added a line for the CM, and a required field for proxy information

.github/ISSUE_TEMPLATE/bug.yaml
sebsoto: Thats true, I'd prefer to leave the option so that if someone is extremely intent on filing a bug for an unsupported platform they still can. Its possible they will select any other platform as the field is required and that will make our lives harder.
I've also added `bare metal` as it is another very popular platform type, and an `other` field just in case.

I've already linked the cluster pre-requisites which clearly state which is supported and which isnt

.github/ISSUE_TEMPLATE/bug.yaml
```diff
@@ -0,0 +1,118 @@
...
+        - OKD
+        - OCP
+    validations:
+      required: true
+  - type: input
+    id: cluster-version
+    attributes:
+      label: Cluster Version
+      description: What is the version of your cluster?
+      placeholder: 4.11
```
mtnbikenc, line 37: This might need quotes.  The validator complains this should be a `string`.

.github/ISSUE_TEMPLATE/bug.yaml
aravindhp: I suggest leaving it out. We should only list supported platforms here.

.github/ISSUE_TEMPLATE/bug.yaml
```diff
@@ -0,0 +1,118 @@
+name: Bug Report
+description: File a bug report
+title: "[Bug]: "
```
aravindhp, line 3: Please add `assignees: @openshift/openshift-team-windows-containers `

aravindhp (changes requested):

.github/ISSUE_TEMPLATE/bug.yaml
```diff
@@ -0,0 +1,116 @@
...
+    attributes:
+      label: Do you have a workaround for this issue?
+    validations:
+      required: false
+  - type: markdown
+    attributes:
+      value: |
+        Please attach any logs and relevant screenshots in a comment.
+        At minimum, please include the output of the following
+        * oc get network.operator cluster -o yaml
```
aravindhp, line 112: Does this tell us that proxy is enabled?

aravindhp (approved):
Thanks for updating this, @sebsoto

.github/ISSUE_TEMPLATE/rfe.yaml
```diff
@@ -0,0 +1,25 @@
...
+  - type: markdown
+    attributes:
+      value: |
+        This form is for requesting a change in expected behavior for WMCO.
+  - type: input
+    id: problem-statement
+    attributes:
+      label: Problem Statement
+      description: A short description of the core problem that needs to be solved
+      placeholder: I want to be able to use private container registries
```
saifshaikh48, line 16: Does this text need to be a double quoted string? I'm not sure a placeholder is needed here anyway

sebsoto: > Does this text need to be a double quoted string?

No it doesn't

I like the placeholder as it guides the issue writer to use a user story approach when describing the problem they have

PR #1091, opened 2022-06-06
WINC-805: Add support for Windows Server 2022 in AWS

jrvaldes:
This PR updates the AWS's EC2 AMI to use Windows Server 2022
Base image without Docker container runtime, as the container runtime
(containerd) is installed by WMCO.

Leverage the environment variable introduced in the release job configuration to use Windows Server 2019 as machine and container images for the `aws-e2e-operator` test
- https://github.com/openshift/release/pull/29791

PR #1197, opened 2022-08-20
WINC-815: Add support for Windows Server 2022 on GCP clusters

sebsoto:
This PR adds multiple commits around providing WMCO support for GCP clusters
* Commits to have GCP to work with our test suite:
   * [controllers] Add ignore label to filter Machines 
   * [e2e] Add GCP provider

*  Commits to add GCP functionality and docs to WMCO:
   * [secrets] Enable Administrator account if needed
   * [docs] Document GCP support

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,7 @@ spec:
 Example MachineSet for other cloud providers:
 - [AWS](docs/machineset-aws.md)
 - [Azure](docs/machineset-azure.md)
+- [GCP](docs/machineset-gcp.md)
 
 Alternatively, the [hack/machineset.sh](hack/machineset.sh) script can be used to generate MachineSets for AWS and Azure platforms.
 The hack script will generate a `MachineSet.yaml` file which can be edited before using or can be used as it is.
--- a/controllers/windowsmachine_controller.go
+++ b/controllers/windowsmachine_controller.go
@@ -54,6 +54,8 @@ const (
 	MachineOSLabel = "machine.openshift.io/os-id"
 	// WindowsMachineController is the name of this controller in logs and other outputs.
 	WindowsMachineController = "windowsmachine"
+	// IgnoreLabel is a label that will cause machines to be ignored by the Windows Machine controller
+	IgnoreLabel = "windowsmachineconfig.openshift.io/ignore"
 )
 
 // WindowsMachineReconciler is used to create a controller which manages Windows Machine objects
@@ -146,7 +148,7 @@ func (r *WindowsMachineReconciler) mapNodeToMachine(object client.Object) []reco
 
 	// Map the Node to the associated Machine through the Node's UID
 	machines, err := r.machineClient.Machines(machineAPINamespace).List(context.TODO(),
-		meta.ListOptions{LabelSelector: MachineOSLabel + "=Windows"})
+		meta.ListOptions{LabelSelector: MachineOSLabel + "=Windows," + IgnoreLabel + "!=true"})
 	if err != nil {
 		r.log.Error(err, "could not get a list of machines")
 	}
@@ -183,6 +185,9 @@ func isWindowsMachine(labels map[string]string) bool {
 
 // isValidMachine returns true if the Machine given object is a Machine with a properly populated status
 func (r *WindowsMachineReconciler) isValidMachine(obj client.Object) bool {
+	if value := obj.GetLabels()[IgnoreLabel]; value == "true" {
+		return false
+	}
 	machine := &mapi.Machine{}
 
 	// If this function is called on an object that equals nil, return false
--- a/docs/HACKING.md
+++ b/docs/HACKING.md
@@ -249,3 +249,8 @@ git submodule update --recursive
 # To update a specific submodule
 git submodule update --remote <path_to_submodule>
 ```
+
+## Preventing Windows Machines from being configured by WMCO
+
+If the Machine spec has the label `windowsmachineconfig.openshift.io/ignore=true`, the Machine will be ignored by WMCO,
+and will not be configured into a Windows node. This can be helpful when debugging userdata changes.
--- a/docs/HACKING.md
+++ b/docs/HACKING.md
@@ -252,5 +252,6 @@ git submodule update --remote <path_to_submodule>
 
 ## Preventing Windows Machines from being configured by WMCO
 
-If the Machine spec has the label `windowsmachineconfig.openshift.io/ignore=true`, the Machine will be ignored by WMCO,
-and will not be configured into a Windows node. This can be helpful when debugging userdata changes.
+If the Machine spec has the label `windowsmachineconfig.openshift.io/ignore=true`, the Machine will be ignored by
+WMCO's Machine controller, and will not be configured into a Windows node. This can be helpful when debugging userdata
+changes.
--- a/docs/machineset-gcp.md
+++ b/docs/machineset-gcp.md
@@ -1,0 +1,74 @@
+# Creating a GCP Windows MachineSet
+
+_\<windows_server_image \>_ should be replaced with the full path to an image of a [supported version](wmco-prerequisites.md#supported-windows-server-versions) of Windows.
+For example, to use the latest image from the `windows-2022-core` family in the `windows-cloud` project, you would use:
+`projects/windows-cloud/global/images/family/windows-2022-core`
+
+
+_\<infrastructure_id\>_ should be replaced with the output of:
+```shell script
+ oc get -o jsonpath='{.status.infrastructureName}{"\n"}' infrastructure cluster
+```
+_\<region\>_ should be replaced with a region, such as `us-central1`.
+
+_\<zone\>_ should be replaced with a zone within the chosen region, such as `us-central1-a`.
+
+_\<zone_suffix\>_ should be replaced with the suffix of the chosen zone, such as `a`.
+
+_\<project_id\>_ should be replaced with the GCP project this cluster was created under.
+```
+apiVersion: machine.openshift.io/v1beta1
+kind: MachineSet
+metadata:
+  labels:
+    machine.openshift.io/cluster-api-cluster: <infrastructure_id>
+  name: <infrastructure_id>-windows-worker-<zone_suffix>
+  namespace: openshift-machine-api
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      machine.openshift.io/cluster-api-cluster: <infrastructure_id>
+      machine.openshift.io/cluster-api-machineset: <infrastructure_id>-windows-worker-<zone_suffix>
+  template:
+    metadata:
+      labels:
+        machine.openshift.io/cluster-api-cluster: <infrastructure_id>
+        machine.openshift.io/cluster-api-machine-role: worker
+        machine.openshift.io/cluster-api-machine-type: worker
+        machine.openshift.io/cluster-api-machineset: <infrastructure_id>-windows-worker-<zone_suffix>
+        machine.openshift.io/os-id: Windows
+    spec:
+      metadata:
+        labels:
+          node-role.kubernetes.io/worker: ""
+      providerSpec:
+        value:
+          apiVersion: machine.openshift.io/v1beta1
+          canIPForward: false
+          credentialsSecret:
+            name: gcp-cloud-credentials
+          deletionProtection: false
+          disks:
+          - autoDelete: true
+            boot: true
+            image: <windows_server_image>
+            sizeGb: 128
+            type: pd-ssd
+          kind: GCPMachineProviderSpec
+          machineType: n1-standard-4
+          networkInterfaces:
+          - network: <infrastructure_id>-network
+            subnetwork: <infrastructure_id>-worker-subnet
+          projectID: <project_id>
+          region: <region>
+          serviceAccounts:
+          - email: <infrastructure_id>-w@<project_id>.iam.gserviceaccount.com
+            scopes:
+            - https://www.googleapis.com/auth/cloud-platform
+          tags:
+          - <infrastructure_id>-worker
+          userDataSecret:
+            name: windows-user-data
+          zone: <zone>
+```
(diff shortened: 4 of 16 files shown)


controllers/controllers.go
```diff
@@ -162,10 +162,15 @@ func (r *instanceReconciler) reconcileKubeletClientCA(ctx context.Context, bundl
 }
 
 // GetAddress returns a non-ipv6 address that can be used to reach a Windows node. This can be either an ipv4
-// or dns address.
+// or dns address. DNS will be preferred, if available.
```
aravindhp, line 165: This change needs to be a separate PR to be on the safe side given it changes behavior across all clouds.

controllers/windowsmachine_controller.go
```diff
@@ -202,7 +201,7 @@ func (r *WindowsMachineReconciler) isValidMachine(obj client.Object) bool {
 		return false
 	}
 
-	_, err := getInternalIPAddress(machine.Status.Addresses)
+	_, err := GetAddress(machine.Status.Addresses)
```
aravindhp, line 209: Where is `GetAddress` defined?

controllers/windowsmachine_controller.go
```diff
@@ -53,6 +53,8 @@ const (
 	MachineOSLabel = "machine.openshift.io/os-id"
 	// WindowsMachineController is the name of this controller in logs and other outputs.
 	WindowsMachineController = "windowsmachine"
+	// IgnoreLabel is a label that will cause machines to be ignored by the Windows Machine controller
+	IgnoreLabel = "windowsmachineconfig.openshift.io/ignore"
```
aravindhp, line 57: You are couching this change as an `e2e` commit but you are actually changing the controller's behavior.

pkg/secrets/secrets.go
```diff
@@ -52,6 +52,22 @@ func GenerateUserData(publicKey ssh.PublicKey) (*core.Secret, error) {
 		},
 		Data: map[string][]byte{
 			"userData": []byte(`<powershell>
+			function Get-RandomPassword {
+				Add-Type -AssemblyName 'System.Web'
+				return [System.Web.Security.Membership]::GeneratePassword(16, 2)
+			}
+
+			# Check if the capi user exists, this will be the case on Azure, and will be used instead of Administrator
```
aravindhp, line 60: As an aside I wonder if we can do this on Azure also? It is also making me wonder if we should just create a new `core` user instead like we wanted to in the past.

pkg/secrets/secrets.go
```diff
@@ -52,6 +52,22 @@ func GenerateUserData(publicKey ssh.PublicKey) (*core.Secret, error) {
...
+			if((Get-LocalUser |Where-Object {$_.Name -eq "capi"})  -eq $null) {
+				# The capi user doesn't exist, ensure the Administrator account is enabled
+				$UserAccount = Get-LocalUser -Name "Administrator"
```
aravindhp, line 63: What happens if there is no Administrator account?

aravindhp (changes requested):
Thanks for working on this, @sebsoto. Looks good. Just some minor comments and splitting up the controller using DNS to another PR.

controllers/windowsmachine_controller.go
sebsoto: Reworked the commit title to correct this

pkg/secrets/secrets.go
sebsoto: It would fail and continue trying to execute the script if both the Administrator account and capi user are missing.
Changed it to only try to enable the Administrator account if it exists.

pkg/secrets/secrets.go
sebsoto: We can re-open https://issues.redhat.com/browse/WINC-430
This could cause issues with people using custom images

controllers/windowsmachine_controller.go
aravindhp: Please document this.

pkg/secrets/secrets.go
aravindhp: >This could cause issues with people using custom images

Why would creating a core user cause issues?

pkg/secrets/secrets.go
aravindhp: We should output an error if there is no capi and Administrator user. This hopefully we will somehow end up in the cloud console log.

aravindhp (changes requested):

pkg/secrets/secrets.go
sebsoto: I'm thinking of a case where the Administrator user has been set up with certain permissions in the image, or theres security software installed special casing the Administrator user. Its nothing we cant doc around.

pkg/secrets/secrets.go
sebsoto: This line does actually write out an error if the user doesn't exist.
Its the reason I used `Where-Object` in the capi query, that one will not write out an error if capi isn't found.

sebsoto: I'll add a comment anyway

pkg/secrets/secrets.go
```diff
@@ -52,6 +52,23 @@ func GenerateUserData(publicKey ssh.PublicKey) (*core.Secret, error) {
...
+				Add-Type -AssemblyName 'System.Web'
+				return [System.Web.Security.Membership]::GeneratePassword(16, 2)
+			}
+
+			# Check if the capi user exists, this will be the case on Azure, and will be used instead of Administrator
+			if((Get-LocalUser |Where-Object {$_.Name -eq "capi"})  -eq $null) {
+				# The capi user doesn't exist, ensure the Administrator account is enabled if it exists
+				# If neither users exist, an error will be written to the console, but the script will still continue
+				$UserAccount = Get-LocalUser -Name "Administrator"
+				if( ($UserAccount -ne $null) -and (!$UserAccount.Enabled) ) {
```
aravindhp, line 65: Nit extra spaces after `if(` and `)`.

controllers/windowsmachine_controller.go
aravindhp: This has not been documented.

aravindhp (changes requested):

controllers/windowsmachine_controller.go
sebsoto: Missed this, done

saifshaikh48 (commented):
Does the e2e [logic to list nodes](https://github.com/sebsoto/windows-machine-config-operator/blob/gcpCIReal/test/e2e/create_test.go#L487-L494) also need to be changed?

I think the ignore label introduction is a big enough change that a short design meeting to introduce it and discuss with the whole team would be helpful.

sebsoto:
@saifshaikh48 Nope that logic does not need to be changed. It is not switching based on the presence of the Windows label

aravindhp (approved):
Nice work @sebsoto

README.md
```diff
@@ -105,6 +105,9 @@ The following label has to be added to the Machine spec within the MachineSet sp
 
 Not having these labels will result in the Windows node not being marked as a worker.
 
+If the Machine spec has the label `windowsmachineconfig.openshift.io/ignore=true`, the Machine will be ignored by WMCO,
```
mansikulkarni96, line 108: should this information be moved to the hacking.md file? Wondering if this would be useful in the README if we don't expect users to use this other than for development and testing.

sebsoto: That's true, I was back and forth about this. I can move it there

test/e2e/validation_test.go
```diff
@@ -310,7 +310,8 @@ func (tc *testContext) sshSetup() error {
 func (tc *testContext) runPowerShellSSHJob(name, command, ip string) (string, error) {
 	// Modify command to work when default shell is the newer Powershell version present on Windows Server 2022.
 	powershellDefaultCommand := command
-	if tc.CloudProvider.GetType() == config.VSpherePlatformType || tc.CloudProvider.GetType() == config.AzurePlatformType {
+	if tc.CloudProvider.GetType() == config.VSpherePlatformType ||
+		tc.CloudProvider.GetType() == config.GCPPlatformType || tc.CloudProvider.GetType() == config.AzurePlatformType {
 		powershellDefaultCommand = strings.ReplaceAll(command, "\\\"", "\"")
```
jrvaldes, line 315: @sebsoto the cloud provider is not the driving factor here. This modification is required due to Windows Server 2022.

Consider using the existing `tc.windowsServerVersion` flag in the condition statement for a more accurate decision. See this [commit](https://github.com/openshift/windows-machine-config-operator/pull/1091/commits/c312195b164d86f351c3a5b5f98ffacd1d795bbf#diff-19a6ecdd2ef106d645523b5e6ddc22fd6ddfd1608477cd0b0fbd85113c8c245d).

For example:
```go
	if tc.windowsServerVersion != "2019" {
		powershellDefaultCommand = strings.ReplaceAll(command, "\\\"", "\"")
	}
```

docs/HACKING.md
```diff
@@ -249,3 +249,8 @@ git submodule update --recursive
 # To update a specific submodule
 git submodule update --remote <path_to_submodule>
 ```
+
+## Preventing Windows Machines from being configured by WMCO
+
+If the Machine spec has the label `windowsmachineconfig.openshift.io/ignore=true`, the Machine will be ignored by WMCO,
+and will not be configured into a Windows node. This can be helpful when debugging userdata changes.
```
saifshaikh48, line 256: > the Machine will be ignored by WMCO and will not be configured into a Windows node

unless it is added as an entry to the windows-instances ConfigMap, right? Since this is hack readme, might be good to call this out.

pkg/secrets/secrets.go
```diff
@@ -52,6 +52,23 @@ func GenerateUserData(publicKey ssh.PublicKey) (*core.Secret, error) {
 		},
 		Data: map[string][]byte{
 			"userData": []byte(`<powershell>
+			function Get-RandomPassword {
+				Add-Type -AssemblyName 'System.Web'
+				return [System.Web.Security.Membership]::GeneratePassword(16, 2)
+			}
+
+			# Check if the capi user exists, this will be the case on Azure, and will be used instead of Administrator
+			if((Get-LocalUser |Where-Object {$_.Name -eq "capi"})  -eq $null) {
```
saifshaikh48, line 61: nits: please put a space after the pipe and remove the extra space before `-eq` (there's 2)

saifshaikh48 (commented):
Good stuff, just a few minor comments

docs/machineset-gcp.md
```diff
@@ -0,0 +1,74 @@
...
+```shell script
+ oc get -o jsonpath='{.status.infrastructureName}{"\n"}' infrastructure cluster
+```
+_\<region\>_ should be replaced with a region, such as `us-central1`.
+
+_\<zone\>_ should be replaced with a zone within the chosen region, such as `us-central1-a`.
+
+_\<zone_suffix\>_ should be replaced with the suffix of the chosen zone, such as `a`.
+
+_\<project_id\>_ should be replaced with the GCP project this cluster was created under.
```
jrvaldes, line 18: Is there a `gcloud`  command to fetch this info? `gcloud config get-value project` maybe?

test/e2e/validation_test.go
sebsoto: @jrvaldes thats a good point, but I don't think that method is usable until the AWS 2022 PR is merged. If I try and do that now, the AWS 2022 jobs will fail

sebsoto: What I mean is the jobs that are set for use with 2022 will fail due to the wrong powershell parsing being used for them

docs/HACKING.md
sebsoto: I'm avoiding calling that out specifically because I dont want anyone to assume thats okay to do.
I'll change the language to specifically call out the Machine controller will ignore the Machine though.

docs/machineset-gcp.md
sebsoto: This could give the wrong value if the default project they have set up on their shell isn't the one that the cluster exists on

test/e2e/validation_test.go
jrvaldes: The environment variable introduced in https://github.com/openshift/release/pull/29791 and is ready to use.

docs/machineset-gcp.md
jrvaldes: I follow, is it safe to use it in the [machineset.sh](https://github.com/openshift/windows-machine-config-operator/blob/258b568146e2e7e2637aa0c28a0062bf34f6d20b/hack/machineset.sh#L189) script?

sebsoto: i think its fine to use there as that script is meant for our use, not for customers

test/e2e/validation_test.go
sebsoto: Yes but aws-e2e-ccm-install and the aws upgrade job have that env var set to ""
this will cause the test suite to detect them as running server 2022 and use the wrong powershell format because of that

jrvaldes: I don't think so. the Windows Server 2019 is [hard-coded](https://github.com/openshift/windows-machine-config-operator/blob/f619b40ec2175f58926547a2e2b94e5b1d8a22c0/test/e2e/providers/aws/aws.go#L128).

My point is that we should stop using `tc.CloudProvider.GetType()` as criteria to tweak the PS commands. It was valid with AWS+WS2022 and is now a reality with GCP+WS2022.

sebsoto: we cannot do `powershellDefaultCommand = strings.ReplaceAll(command, "\\\"", "\"")` on VMs running Windows server 2019. The AWS upgrade and ccm jobs are currently running 2019, but checking that environment variable will tell us that they are running 2022.

sebsoto:
Azure CI failing due to node logs test issues.
Opened https://github.com/openshift/windows-machine-config-operator/pull/1214 to address this

PR #1206, opened 2022-08-23
Prefer using DNS over IP address to connect to Windows instances

sebsoto:
I'm seeing a behavior where Machines will be allocated a new IP address
shortly after WMCO attempts to reconcile them. Using DNS will allow us
to safely retry connecting to the instance, without wasting time by
trying to connect to the wrong IP.

### DIFF
--- a/controllers/controllers.go
+++ b/controllers/controllers.go
@@ -162,10 +162,15 @@ func (r *instanceReconciler) reconcileKubeletClientCA(ctx context.Context, bundl
 }
 
 // GetAddress returns a non-ipv6 address that can be used to reach a Windows node. This can be either an ipv4
-// or dns address.
+// or dns address. DNS will be preferred, if available.
 func GetAddress(addresses []core.NodeAddress) (string, error) {
 	for _, addr := range addresses {
-		if addr.Type == core.NodeInternalIP || addr.Type == core.NodeInternalDNS {
+		if addr.Type == core.NodeInternalDNS {
+			return addr.Address, nil
+		}
+	}
+	for _, addr := range addresses {
+		if addr.Type == core.NodeInternalIP {
 			// filter out ipv6
 			if net.ParseIP(addr.Address) != nil && net.ParseIP(addr.Address).To4() == nil {
 				continue
--- a/controllers/controllers_test.go
+++ b/controllers/controllers_test.go
@@ -12,31 +12,31 @@ func TestGetAddress(t *testing.T) {
 	testCases := []struct {
 		name        string
 		input       []core.NodeAddress
-		expectedOut []string
+		expectedOut string
 		expectedErr bool
 	}{
 		{
 			name:        "no addresses",
 			input:       []core.NodeAddress{{}},
-			expectedOut: []string{""},
+			expectedOut: "",
 			expectedErr: true,
 		},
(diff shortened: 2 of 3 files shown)


aravindhp (approved):
Thanks for splitting this out, @sebsoto

saifshaikh48 (commented):
/lgtm

controllers/controllers_test.go
```diff
@@ -12,39 +12,39 @@ func TestGetAddress(t *testing.T) {
...
+			expectedOut: "localhost",
 			expectedErr: false,
 		},
 		{
 			name: "dns and ipv4",
 			input: []core.NodeAddress{
 				{Type: core.NodeInternalDNS, Address: "localhost"},
 				{Type: core.NodeInternalIP, Address: "127.0.0.1"}},
-			expectedOut: []string{"localhost", "127.0.0.1"},
+			expectedOut: "localhost",
```
jrvaldes, line 47: Consider moving the `NodeInternalDNS` address to the bottom of the array to ensure we pick the preferred one.

controllers/controllers.go
```diff
@@ -162,10 +162,15 @@ func (r *instanceReconciler) reconcileKubeletClientCA(ctx context.Context, bundl
...
+// or dns address. DNS will be preferred, if available.
 func GetAddress(addresses []core.NodeAddress) (string, error) {
 	for _, addr := range addresses {
-		if addr.Type == core.NodeInternalIP || addr.Type == core.NodeInternalDNS {
+		if addr.Type == core.NodeInternalDNS {
+			return addr.Address, nil
+		}
+	}
+	for _, addr := range addresses {
+		if addr.Type == core.NodeInternalIP {
```
jrvaldes, line 173: I feel you can do this in one pass; loop the addresses, save the first valid NodeInternalIP, or as a short-circuit, return the first NodeInternalDNS; 

For example:

```go
internalIP := ""
for _, addr := range addresses {
		if addr.Type == core.NodeInternalDNS {
			return addr.Address, nil
		}
               if addr.Type == core.NodeInternalIP {
		        if net.ParseIP(addr.Address) != nil && net.ParseIP(addr.Address).To4() == nil {
				continue
			 }
                         internalIP = addr.Address
               }
}

if internalIP == "" {
      return "", errors.New("no usable address")
}

return internalIP, nil

```

controllers/controllers.go
```diff
@@ -162,10 +162,15 @@ func (r *instanceReconciler) reconcileKubeletClientCA(ctx context.Context, bundl
 }
 
 // GetAddress returns a non-ipv6 address that can be used to reach a Windows node. This can be either an ipv4
-// or dns address.
+// or dns address. DNS will be preferred, if available.
 func GetAddress(addresses []core.NodeAddress) (string, error) {
```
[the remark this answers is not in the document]
aravindhp: I suggest we stick to `GetAddress()`. There is no real need to call out we are looking at internal addresses.

controllers/controllers.go
aravindhp: This is a good suggestion.

sebsoto:
For some reason this doesn't work on vSphere. Will debug in my free cycles

PR #1207, opened 2022-08-23
[docs] Add Windows Server 2022 for default VXLAN

aravindhp:
Add Windows Server 2022 as a supported version for the default VXLAN port case.


jrvaldes:
/lgtm

Any back-ports?

PR #1209, opened 2022-08-24
Fix issue templates

sebsoto:
Issue templates were not being found due to naming issues. This PR fixes that and forces users to use the issue templates when reporting an issue.


jrvaldes (approved):
/lgtm

aravindhp (approved):

PR #1222, opened 2022-08-29
WINC-874: Pick up 1.25 K8s and upgrade to go1.19

mansikulkarni96:
Manually update k8s libraries to 1.25
Ran commands:
```
go get -u github.com/openshift/client-go@9dfefb19db2e6e1854c389b9e97a784752e2e41c
go mod tidy
go mod vendor
```
Make required changes in Dockerfiles, hack script and the operator.yaml file to update WMCO go version to 1.19.
Ran commands:
`go mod tody && go mod vendor`


jrvaldes:
/lgtm

Thanks @mansikulkarni96  for working on this.

go.mod
```diff
@@ -19,43 +19,43 @@ require (
...
-	golang.org/x/sys v0.0.0-20220712014510-0a85c31ab51e
-	k8s.io/api v0.24.3
-	k8s.io/apimachinery v0.24.3
-	k8s.io/client-go v0.24.3
+	golang.org/x/mod v0.6.0-dev.0.20220419223038-86c51ed26bb4
+	golang.org/x/sys v0.0.0-20220728004956-3c1f35247d10
+	k8s.io/api v0.25.0
+	k8s.io/apimachinery v0.25.0
+	k8s.io/client-go v0.25.0
 	k8s.io/cloud-provider v0.23.1
```
sebsoto, line 27: what about this library?

pkg/certificates/certificates_test.go
```diff
@@ -301,7 +301,8 @@ SupfHrSpy6CQOJNNlDTjGdvJtV7j+uFgrSJhIA==
 
 // testInitialKubeletCACertificate is a fake certificate that contains an initial CA for
 // kube-apiserver-to-kubelet-signer
-//  Issuer: OU=openshift, CN=kube-apiserver-to-kubelet-signer	Not After : Mar 28 14:35:44 2023 GMT
+//
```
sebsoto, line 304: Why does this need to be done?

pkg/nodeconfig/nodeconfig.go
```diff
@@ -165,8 +165,8 @@ func NewNodeConfig(clientset *kubernetes.Clientset, clusterServiceCIDR, vxlanPor
 // For example: https://api-int.abc.devcluster.openshift.com:6443 gets translated to
 // api-int.abc.devcluster.openshift.com
 // TODO: Think if this needs to be removed as this is too restrictive. Imagine apiserver behind
-// 		a loadbalancer.
-// 		Jira story: https://issues.redhat.com/browse/WINC-398
```
sebsoto, line 169: Can this be made to line up with the `Think`?

sebsoto (changes requested):

pkg/certificates/certificates_test.go
sebsoto: Looks like its just an extra space that needs to be corrected?

go.mod
```diff
@@ -6,8 +6,8 @@ require (
 	github.com/apparentlymart/go-cidr v1.1.0
 	github.com/aws/aws-sdk-go v1.38.23
 	github.com/go-logr/logr v1.2.3
-	github.com/openshift/api v0.0.0-20220525145417-ee5b62754c68
-	github.com/openshift/client-go v0.0.0-20220525160904-9e1acff93e4a
+	github.com/openshift/api v0.0.0-20220915134421-1265e9925688
```
aravindhp, line 9: I see newer commits for openshift/api. Any reason to stick to this?

aravindhp (changes requested):

Please fix `go mod tody` in `Update to go1.19` commit message.

pkg/nodeconfig/init.go
```diff
@@ -10,7 +10,7 @@ import (
 // invalidate this cache as of now, since the only entry in this workerIgnitionEndPoint which will be immutable. If
 // someone wants to change it, they've to restart the operator which will invalidate the cache automatically.
 // Note : It is ok to remove this struct in future, if we don't want to continue. As of now, I can think of only
-// 		  worker ignition endpoint being part of this struct.
+// worker ignition endpoint being part of this struct.
```
aravindhp, line 13: You might as well fix up `Note :`

pkg/nodeconfig/nodeconfig.go
aravindhp: We should be consistent about lining up. We are doing things differently for `Note:`.

aravindhp (changes requested):

alinaryan:
LGTM other than the spacing nits

pkg/nodeconfig/nodeconfig.go
mansikulkarni96: no bringing it inline with think is not making linter happy.

Here is an example from: https://go.dev/doc/comment

Even with these heuristics, other existing comments will need manual adjustment to correct their rendering. The most common mistake is indenting a wrapped unindented line of text. For example:

// TODO Revisit this design. It may make sense to walk those nodes
//      only once.

// According to the document:
// "The alignment factor (in bytes) that is used to align the raw data of sections in
//  the image file. The value should be a power of 2 between 512 and 64 K, inclusive."
In both of these, the last line is indented, making it a code block. The fix is to unindent the lines.

sebsoto (approved):

go.mod
```diff
@@ -6,8 +6,8 @@ require (
 	github.com/apparentlymart/go-cidr v1.1.0
 	github.com/aws/aws-sdk-go v1.38.23
 	github.com/go-logr/logr v1.2.3
-	github.com/openshift/api v0.0.0-20220525145417-ee5b62754c68
-	github.com/openshift/client-go v0.0.0-20220525160904-9e1acff93e4a
+	github.com/openshift/api v0.0.0-20220929131120-2bde012f248a
```
aravindhp, line 9: Wouldn't it be `go get -u github.com/openshift/api@2bde012f248a5172dcde2f7104caf0726cf6d93a`?

aravindhp (approved):
/lgtm

I suggest prefixing `[build]` in `Update to go1.19` commit.

go.mod
aravindhp: This is in the commit message

aravindhp (approved):
/lgtm

mtnbikenc:
Failing in GCP deprovision.
/override ci/prow/gcp-e2e-operator

PR #1230, opened 2022-09-01
WINC-879: Introduce Pod Security admission labels

jrvaldes:
This change adds the Pod Security admission enforcement labels to
the test namespace created in the e2e tests and to the WMCO namespace
created at deployment, allowing the automatic synchronization of privileged
pods directly without any intermediate pod controller, so that the serviceaccount
has enough privileges to run it. The WMCO namespace is a system namespace ("openshift-" prefixed)
and the `security.openshift.io/scc.podSecurityLabelSync` label is set as
`true` to ensure the automatic Pod Security label synchronization.

### DIFF
--- a/hack/common.sh
+++ b/hack/common.sh
@@ -122,9 +122,18 @@ run_WMCO() {
       fi
   fi
 
-  # Add the "openshift.io/cluster-monitoring:"true"" label to the operator namespace to enable monitoring
-  if ! oc label ns $WMCO_DEPLOY_NAMESPACE openshift.io/cluster-monitoring=true --overwrite; then
-      return 1
+  # required labels for WMCO namespace
+  declare -a NS_LABELS=(
+    # enable monitoring
+    "openshift.io/cluster-monitoring=true"
+    # turn on the automatic label synchronization required for PodSecurity admission
+    "security.openshift.io/scc.podSecurityLabelSync=true"
+    # set pods security profile to privileged. See https://kubernetes.io/docs/concepts/security/pod-security-admission/#pod-security-levels
+    "pod-security.kubernetes.io/enforce=privileged"
+  )
+  # apply required labels to WMCO namespace
+  if ! oc label ns "${WMCO_DEPLOY_NAMESPACE}" "${NS_LABELS[@]}" --overwrite; then
+    error-exit "error setting labels ${NS_LABELS[@]} in namespace ${WMCO_DEPLOY_NAMESPACE}"
   fi
 
   if [ -n "$PRIVATE_KEY" ]; then
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -74,6 +74,8 @@ type testContext struct {
 	providers.CloudProvider
 	// workloadNamespace is the namespace to deploy our test pods on
 	workloadNamespace string
+	// workloadNamespaceLabels contains the required labels for the test pods' namespace
+	workloadNamespaceLabels map[string]string
 	// toolsImage is the image specified by the  openshift/tools ImageStream, and is the same image used by `oc debug`.
 	// This image is available on all OpenShift Clusters, and has SSH pre-installed.
 	toolsImage string
@@ -94,10 +96,17 @@ func NewTestContext() (*testContext, error) {
 		return nil, errors.Wrap(err, "error getting debug image")
 	}
 
+	workloadNamespaceLabels := map[string]string{
+		// turn off the automatic label synchronization required for PodSecurity admission
+		"security.openshift.io/scc.podSecurityLabelSync": "false",
+		// set pods security profile to privileged. See https://kubernetes.io/docs/concepts/security/pod-security-admission/#pod-security-levels
+		"pod-security.kubernetes.io/enforce": "privileged",
+	}
+
 	// number of nodes, retry interval and timeout should come from user-input flags
 	return &testContext{client: oc, timeout: retry.Timeout, retryInterval: retry.Interval,
 		namespace: "openshift-windows-machine-config-operator", CloudProvider: cloudProvider,
-		workloadNamespace: "wmco-test", toolsImage: toolsImage}, nil
+		workloadNamespace: "wmco-test", workloadNamespaceLabels: workloadNamespaceLabels, toolsImage: toolsImage}, nil
 }
 
 // vmUsername returns the name of the user which can be used to log into each Windows instance
--- a/test/e2e/network_test.go
+++ b/test/e2e/network_test.go
@@ -368,21 +368,31 @@ func getAffinityForNode(node *v1.Node) (*v1.Affinity, error) {
 	}, nil
 }
 
-// ensureNamespace checks if a namespace with the provided name exists and creates one if it does not
-func (tc *testContext) ensureNamespace(name string) error {
+// ensureNamespace checks if a namespace with the provided name exists and creates one if it does not with the given
+// labels
+func (tc *testContext) ensureNamespace(name string, labels map[string]string) error {
 	// Check if the namespace exists
-	_, err := tc.client.K8s.CoreV1().Namespaces().Get(context.TODO(), name, metav1.GetOptions{})
+	ns, err := tc.client.K8s.CoreV1().Namespaces().Get(context.TODO(), name, metav1.GetOptions{})
 	if err != nil && !apierrors.IsNotFound(err) {
 		return err
 	}
 	if err == nil {
+		// ensure namespace was properly created with the required labels
+		for k, expectedValue := range labels {
+			if foundValue, found := ns.Labels[k]; found && expectedValue != foundValue {
+				return fmt.Errorf("labels mismatch for namespace %s label: %s expected: %s found: %s",
+					name, k, expectedValue, foundValue)
+			}
+		}
+		// required labels present in namespace, nothing to do!
 		return nil
 	}
 
 	// The namespace does not exists, so lets create it
-	ns := &v1.Namespace{
+	ns = &v1.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: name,
+			Name:   name,
+			Labels: labels,
 		},
 	}
 	_, err = tc.client.K8s.CoreV1().Namespaces().Create(context.TODO(), ns, metav1.CreateOptions{})
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -2,6 +2,8 @@ apiVersion: v1
 kind: Namespace
 metadata:
   labels:
+    security.openshift.io/scc.podSecurityLabelSync: "true"
+    pod-security.kubernetes.io/enforce: "privileged"
     openshift.io/cluster-monitoring: "true"
     name: windows-machine-config-operator
   name: system
--- a/test/e2e/wmco_test.go
+++ b/test/e2e/wmco_test.go
@@ -34,7 +34,7 @@ func TestWMCO(t *testing.T) {
 	require.NoError(t, err)
 
 	// Create the namespace test resources can be deployed in, as well as required resources within said namespace.
-	require.NoError(t, testCtx.ensureNamespace(testCtx.workloadNamespace), "error creating test namespace")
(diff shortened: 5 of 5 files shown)


hack/common.sh
```diff
@@ -122,10 +122,22 @@ run_WMCO() {
...
   fi
 
-  # Add the "openshift.io/cluster-monitoring:"true"" label to the operator namespace to enable monitoring
-  if ! oc label ns $WMCO_DEPLOY_NAMESPACE openshift.io/cluster-monitoring=true --overwrite; then
-      return 1
-  fi
+  # required labels for WMCO namespace
+  declare -a NS_LABELS=(
+    # enable monitoring
+    "openshift.io/cluster-monitoring=true"
```
aravindhp, line 128: Please do not include this change as part of this PR.

aravindhp (changes requested):

hack/common.sh
```diff
@@ -122,10 +122,22 @@ run_WMCO() {
       fi
   fi
 
-  # Add the "openshift.io/cluster-monitoring:"true"" label to the operator namespace to enable monitoring
-  if ! oc label ns $WMCO_DEPLOY_NAMESPACE openshift.io/cluster-monitoring=true --overwrite; then
-      return 1
-  fi
+  # required labels for WMCO namespace
+  declare -a NS_LABELS=(
```
sebsoto, line 126: Why is this being done here and not in the namespace we are creating when installing the operator:
https://github.com/openshift/windows-machine-config-operator/blob/master/config/manager/manager.yaml#L1

How is this going to work for actual users and not just CI?

sebsoto: I dont understand why in general we create the namespace before attempting to install the operator. The only thing i can see is that we want to create the cloud-private-key secret ahead of time for some reason. I dont see that as necesary

jrvaldes: @sebsoto  you are correct. As per standup, this is out of the scope in this PR and should be tackled in a follow-up story.

test/e2e/network_test.go
```diff
@@ -368,21 +368,24 @@ func getAffinityForNode(node *v1.Node) (*v1.Affinity, error) {
...
+// ensureNamespace checks if a namespace with the provided name exists and creates one if it does not with the given
+// labels
+func (tc *testContext) ensureNamespace(name string, labels map[string]string) error {
 	// Check if the namespace exists
 	_, err := tc.client.K8s.CoreV1().Namespaces().Get(context.TODO(), name, metav1.GetOptions{})
 	if err != nil && !apierrors.IsNotFound(err) {
 		return err
 	}
 	if err == nil {
+		// TODO: ensure required labels
```
jrvaldes, line 380: Will this be a valid corner case? What is the probability that another entity created the `wmco-test` namespace?

test/e2e/main_test.go
```diff
@@ -94,10 +96,17 @@ func NewTestContext() (*testContext, error) {
 		return nil, errors.Wrap(err, "error getting debug image")
 	}
 
+	workloadNamespaceLabels := map[string]string{
+		// turn off the automatic label synchronization required for PodSecurity admission
+		"security.openshift.io/scc.podSecurityLabelSync": "false",
```
mansikulkarni96, line 101: Should this be true or false did we get a clarity on this?

test/e2e/network_test.go
jrvaldes: Makes sense.

hack/common.sh
jrvaldes: > I dont understand why in general we create the namespace before attempting to install the operator. The only thing i can see is that we want to create the cloud-private-key secret ahead of time for some reason. I dont see that as necesary

After attempting to install the operator without creating the namespace, got:
```shell
FATA[0000] Failed to run packagemanifests: create catalog: error creating catalog source: namespaces "openshift-windows-machine-config-operator" not found 
Error from server (NotFound): namespaces "openshift-windows-machine-config-operator" not found

```

hack/common.sh
```diff
@@ -122,10 +122,22 @@ run_WMCO() {
...
+    # enable monitoring
+    "openshift.io/cluster-monitoring=true"
+    # turn on the automatic label synchronization required for PodSecurity admission
+    "security.openshift.io/scc.podSecurityLabelSync=true"
+    # set pods security profile to privileged. See https://kubernetes.io/docs/concepts/security/pod-security-admission/#pod-security-levels
+    "pod-security.kubernetes.io/enforce=privileged"
+  )
```
jrvaldes, line 133: WMCO namespace with labels applied:

```
$oc describe ns openshift-windows-machine-config-operator
Name:         openshift-windows-machine-config-operator
Labels:       kubernetes.io/metadata.name=openshift-windows-machine-config-operator
              olm.operatorgroup.uid/04ec388c-2739-4afb-88e4-50c10642787d=
              olm.operatorgroup.uid/573918ed-b56e-4294-bc58-c7d84338c639=
              openshift.io/cluster-monitoring=true
              pod-security.kubernetes.io/enforce=privileged
              security.openshift.io/scc.podSecurityLabelSync=false
Annotations:  openshift.io/sa.scc.mcs: s0:c27,c4
              openshift.io/sa.scc.supplemental-groups: 1000710000/10000
              openshift.io/sa.scc.uid-range: 1000710000/10000
Status:       Active

No resource quota.

No LimitRange resource.
```

test/e2e/network_test.go
```diff
@@ -368,21 +368,24 @@ func getAffinityForNode(node *v1.Node) (*v1.Affinity, error) {
...
+// ensureNamespace checks if a namespace with the provided name exists and creates one if it does not with the given
+// labels
+func (tc *testContext) ensureNamespace(name string, labels map[string]string) error {
 	// Check if the namespace exists
 	_, err := tc.client.K8s.CoreV1().Namespaces().Get(context.TODO(), name, metav1.GetOptions{})
 	if err != nil && !apierrors.IsNotFound(err) {
 		return err
 	}
 	if err == nil {
+		// at this point; trust the namespace was properly created with the required labels
```
aravindhp, line 380: You can assert that the labels are present by looking at the object returned which is currently ignored.

aravindhp (changes requested):
Thanks for working on this, @jrvaldes please address my comments.

hack/common.sh
jrvaldes: The `openshift.io/cluster-monitoring=true` was present. This change modularizes the logic for adding the required labels, as mentioned in the commit message.

hack/common.sh
jrvaldes: That's correct when deploying the operator using the official deployment method using the CatalogSource along with an operator index. 

However, for the e2e tests ([run-ci-e2e-test.sh](https://github.com/openshift/windows-machine-config-operator/blob/master/hack/run-ci-e2e-test.sh#L94)) the operator is deployed using OLM where the namespace is [created](https://github.com/openshift/windows-machine-config-operator/blob/master/hack/common.sh#L119) first  followed by the creation of the `cloud-private-key` secret. Therefore, the Pod Security Admission labels are required in this step.

test/e2e/network_test.go
jrvaldes: Not sure if is worth spending an API call on such assertion, given the probability is very low or [none](https://github.com/openshift/windows-machine-config-operator/pull/1230#discussion_r960774958).

hack/common.sh
```diff
@@ -122,10 +122,22 @@ run_WMCO() {
...
+  # loop required labels and add them to the WMCO namespace
+  for labelKeyValue in "${NS_LABELS[@]}"
```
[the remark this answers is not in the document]
jrvaldes: Yep, no loop is needed. Addressed in https://github.com/openshift/windows-machine-config-operator/pull/1230/commits/a681289c056acea47571c3bc4720fa88e85c7cab

aravindhp (approved):

hack/common.sh
saifshaikh48: The openshift-wmco namespace should have `podSecurityLabelSync=true` right? Just want to confirm since the posted output says `false`

test/e2e/network_test.go
```diff
@@ -368,21 +368,31 @@ func getAffinityForNode(node *v1.Node) (*v1.Affinity, error) {
...
 	}
 	if err == nil {
+		// ensure namespace was properly created with the required labels
+		for k, expectedValue := range labels {
+			if foundValue, found := ns.Labels[k]; found && expectedValue != foundValue {
+				return fmt.Errorf("labels mismatch for namespace %s label: %s expected: %s found: %s",
+					name, k, expectedValue, foundValue)
+			}
+		}
+		// required labels present in namespace, nothing to do!
```
saifshaikh48, line 387: I'm not sure this logic is complete. What if the namespace already exists but without a required label? We should by applying the missing ones

saifshaikh48 (commented):
/hold 

one comment in the e2e suite

hack/common.sh
jrvaldes: @saifshaikh48 you are correct. I updated the above output.

test/e2e/network_test.go
jrvaldes: >  What if the namespace already exists but without a required label?

Throws an error. The probability for this edge case to happen is very low since the namespace is maintained by the e2e test suite.

I was overthinking this, and now you are too 😃 .

jrvaldes:
/test aws-e2e-operator

PR merged in the release repo to fix CI issues with AWS compute capacity
- https://github.com/openshift/release/pull/31828

test/e2e/network_test.go
saifshaikh48: Fair enough

saifshaikh48:
/lgtm

Feel free to remove the hold when we can get passed the AWS cluster issues

test/e2e/network_test.go
aravindhp: The only use case I see is in our local testing where we are playing around with the labels and don't want the tests to fail if they don't exist.

jrvaldes:
/hold cancel

Infrastructure resources successfully provisioned in AWS CI [job](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/1230/pull-ci-openshift-windows-machine-config-operator-master-aws-e2e-operator/1565792314415124480)

jrvaldes:
/hold

to investigate CI failure in `aws-e2e-ccm-install` [jobs](https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-openshift-windows-machine-config-operator-master-aws-e2e-ccm-install)

```
time="2022-09-02T22:24:49Z" level=error msg="Bootstrap failed to complete: timed out waiting for the condition"
time="2022-09-02T22:24:49Z" level=error msg="Failed to wait for bootstrapping to complete. This error usually happens when there is a problem with control plane hosts that prevents the control plane operators from creating the control plane."
```

aravindhp:
/hold cancel

[Looks](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/1230/pull-ci-openshift-windows-machine-config-operator-master-aws-e2e-ccm-install/1567221559209234432) like the cluster came up and the WMCO tests are running
```
INFO[2022-09-06T19:41:08Z] Running step aws-e2e-ccm-install-ipi-install-install. 
INFO[2022-09-06T20:26:08Z] Step aws-e2e-ccm-install-ipi-install-install succeeded after 45m0s. 
INFO[2022-09-06T20:26:08Z] Running step aws-e2e-ccm-install-ipi-install-times-collection. 
INFO[2022-09-06T20:26:28Z] Step aws-e2e-ccm-install-ipi-install-times-collection succeeded after 20s. 
INFO[2022-09-06T20:26:28Z] Step phase pre succeeded after 47m10s.       
INFO[2022-09-06T20:26:28Z] Running multi-stage phase test               
INFO[2022-09-06T20:26:28Z] Running step aws-e2e-ccm-install-windows-e2e-operator-test.
```

PR #1234, opened 2022-09-02
[docs] Adjust community readme

alinaryan:
Adds specifics to the community readme.md doc on how testing works,
the benefit of the new testing process, and the steps to release a community
WMCO in order for anyone on the team to release a new community WMCO.


hack/community/readme.md
```diff
@@ -1,21 +1,50 @@
...
+  - Add 'community-' prefix to 
+    - 'metadata.name' 
+    - 'spec.install.spec.deployments.spec.template.spec.containers.env.value'
+  - Add image path 
+  'quay.io/openshift-windows/community-windows-machine-config-operator:community-4.X-<PR-commit-hash>' 
+  to
+    - 'metadata.annotations.containerImage:'
+    - 'spec.install.spec.deployments.spec.template.spec.containers.image:'
+  - If this is a minor release:
+    - remove 'metadata.annotations.olm.skipRange'
```
mansikulkarni96, line 39: we don't need to remove skipRange correct me if we have been doing so.

hack/community/readme.md
```diff
@@ -1,21 +1,50 @@
...
+Red Hat operators WMCO manifests were brought in while maintaining the
+differences needed by the community operator. This new process automates the
+testing portion of a community WMCO release by reducing the manual intervention
+done in the bundle and testing the bundle directly in CI.
+
+### Process
 - Fork and clone [community-operators-prod](https://github.com/redhat-openshift-ecosystem/community-operators-prod)
 repo.
-- Copy the community bundle from the ARTIFACT_DIR.
+- Copy the manifests/ and metadata/ directories from the aws-e2e-operator test
```
jrvaldes, line 20: Given this is an ordered process, I suggest including numbering bullet points. For example:
1. Copy the manifests ...
2. Paste the bundle contents ...
3. ...
N. The last step

hack/community/readme.md
alinaryan: Just double-checked. You're correct.

saifshaikh48 (commented):
/lgtm

mansikulkarni96:
LGTM

hack/community/readme.md
```diff
@@ -1,21 +1,49 @@
 # Cut a new community release
-The generate.sh script is called in the pre step to set up for the 
-aws-e2e-operator-test. The script copies over the WMCO community 
-bundle to the artifact directory while also replacing csv fields such as the 
-name, description, and date. Once generated into the artifact directory, the 
-bundle is used in the windows-e2e-operator test. After it passes, the bundle can
-be used to release a new community WMCO.
+CI contains a COMMUNITY environment variable that defaults to false, and is only
+set to true when a community-4.x branch PR is opened. When true, CI is
```
sebsoto, line 3: `contains a COMMUNITY environment variable` should be reworded, as thats not using fully accurate terminology.

consider something like this: 
```
In CI jobs the `COMMUNITY` environment variable is set to false by default, and is set true when testing PRs against community-4.x branches.
```

hack/community/readme.md
```diff
@@ -1,21 +1,49 @@
...
-bundle to the artifact directory while also replacing csv fields such as the 
-name, description, and date. Once generated into the artifact directory, the 
-bundle is used in the windows-e2e-operator test. After it passes, the bundle can
-be used to release a new community WMCO.
+CI contains a COMMUNITY environment variable that defaults to false, and is only
+set to true when a community-4.x branch PR is opened. When true, CI is 
+signaled to copy the repo's bundle into the test artifact directory, edit the 
+bundle, and then use the revised bundle in the PR's aws-e2e-operator test. After
+the test passes, the bundle can be copied from the PR's artifact directory and 
+used to release a new community WMCO.
```
sebsoto, line 7: Would "bundle manifests", or "manifests", be more accurate here, instead of bundle?

hack/community/readme.md
```diff
@@ -1,21 +1,49 @@
...
+operators/community-windows-machine-config-operator/<x.y.z>. The folder title 
+should be the next semantic version of the community WMCO that is to be 
+released.
+4. Scan over the CSV and annotation files to ensure all the necessary fields
+were replaced by the generate.sh script. The script handles replacing the date, 
+operator description, display name, and maturity.
+5. Manual changes needed in the CSV:
+   - Add 'community-' prefix to 
+     - 'metadata.name' field
+     - 'spec.install.spec.deployments.spec.template.spec.containers.env.value'
```
sebsoto, line 31: This needs to specify the `OPERATOR_NAME` env var

hack/community/readme.md
```diff
@@ -1,21 +1,49 @@
...
+     - 'spec.install.spec.deployments.spec.template.spec.containers.env.value' 
+     field
+   - Add image path 
+   'quay.io/openshift-windows/community-windows-machine-config-operator:community-4.x-<PR-commit-hash>' 
+   to
+     - 'metadata.annotations.containerImage'
+     - 'spec.install.spec.deployments.spec.template.spec.containers.image'
+   - If this is a minor release:
+     - add
+       'spec.replaces: community-windows-machine-config-operator.v<x.y.z>' where
```
sebsoto, line 40: On minor versions `olm.skipRange` should be removed as well

sebsoto (changes requested):
Thanks @alinaryan, mostly looks good, a few comments

hack/community/readme.md
mansikulkarni96: we don't do it on RH operator side and we haven't done it for all releases on community side, is that required?
https://github.com/redhat-openshift-ecosystem/community-operators-prod/blob/main/operators/community-windows-machine-config-operator/5.1.1/manifests/community-windows-machine-config-operator.5.1.1.clusterserviceversion.yaml#L11

sebsoto: [we used to do it](https://github.com/redhat-openshift-ecosystem/community-operators-prod/blob/main/operators/community-windows-machine-config-operator/3.1.1/manifests/community-windows-machine-config-operator.3.1.1.clusterserviceversion.yaml) and I think we should be doing it moving forward. Its not 'required' for things to work, but it doesn't make sense when considering WMCO's upgrade graph, and makes it unclear as to where WMCO will be upgraded to from a previous major version

sebsoto (approved):

saifshaikh48 (commented):
/lgtm

PR #1260, opened 2022-09-22
WINC-738: Define Kubelet service in services ConfigMap

sebsoto:
This PR moves the responsibility of managing the kubelet service into WICD. This required WMCO to pick up the responsibility of getting and parsing the ignition file for kubelet arguments. With this commit, WICD will reconcile an incorrectly configured kubelet service on Windows nodes.


cmd/operator/main.go
```diff
@@ -174,6 +178,17 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Get and parse the worker ignition spec. Done once on operator start.
```
aravindhp, line 181: We should call out somewhere that we are changing the behavior when compared to WMCB when we were doing it once per node.

pkg/ignition/ignition.go
```diff
@@ -0,0 +1,118 @@
...
+	ignCfgTypes.Config
+	// KubeletUnit is the kubelet systemd service spec in ignition file
+	KubeletUnit ignCfgTypes.Unit
+}
+
+// Initialize populates the RenderedWorker global variable with the worker ignition file data.
+func Initialize(c client.Client) error {
+	log := ctrl.Log.WithName("ignition")
+
+	mcList := &mcfg.MachineConfigList{}
```
aravindhp, line 44: `machineConfigs`?

pkg/ignition/ignition.go
```diff
@@ -0,0 +1,118 @@
...
+	err := c.List(context.TODO(), mcList)
+	if err != nil {
+		return err
+	}
+	renderedWorkerPrefix := "rendered-worker-"
+	var rawIgnition []byte
+	for _, mc := range mcList.Items {
+		if strings.HasPrefix(mc.Name, renderedWorkerPrefix) {
```
aravindhp, line 52: Is there a way to pick the latest? Or add a doc here saying why we are picking the first one we find.

pkg/ignition/ignition.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+	}
+	if renderedWorker.KubeletUnit.Contents == nil {
+		return errors.Errorf("ignition missing kubelet systemd unit file")
+	}
+	log.Info("Extracted", "kubelet systemd unit", renderedWorker.KubeletUnit)
+	return nil
+}
+
+// GetKubeletArgs returns a set of arguments for kubelet.exe, as specified in the ignition file
```
aravindhp, line 79: There is an opportunity to make this more oops like. Can you take a stab?

pkg/ignition/ignition.go
```diff
@@ -0,0 +1,118 @@
...
+		if len(keyValue) != 2 {
+			// Not a key value pair, continue
+			continue
+		}
+		for _, windowsArg := range windowsArgs {
+			if windowsArg == keyValue[0] {
+				kubeletArgs[keyValue[0]] = keyValue[1]
+			}
+		}
+	}
```
aravindhp, line 116: Please add some docs around what you are doing here

pkg/ignition/ignition.go
```diff
@@ -0,0 +1,118 @@
...
+		return nil, fmt.Errorf("unit missing ExecStart")
+	}
+	cmdEndSplit := strings.SplitN(execSplit[1], "\n\n", 2)
+	argumentSplit := strings.Split(cmdEndSplit[0], "\\\n")
+	kubeletArgs := make(map[string]string)
+	for _, arg := range argumentSplit[1:] {
+		arg = strings.TrimSpace(arg)
+		arg = strings.TrimPrefix(arg, "--")
+		keyValue := strings.SplitN(arg, "=", 2)
+		if len(keyValue) != 2 {
```
aravindhp, line 107: Shouldn't we also include these?

aravindhp (changes requested):

>This commit moves the responsibility of managing the kubelet service into
WICD. This required WMCO to pick up the responsibility of getting and
parsing the ignition file for kubelet arguments.

Is the above accurate? WMCO is still actually managing the kubelet. It is the follow up PR that will move the responsibility. You are just laying the groundwork.

>With this commit, WICD will reconcile an incorrectly configured kubelet service on Windows nodes.

I find the above statement confusing. We are not going to fix an incorrect kubelet service once we have already prepped the node.

pkg/daemon/controller/controller.go
```diff
@@ -352,6 +354,29 @@ func (sc *ServiceController) expectedServiceCommand(expected servicescm.Service)
 	for key, value := range psVars {
 		expectedCmd = strings.ReplaceAll(expectedCmd, key, value)
 	}
+	if strings.Contains(expectedCmd, services.NodeIPVar) {
```
aravindhp, line 357: Please mention that this goes against the WICD design principle add a link to a Jira story where you will address this.

pkg/ignition/ignition.go
sebsoto: Added comment explaining why thats not needed

cmd/operator/main.go
sebsoto: Changed this as it was unnecessary.

pkg/ignition/ignition.go
sebsoto: Picking the latest now

pkg/ignition/ignition.go
```diff
@@ -0,0 +1,120 @@
...
+	argsFromIgnition, err := parseKubeletArgs(*kubeletUnit.Contents)
+	if err != nil {
+		return nil, errors.Wrap(err, "error parsing kubelet systemd unit args")
+	}
+	return argsFromIgnition, nil
+}
+
+//parseKubeletArgs parses a systemd unit file, returning the kubelet args WMCO is interested in
+func parseKubeletArgs(unitContents string) (map[string]string, error) {
+	windowsArgs := []string{CloudProviderOption, CloudConfigOption}
```
saifshaikh48, line 89: can be moved closer to where it is used

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
```diff
@@ -318,6 +318,13 @@ spec:
           - get
           - list
           - watch
+        - apiGroups:
+          - machineconfiguration.openshift.io
+          resources:
+          - machineconfigs
+          verbs:
+          - list
+          - watch
```
saifshaikh48, line 327: Curious if you dug into [why watch is needed](https://github.com/openshift/windows-machine-config-operator/pull/1235#discussion_r964225221)?

pkg/ignition/ignition.go
```diff
@@ -0,0 +1,120 @@
...
+			break
+		}
+	}
+	if kubeletUnit.Contents == nil {
+		return nil, errors.Errorf("ignition missing kubelet systemd unit file")
+	}
+	log.V(1).Info("Extracted", "kubelet systemd unit", kubeletUnit)
+	if kubeletUnit.Contents == nil {
+		return nil, errors.New("kubelet unit contents nil")
+	}
```
saifshaikh48, line 79: can delete these lines

pkg/ignition/ignition.go
```diff
@@ -0,0 +1,120 @@
...
+	sort.Slice(machineConfigs.Items, func(i, j int) bool {
+		iTimestamp := machineConfigs.Items[i].GetCreationTimestamp()
+		jTimestamp := machineConfigs.Items[j].GetCreationTimestamp()
+		return jTimestamp.Before(&iTimestamp)
+	})
+	var rawIgnition []byte
+	for _, mc := range machineConfigs.Items {
+		if strings.HasPrefix(mc.Name, RenderedWorkerPrefix) {
+			rawIgnition = mc.Spec.Config.Raw
+			log.Info("Received ignition spec", "MachineConfig", mc.Name)
```
saifshaikh48, line 52: do we need this log?

saifshaikh48 (commented):
Nice splitting of the work and looks good overall!

I assume the followup PR to create all the kubelet-required files from WMCO will churn some of this code when trying to avoid getting and parsing ignition every time we reconcile a node.

aravindhp (changes requested):

aravindhp:
> I assume the followup PR to create all the kubelet-required files from WMCO will churn some of this code when trying to avoid getting and parsing ignition every time we reconcile a node.

Really? I thought @sebsoto wanted to retain the WMCB behavior of getting and parsing ignition every a node is reconciled.

pkg/ignition/ignition.go
```diff
@@ -0,0 +1,120 @@
...
+	// Grab the latest rendered-worker MachineConfig by sorting the MachineConfig list by the latest creation
+	// timestamp first.
+	sort.Slice(machineConfigs.Items, func(i, j int) bool {
+		iTimestamp := machineConfigs.Items[i].GetCreationTimestamp()
+		jTimestamp := machineConfigs.Items[j].GetCreationTimestamp()
+		return jTimestamp.Before(&iTimestamp)
+	})
+	var rawIgnition []byte
+	for _, mc := range machineConfigs.Items {
+		if strings.HasPrefix(mc.Name, RenderedWorkerPrefix) {
```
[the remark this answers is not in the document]
sebsoto: I wasn't able to find a way to do that with a name field selector, I'm only aware of exact matching for a field.
Additionally, theres no annotation or label which denotes it being a rendered worker type

pkg/ignition/ignition.go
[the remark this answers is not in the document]
sebsoto: I'm not sure the value of this considering this would only be used when bootstrapping, its a lot more valuable for linux nodes. I've added [another story](https://issues.redhat.com/browse/WINC-897) in a TODO which is semi-related.

pkg/services/services.go
```diff
@@ -96,3 +117,116 @@ func kubeProxyConfiguration(debug bool) servicescm.Service {
...
+		"--container-runtime=remote",
+		"--container-runtime-endpoint=" + containerdEndpointValue,
+		"--resolv-conf=",
+	}
+	if debug {
+		kubeletArgs = append(kubeletArgs, "--v="+debugLogLevel)
+	} else {
+		// In case the verbosity argument is missing, use a default value
+		kubeletArgs = append(kubeletArgs, "--v="+standardLogLevel)
+	}
```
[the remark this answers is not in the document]
sebsoto: Made this into a function

controllers/configmap_controller.go
```diff
@@ -387,9 +393,21 @@ func (r *ConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			builder.WithPredicates(outdatedWindowsNodePredicate(true))).
 		Watches(&source.Kind{Type: &core.Node{}}, handler.EnqueueRequestsFromMapFunc(r.mapToServicesConfigMap),
 			builder.WithPredicates(windowsNodeVersionChangePredicate())).
+		// An update to the rendered worker MachineConfig could include changes to the kubelet args, which must be
+		// reflected in the services ConfigMap.
+		Watches(&source.Kind{Type: &mcfg.MachineConfig{}}, handler.EnqueueRequestsFromMapFunc(r.mapToServicesConfigMap),
```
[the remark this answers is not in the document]
sebsoto: Yes, that is me thinking too far ahead. I've changed the predicate to filter out all MachineConfig events, leaving it in the `Watches` is important as it allows the cached client to pull the MachineConfig object. Added a [TODO](https://issues.redhat.com/browse/WINC-897) to stop filtering events and act upon changes.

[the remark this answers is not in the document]
aravindhp: > leaving it in the Watches is important as it allows the cached client to pull the MachineConfig object

That sounds really weird. This shouldn't be the reason to Watch the object. 

>Added a [TODO](https://issues.redhat.com/browse/WINC-897) to stop filtering events and act upon changes.

This sounds like a risky TODO. Do we know how often these rendered config changes? We don't want a situation where we are unnecessarily re-configuring Windows nodes.

pkg/ignition/ignition.go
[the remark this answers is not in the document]
aravindhp: The value would be when debugging a customer issue. We can easily look at the node object and figure out where we got the kubelet args from.

pkg/daemon/controller/controller.go
```diff
@@ -352,6 +354,31 @@ func (sc *ServiceController) expectedServiceCommand(expected servicescm.Service)
...
+			return "", err
+		}
+		addrs, err := LocalInterfaceAddresses()
+		if err != nil {
+			return "", err
+		}
+		nodeIPValue := ""
+		for _, addr := range addrs {
+			for _, instance := range instances {
+				if instance.IPv4Address == addr.String() {
```
saifshaikh48, line 372: Shouldn't we also compare the address listed in the instances ConfigMap instead? Or am I missing something that makes this check unnecessary

`if instance.Address == addr.String() || instance.IPv4Address == addr.String() {`

saifshaikh48: I suspect this is why the platform-none test is failing with `unable to find IPv4 address to use as Node IP for this instance` -- there is no guarantee the interface addresses for a specific interface will contain an IPv4 address, right?

saifshaikh48: It looks like your latest debug commit makes it clear. `addr.String()` returns a value with subnet appended

```
I0928 22:14:47.569874    4660 controller.go:372] Checking if 192.168.59.27 == fe80::f127:441b:58:4b46/64
I0928 22:14:47.569874    4660 controller.go:372] Checking if 192.168.59.27 == 192.168.59.27/27
I0928 22:14:47.569874    4660 controller.go:372] Checking if 192.168.59.27 == ::1/128
I0928 22:14:47.569874    4660 controller.go:372] Checking if 192.168.59.27 == 127.0.0.1/8
I0928 22:14:47.569874    4660 controller.go:372] Checking if 192.168.59.27 == fe80::d156:975f:69ef:f1c4/64
I0928 22:14:47.569874    4660 controller.go:372] Checking if 192.168.59.27 == 172.27.0.1/20
E0928 22:14:47.569874    4660 controller.go:252] unable to find IPv4 address to use as Node IP for this instance
```

sebsoto: The test was failing because the local address wasn't being converted into an ipv4 address before the comparison.
String-ing a net.Addr will result in an address like `192.168.0.1/32 `, stringing a net.IP will give `192.168.0.1`

sebsoto: > there is no guarantee the list of all interface addresses on a specific instance will contain an IPv4 address, right?

As part of `GetInstances()` .IPv4Address needs to be populated or else the CM parse will fail.

saifshaikh48: I guess my question is related to the instance -- will the addr list returned by `LocalInterfaceAddresses()` always contain an IPv4 value? If not, then IMO we should also compare with `instance.Address`

controllers/configmap_controller.go
```diff
@@ -387,9 +401,24 @@ func (r *ConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			builder.WithPredicates(renderedWorkerPredicate())).
 		Complete(r)
 }
 
+// renderedWorkerPredicate returns a predicate which filters out changes to any MachineConfig except for the rendered
+// worker config
+func renderedWorkerPredicate() predicate.Funcs {
+	// TODO: For now, filter out all rendered worker config changes. In the future we should be updating the kubelet
+	//       config in the windows-services ConfigMap if there is a kubelet arg change.
```
mansikulkarni96, line 415: I suggest we follow the new gofmt conventions keeping non-code blocks unindented.

pkg/services/services.go
```diff
@@ -1,17 +1,37 @@
...
+	"github.com/pkg/errors"
+
+	"github.com/openshift/windows-machine-config-operator/pkg/ignition"
 	"github.com/openshift/windows-machine-config-operator/pkg/nodeconfig"
 	"github.com/openshift/windows-machine-config-operator/pkg/servicescm"
 	"github.com/openshift/windows-machine-config-operator/pkg/windows"
 )
 
+const (
+	//See doc link for explanation of log levels:
```
mansikulkarni96, line 22: space needed after //

pkg/daemon/controller/controller.go
```diff
@@ -352,6 +354,35 @@ func (sc *ServiceController) expectedServiceCommand(expected servicescm.Service)
 	for key, value := range psVars {
 		expectedCmd = strings.ReplaceAll(expectedCmd, key, value)
 	}
+	// TODO: This goes against WICD design principles and needs to be changed https://issues.redhat.com/browse/WINC-896
```
mansikulkarni96, line 357: same applies here for unindented comments.

pkg/daemon/controller/controller.go
sebsoto: It would as per the requirements of WMCO. We don't support IPv6.

controllers/configmap_controller.go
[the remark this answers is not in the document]
saifshaikh48: +1 on removing this watch and `renderedWorkerPredicate`. You can put the TODO with issue link here, and if you'd like update the issue engineering details to capture the watch + predicate.

test/e2e/create_test.go
```diff
@@ -176,6 +196,24 @@ func (tc *testContext) testBYOHConfiguration(t *testing.T) {
...
+// byohLogCollection kicks off the collection of logs for instances listed in the windows-instances configmap
+func (tc *testContext) byohLogCollection() {
+	windowsInstances, err := tc.client.K8s.CoreV1().ConfigMaps(tc.namespace).Get(context.TODO(),
+		wiparser.InstanceConfigMap, metav1.GetOptions{})
+	if err != nil {
+		log.Printf("unable to collect logs, error retrieving instances ConfigMap: %v", err)
+		return
+	}
+	for addr := range windowsInstances.Data {
+		err = tc.nodelessLogCollection("byoh", addr)
```
saifshaikh48, line 212: Should we append the address to the job name instead of just `print-logs-byoh`?

saifshaikh48 (commented):
Nice job getting this working. Just a couple comments

controllers/configmap_controller.go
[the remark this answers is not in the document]
aravindhp: I continue to find this strange. It is one thing to say we need watch for changes in the future but another to add it here because it allows the cached client to pull the MachineConfig object without knowing why. Are you sure you are not missing adding it to a scheme somewhere in the code?

pkg/ignition/ignition.go
```diff
@@ -0,0 +1,134 @@
...
+	// CloudConfigOption is the kubelet CLI option for the cloud configuration file
+	CloudConfigOption = "cloud-config"
+	// CloudProviderOption is the kubelet CLI option for cloud provider
+	CloudProviderOption = "cloud-provider"
+	// RenderedWorkerPrefix allows identification of the rendered worker MachineConfig, the combination of all worker
+	// MachineConfigs.
+	RenderedWorkerPrefix = "rendered-worker-"
+)
+
+// Ignition is a representation of an Ignition file
```
aravindhp, line 32: Nice work

pkg/ignition/ignition.go
```diff
@@ -0,0 +1,134 @@
...
+		iTimestamp := machineConfigs[i].GetCreationTimestamp()
+		jTimestamp := machineConfigs[j].GetCreationTimestamp()
+		return jTimestamp.Before(&iTimestamp)
+	})
+	for _, mc := range machineConfigs {
+		if strings.HasPrefix(mc.Name, RenderedWorkerPrefix) {
+			if len(mc.Spec.Config.Raw) == 0 {
+				continue
+			}
+			return &mc, nil
```
aravindhp, line 95: Please add an info message here saying which MC we picked. This will help debugging customer issues.

pkg/services/services.go
```diff
@@ -67,13 +88,8 @@ func kubeProxyConfiguration(debug bool) servicescm.Service {
...
 		windows.KubeProxyLogDir, windows.OVNKubeOverlayNetwork)
-	// Set log level. See doc link for explanation of log levels:
-	// https://docs.openshift.com/container-platform/latest/rest_api/editing-kubelet-log-level-verbosity.html#log-verbosity-descriptions_editing-kubelet-log-level-verbosity
-	if debug {
-		cmd = fmt.Sprintf("%s --v=4", cmd)
-	} else {
-		cmd = fmt.Sprintf("%s --v=2", cmd)
-	}
+	// Set log level
+	cmd = fmt.Sprintf("%s %s", cmd, klogVerbosityArg(debug))
```
aravindhp, line 93: Thanks for modularizing this but my question still remains. Why do you have to add these args in two places?

test/e2e/create_test.go
aravindhp: +1

aravindhp (changes requested):

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
sebsoto: I was able to remove this. not sure why you were seeing this

controllers/configmap_controller.go
```diff
@@ -390,6 +403,17 @@ func (r *ConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Complete(r)
 }
 
+// renderedWorkerPredicate returns a predicate which filters out changes to any MachineConfig except for the rendered
+// worker config
+func renderedWorkerPredicate() predicate.Funcs {
```
aravindhp, line 408: This function can be removed now.

test/e2e/create_test.go
```diff
@@ -150,6 +159,18 @@ func (tc *testContext) testMachineConfiguration(t *testing.T) {
...
+		addr, err := controllers.GetAddress(machine.Status.Addresses)
+		if err != nil {
+			log.Printf("Machine %s does not have a valid address, unable to get logs", machine.GetName())
+			continue
+		}
+		err = tc.nodelessLogCollection(machine.GetName(), addr)
+		if err != nil {
+			log.Printf("failed to collect logs from %s: %v", machine.GetName(), err)
+		}
+	}
```
aravindhp, line 173: I suggest making this a helper function.

aravindhp (approved):

controllers/configmap_controller.go
saifshaikh48: +1
the `TODO` can be moved to `SetupWithManager` (line 403)

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
saifshaikh48: Thanks for investigating

saifshaikh48 (commented):
/lgtm

saifshaikh48 (commented):
/lgtm

aravindhp:
Blocked by OCPBUGS-2175

pkg/services/services.go
```diff
@@ -96,3 +111,89 @@ func kubeProxyConfiguration(debug bool) servicescm.Service {
...
+	if err != nil {
+		return servicescm.Service{}, err
+	}
+	var powershellVars []servicescm.PowershellCmdArg
+	hostnameArg := getKubeletHostnameOverride(platform)
+	if hostnameArg != "" {
+		kubeletArgs = append(kubeletArgs, hostnameArg)
+		powershellVars = append(powershellVars, servicescm.PowershellCmdArg{
+			Name: ec2HostnameVar,
+			Path: "Get-EC2InstanceMetadata -Category LocalHostname",
```
jrvaldes, line 128: @sebsoto, with the need for the hostname override in GCP, I feel there is a disconnection here between `getKubeletHostnameOverride()`  and the way the variable is getting resolved.  `Get-EC2InstanceMetadata` is platform specific and will only work in AWS. 

Can this be done in a way that could be extended based on the platform type?

jrvaldes: The `Get-EC2InstanceMetadata` is part of the [AWS.Tools for PowerShell](https://docs.aws.amazon.com/powershell/latest/userguide/pstools-welcome.html), and it comes pre-installed on Windows machines. What about BYOH instances in AWS, will the AWS.Tools for PowerShell be available?

pkg/services/services.go
```diff
@@ -2,16 +2,35 @@ package services
...
 	"github.com/openshift/windows-machine-config-operator/pkg/nodeconfig"
 	"github.com/openshift/windows-machine-config-operator/pkg/servicescm"
 	"github.com/openshift/windows-machine-config-operator/pkg/windows"
 )
 
+const (
+	// See doc link for explanation of log levels:
+	// https://docs.openshift.com/container-platform/latest/rest_api/editing-kubelet-log-level-verbosity.html#log-verbosity-descriptions_editing-kubelet-log-level-verbosity
+	debugLogLevel    = "4"
+	standardLogLevel = "2"
```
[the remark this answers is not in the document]
sebsoto: These are not, they are specific to klog

sebsoto:
@jrvaldes considering the length of time this PR was approved and LGTM'd before you reviewed, I'd prefer to skip the comments you've left, if you don't consider them blockers for this PR.

jrvaldes:
> @jrvaldes considering the length of time this PR was approved and LGTM'd before you reviewed, I'd prefer to skip the comments you've left, if you don't consider them blockers for this PR.

Agree. Most of the comments will be addressed in the follow-up PRs that are based on these changes.

aravindhp:
@sebsoto just be aware that the PR will merge even if `platform-none-vsphere-e2e-operator` fails as it is still marked as optional. You would need the hold to ensure it passes.

sebsoto:
Platform none test passed, removing hold
/hold cancel

aravindhp:
@sebsoto I am hoping [this](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/1260/pull-ci-openshift-windows-machine-config-operator-master-azure-e2e-ccm-install/1585695398150803456) is a flake:

```
2022/10/27 20:35:47 Unable to get logs associated with pod: job-name=print-logs-e2e-wm-5xh9r-job-gbpbj
2022/10/27 20:35:47 failed to collect logs from e2e-wm-5xh9r: error waiting for job to succeed:
```

sebsoto:
@aravindhp That job doesn't make any sense to me. I'm not sure why its appearing on this PR. Its running using test code from another branch

aravindhp:
> @aravindhp That job doesn't make any sense to me. I'm not sure why its appearing on this PR. Its running using test code from another branch

Whoa! How is that even possible? Please check with @Fedosin who I think added the job.

PR #1261, opened 2022-09-22
WINC-830: [test] Use lts-nanoserver-ltsc2022 image

aravindhp:
Use the lts-nanoserver-ltsc2022 image for all providers except AWS now that we have moved to using Windows Server 2022.
Ref: https://github.com/openshift/release/pull/32500

### DIFF
--- a/test/e2e/network_test.go
+++ b/test/e2e/network_test.go
@@ -450,13 +450,8 @@ func (tc *testContext) getPodIP(selector metav1.LabelSelector) (string, error) {
 	return podList.Items[0].Status.PodIP, nil
 }
 
-// getWindowsServerContainerImage gets the appropriate WindowsServer image based on VXLAN port
+// getWindowsServerContainerImage gets the appropriate WindowsServer image based on the cloud provider
 func (tc *testContext) getWindowsServerContainerImage() string {
-	if tc.CloudProvider.GetType() == config.NonePlatformType {
-		// TODO: On platform=none we have to use 2004 since the job points to the WS2004 golden image via release repo
-		// When the release repo is updated to use 2022, this clause should be squashed with the above to use 2022
-		return "mcr.microsoft.com/powershell:lts-nanoserver-2004"
-	}
 	if tc.CloudProvider.GetType() == config.AWSPlatformType {
 		// On AWS we are currently testing 2019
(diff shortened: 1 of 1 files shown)


aravindhp:
/retitle WINC-830: [docs] Re-add Docker install in autounattend.xml

aravindhp:
/retitle WINC-830: [docs] Re-add Docker for Windows Server 2022 platform:none

sebsoto:
> Given we still need to use Docker in 5.x and lower it is best we retain this section until we have 0 releases where we use Docker.

Since this is not true for the current release, shouldn't this change be made to release 4.11 and below only?

aravindhp:
> > Given we still need to use Docker in 5.x and lower it is best we retain this section until we have 0 releases where we use Docker.
> 
> Since this is not true for the current release, shouldn't this change be made to release 4.11 and below only?

This is what I was hinting at in today's standup. i.e. split this into multiple PRs. The master PR will only need [[test] Use lts-nanoserver-ltsc2022 image](https://github.com/openshift/windows-machine-config-operator/pull/1261/commits/8ba432d4e759fe3de2a93e89fafc14ee34039091)

saifshaikh48:
/lgtm

Please update the PR description

test/e2e/network_test.go
```diff
@@ -452,11 +452,6 @@ func (tc *testContext) getPodIP(selector metav1.LabelSelector) (string, error) {
 
 // getWindowsServerContainerImage gets the appropriate WindowsServer image based on VXLAN port
```
jrvaldes, line 453: The function's description should be updated given the `VXLAN port` is no longer the driving condition for the selection of the container image.

aravindhp: Fixed.

aravindhp:
> > Given we still need to use Docker in 5.x and lower it is best we retain this section until we have 0 releases where we use Docker.
> 
> Since this is not true for the current release, shouldn't this change be made to release 4.11 and below only?

@sebsoto shouldn't this be release-4.10 and below given we have removed Docker support in 4.11?

aravindhp:
/test platform-none-vsphere-e2e-operator

Hopefully https://github.com/openshift/release/pull/33411 did the trick

aravindhp:
/test platform-none-vsphere-e2e-operator

Cluster failed to come up with what looks like a flake

PR #1269, opened 2022-09-26
WINC-873 : Upgrade to go1.19

mansikulkarni96:
This PR includes all the required changes in Dockerfiles, hack script and the operator.yaml file to update the WMCO go version to 1.19.
Release image has been updated to one that's available for go1.19.
Ran commands:
go mod tidy && go mod vendor


build/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.ci.openshift.org/openshift/release:golang-1.18 as build
+FROM registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.19-openshift-4.12 as build
```
aravindhp, line 1: Is image pullable without credentials? I feel we keep going back and forth between `release:golang-1.n` and `release:rhel-8-release-golang-1.n-openshift-4.y`. What is the image we use in brew?

mansikulkarni96: I am not using release:golang-1.n because it is not available yet, there have been delays from the compiler side. yes I am able to pull it locally without using login credentials.
The one we use in brew now is openshift/golang-builder:rhel_8_1.18

aravindhp: Is it possible to use `openshift/golang-builder:rhel_8_1.18` here? I am guessing no as it won't be part of the CI imagestream. Ideally we should be using the same compiler version everywhere.

mansikulkarni96:
Closing in favor of #1222

PR #1273, opened 2022-10-12
WINC-739: Azure node manager added to services ConfigMap

sebsoto:
Adds the azure cloud node manager into the windows-services ConfigMap
when an azure cluster is set to use CCM. This causes WICD to start and
manage the service. WMCO will no longer start the service.

### DIFF
--- a/controllers/configmap_controller.go
+++ b/controllers/configmap_controller.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 
 	config "github.com/openshift/api/config/v1"
+	openshiftClient "github.com/openshift/client-go/config/clientset/versioned"
 	"github.com/pkg/errors"
 	core "k8s.io/api/core/v1"
 	k8sapierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -100,8 +101,16 @@ func NewConfigMapReconciler(mgr manager.Manager, clusterConfig cluster.Config, w
 	if err != nil {
 		return nil, err
 	}
+	oc, err := openshiftClient.NewForConfig(mgr.GetConfig())
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to create openshift client-go client")
+	}
+	ccmEnabled, err := cluster.IsCloudControllerOwnedByCCM(oc)
+	if err != nil {
+		return nil, errors.Wrap(err, "error determining if CCM owns cloud controller")
+	}
 	svcData, err := services.GenerateManifest(argsFromIgnition, clusterConfig.Network().VXLANPort(),
-		clusterConfig.Platform(), ctrl.Log.V(1).Enabled())
+		clusterConfig.Platform(), ccmEnabled, ctrl.Log.V(1).Enabled())
 	if err != nil {
 		return nil, errors.Wrap(err, "error generating expected Windows service state")
 	}
--- a/controllers/configmap_controller.go
+++ b/controllers/configmap_controller.go
@@ -101,7 +101,7 @@ func NewConfigMapReconciler(mgr manager.Manager, clusterConfig cluster.Config, w
 		return nil, err
 	}
 	svcData, err := services.GenerateManifest(argsFromIgnition, clusterConfig.Network().VXLANPort(),
-		string(clusterConfig.Platform()), ctrl.Log.V(1).Enabled())
+		clusterConfig.Platform(), ctrl.Log.V(1).Enabled())
 	if err != nil {
 		return nil, errors.Wrap(err, "error generating expected Windows service state")
 	}
--- a/pkg/nodeconfig/nodeconfig.go
+++ b/pkg/nodeconfig/nodeconfig.go
@@ -225,23 +225,6 @@ func (nc *nodeConfig) Configure() error {
 			return errors.Wrap(err, "unable to check if cloud controller owned by cloud controller manager")
 		}
 
-		// Configuring cloud node manager for Azure platform with CCM support
-		if ownedByCCM && nc.platformType == configv1.AzurePlatformType {
-			if err := nc.Windows.ConfigureAzureCloudNodeManager(nc.node.GetName()); err != nil {
-				return errors.Wrap(err, "configuring azure cloud node manager failed")
-			}
-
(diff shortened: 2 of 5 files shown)


pkg/services/services.go
```diff
@@ -46,11 +46,33 @@ func GenerateManifest(kubeletArgsFromIgnition map[string]string, vxlanPort strin
...
+		Name:    windows.AzureCloudNodeManagerServiceName,
+		Command: cmd,
+		NodeVariablesInCommand: []servicescm.NodeCmdArg{{
+			Name:               "NODE_NAME",
+			NodeObjectJsonPath: "{.metadata.name}",
+		}},
+		PowershellVariablesInCommand: nil,
+		Dependencies:                 nil,
+		Bootstrap:                    false,
+		Priority:                     0,
```
saifshaikh48, line 72: kubelet has priority 0 already. The enhancement specifies each service that has the bootstrap flag set as true must have a higher priority than all non-bootstrap services.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -227,13 +227,13 @@ func (nc *nodeConfig) Configure() error {
 
 		// Configuring cloud node manager for Azure platform with CCM support
 		if ownedByCCM && nc.platformType == configv1.AzurePlatformType {
```
saifshaikh48, line 229: Should this taint removal `if` block be moved after the `nc.configureWICD()`? This will ensure the taint is removed only after the CCM service has successfully started.

mansikulkarni96:
LGTM

pkg/services/services.go
```diff
@@ -46,11 +46,33 @@ func GenerateManifest(kubeletArgsFromIgnition map[string]string, vxlanPort strin
...
+	if platform == config.AzurePlatformType && ccmEnabled {
+		*services = append(*services, azureCloudNodeManagerConfiguration())
+	}
 	// TODO: All payload filenames and checksums must be added here https://issues.redhat.com/browse/WINC-847
 	files := &[]servicescm.FileInfo{}
 	return servicescm.NewData(services, files)
 }
 
+// azureCloudNodeManagerConfiguration returns the service specification for azure-cloud-node-manager.exe
+func azureCloudNodeManagerConfiguration() servicescm.Service {
```
jrvaldes, line 58: Does it make sense to generalize this, so that it can be extended as more platforms use CCM? I'm thinking of something like:
```go
func getCloudNodeManagerConfiguration(serviceName, servicePath) servicescm.Service {
   ...
}
```
where the `windows.AzureCloudNodeManagerServiceName` and  `windows.AzureCloudNodeManagerPath` are the parameter values in this case.

saifshaikh48:
LGTM

pkg/services/services.go
sebsoto: Each platform will handle its integration with CCM differently,

mansikulkarni96:
LGTM

pkg/services/services.go
mansikulkarni96: A bit confused about this. Are we going to do this for other platforms then?

controllers/configmap_controller.go
```diff
@@ -100,8 +101,16 @@ func NewConfigMapReconciler(mgr manager.Manager, clusterConfig cluster.Config, w
...
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to create openshift client-go client")
+	}
+	ccmEnabled, err := cluster.IsCloudControllerOwnedByCCM(oc)
+	if err != nil {
+		return nil, errors.Wrap(err, "error determining if CCM owns cloud controller")
+	}
 	svcData, err := services.GenerateManifest(argsFromIgnition, clusterConfig.Network().VXLANPort(),
-		clusterConfig.Platform(), ctrl.Log.V(1).Enabled())
+		clusterConfig.Platform(), ccmEnabled, ctrl.Log.V(1).Enabled())
```
aravindhp, line 113: Nit: I think I had mentioned this before but this parameter list is growing and it all has to do with cluster configuration, so a struct encapsulating that will be more succinct. But I am not blocking on that change.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -267,6 +250,22 @@ func (nc *nodeConfig) Configure() error {
 		if err := nc.applyLabelsAndAnnotations(nil, map[string]string{metadata.VersionAnnotation: version.Get()}); err != nil {
 			return errors.Wrapf(err, "error updating version annotation on node %s", nc.node.GetName())
 		}
+		// If we deploy on Azure with CCM support, we have to explicitly remove the cloud taint, because cloud node
+		// manager running on the node can't do it itself, due to lack of RBAC permissions given by the node
+		// kubeconfig it uses.
+		if ownedByCCM && nc.platformType == configv1.AzurePlatformType {
```
aravindhp, line 256: - Is this how it is done on the Linux side?
- Shouldn't this be done before applying `nc.applyLabelsAndAnnotations` because at that point we still have not finished configuring and this is now WMCO's responsibility?

aravindhp (approved):

pkg/services/services.go
sebsoto: No, we will not

pkg/nodeconfig/nodeconfig.go
aravindhp: >On linux cloud-node-manager removes the taint itself, which it can do as it runs in a pod.

My concern is we could be removing the taint while the `cloud-node-manager` has a failure and we really shouldn't be removing the taint. IOW, we seem to be removing the taint without any feedback from the `cloud-node-manager`.

saifshaikh48 (commented):
/lgtm

pkg/nodeconfig/nodeconfig.go
sebsoto: That definitely could be an issue. That behavior still exists before this PR, in that as long as the service starts successfully, we have no idea if it is doing the correct thing before we remove the annotation. The only real solution to this is to run it as a pod, as it is meant to be.

alinaryan:
LGTM

PR #1198, opened 2022-08-22
[windows] Increase kubelet verbosity in debug

jrvaldes:
This change increases the kubelet log verbosity according
to the configured logging level.

PR #1279, opened 2022-10-13
WINC-732: Nodes are bootstrapped with WICD bootstrap command

saifshaikh48:
- Extract kubelet bootstrap files from ignition & copy to Windows instances
- Bootstrap nodes using WICD
- Drop all reference to WMCB


config/rbac/role.yaml
```diff
@@ -177,6 +177,7 @@ rules:
   - machineconfigs
   verbs:
   - list
+  - watch
```
saifshaikh48, line 180: I am seeing this again

> I was getting errors in the WMCO log around failing to get the machine config resource because it was unable to watch it. I didn't really understand why I was getting the error honestly, but giving watch permissions "fixed" it.

saifshaikh48 (commented):
Will likely have to rebase if https://github.com/openshift/windows-machine-config-operator/pull/1273 merges first. Both are held up by https://github.com/openshift/windows-machine-config-operator/pull/1260/

config/rbac/role.yaml
aravindhp, line 180: Why do you need this?

pkg/ignition/ignition.go
```diff
@@ -132,3 +166,178 @@ func parseKubeletArgs(unitContents string) (map[string]string, error) {
...
+	for _, ignFile := range ign.config.Storage.Files {
+		if ignFile.Path == cloudConfigPath {
+			files = append(files, cloudConfigPath)
+		}
+	}
+	return files
+}
+
+// CreateBootstrapFiles creates all prerequisite files required to start kubelet in the WMCO payload. Returns file paths
+func CreateBootstrapFiles(ign *Ignition, c client.Client, apiServerURL, clusterServiceCIDR string) ([]string, error) {
```
aravindhp, line 182: Shouldn't this be part of the `Ignition` struct?

pkg/windows/windows.go
```diff
@@ -136,15 +139,18 @@ var (
...
 		ContainerdDir,
-		ContainerdLogDir}
+		ContainerdLogDir,
+		podManifestDirectory,
+	}
 )
 
 // getFilesToTransfer returns the properly populated filesToTransfer map
-func getFilesToTransfer() (map[*payload.FileInfo]string, error) {
+func (vm *windows) getFilesToTransfer() (map[*payload.FileInfo]string, error) {
```
aravindhp, line 157: Why is this part of the `windows` struct? Seems to be an independent helper.

aravindhp (changes requested):
Nice work, @saifshaikh48. Just some comments around object structuring.

pkg/ignition/ignition.go
```diff
@@ -27,16 +41,37 @@ const (
...
+	// ClientCAFile specifies location to client certificate
+	ClientCAFile string
+	// ClusterDNS is the IP address of the DNS server used for all containers
+	ClusterDNS string
+}
+
 // New returns a new instance of Ignition
-func New(c client.Client) (*Ignition, error) {
-	log := ctrl.Log.WithName("ignition")
+func New(c client.Client, apiServerURL, clusterServiceCIDR string) (*Ignition, error) {
```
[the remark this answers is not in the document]
saifshaikh48: Leftover from a refactor. Will remove them, thanks

config/rbac/role.yaml
saifshaikh48: Not sure still https://github.com/openshift/windows-machine-config-operator/pull/1279#discussion_r994898616

pkg/windows/windows.go
saifshaikh48: Bootstrap files field was added to the `windows` struct and are accessed within this method

pkg/ignition/ignition.go
saifshaikh48: I wasn't sure since some of the files being created (kubelet config file and bootstrap kubeconfig) don't rely on ignition. I'll reconsider

config/rbac/role.yaml
aravindhp: I thought @sebsoto figured this out,

pkg/windows/windows.go
```diff
@@ -240,11 +251,13 @@ type windows struct {
 	defaultShellPowerShell bool
 	// platformType overrides default hostname in bootstrapper
 	platformType string
+	// bootstrapFiles contains the files to be transferred to the VM, required to start the kubelet
+	bootstrapFiles []string
```
aravindhp, line 256: Not sure why this has been added. Payload information should not be in here.

pkg/windows/windows.go
aravindhp: I am not sure why that is needed.

pkg/windows/windows.go
```diff
@@ -165,6 +172,13 @@ func getFilesToTransfer() (map[*payload.FileInfo]string, error) {
 		payload.HcsshimPath:                ContainerdDir,
 		payload.ContainerdConfPath:         ContainerdDir,
 		payload.NetworkConfigurationScript: remoteDir,
+		payload.BootstrapKubeconfigPath:    K8sDir,
+		payload.KubeletConfigPath:          K8sDir,
+		payload.KubeletCACertPath:          K8sDir,
+	}
+	// The cloud config file is only present for Azure and hence we cannot assume it requires translation across clouds.
+	if platform == configv1.AzurePlatformType {
+		srcDestPairs[payload.CloudConfigPath] = K8sDir
```
saifshaikh48, line 183: Is there a better way to check for this? Open to ideas

pkg/ignition/ignition.go
```diff
@@ -132,3 +166,178 @@ func parseKubeletArgs(unitContents string) (map[string]string, error) {
...
+	kubeletConfFile, err := os.Create(payload.KubeletConfigPath)
+	defer kubeletConfFile.Close()
+	if err != nil {
+		return "", fmt.Errorf("error creating %s: %v", payload.KubeletConfigPath, err)
+	}
+	log.V(1).Info("created", "file", payload.KubeletConfigPath)
+	err = kubeletConfTmpl.Execute(kubeletConfFile, variableFields)
+	if err != nil {
+		return "", fmt.Errorf("error writing data to %v file: %v", payload.KubeletConfigPath, err)
+	}
```
sebsoto, line 341: Can get rid of the os.Create when you switch to marshalling and os.Writefile

pkg/ignition/ignition.go
```diff
@@ -132,3 +166,178 @@ func parseKubeletArgs(unitContents string) (map[string]string, error) {
 	}
 	return kubeletArgs, nil
 }
+
+// GetBootstrapFiles returns the paths of all the files needed to bootstrap kubelet
+func (ign *Ignition) GetBootstrapFiles() []string {
```
sebsoto, line 171: I dont see this function being used in this commit, please hold off on adding it until it is used

pkg/ignition/ignition.go
```diff
@@ -132,3 +166,178 @@ func parseKubeletArgs(unitContents string) (map[string]string, error) {
...
+
+// CreateBootstrapFiles creates all prerequisite files required to start kubelet in the WMCO payload. Returns file paths
+func (ign *Ignition) CreateBootstrapFiles(c client.Client, apiServerURL, clusterServiceCIDR string) ([]string, error) {
+	// Ensure generated dir exists
+	if _, err := os.Stat(payload.GeneratedDir); errors.Is(err, os.ErrNotExist) {
+		err := os.Mkdir(payload.GeneratedDir, os.ModePerm)
+		if err != nil {
+			return nil, err
+		}
+	}
```
sebsoto, line 189: This shouldn't be within the responsibility of the ignition package.

pkg/ignition/ignition.go
```diff
@@ -132,3 +166,178 @@ func parseKubeletArgs(unitContents string) (map[string]string, error) {
...
+	bootstrapFiles = append(bootstrapFiles, bootstrapKubeconfigPath)
+
+	clusterDNS, err := cluster.GetDNS(clusterServiceCIDR)
+	if err != nil {
+		return nil, err
+	}
+	initKubeletConfigPath, err := createKubeletConf(clusterDNS)
+	if err != nil {
+		return nil, err
+	}
```
sebsoto, line 209: I'm a bit concerned about this direction, the files within the payload should be static during the lifetime of the operator, but the files within this function, especially the bootstrap kubeconfig need to be current, using the secret as it is at the time of node bootstrap.

Is this an effort to save cycles? unless theres something I'm missing here, I've brought my concerns about this before. This is a change in behavior that has the potential to prevent nodes from being bootstrapped.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -120,12 +121,12 @@ func discoverKubeAPIServerEndpoint() (string, error) {
 
 // NewNodeConfig creates a new instance of nodeConfig to be used by the caller.
 // hostName having a value will result in the VM's hostname being changed to the given value.
-func NewNodeConfig(clientset *kubernetes.Clientset, clusterServiceCIDR, vxlanPort string,
+func NewNodeConfig(c client.Client, clientset *kubernetes.Clientset, clusterServiceCIDR, vxlanPort string,
```
sebsoto, line 124: This is being introduced and not used in this commit

saifshaikh48: Being used now

pkg/ignition/ignition.go
```diff
@@ -132,3 +167,186 @@ func parseKubeletArgs(unitContents string) (map[string]string, error) {
 	}
 	return kubeletArgs, nil
 }
+
+// GetBootstrapFiles returns the paths of all the files needed to bootstrap kubelet
+func GetBootstrapFiles(platformType config.PlatformType) []string {
+	files := []string{payload.BootstrapKubeconfigPath, payload.KubeletConfigPath, payload.KubeletCACertPath}
+	// The cloud config file is only present for Azure and hence we cannot assume it requires translation across clouds.
+	if platformType == config.AzurePlatformType {
```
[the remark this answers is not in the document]
saifshaikh48: Newest changes avoid platform dependent logic

pkg/ignition/ignition.go
saifshaikh48: Removed it altogether

alinaryan (commented):
Great work!! I just had one question:

pkg/ignition/ignition.go
```diff
@@ -0,0 +1,134 @@
...
+		return nil, errors.Errorf("ignition missing kubelet systemd unit file")
+	}
+	argsFromIgnition, err := parseKubeletArgs(*kubeletUnit.Contents)
+	if err != nil {
+		return nil, errors.Wrap(err, "error parsing kubelet systemd unit args")
+	}
+	return argsFromIgnition, nil
+}
+
+// getLatestRenderedWorker returns the most recently created rendered worker MachineConfig
```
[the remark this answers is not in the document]
saifshaikh48: We create some of the bootstrap files from the contents of the rendered-worker MachineConfig (ignition). We want to always grab the latest one so that if something changes in the ignition spec, we can pick these changes up when configuring next

pkg/services/services.go
```diff
@@ -168,6 +168,7 @@ func generateKubeletArgs(argsFromIgnition map[string]string, debug bool) ([]stri
 		"--container-runtime=remote",
 		"--container-runtime-endpoint=" + containerdEndpointValue,
 		"--resolv-conf=",
+		"--enforce-node-allocatable=",
```
saifshaikh48, line 171: We have to explicitly set this to empty in order to run kubelet on Windows. 

This is need as a command line flag instead of in the `KubeletConfiguration` object because JSON marshalling ignores the argument if the value is empty, leaving this option out of the config file. Leaving it out of the config causes kubelet to assign a default value at runtime (`[pods]`), resulting in the following error.
```
"command failed" err="failed to validate kubelet configuration, 
error: [
invalid configuration: enforceNodeAllocatable (--enforce-node-allocatable) is not supported unless cgroupsPerQOS (--cgroups-per-qos) is set to true, 
invalid configuration: EnforceNodeAllocatable (--enforce-node-allocatable) [pods] is not supported on Windows
],
path: &TypeMeta{Kind:,APIVersion:,}"
```

pkg/ignition/ignition.go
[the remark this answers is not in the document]
aravindhp: I suggest capturing the "why` in the comment also.

pkg/services/services.go
jrvaldes: The `--enforce-node-allocatable` flag is [depreacted](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/#options), I suggest to leverage the [KubeletConfiguration object](https://github.com/openshift/windows-machine-config-operator/pull/1279/files#diff-d5ff1964cc140f787c316cc3b809cd4bf63566860d34d4716fb24f1b171d30b7R411)  proposed in https://github.com/openshift/windows-machine-config-operator/pull/1279

pkg/nodeconfig/nodeconfig.go
```diff
@@ -288,6 +304,186 @@ func (nc *nodeConfig) Configure() error {
...
+			Anonymous: kubeletconfig.KubeletAnonymousAuthentication{
+				Enabled: &falseBool,
+			},
+		},
+		ClusterDomain:         "cluster.local",
+		ClusterDNS:            []string{clusterDNS},
+		CgroupsPerQOS:         &falseBool,
+		RuntimeRequestTimeout: meta.Duration{Duration: 10 * time.Minute},
+		MaxPods:               250,
+		KubeAPIQPS:            &kubeAPIQPS,
```
sebsoto, line 431: Theres a lot of magic numbers used in this config. If the answer for why these numbers are being used is, thats what is done for linux workers, thats fine, just comment that at the start of the function.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -288,6 +304,186 @@ func (nc *nodeConfig) Configure() error {
...
+}
+
+// createKubeletConf creates config file for kubelet, with Windows specific configuration.
+func (nc *nodeConfig) createKubeletConf(clusterDNS string) error {
+	kubeletConfig := generateKubeletConfiguration(clusterDNS)
+	kubeletConfigData, err := json.Marshal(kubeletConfig)
+	if err != nil {
+		return err
+	}
+	kubletConfigFileName := "kubelet.conf"
```
sebsoto, line 399: Please define this in the `Windows` package with the rest of the on-node filepaths

pkg/nodeconfig/nodeconfig.go
```diff
@@ -288,6 +304,186 @@ func (nc *nodeConfig) Configure() error {
...
+	if err := nc.Windows.EnsureFileContent(kubeletConfigData, kubletConfigFileName, windows.K8sDir); err != nil {
+		return err
+	}
```
sebsoto, line 402: With the full path defined in the windows package you should use filepath.Base() and filepath.Dir() for the arguments to this function

pkg/nodeconfig/nodeconfig.go
sebsoto: Oh it looks like it already is there

pkg/nodeconfig/nodeconfig.go
```diff
@@ -288,6 +304,186 @@ func (nc *nodeConfig) Configure() error {
...
+	kubeconfigData, err := json.Marshal(kubeconfig)
+	if err != nil {
+		return err
+	}
+	kubeconfigFileName := "bootstrap-kubeconfig"
+	if err := nc.Windows.EnsureFileContent(kubeconfigData, kubeconfigFileName, windows.K8sDir); err != nil {
+		return err
+	}
```
sebsoto, line 350: Please use filepath.Base and filepath.Dir to get the arguments for EnsureFileContent instead of `kubeconfigFileName`

pkg/nodeconfig/nodeconfig.go
```diff
@@ -288,6 +304,186 @@ func (nc *nodeConfig) Configure() error {
...
+	return nil
+}
+
+// createFilesFromIgnition creates any files it can from ignition: kubelet CA cert, cloud-config file
+func (nc *nodeConfig) createFilesFromIgnition() error {
+	// For each new file in the ignition file check if is a file we are interested in, if so, decode it
+	// and write the contents to a temporary destination path
+	filesToTranslate := map[string]struct{}{
```
sebsoto, line 295: I suggest renaming this to `filesToTransfer`

pkg/nodeconfig/nodeconfig.go
```diff
@@ -288,6 +304,186 @@ func (nc *nodeConfig) Configure() error {
...
+	if _, ok := kubeletArgs[ignition.CloudConfigOption]; ok {
+		filesToTranslate[ignition.CloudConfigPath] = struct{}{}
+	}
+
+	for _, ignFile := range ign.Config.Storage.Files {
+		if _, ok := filesToTranslate[ignFile.Node.Path]; ok {
+			if ignFile.Contents.Source == nil {
+				return errors.Errorf("could not process %s: File is empty", ignFile.Node.Path)
+			}
+			contents, err := dataurl.DecodeString(*ignFile.Contents.Source)
```
sebsoto, line 316: It looks like this library is being used within the coreos code so i guess its okay... but it feels not great using a random library like this.

pkg/services/services.go
saifshaikh48: `[none]` doesn't work for Windows. The value we're looking for is `[]` and the only way to set that is through the command line arg or explicitly appending the string `"enforceNodeAllocatable":[]` to the kubelet config file

pkg/ignition/ignition.go
```diff
@@ -27,11 +27,15 @@ const (
...
+	// KubeletCACertPath is the path to the kubelet's client CA certificate as defined in ignition
+	KubeletCACertPath = "/etc/kubernetes/kubelet-ca.crt"
+	// CloudConfigPath is the path to the cloud config file as defined in ignition
+	CloudConfigPath = "/etc/kubernetes/cloud.conf"
 )
 
 // Ignition is a representation of an Ignition file
 type Ignition struct {
-	config ignCfgTypes.Config
+	Config ignCfgTypes.Config
```
sebsoto, line 38: Instead of making this public, please add a method such as `GetFiles` or `GetFileList`

pkg/windows/windows.go
```diff
@@ -456,7 +464,7 @@ func (vm *windows) Deconfigure() error {
 	return nil
 }
 
-func (vm *windows) Configure() error {
+func (vm *windows) Prepare() error {
```
sebsoto, line 455: I don't see the reason to separate this into `Prepare` and `Bootstrap`. Is there ever a time we'd need to prepare a node and not bootstrap it?

This change makes the Windows interface more clunky if that is not the case

pkg/nodeconfig/nodeconfig.go
```diff
@@ -664,13 +664,34 @@ func (nc *nodeConfig) UpdateKubeletClientCA(contents []byte) error {
...
 func (nc *nodeConfig) configureWICD() error {
+	saCA, saToken, err := getWICDCertAndToken(nc.k8sclientset)
+	if err != nil {
+		return err
+	}
+	return nc.Windows.ConfigureWICD(nodeConfigCache.apiServerEndpoint, saCA, saToken)
+}
+
+// getWICDTokenAndCert returns the token and CA cert associated with the WICD service account
+func getWICDCertAndToken(clientset *kubernetes.Clientset) ([]byte, []byte, error) {
```
sebsoto, line 653: This can be made a method of nodeConfig.

Can the cached client be used for this now that you've added that to the nodeConfig struct?

pkg/nodeconfig/nodeconfig.go
```diff
@@ -664,13 +664,34 @@ func (nc *nodeConfig) UpdateKubeletClientCA(contents []byte) error {
 	return nil
 }
 
+// bootstrap starts all required services to bootstrap a node object
+func (nc *nodeConfig) bootstrap() error {
+	if err := nc.Windows.Prepare(); err != nil {
+		return err
+	}
+	saCA, saToken, err := getWICDCertAndToken(nc.k8sclientset)
```
sebsoto, line 636: Please create a struct for the `saCA, saToken` as you are passing this pair around a lot.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -45,7 +44,7 @@ const (
 	HybridOverlaySubnet = "k8s.ovn.org/hybrid-overlay-node-subnet"
 	// HybridOverlayMac is an annotation applied by the hybrid-overlay
 	HybridOverlayMac = "k8s.ovn.org/hybrid-overlay-distributed-router-gateway-mac"
-	// WindowsOSLabel is the label that is applied by WMCB to identify the Windows nodes bootstrapped via WMCB
+	// WindowsOSLabel is the label applied when kubelet is ran to identify Windows nodes that have been bootstrapped
```
sebsoto, line 47: Can shorten to:
```
 	// WindowsOSLabel is the label applied when kubelet is ran to identify Windows nodes
 ```

**Review by sebsoto — CHANGES_REQUESTED** [review]
Great work. PTAL at my comments

File: pkg/services/services.go (hunk 14 above)
Comment (jrvaldes) [reply pkg/services/services.go:171]: I follow.

File: pkg/nodeconfig/nodeconfig.go (hunk 18 above)
Comment (saifshaikh48) [reply pkg/nodeconfig/nodeconfig.go:350]: Having trouble with this suggestion. Since WMCO runs as a linux process, `filepath` is not handling Windows paths properly.
```golang
func test() {
	path := "C:\\k\\bootstrap-kubeconfig"
	fileName := filepath.Base(path)
	dir := filepath.Dir(path)

	fmt.Println("Dir:", dir)
	fmt.Println("File:", fileName)
}
```
Output: 
```
Dir: .
File: C:\k\bootstrap-kubeconfig
```

pkg/nodeconfig/nodeconfig.go
saifshaikh48: Sure, I took a look at the `kubelet.conf` on a linux worker and confirmed this

pkg/windows/windows.go
saifshaikh48: > Is there ever a time we'd need to prepare a node and not bootstrap it?

Not that I can think of. You're right, I consolidated them

jrvaldes:
/hold

- https://github.com/openshift/windows-machine-config-operator/pull/1198 needs to merge before we move the bootstrapping completely to WICD (this PR)  cc @saifshaikh48

config/rbac/role.yaml
aravindhp: @saifshaikh48 please add the above to the commit message. If possible please add a comment here too.

pkg/nodeconfig/nodeconfig.go
aravindhp: +1. 

>It looks like this library is being used within the coreos code

That doesn't automatically make it a good pattern. I suggest looking for an alternative if available.

pkg/nodeconfig/nodeconfig.go
aravindhp: Would [FromSlash](https://cs.opensource.google/go/go/+/go1.19.2:src/path/filepath/path.go;l=188) [¶](https://pkg.go.dev/path/filepath#FromSlash) or [ToSlash](https://cs.opensource.google/go/go/+/go1.19.2:src/path/filepath/path.go;l=178) [¶](https://pkg.go.dev/path/filepath#ToSlash) help?

pkg/nodeconfig/nodeconfig.go
```diff
@@ -288,6 +304,185 @@ func (nc *nodeConfig) Configure() error {
...
+	kubletConfigFileName := "kubelet.conf"
+	if err := nc.Windows.EnsureFileContent(kubeletConfigData, kubletConfigFileName, windows.K8sDir); err != nil {
+		return err
+	}
+	nc.log.V(1).Info("created", "file", windows.KubeletConfigPath)
+	return nil
+}
+
+// generateKubeletConfiguration returns the configuration spec for the kubelet Windows service
+func generateKubeletConfiguration(clusterDNS string) kubeletconfig.KubeletConfiguration {
```
aravindhp, line 418: Please try bunch helpers and structs separately.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -288,6 +304,185 @@ func (nc *nodeConfig) Configure() error {
...
+// createKubeletConf creates config file for kubelet, with Windows specific configuration.
+func (nc *nodeConfig) createKubeletConf(clusterDNS string) error {
+	kubeletConfig := generateKubeletConfiguration(clusterDNS)
+	kubeletConfigData, err := json.Marshal(kubeletConfig)
+	if err != nil {
+		return err
+	}
+	kubletConfigFileName := "kubelet.conf"
+	if err := nc.Windows.EnsureFileContent(kubeletConfigData, kubletConfigFileName, windows.K8sDir); err != nil {
+		return err
```
aravindhp, line 411: I see this pattern being repeated in  createBootstrapKubeconfig(). 
```golang
   config := generate()
   data := json.Marshal(config)
   nc.Windows.EnsureFileContent(data, filename, dir)
```
Calls for modularization. Try making this an object into which you pass the data you want to marshal and where you want the file to be generated.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -288,6 +304,185 @@ func (nc *nodeConfig) Configure() error {
...
+		Contexts: []clientcmdv1.NamedContext{{
+			Name: "kubelet",
+			Context: clientcmdv1.Context{
+				Cluster:  "local",
+				AuthInfo: "kubelet",
+			},
+		}},
+		CurrentContext: "kubelet",
+	}
+	return &kubeconfig, nil
```
aravindhp, line 399: Here you are using a pointer but not in generateKubeletConfiguration. What is the reasoning either way?

pkg/services/services.go
aravindhp: Given we want to move away from command line args, why don't you want to append to the kubelet config file?

pkg/windows/windows.go
```diff
@@ -474,7 +480,13 @@ func (vm *windows) Configure() error {
 		return errors.Wrapf(err, "error configuring containerd")
 	}
 
-	return vm.runBootstrapper()
+	if err := vm.ensureWICDSecretContent(serviceAccountCA, serviceAccountToken); err != nil {
+		return err
+	}
+	wicdBootstrapCmd := fmt.Sprintf("%s bootstrap --desired-version %s --api-server %s --sa-ca %s%s --sa-token %s%s",
+		wicdPath, desiredVer, apiServerURL, K8sDir, wicdCAFile, K8sDir, wicdTokenFile)
+	_, err := vm.Run(wicdBootstrapCmd, true)
```
aravindhp, line 476: Why is the output being ignored? Shouldn't we display it as a debug? Won't it be useful to include it in the error message?

aravindhp (changes requested):

pkg/nodeconfig/nodeconfig.go
saifshaikh48: The idea here was that this function can return an error, and on error returning `nil` makes more sense than returning an empty/defaulted `Config` struct. In `generateKubeletConfiguration` there's no chance to return error so I decided to forgo the pointer. Not sure if this is valid reasoning :)

pkg/nodeconfig/nodeconfig.go
saifshaikh48: Nope, the separator in the `filepath` library is based on the OS. Since WMCO does not run on Windows, it can't handle Windows paths.

pkg/nodeconfig/nodeconfig.go
aravindhp: Not sure if that is applicable here. Given you are generating something I don't see a need to use a pointer. The contract should be that if there is an error, the returned config is invalid and should not be used.

pkg/nodeconfig/nodeconfig.go
saifshaikh48: I can do some string manipulation here instead. Find last occurrence of `\` and split string at that index to get the dir and fileName

pkg/nodeconfig/nodeconfig.go
```diff
@@ -207,14 +215,18 @@ func (nc *nodeConfig) Configure() error {
 	if err := nc.createBootstrapFiles(); err != nil {
 		return err
 	}
-
-	// Perform the basic kubelet configuration using WMCB
-	if err := nc.Windows.Configure(); err != nil {
+	wicdKey, err := nc.getWICDCredentials()
```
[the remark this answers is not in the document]
saifshaikh48: I moved this to the nodeconfig cache, thanks

pkg/nodeconfig/nodeconfig.go
saifshaikh48: I tried my hand here and I could not see the benefit. Happy to discuss more over a call if needed.

pkg/nodeconfig/nodeconfig.go
saifshaikh48: Moved to using `url.QueryUnescape`, looks like that does the trick

pkg/services/services.go
saifshaikh48: It requires some funky manipulation. Take a look at [the implementation](https://github.com/saifshaikh48/windows-machine-config-operator/blob/bootstrap-using-wicd/pkg/nodeconfig/nodeconfig.go#L395-L400)

config/rbac/role.yaml
saifshaikh48: `make bundle` kicks back any comments in this file. I added the explanation to where to watch is specified in `ignition.go`

pkg/nodeconfig/nodeconfig.go
```diff
@@ -496,6 +616,98 @@ func (nc *nodeConfig) configureWICD() error {
 	return nc.Windows.ConfigureWICD(nodeConfigCache.apiServerEndpoint, saCA, saToken)
 }
 
+// generateKubeconfig creates a kubeconfig spec with the certificate and token data from the given secret
+func generateKubeconfig(secret *core.Secret, apiServerURL string) (clientcmdv1.Config, error) {
+	// extract ca.crt and token data fields
+	caCert := secret.Data[core.ServiceAccountRootCAKey]
+	if caCert == nil {
+		return clientcmdv1.Config{}, errors.New("unable to find CA cert")
```
jrvaldes, line 610: Consider adding more context to the error message, for example:

```go
 errors.errorf("unable to find %s key in secret %s", core.ServiceAccountRootCAKey, secret.GetName())
```

pkg/nodeconfig/nodeconfig.go
```diff
@@ -496,6 +616,98 @@ func (nc *nodeConfig) configureWICD() error {
...
+	}
+	token := secret.Data[core.ServiceAccountTokenKey]
+	if token == nil {
+		return clientcmdv1.Config{}, errors.New("unable to find token")
```
jrvaldes, line 614: ditto.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -288,6 +303,110 @@ func (nc *nodeConfig) Configure() error {
...
+	kubeletConfigData, err := json.Marshal(kubeletConfig)
+	if err != nil {
+		return err
+	}
+	// Appending this option is needed here instead of in the kubelet configuration object. Otherwise, when marshalling,
+	// the empty value will be omitted, so it would end up being incorrectly populated at service start time.
+	enforceNodeAllocatable := "\"enforceNodeAllocatable\":[]}"
+	// Replace last character ('}') with comma
+	kubeletConfigData[len(kubeletConfigData)-1] = ','
+	kubeletConfigData = []byte(string(kubeletConfigData) + enforceNodeAllocatable)
```
jrvaldes, line 415: Consider using "..." as a suffix to append directly and avoid the string conversion.

```go
	kubeconfigData[len(kubeconfigData)-1] = ','
        kubeconfigData = append(kubeconfigData, enforceNodeAllocatable...)

```

pkg/nodeconfig/nodeconfig.go
```diff
@@ -288,6 +303,110 @@ func (nc *nodeConfig) Configure() error {
...
+
+	dir, fileName := splitWindowsPath(windows.KubeletConfigPath)
+	if err := nc.Windows.EnsureFileContent(kubeletConfigData, fileName, dir); err != nil {
+		return err
+	}
+	nc.log.V(1).Info("created", "file", windows.KubeletConfigPath)
+	return nil
+}
+
```
saifshaikh48, line 424: object-oriented version
```suggestion
// createBootstrapKubeconfig creates a kubeconfig for kubelet to initially communicate with the API server
func (nc *nodeConfig) createBootstrapKubeconfig() error {
	nodeBootstrapToken, err := nc.k8sclientset.CoreV1().Secrets(mcoNamespace).Get(context.TODO(), mcoBootstrapSecret,
		meta.GetOptions{})
	if err != nil {
		return err
	}
	kubeconfig, err := generateKubeconfig(nodeBootstrapToken, nodeConfigCache.apiServerEndpoint)
	if err != nil {
		return err
	}
	kubeconfigData, err := json.Marshal(kubeconfig)
	if err != nil {
		return err
	}
	return nc.createFile(kubeconfigData, windows.BootstrapKubeconfig)
}

// createKubeletConf creates config file for kubelet, with Windows specific configuration.
func (nc *nodeConfig) createKubeletConf(clusterDNS string) error {
	kubeletConfig := generateKubeletConfiguration(clusterDNS)
	kubeletConfigData, err := json.Marshal(kubeletConfig)
	if err != nil {
		return err
	}
	// Appending this option is needed here instead of in the kubelet configuration object. Otherwise, when marshalling,
	// the empty value will be omitted, so it would end up being incorrectly populated at service start time.
	enforceNodeAllocatable := "\"enforceNodeAllocatable\":[]}"
	// Replace last character ('}') with comma
	kubeletConfigData[len(kubeletConfigData)-1] = ','
	kubeletConfigData = []byte(string(kubeletConfigData) + enforceNodeAllocatable)
	return nc.createFile(kubeletConfigData, windows.KubeletConfigPath)
}

func (nc *nodeConfig) createFile(jsonData []byte, path string) error {
	writer := JSONFileWriter{
		data:     jsonData,
		path:     windows.KubeletConfigPath,
		instance: nc.Windows,
	}
	if err := writer.write(); err != nil {
		return err
	}
	nc.log.V(1).Info("created", "file", windows.KubeletConfigPath)
	return nil
}

type JSONFileWriter struct {
	// data is the marshalled data to be used as file contents.
	data []byte
	// path is a Windows file path where the data will be written.
	path string
	// instance is the Windows instance to write the file to.
	instance windows.Windows
}

// write writes the data within the JSONFileWriter to the path on the given Windows instance. Creates the file if needed
func (w *JSONFileWriter) write() error {
	dir, fileName := splitWindowsPath(w.path)
	return w.instance.EnsureFileContent(w.data, fileName, dir)
}
```

pkg/services/services.go
jrvaldes: Added a [suggestion](https://github.com/openshift/windows-machine-config-operator/pull/1279#discussion_r1000946730) there. 

Wondering if there is a way to override the `omitempty` for `EnforceNodeAllocatable`  in the [KubeletConfiguration](https://github.com/jrvaldes/windows-machine-config-operator/blob/3bac69ceb49f8f20a0327de58b5c55831e3a7993/vendor/k8s.io/kubelet/config/v1beta1/types.go#L651) definition. Given `[none]` is not supported for Windows, `[]` should be a valid value.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -496,6 +615,98 @@ func (nc *nodeConfig) configureWICD() error {
...
+			"cpu":               "500m",
+			"ephemeral-storage": "1Gi",
+			"memory":            "1Gi",
+		},
+	}
+}
+
+// splitWindowsPath splits a Windows file path into the directory and base file name.
+// Example: 'C:\\k\\bootstrap-kubeconfig' --> dir: 'C:\k\', fileName: 'bootstrap-kubeconfig
+func splitWindowsPath(filepath string) (dir string, fileName string) {
```
jrvaldes, line 689: Consider dropping the named return parameters (`dir` and `fileName`) as they are not in use.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -496,6 +615,98 @@ func (nc *nodeConfig) configureWICD() error {
...
+	if filepath == "" {
```
jrvaldes, line 690: No need for this check, the logic below can handle this scenario.

pkg/nodeconfig/nodeconfig.go
saifshaikh48: decided not to use the OO approach

pkg/nodeconfig/nodeconfig.go
saifshaikh48: I think the names add useful context to the return values

pkg/windows/windows.go
```diff
@@ -233,9 +229,6 @@ type Windows interface {
 
 // windows implements the Windows interface
 type windows struct {
-	// workerIgnitionEndpoint is the Machine Config Server(MCS) endpoint from which we can download the
-	// the OpenShift worker ignition file.
-	workerIgnitionEndpoint string
```
jrvaldes, line 215: Consider mentioning in the commit message why `workerIgnitionEndpoint` is no longer relevant.

jrvaldes (commented):
Great work @saifshaikh48! Left a few comments. PTAL

pkg/nodeconfig/nodeconfig.go
jrvaldes: True, and you can also achieve that by improving the documentation of the function.

build/Dockerfile.ci
```diff
@@ -108,19 +103,13 @@ RUN make build-daemon
...
 #├── windows_exporter.exe
-#├── windows-instance-config-daemon.exe
-#└── wmcb.exe
+#└── windows-instance-config-daemon.exe
 
 FROM registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.19-openshift-4.12
 LABEL stage=operator
 
-# Copy wmcb.exe
-WORKDIR /payload/
```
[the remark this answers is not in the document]
saifshaikh48: Good catch, this would probably have caused complications when run in CI

pkg/nodeconfig/nodeconfig.go
```diff
@@ -129,6 +137,44 @@ func discoverKubeAPIServerEndpoint() (string, error) {
...
+	cfg, err := crclientcfg.GetConfig()
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to get config to talk to kubernetes api server")
+	}
+	clientset, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		return nil, errors.Wrap(err, "error creating kubernetes clientset")
+	}
+
+	secrets, err := clientset.CoreV1().Secrets("openshift-windows-machine-config-operator").List(context.TODO(),
```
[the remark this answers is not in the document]
saifshaikh48: Yep, added a commit so we can reuse this

pkg/nodeconfig/nodeconfig.go
```diff
@@ -84,6 +84,14 @@ type nodeConfig struct {
 	platformType configv1.PlatformType
 }
 
+// authentication holds credential information used to communicate with a kubernetes cluster
+type authentication struct {
```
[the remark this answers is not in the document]
saifshaikh48: That makes sense, I moved it there. Allows for the struct to be used in windows pkg too

pkg/nodeconfig/nodeconfig.go
```diff
@@ -496,6 +615,98 @@ func (nc *nodeConfig) configureWICD() error {
...
+			},
+		}},
+		CurrentContext: "kubelet",
+	}
+	return kubeconfig, nil
+}
+
+// generateKubeletConfiguration returns the configuration spec for the kubelet Windows service
+func generateKubeletConfiguration(clusterDNS string) kubeletconfig.KubeletConfiguration {
+	// default values chosen based on the config on Linux worker nodes
```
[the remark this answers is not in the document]
saifshaikh48: adjusted wording

pkg/nodeconfig/nodeconfig.go
```diff
@@ -129,6 +137,44 @@ func discoverKubeAPIServerEndpoint() (string, error) {
...
+	}
+
+	secrets, err := clientset.CoreV1().Secrets("openshift-windows-machine-config-operator").List(context.TODO(),
+		meta.ListOptions{})
+	if err != nil {
+		return nil, errors.Wrap(err, "error listing secrets")
+	}
+
+	tokenSecretPrefix := "windows-instance-config-daemon-token-"
+	var filteredSecrets []core.Secret
```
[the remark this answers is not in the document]
saifshaikh48: current name seems more accurate

saifshaikh48:
/test vsphere-operator-e2e
/test aws-upgrade-e2e

Rebased to pick up latest commit from https://github.com/openshift/windows-machine-config-operator/pull/1260. Running this subset of tests since azure and gcp are blocked

saifshaikh48:
/test vsphere-e2e-operator
/test aws-e2e-upgrade

Properly removed the wmcb submodule to prevent it from being built.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -39,6 +39,11 @@ import (
 	"github.com/openshift/windows-machine-config-operator/version"
 )
 
+const (
+	// WMCONamespace defines the namespace WMCO-managed resourced are expected to be in
+	WMCONamespace = "openshift-windows-machine-config-operator"
```
sebsoto, line 44: I don't see how this belongs in nodeconfig.
In any case, you should use the watchNamespace value thats present in each controller instead of using a constant.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -288,6 +303,108 @@ func (nc *nodeConfig) Configure() error {
 	return err
 }
 
+// CreateBootstrapFiles creates all prerequisite files on the node required to start kubelet using latest ignition spec
```
jrvaldes, line 316: s/CreateBootstrapFiles/createBootstrapFiles

pkg/nodeconfig/nodeconfig.go
```diff
@@ -134,6 +134,44 @@ func discoverKubeAPIServerEndpoint() (string, error) {
...
+	cfg, err := crclientcfg.GetConfig()
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to get config to talk to kubernetes api server")
+	}
+	clientset, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		return nil, errors.Wrap(err, "error creating kubernetes clientset")
+	}
+
+	secrets, err := clientset.CoreV1().Secrets(WMCONamespace).List(context.TODO(),
```
sebsoto, line 148: This should be a method of nodeconfig and use nc.getNamespace

pkg/nodeconfig/init.go
```diff
@@ -17,6 +19,8 @@ type cache struct {
 	workerIgnitionEndPoint string
 	// apiServerEndpoint is the address which clients can interact with the API server through
 	apiServerEndpoint string
+	// credentials holds a certificate and token needed to interact with the API server
+	credentials *cluster.Authentication
```
sebsoto, line 16: I don't think this fits in here. The other members of this are immutable. The cert can be rotated.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -212,13 +250,14 @@ func (nc *nodeConfig) Configure() error {
 		return err
 	}
 
-	// Perform the basic kubelet configuration using WMCB
-	if err := nc.Windows.Configure(); err != nil {
+	// Start all required services to bootstrap a node object using WICD
+	err := nc.Windows.Bootstrap(version.Get(), nodeConfigCache.apiServerEndpoint, nodeConfigCache.credentials)
+	if err != nil {
 		return errors.Wrap(err, "configuring the Windows VM failed")
```
sebsoto, line 221: s/configuring/bootstrapping ?

pkg/windows/windows.go
```diff
@@ -114,6 +117,10 @@ const (
 	ManagedTag = "OpenShift managed"
 	// containersFeatureName is the name of the Windows feature that is required to be enabled on the Windows instance.
 	containersFeatureName = "Containers"
+	// wicdCAFile is the name of the file that holds the WICD service account CA certificate
+	wicdCAFile = "sa-ca.crt"
+	// wicdTokenFile is the name of the file that holds the WICD service account secret token
+	wicdTokenFile = "sa-token"
```
sebsoto, line 121: Please make these be the full path like the rest of file constants

sebsoto (changes requested):

jrvaldes (commented):
LGTM, just a minor suggestion. Will build an image from this branch and test it with a dev cluster.

saifshaikh48:
/test azure-e2e-operator

kicking off Azure job since cloud.conf should be in use there as well – if I see the same issue here as in vSphere, we'll know what the problem is. AWS doesn't have cloud.conf and kubelet starts fine.

pkg/nodeconfig/nodeconfig.go
saifshaikh48: @aravindhp @sebsoto After some testing, I cannot find an alternative standard library to get clean file contents. There's issues translating cloud.config when using `net/url`.

Using the `vincent-petithory/dataurl` library works for all files embedded in ignition so I lean towards using it. The only other alternative is some hacky string manipulation.

aravindhp: > There's issues translating cloud.config when using net/url

What is the issue? I suggest opening a bug.

>Using the vincent-petithory/dataurl library works for all files embedded in ignition so I lean towards using it.

I am fine with that approach for now.

pkg/nodeconfig/nodeconfig.go
saifshaikh48: This is to used to initialize the nodeconfig cache. Take a look at the [other functions](https://github.com/openshift/windows-machine-config-operator/blob/master/pkg/nodeconfig/nodeconfig.go#L99) we use there.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -288,6 +303,108 @@ func (nc *nodeConfig) Configure() error {
...
+	return nc.write(kubeletConfigData, windows.KubeletConfigPath)
+}
+
+// write outputs the given data to the given path on the underlying Windows instance. Creates the file if needed.
+func (nc *nodeConfig) write(data []byte, path string) error {
+	dir, fileName := splitWindowsPath(path)
+	if err := nc.Windows.EnsureFileContent(data, fileName, dir); err != nil {
+		return err
+	}
+	nc.log.V(1).Info("created", "file", path)
```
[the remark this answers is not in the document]
saifshaikh48: fair enough, a log message is already printed by EnsureFileContent so I'll remove this

pkg/nodeconfig/nodeconfig.go
saifshaikh48: It is not stripping the leading `data:,` tag of the contents which are in [data url scheme](https://www.rfc-editor.org/rfc/rfc2397.html). But it does strip this when parsing kubelet.conf contents. I'll look into opening a bug

saifshaikh48:
/test vsphere-e2e-operator

Did local testing of the fix on an Azure cluster, want to make sure it works for vSphere as well

saifshaikh48:
/test vsphere-e2e-operator

fixed unused import

pkg/nodeconfig/nodeconfig.go
```diff
@@ -129,6 +129,43 @@ func discoverKubeAPIServerEndpoint() (string, error) {
...
+func getWICDCredentials() (*windows.Authentication, error) {
+	cfg, err := crclientcfg.GetConfig()
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to get config to talk to kubernetes api server")
+	}
+	clientset, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		return nil, errors.Wrap(err, "error creating kubernetes clientset")
+	}
+	operatorNamespace := "openshift-windows-machine-config-operator"
```
aravindhp, line 142: Can you hard code this? It is possible that the operator get deployed to another namespace. Isn't there some env var that you can get this from?

pkg/nodeconfig/nodeconfig.go
```diff
@@ -129,6 +129,43 @@ func discoverKubeAPIServerEndpoint() (string, error) {
...
+	for _, secret := range secrets.Items {
+		if strings.HasPrefix(secret.Name, tokenSecretPrefix) {
+			filteredSecrets = append(filteredSecrets, secret)
+		}
+	}
+	if len(filteredSecrets) != 1 {
+		return nil, fmt.Errorf("expected 1 secret with '%s' prefix, found %d", tokenSecretPrefix, len(filteredSecrets))
+	}
+	caCert := filteredSecrets[0].Data[core.ServiceAccountRootCAKey]
+	if len(caCert) == 0 {
```
aravindhp, line 159: What if `core.ServiceAccountRootCAKey` is not present?

aravindhp (commented):
Mostly lgtm. I will let @sebsoto approve.

pkg/nodeconfig/nodeconfig.go
saifshaikh48: Then caCert would be initialized to an empty byte array (len == 0) so this if statement would catch it.

pkg/nodeconfig/nodeconfig.go
saifshaikh48: > It is possible that the operator get deployed to another namespace

Valid point. I'm leaning towards adding the namespace to the nodeconfig struct.

I'm also getting less convinced the WICD service account credentials should be part of the nodeconfig cache, given they can change throughout the life of the cluster. I'll make this function upon the nodeconfig struct instead.

aravindhp: >I'm also getting less convinced the WICD service account credentials should be part of the nodeconfig cache, given they can change throughout the life of the cluster.

While it can change through the life of a cluster,  but can it change without restarting or re-deploying the operator?

pkg/nodeconfig/nodeconfig.go
aravindhp: It would help debugging if we can differentiate between the two cases.

pkg/nodeconfig/nodeconfig.go
saifshaikh48: I don't think so unless a cluster admin edits somehow

aravindhp: So then why can't it be part of the nodeconfig cache?

pkg/nodeconfig/nodeconfig.go
```diff
@@ -288,6 +304,106 @@ func (nc *nodeConfig) Configure() error {
...
+	}
+	// Replace last character ('}') with comma
+	kubeletConfigData[len(kubeletConfigData)-1] = ','
+	// Appending this option is needed here instead of in the kubelet configuration object. Otherwise, when marshalling,
+	// the empty value will be omitted, so it would end up being incorrectly populated at service start time.
+	// Can be moved to kubelet configuration object with https://issues.redhat.com/browse/WINC-926
+	enforceNodeAllocatable := []byte("\"enforceNodeAllocatable\":[]}")
+	kubeletConfigData = append(kubeletConfigData, enforceNodeAllocatable...)
+
+	return nc.write(kubeletConfigData, windows.KubeletConfigPath)
```
jrvaldes, line 403: I'm thinking of returning the `kubeletConfigData` here and leaving the responsibility of writing the file to the `createBootstrapFiles` function. It may also help to unit test this function.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -496,6 +612,97 @@ func (nc *nodeConfig) configureWICD() error {
...
+				AuthInfo: "kubelet",
+			},
+		}},
+		CurrentContext: "kubelet",
+	}
+	return kubeconfig, nil
+}
+
+// generateKubeletConfiguration returns the configuration spec for the kubelet Windows service
+func generateKubeletConfiguration(clusterDNS string) kubeletconfig.KubeletConfiguration {
```
jrvaldes, line 635: Consider adding a unit test for this function.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -496,6 +612,97 @@ func (nc *nodeConfig) configureWICD() error {
 	return nc.Windows.ConfigureWICD(nodeConfigCache.apiServerEndpoint, saCA, saToken)
 }
 
+// generateKubeconfig creates a kubeconfig spec with the certificate and token data from the given secret
+func generateKubeconfig(secret *core.Secret, apiServerURL string) (clientcmdv1.Config, error) {
```
jrvaldes, line 596: Consider adding a unit test for this function.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -496,6 +612,97 @@ func (nc *nodeConfig) configureWICD() error {
...
+			"memory":            "1Gi",
+		},
+	}
+}
+
+// splitWindowsPath splits a Windows file path into the directory and base file name.
+// Example: 'C:\\k\\bootstrap-kubeconfig' --> dir: 'C:\\k\\', fileName: 'bootstrap-kubeconfig'
+func splitWindowsPath(filepath string) (dir string, fileName string) {
+	splitIndex := strings.LastIndexByte(filepath, '\\') + 1
+	return filepath[:splitIndex], filepath[splitIndex:]
```
jrvaldes: Would be nice to cover this string-handling logic in a unit test.

pkg/windows/windows.go
```diff
@@ -1055,3 +1081,10 @@ func rmDirCmd(dirName string) string {
 func getHNSNetworkCmd(networkName string) string {
 	return "Get-HnsNetwork | where { $_.Name -eq '" + networkName + "'}"
 }
+
+// SplitPath splits a Windows file path into the directory and base file name.
+// Example: 'C:\\k\\bootstrap-kubeconfig' --> dir: 'C:\\k\\', fileName: 'bootstrap-kubeconfig'
+func SplitPath(filepath string) (dir string, fileName string) {
```
jrvaldes, line 1026: Looks like this logic is duplicated in [splitWindowsPath](https://github.com/openshift/windows-machine-config-operator/pull/1279/commits/895a426f5807419fc181bf04d7996a22ea185166#diff-d5ff1964cc140f787c316cc3b809cd4bf63566860d34d4716fb24f1b171d30b7R701). Could this be modularized?

pkg/windows/windows.go
```diff
@@ -474,7 +488,16 @@ func (vm *windows) Configure() error {
 		return errors.Wrapf(err, "error configuring containerd")
 	}
 
-	return vm.runBootstrapper()
+	if err := vm.ensureWICDSecretContent(credentials); err != nil {
+		return err
+	}
+	wicdBootstrapCmd := fmt.Sprintf("%s bootstrap --desired-version %s --api-server %s --sa-ca %s --sa-token %s",
+		wicdPath, desiredVer, apiServerURL, wicdCAFile, wicdTokenFile)
```
jrvaldes, line 483: Could `desiredVer` and/or `apiServerURL` be empty? Is it safe passing an empty value, i.e. `--desired-version ` ?

pkg/windows/windows.go
saifshaikh48: Yep, splitWindowsPath was supposed to be removed in favor of SplitPath (bad rebase). I'll fix next time I push.

pkg/nodeconfig/nodeconfig.go
saifshaikh48: I'm not convinced on the unit testing of this or the kubeconfig. The test for this would be reconstructing the `KubeletConfiguration` object with the same values as in this function, then asserting they are equal. Don't see much benefit

pkg/nodeconfig/nodeconfig.go
saifshaikh48: I explored this option and decided against it. I'd like all the functions called from `createBootstrapFiles` to do the same thing -- if `createFilesFromIgnition` is writing data to a file, then `createBootstrapKubeconfig` and `createKubeletConf` should as well.

saifshaikh48:
/unhold

https://github.com/openshift/windows-machine-config-operator/pull/1198 merged
cc: @jrvaldes

jrvaldes:
> /unhold
> 
> #1198 merged cc: @jrvaldes

@saifshaikh48, I think you should hold this PR until [WINC-738: Define Kubelet service in services ConfigMap](https://github.com/openshift/windows-machine-config-operator/pull/1260#top)  merges.

saifshaikh48:
/hold

Fair, looks like it (https://github.com/openshift/windows-machine-config-operator/pull/1260) will merge soon enough but I'll throw a hold just in case.

saifshaikh48:
/unhold 

https://github.com/openshift/windows-machine-config-operator/pull/1260 has merged

pkg/nodeconfig/nodeconfig.go
aravindhp: We are adding a new component. If you find an existing pattern that does not look good, please change it though changing `createBootstrapKubeconfig` and `createKubeletConf` should be done in a separate PR or a separate commit at least.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -496,6 +612,90 @@ func (nc *nodeConfig) configureWICD() error {
...
+		}},
+		CurrentContext: "kubelet",
+	}
+	return kubeconfig, nil
+}
+
+// generateKubeletConfiguration returns the configuration spec for the kubelet Windows service
+func generateKubeletConfiguration(clusterDNS string) kubeletconfig.KubeletConfiguration {
+	// default numeric values chosen based on the OpenShift kubelet config recommendations for Linux worker nodes
+	falseBool := false
```
mansikulkarni96, line 637: Is there a need for a false variable?

pkg/nodeconfig/nodeconfig.go
saifshaikh48: I weighed the two patterns and decided which was better...
However, if the team thinks `leaving the responsibility of writing the file to the createBootstrapFiles function` than the approach here, please take a look at the code in the next commit (0e2799bc0ce7bd2f7386d2d36779f4bb1eb2c1d4) and let me know.

pkg/nodeconfig/nodeconfig.go
saifshaikh48: Yes, because the kubelet conf object expects a bool pointer and there's no way to initialize one inline.

pkg/nodeconfig/nodeconfig.go
jrvaldes: LGTM

pkg/nodeconfig/nodeconfig.go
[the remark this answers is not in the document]
saifshaikh48: Took the approach to read namespace from environment var since we can't always assume operator is deployed in `openshift-windows-machine-config-operator` namespace

pkg/nodeconfig/nodeconfig.go
saifshaikh48: I moved it back as part of the nodeconfig cache after this discussion

pkg/nodeconfig/nodeconfig.go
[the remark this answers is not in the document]
saifshaikh48: I ended up moving this to the windows pkg due to https://github.com/openshift/windows-machine-config-operator/pull/1279#discussion_r1006037821

pkg/nodeconfig/nodeconfig.go
saifshaikh48: Unit test added after further discussion

pkg/nodeconfig/nodeconfig.go
```diff
@@ -130,6 +131,48 @@ func discoverKubeAPIServerEndpoint() (string, error) {
...
+	cfg, err := crclientcfg.GetConfig()
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to get config to talk to kubernetes api server")
+	}
+	client, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to get client from the given config")
+	}
+
+	operatorNamespaceVar := "WATCH_NAMESPACE"
```
aravindhp, line 144: Isn't there an `OPERATOR_NAMESPACE` variable also? If yes, wouldn't that be more appropriate?

saifshaikh48: I don't think `OPERATOR_NAMESPACE` gets set, closest thing I see is `OPERATOR_NAME`.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -130,6 +131,48 @@ func discoverKubeAPIServerEndpoint() (string, error) {
 	return host.Status.APIServerInternalURL, nil
 }
 
+// getWICDCredentials returns the CA cert and access token associated with the WICD service account
```
sebsoto, line 133: I suggest moving this function into `init.go` as that is the only place it is used

sebsoto (approved):

saifshaikh48:
/retest-required

couple cluster install failures

PR #1281, opened 2022-10-13
[release-4.10] WINC-830: Packer refactor

aravindhp:
- [docs] Refactor Packer build steps
- [docs] Migrate Packer build to HCL format
- [docs] Update golden image creation


docs/vsphere_ci/windows-server-2022.pkr.hcl
```diff
@@ -0,0 +1,150 @@
...
+    elevated_password = "${var.vm-elevated-password}"
+    elevated_user     = "${var.vm-elevated-user}"
+    script            = "scripts/install-firewall-rules.ps1"
+  }
+
+  provisioner "powershell" {
+    elevated_password = "${var.vm-elevated-password}"
+    elevated_user     = "${var.vm-elevated-user}"
+    script            = "scripts/install-docker.ps1"
+  }
```
saifshaikh48, line 117: It looks like the order of the provisioner steps changed in the translation to HCL. In the HCL code, Docker is installed before any run of `install-updates.ps1`. But in build.json, Docker was installed after the running `install-updates.ps1` twice.

Was the golden image built using the HCL config? Just want to confirm in case we have to rebuild the image.

aravindhp: > But in build.json, Docker was installed after the running install-updates.ps1 twice

Good catch. I can fix that just to make things consistent.

>Was the golden image built using the HCL config? Just want to confirm in case we have to rebuild the image.

Yes the final image was built using the HCL config.

aravindhp: @saifshaikh48 fixed. To be clear I did most of my testing with the HCL config given that was the way forward. I only made the `build.json` changes to make it easier to review.

saifshaikh48 (commented):
/lgtm

PR #1283, opened 2022-10-19
WINC-818: [test] Fix upgradeable condition test 

mansikulkarni96:
This PR fixes an issue where the upgradeable condition was not being tested in the e2e test suite. This was the case because the OPERATOR_CONDITION_NAME env var is not being set in the CI env when using the OLM upgrade workflow. The recommendation to fix this was to instead get it from the deployment spec as the condition is present once the operator is deployed. The cache client also had to be replaced with the OLM client to get the operator condition using current context.


mansikulkarni96:
/test aws-e2e-upgrade
cluster install failed

test/e2e/validation_test.go
```diff
@@ -667,22 +665,24 @@ func (tc *testContext) findNodeCSRs(nodeName string) ([]certificates.Certificate
...
-		oc := &operators.OperatorCondition{}
-		err := tc.client.Cache.Get(context.TODO(), types.NamespacedName{Namespace: tc.namespace, Name: ocName}, oc)
+	deployment, err := tc.client.K8s.AppsV1().Deployments(resourceNamespace).Get(context.TODO(), resourceName,
+		meta.GetOptions{})
+	if err != nil {
+		log.Printf("error getting operator deployment: %v", err)
+		return err
+	}
+	// Get the operator condition name using the deployment spec
+	ocName := deployment.Spec.Template.Spec.Containers[0].Env[3].Value
```
jrvaldes, line 675: Is it safe to get these values by index? Will the operator condition name always be in the third position?

mansikulkarni96: that has been my observation but I think it will be safer to loop through and find the env var by key, will update.

saifshaikh48: Also the `Container[0]` -- is it worth looping over containers and filtering for the the desired one by name?

test/e2e/validation_test.go
```diff
@@ -667,22 +665,24 @@ func (tc *testContext) findNodeCSRs(nodeName string) ([]certificates.Certificate
...
+		oc, err := tc.client.Olm.OperatorsV1().OperatorConditions(tc.namespace).Get(context.TODO(), ocName, meta.GetOptions{})
 		if err != nil {
 			log.Printf("unable to get OperatorCondition %s from namespace %s", ocName, tc.namespace)
 			return false, nil
 		}
-
-		specCheck := condition.Validate(oc.Spec.Conditions, operators.Upgradeable, expected)
-		statusCheck := condition.Validate(oc.Status.Conditions, operators.Upgradeable, expected)
-		return specCheck && statusCheck, nil
+		return condition.Validate(oc.Status.Conditions, operators.Upgradeable, expected), nil
```
saifshaikh48, line 685: Any reason to move to away from checking both spec and status? I felt that checking both was important since the upgradeable condition changes so often, we should ensure it has settled at the expected state (i.e. `spec == status == expected`).

mansikulkarni96: good point, I was thinking why it wasn't available in the object. I realised I was using the v1 client, switched to v2 now and I see conditions in spec now. Updating this as well.

test/e2e/validation_test.go
```diff
@@ -689,3 +685,22 @@ func (tc *testContext) validateUpgradeableCondition(expected meta.ConditionStatu
...
+	// Get the operator condition name using the deployment spec
+	for i := range deployment.Spec.Template.Spec.Containers {
+		envVars := deployment.Spec.Template.Spec.Containers[i].Env
+		for j, each := range envVars {
+			if each.Name == condition.OperatorConditionName {
+				return envVars[j].Value, nil
+			}
+		}
+	}
+	return "", nil
```
jrvaldes, line 705: As the `OperatorConditionName` var is a hard requirement, I suggest returning an exception here.

test/e2e/validation_test.go
```diff
@@ -667,14 +665,12 @@ func (tc *testContext) findNodeCSRs(nodeName string) ([]certificates.Certificate
...
-	if !present {
-		// Implies operator is not OLM-managed
-		return nil
-	}
-	err := wait.Poll(retry.Interval, retry.ResourceChangeTimeout, func() (bool, error) {
-		oc := &operators.OperatorCondition{}
-		err := tc.client.Cache.Get(context.TODO(), types.NamespacedName{Namespace: tc.namespace, Name: ocName}, oc)
+	ocName, err := tc.getOperatorConditionName()
+	if err != nil || ocName == "" {
+		return errors.Errorf("unable to get operatorCondition name from namespace %s", tc.namespace)
```
jrvaldes, line 670: s/`tc.namespace`/`resourceNamespace`

To align with the value used in the API [call](https://github.com/openshift/windows-machine-config-operator/pull/1283/commits/f94484220f4198182a545297028b9fe10357c1c2#diff-19a6ecdd2ef106d645523b5e6ddc22fd6ddfd1608477cd0b0fbd85113c8c245dR691).

test/e2e/validation_test.go
```diff
@@ -667,14 +665,12 @@ func (tc *testContext) findNodeCSRs(nodeName string) ([]certificates.Certificate
...
+	}
+	err = wait.Poll(retry.Interval, retry.ResourceChangeTimeout, func() (bool, error) {
```
jrvaldes, line 672: Is it required to wait first? Can we use `wait.PollImmediate` instead to avoid the idle time?

test/e2e/validation_test.go
```diff
@@ -689,3 +685,22 @@ func (tc *testContext) validateUpgradeableCondition(expected meta.ConditionStatu
...
+// getOperatorConditionName returns the operator condition name using the env var present in the deployment
+func (tc *testContext) getOperatorConditionName() (string, error) {
+	deployment, err := tc.client.K8s.AppsV1().Deployments(resourceNamespace).Get(context.TODO(), resourceName,
+		meta.GetOptions{})
+	if err != nil {
+		return "", errors.Wrap(err, "error getting operator deployment")
+	}
+	// Get the operator condition name using the deployment spec
+	for i := range deployment.Spec.Template.Spec.Containers {
+		envVars := deployment.Spec.Template.Spec.Containers[i].Env
```
jrvaldes, line 698: Do we know the name of the container beforehand, so that we can filter out the one we are not interested in? Otherwise,  this logic with loop all the `envVars` for all containers.

jrvaldes (commented):
@mansikulkarni96  thanks for working on this. Added a few comments. PTAL

test/e2e/validation_test.go
mansikulkarni96: Error is returned from the method that calls getOperatorCondition name [here](https://github.com/openshift/windows-machine-config-operator/blob/f94484220f4198182a545297028b9fe10357c1c2/test/e2e/validation_test.go#L669)

test/e2e/validation_test.go
```diff
@@ -689,3 +685,22 @@ func (tc *testContext) validateUpgradeableCondition(expected meta.ConditionStatu
...
+		for j, each := range envVars {
+			if each.Name == condition.OperatorConditionName {
+				return envVars[j].Value, nil
```
saifshaikh48, line 701: nit: `each.Value` instead of `envVars[j].Value`

test/e2e/validation_test.go
saifshaikh48: We should use` tc.namespace` in all cases. The operator can technically be deployed in any namespace and it's best if we get this value from the test context rather than a constant.

test/e2e/validation_test.go
mansikulkarni96: that's correct, updating it now.

test/e2e/validation_test.go
mansikulkarni96: yes it is required to wait, we are not exactly sure at what point OLM injects the variable. I tried pollImmediate and it did not work in this case.

test/e2e/validation_test.go
```diff
@@ -36,6 +34,10 @@ import (
 	"github.com/openshift/windows-machine-config-operator/pkg/windows"
 )
 
+const (
```
mansikulkarni96, line 37: Will add a comment here in the next push

saifshaikh48 (commented):
/lgtm

aside from a nit for cleaner loops

test/e2e/validation_test.go
saifshaikh48: Also consider naming it `wmcoContainerName` or similar as it is not clear which deployment is being referred to

test/e2e/validation_test.go
```diff
@@ -689,3 +689,25 @@ func (tc *testContext) validateUpgradeableCondition(expected meta.ConditionStatu
...
+	}
+	// Get the operator condition name using the deployment spec
+	for i := range deployment.Spec.Template.Spec.Containers {
+		if deployment.Spec.Template.Spec.Containers[i].Name != deploymentContainerName {
+			continue
+		}
+		envVars := deployment.Spec.Template.Spec.Containers[i].Env
+		for _, each := range envVars {
+			if each.Name == condition.OperatorConditionName {
+				return each.Value, nil
```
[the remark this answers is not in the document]
mansikulkarni96: Thanks for the suggestion, looks much better.

test/e2e/validation_test.go
jrvaldes: I was thinking to avoid the `ocName == ""` comparison. Unless `each.Value` could return an empty string and you are also accounting for that.

aravindhp (approved):

jrvaldes (approved):
/lgtm

saifshaikh48:
LGTM

saifshaikh48 (commented):
/lgtm

I think it's worth noting that a side effect of this PR is that the e2e suite will fail if the operator is [not deployed through OLM](https://github.com/openshift/windows-machine-config-operator/blob/master/docs/HACKING.md#deploying-the-operator-directly) (this shouldn't affect end users though).

mansikulkarni96:
> I think it's worth noting that a side effect of this PR is that the e2e suite will fail if the operator is [not deployed through OLM](https://github.com/openshift/windows-machine-config-operator/blob/master/docs/HACKING.md#deploying-the-operator-directly) (this shouldn't affect end users though).

I would like to clarify though the test modified through this PR is testable when running the test suite locally as well as CI through the hack script.

mansikulkarni96:
> @mansikulkarni96 should this be backported to 4.9 as well (since #618 is also in 4.9)?

yes it should looking into the failed backports as well.

mansikulkarni96:
/cherry-pick release-4.11
not seeing issue backporting locally.

PR #1284, opened 2022-10-19
OCPBUGS-2028: Set kubelet hostname for GCP platform

jrvaldes:
This change fetches the hostname from the GCE instance metadata service
and overrides the kubelet hostname so that the name of the Windows node
matches the hostname. The hostname is truncated if the length is longer
than 63 characters.

Should go after:
- https://github.com/openshift/windows-machine-config-operator/pull/1279

### DIFF
--- a/pkg/internal/gcp-get-hostname.ps1
+++ b/pkg/internal/gcp-get-hostname.ps1
@@ -1,0 +1,32 @@
+# This script returns a mutated hostname if the length is longer than 63 characters for GCP. The hostname will be
+# the lesser of 63 characters after the first dot in the FQDN.
+#
+# To get the original hostname for a GCP compute instance, the instance metadata service is invoked using its
+# DNS address (metadata.google.internal) as the prefered method over the IPv4 address or other alternatives like
+# the GoogleCloud Tools for PowerShell library (https://cloud.google.com/tools/powershell/docs/quickstart) since
+# they may not be available, specially in the BYOH use case.
+#
+# See instance metadata documentaion for GCP https://cloud.google.com/compute/docs/metadata/default-metadata-values
+
+# MAX_LENGTH is the maximum number of character allowed for the instance's hostname in GCP
+$MAX_LENGTH = 63
+
+# get hostname from the instance metadata service
+$hostname=(Invoke-RestMethod -Headers @{'Metadata-Flavor'='Google'} `
+            -Uri "http://metadata.google.internal/computeMetadata/v1/instance/hostname")
+
+# check hostname length
+if ($hostname.Length -le $MAX_LENGTH) {
+    # up to 63 characters is good, nothing to do!
+    return $hostname
+}
+
+# find the index of first dot in the FQDN
+$firstDotIndex=$hostname.IndexOf(".")
+if (($firstDotIndex -gt 0) -and ($firstDotIndex -le $MAX_LENGTH) ) {
+    # and return first part of the FQDN
+    return $hostname.Substring(0, $firstDotIndex)
+}
+
+# otherwise, return the first 63 characters of the hostname
+return $hostname.Substring(0, $MAX_LENGTH)
--- a/pkg/services/services.go
+++ b/pkg/services/services.go
@@ -19,10 +19,9 @@ const (
 	// https://docs.openshift.com/container-platform/latest/rest_api/editing-kubelet-log-level-verbosity.html#log-verbosity-descriptions_editing-kubelet-log-level-verbosity
 	debugLogLevel    = "4"
 	standardLogLevel = "2"
-	// ec2HostnameVar is a variable that should be replaced with the value of `local-hostname` in the ec2 instance
-	// metadata
-	ec2HostnameVar = "EC2_HOSTNAME"
-	NodeIPVar      = "NODE_IP"
+	// hostnameOverrideVar is the variable that should be replaced with the value of the desired instance hostname
+	hostnameOverrideVar = "HOSTNAME_OVERRIDE"
+	NodeIPVar           = "NODE_IP"
 )
 
 // GenerateManifest returns the expected state of the Windows service configmap. If debug is true, debug logging
@@ -144,13 +143,16 @@ func getKubeletServiceConfiguration(argsFromIginition map[string]string, debug b
 		return servicescm.Service{}, err
 	}
 	var powershellVars []servicescm.PowershellCmdArg
-	hostnameArg := getKubeletHostnameOverride(platform)
-	if hostnameArg != "" {
-		kubeletArgs = append(kubeletArgs, hostnameArg)
-		powershellVars = append(powershellVars, servicescm.PowershellCmdArg{
-			Name: ec2HostnameVar,
-			Path: "Get-EC2InstanceMetadata -Category LocalHostname",
-		})
+
+	hostnameOverrideCmd := getHostnameCmd(platform)
+	if hostnameOverrideCmd != "" {
+		hostnameOverrideArg := "--hostname-override=" + hostnameOverrideVar
+		hostnameOverridePowershellVar := servicescm.PowershellCmdArg{
+			Name: hostnameOverrideVar,
+			Path: hostnameOverrideCmd,
+		}
+		kubeletArgs = append(kubeletArgs, hostnameOverrideArg)
+		powershellVars = append(powershellVars, hostnameOverridePowershellVar)
 	}
 
 	kubeletServiceCmd := windows.KubeletPath
@@ -217,12 +219,15 @@ func klogVerbosityArg(debug bool) string {
 	}
 }
 
-// getKubeletHostnameOverride returns the hostname override arg that should be used
-func getKubeletHostnameOverride(platformType config.PlatformType) string {
+// getHostnameCmd returns the hostname override command for the given platform as needed
+func getHostnameCmd(platformType config.PlatformType) string {
 	switch platformType {
 	case config.AWSPlatformType:
-		return "--hostname-override=" + ec2HostnameVar
+		return "Get-EC2InstanceMetadata -Category LocalHostname"
+	case config.GCPPlatformType:
+		return windows.GcpGetHostnameScriptRemotePath
 	default:
+		// by default use the original hostname
 		return ""
 	}
 }
--- a/pkg/services/services_test.go
+++ b/pkg/services/services_test.go
@@ -1,0 +1,38 @@
+package services
+
+import (
+	"testing"
+
+	config "github.com/openshift/api/config/v1"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetHostnameCmd(t *testing.T) {
+	tests := []struct {
+		name         string
+		platformType config.PlatformType
+		expected     string
+	}{
+		{
+			name:         "any platform",
+			platformType: "",
+			expected:     "",
+		},
+		{
+			name:         "AWS platform",
+			platformType: config.AWSPlatformType,
+			expected:     "Get-EC2InstanceMetadata -Category LocalHostname",
+		},
+		{
+			name:         "GCP platform",
+			platformType: config.GCPPlatformType,
+			expected:     "C:\\Temp\\gcp-get-hostname.ps1",
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			actual := getHostnameCmd(test.platformType)
+			assert.Equal(t, test.expected, actual)
+		})
+	}
+}
--- a/pkg/windows/windows.go
+++ b/pkg/windows/windows.go
@@ -23,6 +23,9 @@ import (
 const (
 	// remoteDir is the remote temporary directory created on the Windows VM
 	remoteDir = "C:\\Temp\\"
+	// GcpGetHostnameScriptRemotePath is the remote location of the PowerShell script that resolves the hostname
+	// for GCP instances
+	GcpGetHostnameScriptRemotePath = remoteDir + payload.GcpGetHostnameScriptName
 	// HNSPSModule is the remote location of the hns.psm1 module
 	HNSPSModule = remoteDir + "hns.psm1"
 	// K8sDir is the remote kubernetes executable directory
@@ -157,20 +160,21 @@ func getFilesToTransfer() (map[*payload.FileInfo]string, error) {
 		return filesToTransfer, nil
 	}
 	srcDestPairs := map[string]string{
-		payload.WICDPath:                   K8sDir,
-		payload.HybridOverlayPath:          K8sDir,
-		payload.HNSPSModule:                remoteDir,
-		payload.WindowsExporterPath:        K8sDir,
-		payload.WinBridgeCNIPlugin:         cniDir,
-		payload.HostLocalCNIPlugin:         cniDir,
-		payload.WinOverlayCNIPlugin:        cniDir,
-		payload.KubeProxyPath:              K8sDir,
-		payload.KubeletPath:                K8sDir,
-		payload.AzureCloudNodeManagerPath:  K8sDir,
-		payload.ContainerdPath:             ContainerdDir,
-		payload.HcsshimPath:                ContainerdDir,
-		payload.ContainerdConfPath:         ContainerdDir,
-		payload.NetworkConfigurationScript: remoteDir,
+		payload.GcpGetValidHostnameScriptPath: remoteDir,
+		payload.WICDPath:                      K8sDir,
+		payload.HybridOverlayPath:             K8sDir,
+		payload.HNSPSModule:                   remoteDir,
+		payload.WindowsExporterPath:           K8sDir,
+		payload.WinBridgeCNIPlugin:            cniDir,
+		payload.HostLocalCNIPlugin:            cniDir,
+		payload.WinOverlayCNIPlugin:           cniDir,
+		payload.KubeProxyPath:                 K8sDir,
+		payload.KubeletPath:                   K8sDir,
+		payload.AzureCloudNodeManagerPath:     K8sDir,
+		payload.ContainerdPath:                ContainerdDir,
+		payload.HcsshimPath:                   ContainerdDir,
+		payload.ContainerdConfPath:            ContainerdDir,
+		payload.NetworkConfigurationScript:    remoteDir,
 	}
 	files := make(map[*payload.FileInfo]string)
 	for src, dest := range srcDestPairs {
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -93,6 +93,7 @@ RUN make build-daemon
 #│   ├── kubelet.exe
 #│   └── kube-proxy.exe
 #├── powershell/
+#│   ├── gcp-get-hostname.ps1
 #│   └── hns.psm1
 #├── windows_exporter.exe
 #└── windows-instance-config-daemon.exe
@@ -136,6 +137,7 @@ RUN chmod 0777 /payload/generated
 
 # Copy required powershell scripts
 WORKDIR /payload/powershell/
+COPY pkg/internal/gcp-get-hostname.ps1 .
 COPY pkg/internal/hns.psm1 .
 
 WORKDIR /
--- a/build/Dockerfile.ci
+++ b/build/Dockerfile.ci
@@ -103,6 +103,7 @@ RUN make build-daemon
 #│   ├── kubelet.exe
 #│   └── kube-proxy.exe
 #├── powershell/
+#│   ├── gcp-get-hostname.ps1
 #│   └── hns.psm1
 #├── windows_exporter.exe
 #└── windows-instance-config-daemon.exe
@@ -146,6 +147,7 @@ RUN chmod 0777 /payload/generated
 
 # Copy required powershell scripts
 WORKDIR /payload/powershell/
+COPY --from=build /build/windows-machine-config-operator/pkg/internal/gcp-get-hostname.ps1 .
 COPY --from=build /build/windows-machine-config-operator/pkg/internal/hns.psm1 .
 
 WORKDIR /
--- a/build/Dockerfile.wmco
+++ b/build/Dockerfile.wmco
@@ -35,6 +35,7 @@ RUN chmod 0777 generated
 
 # Copy required powershell scripts
 WORKDIR /payload/powershell/
+COPY pkg/internal/gcp-get-hostname.ps1 .
 COPY pkg/internal/hns.psm1 .
 
 WORKDIR /
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -115,6 +115,7 @@ func main() {
 		payload.HybridOverlayPath,
 		payload.KubeletPath,
 		payload.KubeProxyPath,
+		payload.GcpGetValidHostnameScriptPath,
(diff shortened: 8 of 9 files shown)


pkg/services/services.go
```diff
@@ -194,12 +190,57 @@ func klogVerbosityArg(debug bool) string {
 }
 
 // getKubeletHostnameOverride returns the hostname override arg that should be used
-func getKubeletHostnameOverride(platformType string) string {
+func getKubeletHostnameOverride(platformType string) (string, servicescm.PowershellCmdArg) {
```
jrvaldes, line 193: This addresses https://github.com/openshift/windows-machine-config-operator/pull/1260#discussion_r997390403

pkg/services/services.go
```diff
@@ -194,12 +190,57 @@ func klogVerbosityArg(debug bool) string {
...
+}
+
+// getGCPMetadataHostnamePowershellCmdArg returns the PowershellCmdArg to resolve the hostname using the instance
+// metadata service in GCP. Uses GoogleCloud Tools for PowerShell (https://cloud.google.com/tools/powershell/docs/quickstart)
+// getGCPMetadataHostname returns the GCP instance hostname from the metadata service.  GCP is an especial case,
+// and the resulting hostname is truncated if the length is longer than 63 characters.
+func getGCPMetadataHostnamePowershellCmdArg(variableName string) servicescm.PowershellCmdArg {
+	return servicescm.PowershellCmdArg{
+		Name: variableName,
+		Path: `
```
jrvaldes, line 223: As an alternative, I can move this PowerShell script into a file under the [pkg/internal](https://github.com/openshift/windows-machine-config-operator/tree/master/pkg/internal) directory, add the file to the payload and here just declare the path. Thoughts?

pkg/services/services.go
```diff
@@ -194,12 +190,57 @@ func klogVerbosityArg(debug bool) string {
...
+# MAX_LENGTH is the maximum number of character allowed for the instance's hostname in GCP
+$MAX_LENGTH = 63
+
+# get hostname from the instance metadata service
+$hostname=Get-GceMetadata -Path "instance/hostname"
```
jrvaldes, line 228: For consistency, I followed the existing approach and am still concerned about https://github.com/openshift/windows-machine-config-operator/pull/1260#discussion_r997394197 .

I feel querying the instance metadata endpoint directly may be safer, thinking of the BYOH use case. For example in GCP:
```powershell
$HOSTNAME = (Invoke-RestMethod ` -Headers @{'Metadata-Flavor' = 'Google'} `  -Uri "http://metadata.google.internal/computeMetadata/v1/instance/hostname")
```

jrvaldes (commented):
Added a [comment](https://issues.redhat.com/browse/OCPBUGS-2028?focusedCommentId=21122869&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-21122869) in the ticket capturing the machine and node outputs using the proposed logic to truncate longer hostnames.

I tested this with short and long machine names, and the resulting node name is adjusted accordingly. 

```
[windows-machine-config-operator]$ date && oc get machines -A && oc get nodes
Wed Oct 19 02:31:57 PM EDT 2022
NAMESPACE               NAME                                                              PHASE     TYPE            REGION        ZONE            AGE
openshift-machine-api   jvaldes-vcjw5-master-0                                            Running   n2-standard-4   us-central1   us-central1-a   4h14m
openshift-machine-api   jvaldes-vcjw5-master-1                                            Running   n2-standard-4   us-central1   us-central1-b   4h14m
openshift-machine-api   jvaldes-vcjw5-master-2                                            Running   n2-standard-4   us-central1   us-central1-c   4h14m
openshift-machine-api   jvaldes-vcjw5-windows-worker-with-the-max-length-in-name-6vnsk7   Running   n1-standard-4   us-central1   us-central1-a   11m
openshift-machine-api   jvaldes-vcjw5-winworker-a-ssgrf                                   Running   n1-standard-4   us-central1   us-central1-a   3h6m
openshift-machine-api   jvaldes-vcjw5-worker-a-lttk7                                      Running   n2-standard-4   us-central1   us-central1-a   4h7m
openshift-machine-api   jvaldes-vcjw5-worker-b-gxtjn                                      Running   n2-standard-4   us-central1   us-central1-b   4h7m
openshift-machine-api   jvaldes-vcjw5-worker-c-xmbsk                                      Running   n2-standard-4   us-central1   us-central1-c   4h7m
NAME                                                              STATUS   ROLES                  AGE     VERSION
jvaldes-vcjw5-master-0.c.openshift-gce-devel.internal             Ready    control-plane,master   4h11m   v1.25.0+3ef6ef3
jvaldes-vcjw5-master-1.c.openshift-gce-devel.internal             Ready    control-plane,master   4h11m   v1.25.0+3ef6ef3
jvaldes-vcjw5-master-2.c.openshift-gce-devel.internal             Ready    control-plane,master   4h11m   v1.25.0+3ef6ef3
jvaldes-vcjw5-windows-worker-with-the-max-length-in-name-6vnsk7   Ready    worker                 4m54s   v1.25.0-2492+3ef6ef3cfbd273
jvaldes-vcjw5-winworker-a-ssgrf.c.openshift-gce-devel.internal    Ready    worker                 3h      v1.25.0-2492+3ef6ef3cfbd273
jvaldes-vcjw5-worker-a-lttk7.c.openshift-gce-devel.internal       Ready    worker                 3h57m   v1.25.0+3ef6ef3
jvaldes-vcjw5-worker-b-gxtjn.c.openshift-gce-devel.internal       Ready    worker                 3h57m   v1.25.0+3ef6ef3
jvaldes-vcjw5-worker-c-xmbsk.c.openshift-gce-devel.internal       Ready    worker                 3h57m   v1.25.0+3ef6ef3

```

pkg/services/services.go
saifshaikh48: I agree with the proposed alternative. Having the script live in a separate file might be a cleaner approach.

aravindhp: +1. That is is the pattern we have been following so far.

pkg/services/services.go
aravindhp: Can we be sure that the endpoint won't change?

jrvaldes: I feel we can trust it. The `metadata.google.internal` DNS name points to the IPv4 address of the instance metadata service: `169.254.169.254`. 

```
PS C:\Users\administrator> cat C:\Windows\System32\drivers\etc\hosts   
# Copyright (c) 1993-2009 Microsoft Corp.
#
# This is a sample HOSTS file used by Microsoft TCP/IP for Windows.
#
# This file contains the mappings of IP addresses to host names. Each
# entry should be kept on an individual line. The IP address should
# be placed in the first column followed by the corresponding host name.
# The IP address and the host name should be separated by at least one
# space.
#
# Additionally, comments (such as these) may be inserted on individual
# lines or following the machine name denoted by a '#' symbol.
#
# For example:
#
#      102.54.94.97     rhino.acme.com          # source server
#       38.25.63.10     x.acme.com              # x client host

# localhost name resolution is handled within DNS itself.
#       127.0.0.1       localhost
#       ::1             localhost

# Google Compute Engine metadata server
    169.254.169.254    metadata.google.internal metadata

```

aravindhp: My point is if we can trust that `metadata.google.internal` itself won't change. Just because it is in a `/etc/hosts` file does not mean that the content cannot change.

jrvaldes: If we are trusting the `Get-GceMetadata` command from the GCP Cloud Tools for PowerShell, we can trust the `metadata.google.internal` endpoint as this is what the library is [using](https://github.com/GoogleCloudPlatform/google-cloud-powershell/blob/b0549f98b2a150830ff6ddf10c66c3252df7626d/Google.PowerShell/Compute/GceMetadataCmdlet.cs#L131).

pkg/services/services.go
```diff
@@ -193,12 +189,57 @@ func klogVerbosityArg(debug bool) string {
...
+
+// getGCPMetadataHostnamePowershellCmdArg returns the PowershellCmdArg to resolve the hostname using the instance
+// metadata service in GCP. Uses GoogleCloud Tools for PowerShell (https://cloud.google.com/tools/powershell/docs/quickstart)
+// getGCPMetadataHostname returns the GCP instance hostname from the metadata service.  GCP is an especial case,
+// and the resulting hostname is truncated if the length is longer than 63 characters.
+func getGCPMetadataHostnamePowershellCmdArg(variableName string) servicescm.PowershellCmdArg {
+	return servicescm.PowershellCmdArg{
+		Name: variableName,
+		Path: `
+# MAX_LENGTH is the maximum number of character allowed for the instance's hostname in GCP
```
alinaryan, line 223: Is this/should this be doc'd somewhere so no one accidentally creates an instance more than 63 chars?

pkg/services/services.go
alinaryan: +1

alinaryan (commented):
Looks good so far: Just a comment

pkg/services/services.go
aravindhp: Point of using a library is that it abstracts such details from us.

aravindhp: What I am looking for is a reference to a GCE doc that guarantees this endpoint won't change or a way for us to easily catch the change in an e2e test.

pkg/services/services.go
aravindhp: To be clear the request is to include this as a file in the internal directory

saifshaikh48: You could check platform before adding it to the files to be transferred in `windows.go`. We currently transfer `azure-cloud-node-manager.exe` to every instance on every platform though, so this may warrant a team-wide consensus on whether everything in the payload should always be transferred as is.

aravindhp: +1

And this should not be the reason to not make this file

pkg/services/services.go
```diff
@@ -236,29 +236,11 @@ func getAWSMetadataHostnamePowershellCmdArg(variableName string) servicescm.Powe
 }
 
 // getGCPMetadataHostnamePowershellCmdArg returns the PowershellCmdArg to resolve the hostname using the instance
-// metadata service in GCP. Uses GoogleCloud Tools for PowerShell (https://cloud.google.com/tools/powershell/docs/quickstart)
-// getGCPMetadataHostname returns the GCP instance hostname from the metadata service.  GCP is an especial case,
-// and the resulting hostname is truncated if the length is longer than 63 characters.
+// metadata service in GCP.
+// GCP is an especial case, and the resulting hostname is truncated if the length is longer than 63 characters.
```
saifshaikh48, line 240: `especial` --> `special`

pkg/services/services.go
```diff
@@ -144,13 +143,10 @@ func getKubeletServiceConfiguration(argsFromIginition map[string]string, debug b
 		return servicescm.Service{}, err
 	}
 	var powershellVars []servicescm.PowershellCmdArg
-	hostnameArg := getKubeletHostnameOverride(platform)
+	hostnameArg, hostnamePowershellVars := getKubeletHostnameOverride(platform)
```
saifshaikh48, line 146: nit: `hostnamePowershellVar` instead of the plural `hostnamePowershellVars`, I had to re-read to make sure `getKubeletHostnameOverride()` only return one PowershellCmdArg

saifshaikh48 (commented):
My vote on [this discussion](https://github.com/openshift/windows-machine-config-operator/pull/1284/commits/13e9b4c8f0d13577a9ad574ed20d0bc7f9b566c6#r1012929387) is:

- I'm a proponent for minimizing platform-specific logic. Therefore, I am fine if we always transfer the `gcp-get-valid-hostname.ps1` script as part of the payload (the approach proposed in the second commit [ee0c52e](https://github.com/openshift/windows-machine-config-operator/pull/1284/commits/ee0c52e372f54ed69b004ef719aae576a9e5ef9f#diff-4d8fe3bb66a872f0e4a277f820d60d2efa722f09aa89516be74b98c06b57f3bdR155))

pkg/services/services.go
jrvaldes: Makes sense, will squash the commits. 

Thanks you all for the feedback.

pkg/services/services.go
jrvaldes: The proposed implementation will take care of this so that the naming is consistent, even if the length of the instance's name is greater than 63 chars.

pkg/services/services.go
jrvaldes: Addressed.

pkg/services/services.go
jrvaldes: >  a reference to a GCE doc that guarantees this endpoint won't change

Couldn't find such a statement explicitly in the docs or source code.

> way for us to easily catch the change in an e2e test.

Sure, it can be added in case we decided to consume the endpoint directly.


As mentioned, using the Cloud Tools for PowerShell library provides a safe abstraction from maintaining the instance metadata endpoint as the library is installed in the instance by the cloud provider.

Back to https://github.com/openshift/windows-machine-config-operator/pull/1260#discussion_r997394197  about the  BYOH use case, it may be an issue if the instance does not have the library. I suggest the following options:
 - Add the `Cloud Tools for PowerShell` library as a BYOH requirement for AWS and GCP (at the moment), or
 - Invoke the instance metadata endpoint directly to eliminate the dependency on the  `Cloud Tools for PowerShell` library

aravindhp: > Couldn't find such a statement explicitly in the docs or source code.

[Default VM metadata values](https://cloud.google.com/compute/docs/metadata/default-metadata-values) should suffice.

> Invoke the instance metadata endpoint directly to eliminate the dependency on the Cloud Tools for PowerShell library

I am not against this approach and it is the best path forward.

pkg/internal/gcp-get-valid-hostname.ps1
```diff
@@ -0,0 +1,25 @@
+# This script returns a mutated hostname if the length is longer than 63 characters. The hostname will be the
```
aravindhp, line 1: `/gcp-get-hostname` will suffice. Valid is implied and obvious. If the script is returning an invalid hostname then it is a bug.

pkg/internal/gcp-get-valid-hostname.ps1
```diff
@@ -0,0 +1,25 @@
...
+# lesser of 63 characters after the first dot in the FQDN. Uses the GoogleCloud Tools for PowerShell (https://cloud.google.com/tools/powershell/docs/quickstart)
+# to get the original hostname from the GCP instance metadata service.
+
+# MAX_LENGTH is the maximum number of character allowed for the instance's hostname in GCP
+$MAX_LENGTH = 63
+
+# get hostname from the instance metadata service
+$hostname=Get-GceMetadata -Path "instance/hostname"
```
aravindhp, line 9: As discussed it is best to go the internal endpoint due to the BYOH case you have pointed out. Please link to the doc I pointed you at.

pkg/services/services.go
```diff
@@ -144,13 +143,11 @@ func getKubeletServiceConfiguration(argsFromIginition map[string]string, debug b
 		return servicescm.Service{}, err
 	}
 	var powershellVars []servicescm.PowershellCmdArg
-	hostnameArg := getKubeletHostnameOverride(platform)
-	if hostnameArg != "" {
-		kubeletArgs = append(kubeletArgs, hostnameArg)
-		powershellVars = append(powershellVars, servicescm.PowershellCmdArg{
-			Name: ec2HostnameVar,
-			Path: "Get-EC2InstanceMetadata -Category LocalHostname",
```
aravindhp, line 152: This is the only thing that changes across cloud providers, so your helper should just focus on that.

pkg/services/services.go
```diff
@@ -218,11 +215,34 @@ func klogVerbosityArg(debug bool) string {
...
+		return "", servicescm.PowershellCmdArg{}
+	}
+}
+
+// getAWSMetadataHostnamePowershellCmdArg returns the PowershellCmdArg to resolve the hostname using the instance
+// metadata service in AWS. Uses AWS Tools for PowerShell (https://docs.aws.amazon.com/powershell/latest/userguide/pstools-welcome.html)
+func getAWSMetadataHostnamePowershellCmdArg(variableName string) servicescm.PowershellCmdArg {
+	return servicescm.PowershellCmdArg{
+		Name: variableName,
+		Path: "Get-EC2InstanceMetadata -Category LocalHostname",
```
aravindhp, line 236: The helper should only deal with the command not the whole `PowershellCmdArg` object. 
`GetAWSHostnameCmd` is succinct and sufficiently informs us what the helper does.

pkg/services/services.go
```diff
@@ -218,11 +215,34 @@ func klogVerbosityArg(debug bool) string {
...
+	}
+}
+
+// getGCPMetadataHostnamePowershellCmdArg returns the PowershellCmdArg to resolve the hostname using the instance
+// metadata service in GCP.
+// GCP is a special case, and the resulting hostname is truncated if the length is longer than 63 characters.
+func getGCPMetadataHostnamePowershellCmdArg(variableName string) servicescm.PowershellCmdArg {
```
aravindhp, line 243: `GetGCPHostnameCmd`

aravindhp (changes requested):

pkg/internal/gcp-get-valid-hostname.ps1
saifshaikh48: +1, applies throughout PR

pkg/services/services.go
jrvaldes: Makes sense.

pkg/windows/windows.go
```diff
@@ -25,6 +25,8 @@ const (
 	remoteDir = "C:\\Temp\\"
 	// winTemp is the default Windows temporary directory
 	winTemp = "C:\\Windows\\Temp\\"
+	// GcpGetValidHostnameScriptRemotePath a
```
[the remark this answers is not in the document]
jrvaldes: Fixed.

test/e2e/create_test.go
```diff
@@ -589,3 +593,46 @@ func (tc *testContext) scaleDeployment(namespace, name, selector string, expecte
...
 	return nil
 }
+
+// checkMachineInstanceMetadata checks the instance metadata on the Windows machines and returns with success
+// after finding the first machine where the instance metadata endpoint is available. Otherwise, returns an error.
+// This check is platform specific, as the instance metadata endpoint is required in a subset of the supported
+// platforms (i.e. GCP).
+// The BYOH use case is also covered by this check, since both machine and BYOH instances runs on the same cloud
+// provider.
+func (tc *testContext) checkMachineInstanceMetadata(machines []mapi.Machine) (err error) {
```
sebsoto, line 603: I dont see the value in this helper.
Its not testing the functionality of our code.
If the metadata endpoint isnt reachable, WMCO will fail to configure the VM.

pkg/services/services_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetHostnameCmd(t *testing.T) {
+	tests := []struct {
+		name         string
+		platformType config.PlatformType
+		expected     string
+	}{
+		// TODO: Add test cases.
```
sebsoto, line 16: This TODO isn't helpful as it doesn't describe what test cases are missing.
Additionally if you know what test cases are missing that are required, please just add them

sebsoto (changes requested):
Thanks @jrvaldes PTAL at my comments

pkg/services/services_test.go
jrvaldes: Missed removing this line. Good catch.

test/e2e/create_test.go
jrvaldes: Correct. This test is redundant as WMCO won't be able to configure the Windows instance if the address of the instance metadata endpoint changes. 

Removed the test.

saifshaikh48 (commented):
/lgtm

:rocket:

alinaryan:
LGTM

test/e2e/create_test.go
aravindhp: I am fine with not having a test as long as WMCO failing to configuring the VM will point to an issue with the endpoint.

pkg/internal/gcp-get-hostname.ps1
```diff
@@ -0,0 +1,33 @@
...
+#
+# The IPv4 address of the instance metadata service is defaulted to 169.254.169.254. See instance metadata
+# documentaion for GCP https://cloud.google.com/compute/docs/metadata/default-metadata-values.
+
+# MAX_LENGTH is the maximum number of character allowed for the instance's hostname in GCP
+$MAX_LENGTH = 63
+
+# get hostname from the instance metadata service
+$hostname=(Invoke-RestMethod -Headers @{'Metadata-Flavor'='Google'} `
+            -Uri "http://169.254.169.254/computeMetadata/v1/instance/hostname")
```
aravindhp, line 17: Going with the doc this should be http://metadata.google.internal and not the IP address.

pkg/services/services.go
```diff
@@ -19,10 +19,9 @@ const (
 	// https://docs.openshift.com/container-platform/latest/rest_api/editing-kubelet-log-level-verbosity.html#log-verbosity-descriptions_editing-kubelet-log-level-verbosity
 	debugLogLevel    = "4"
 	standardLogLevel = "2"
-	// ec2HostnameVar is a variable that should be replaced with the value of `local-hostname` in the ec2 instance
-	// metadata
-	ec2HostnameVar = "EC2_HOSTNAME"
-	NodeIPVar      = "NODE_IP"
+	// hostnameOverrideVarName is a variable that should be replaced with the value of the desired instance hostname
+	hostnameOverrideVarName = "HOSTNAME_OVERRIDE"
```
aravindhp, line 23: This should be `hostNameOverrideVar`

aravindhp (changes requested):

pkg/internal/gcp-get-hostname.ps1
jrvaldes: There is a [comment](https://github.com/openshift/windows-machine-config-operator/pull/1284/files#diff-e5bc5d4c1a111c9f29a32845d5d34069f73857555f955fe898efeed077cdcba7R4) up in this file with details about this. 

TL;DR
Using the IP address is safer, especially in the BYOH use case where there is no control over the host file and the resolution of the  `metadata.google.internal` DNS name may not be available.

aravindhp: Won't images be created from existing master images which will have this DNS setup? Now we could run into issues if the IP has changed. Or is there some way to import images created on prem where this need not have setup?

aravindhp (approved):

saifshaikh48 (commented):
/lgtm

nice work @jrvaldes

aravindhp:
@jrvaldes please don't manually cherry-pick this. This should apply cleanly to 4.12 and most likely is missing some other backport.

jrvaldes:
> @jrvaldes please don't manually cherry-pick this. This should apply cleanly to 4.12 and most likely is missing some other backport.

I don't plan to. https://github.com/openshift/windows-machine-config-operator/pull/1314 must go first.

PR #1306, opened 2022-11-01
WINC-927: Enable stopping a service with dependent services

sebsoto:
Enables stopping a service which other services depend on


pkg/daemon/manager/manager.go
```diff
@@ -98,6 +115,26 @@ func (m *manager) EnsureServiceState(service winsvc.Service, state svc.State) er
...
+		// The service must be cast to the actual type so we can get its handle
+		winSvc, ok := service.(*mgr.Service)
+		if !ok {
+			return fmt.Errorf("service is not correct type")
+		}
+		dependentServices, err := m.getDependentServices(winSvc.Handle)
+		if err != nil {
+			return errors.Wrap(err, "error finding dependent services")
+		}
+		for _, dependentServiceName := range dependentServices {
```
saifshaikh48, line 128: What is preventing us from just using the service object as is? 

``` suggestion
		config, err := service.Config()
		if err != nil {
			return errors.Wrap(err, "error finding dependent services")
		}
		for _, dependentServiceName := range config.Dependencies {
```

saifshaikh48, line 128: What is preventing us from just using the service object as is? This would majorly simplify things

``` suggestion
		config, err := service.Config()
		if err != nil {
			return errors.Wrap(err, "error finding dependent services")
		}
		for _, dependentServiceName := range config.Dependencies {
```

pkg/daemon/manager/manager.go
```diff
@@ -109,6 +146,60 @@ func (m *manager) EnsureServiceState(service winsvc.Service, state svc.State) er
...
+// getDependentServices returns a list of names of all services dependent on the given service
+func (m *manager) getDependentServices(serviceHandle windows.Handle) ([]string, error) {
+	// The documentation for this syscall says it should be ran twice. First to determine the size of the buffer it will
+	// return, and then to actually capture the data with an allocated buffer. In practice, I'm finding it succeeds on
+	// the first run.
+
+	// Borrowing the main steps done here from the golang windows/mgr library's ListServices() function, as the
+	// EnumServicesStatusEx syscall has a very similar way of being called.
+	var buf []byte
+	var bytesNeeded, servicesReturned uint32
```
jrvaldes, line 165: I suggest naming `servicesReturned` as `servicesCount` to better reflect the number of returned services.

jrvaldes, line 165: I suggest naming `servicesReturned` as `servicesCount` to better reflect the number of returned services.

pkg/daemon/manager/manager.go
```diff
@@ -109,6 +146,60 @@ func (m *manager) EnsureServiceState(service winsvc.Service, state svc.State) er
...
+	for {
+		var p *byte
+		if len(buf) > 0 {
+			p = &buf[0]
+		}
+		// Returned error from `Call` will always be non-nil
+		success, _, err := enumDependentServicesSyscall(serviceHandle, windows.SERVICE_STATE_ALL, p, uint32(len(buf)),
+			&bytesNeeded, &servicesReturned)
+		if success != 0 {
+			break
```
jrvaldes, line 175: Is it safe to `break` the loop here? or should `continue` for another attempt?  Consider adding an inline comment here.

jrvaldes, line 175: Is it safe to `break` the loop here? or should `continue` for another attempt?  Consider adding an inline comment here.

jrvaldes (commented):
@sebsoto  thanks for working on this, left a few comments. PTAL.

pkg/daemon/manager/manager.go
sebsoto: What you've suggested is looking at the services this service is dependent on, not the services that are dependent on this services.

Handling the case of kubelet, my code will stop hybrid-overlay, kube-proxy before stopping kubelet, the suggested code will stop containerd which isnt necessary.

pkg/daemon/manager/manager.go
```diff
@@ -98,6 +115,26 @@ func (m *manager) EnsureServiceState(service winsvc.Service, state svc.State) er
 			return err
 		}
 	case svc.Stopped:
+		// Before we can stop this service, we need to make sure all services dependent on this service are stopped
+		// The service must be cast to the actual type so we can get its handle
+		winSvc, ok := service.(*mgr.Service)
```
mansikulkarni96, line 120: should we have a check before if the mgr object is nil?

pkg/daemon/manager/manager.go
saifshaikh48: Ah I see, thought I was missing something. An alternative to the `EnumDependentServicesW` approach would be listing all Windows services and filtering them based on if our `service` appears in their `Dependencies` list. Simpler code but definitely more syscalls.
```
svcConfig, err := service.Config()
if err != nil {
	return nil
}
allSvcs, err := m.GetServices()
if err != nil {
	return nil
}
dependentServices := []string{}
for potentialDependent := range allSvcs {
	potentialDependentHandle, err := m.OpenService(potentialDependent)
	if err != nil {
		return nil
	}
	potentialDependentConfig, err := potentialDependentHandle.Config()
	if err != nil {
		return nil
	}
	for _, dep := range potentialDependentConfig.Dependencies {
		if dep == svcConfig.DisplayName {
			dependentServices = append(dependentServices, potentialDependent)
		}
	}
}
```

pkg/daemon/manager/manager.go
```diff
@@ -82,6 +83,32 @@ func (m *manager) DeleteService(name string) error {
 	return nil
 }
 
+func (m *manager) EnsureServiceState(service winsvc.Service, state svc.State) error {
```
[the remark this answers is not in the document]
mansikulkarni96: can this function be used inside [EnsureRequiredServicesStopped](https://github.com/openshift/windows-machine-config-operator/blob/39571269183f58c75d68c08e8954da5937ea5ebc/pkg/windows/windows.go#L387) now?

pkg/daemon/manager/manager.go
```diff
@@ -109,6 +146,60 @@ func (m *manager) EnsureServiceState(service winsvc.Service, state svc.State) er
...
+			p = &buf[0]
+		}
+		// Returned error from `Call` will always be non-nil
+		success, _, err := enumDependentServicesSyscall(serviceHandle, windows.SERVICE_STATE_ALL, p, uint32(len(buf)),
+			&bytesNeeded, &servicesReturned)
+		if success != 0 {
+			break
+		}
+		if err != windows.ERROR_MORE_DATA {
+			return nil, errors.Wrapf(err, "received unexpected error from EnumDependentServicesW")
```
aravindhp, line 178: Given where this is, it should be `enumDependentServicesSyscall`.

pkg/daemon/manager/manager.go
```diff
@@ -109,6 +146,60 @@ func (m *manager) EnsureServiceState(service winsvc.Service, state svc.State) er
...
+		}
+		buf = make([]byte, bytesNeeded)
+	}
+	// If no services are dependent on this service, return successfully
+	if servicesReturned == 0 {
+		return nil, nil
+	}
+	// create a slice based on the buffer that was returned to us, so that we can iterate through it
+	var services []enumServiceStatus
+	hdr := (*reflect.SliceHeader)(unsafe.Pointer(&services))
```
aravindhp, line 191: Can this cast fail?

pkg/daemon/manager/manager.go
```diff
@@ -109,6 +146,60 @@ func (m *manager) EnsureServiceState(service winsvc.Service, state svc.State) er
...
+	var buf []byte
+	var bytesNeeded, servicesReturned uint32
+	for {
+		var p *byte
+		if len(buf) > 0 {
+			p = &buf[0]
+		}
+		// Returned error from `Call` will always be non-nil
+		success, _, err := enumDependentServicesSyscall(serviceHandle, windows.SERVICE_STATE_ALL, p, uint32(len(buf)),
+			&bytesNeeded, &servicesReturned)
```
aravindhp, line 173: I don't follow how `servicesReturned` will contain all the services if this is loop runs more than once.

test/e2e/upgrade_test.go
```diff
@@ -74,6 +79,7 @@ func upgradeTestSuite(t *testing.T) {
 	require.NoError(t, err, "wrong number of ConfigMap controller nodes found")
 
 	t.Run("Operator version upgrade", testUpgradeVersion)
+	t.Run("Dependent service reconcilation", testCtx.testDependentServiceChanges)
```
aravindhp, line 82: Why is this test in the upgrade test suite?

test/e2e/upgrade_test.go
```diff
@@ -105,6 +111,93 @@ func testUpgradeVersion(t *testing.T) {
...
+		[]byte(patchData), metav1.PatchOptions{})
+	if err != nil {
+		return errors.Wrap(err, "error patching node annotation back to expected")
+	}
+	return nil
+}
+
+// changeHybridOverlayCommandVerbosity will change the loglevel argument in the hybrid-overlay-node command to a
+// different level
+func changeHybridOverlayCommandVerbosity(in string) (string, error) {
```
aravindhp, line 180: Please be more descriptive about the input and output.

aravindhp (changes requested):
Nice work, @sebsoto

pkg/daemon/manager/manager.go
```diff
@@ -109,6 +146,60 @@ func (m *manager) EnsureServiceState(service winsvc.Service, state svc.State) er
...
+	// the first run.
+
+	// Borrowing the main steps done here from the golang windows/mgr library's ListServices() function, as the
+	// EnumServicesStatusEx syscall has a very similar way of being called.
+	var buf []byte
+	var bytesNeeded, servicesReturned uint32
+	for {
+		var p *byte
+		if len(buf) > 0 {
+			p = &buf[0]
```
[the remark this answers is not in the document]
sebsoto: p is the start of the array contained within buf, which is not guaranteed to be allocated with enough space for the returned services. This is why the syscall may need to be called multiple times, as we can only correctly alloc buf if we know how many bytes are required to store the returned data.

Why we need to do it every time, is on the first run, buf has size 0 there will not be a 0th element to point to, and on subsequent runs if buf is realloc'd the location of the array will change.

I'm fairly confident in whats being done in this function, due to it being borrowed from the [official golang windows package](https://cs.opensource.google/go/x/sys/+/refs/tags/v0.1.0:windows/svc/mgr/mgr.go;l=176), who's syscall behaves in the same way.

pkg/daemon/manager/manager.go
sebsoto: Commented

pkg/daemon/manager/manager.go
sebsoto: servicesReturned has been renamed to returnedServiceCount to help confusion.
buf is what will contain the returned services. Renaming it to `serviceBuffer`


Added comment below as well:
```
			// a non-zero return value indicates the syscall completed successfully, and buf has been filled with the
			// requested data.
```

pkg/daemon/manager/manager.go
sebsoto: That would be caught by the compiler and cause a build failure

test/e2e/upgrade_test.go
sebsoto: Somewhat arbitrary, but I figured this specific situation is more likely in an upgrade scenario. I'm fine moving it to elsewhere if you think it should be run on all platforms.

pkg/daemon/manager/manager.go
```diff
@@ -109,6 +146,60 @@ func (m *manager) EnsureServiceState(service winsvc.Service, state svc.State) er
 	return winsvc.WaitForState(service, state)
 }
 
+// enumDependentServicesSyscall is a wrapper around enumDependentServicesW.Call with the correct argument casting
```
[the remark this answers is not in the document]
sebsoto: Added doc link, I'd prefer to not rewrite the API docs

pkg/daemon/manager/manager.go
```diff
@@ -109,6 +146,55 @@ func (m *manager) EnsureServiceState(service winsvc.Service, state svc.State) er
 	return winsvc.WaitForState(service, state)
 }
 
+// listDependentServices returns a list of names of all services dependent on the given service
+func (m *manager) listDependentServices(serviceHandle windows.Handle) ([]string, error) {
+	// Borrowing the main steps done here from the golang windows/mgr library's ListServices() function, as the
```
saifshaikh48, line 151: Could you add the link (https://cs.opensource.google/go/x/sys/+/refs/tags/v0.1.0:windows/svc/mgr/mgr.go;l=176) within the comment?

alinaryan:
LGTM

test/e2e/upgrade_test.go
```diff
@@ -105,6 +111,94 @@ func testUpgradeVersion(t *testing.T) {
...
+	if err != nil {
+		return "", err
+	}
+	var newLogLevel int
+	if originalLogLevel == 0 {
+		newLogLevel = originalLogLevel + 1
+	} else {
+		newLogLevel = originalLogLevel - 1
+	}
+	return fmt.Sprintf("%s%d%s", loglevelSplit[0], newLogLevel, loglevelSplit[1][1:]), nil
```
saifshaikh48, line 199: Looks like the logic depends on `loglevel` always being the last arg and a single digit. Consider using a regex to simplify all this. Example: https://go.dev/play/p/_6LYdxD2-en

test/e2e/upgrade_test.go
saifshaikh48: I'm fine with having this as part of the upgrade tests as long as it doesn't greatly increase the time the suite takes to complete. No need to run on all platforms IMO

pkg/daemon/manager/manager.go
```diff
@@ -23,6 +29,17 @@ type Manager interface {
...
+// enumServiceStatus implements the ENUM_SERVICE_STATUS type as defined in the Windows API
+type enumServiceStatus struct {
+	ServiceName   *uint16
+	DisplayName   *uint16
+	ServiceStatus windows.SERVICE_STATUS
+}
+
+// enumDependentServicesW is a handle to the EnumDependentServicesW syscall
+// https://learn.microsoft.com/en-us/windows/win32/api/winsvc/nf-winsvc-enumdependentservicesw
+var enumDependentServicesW = windows.NewLazySystemDLL("Advapi32.dll").NewProc("EnumDependentServicesW")
```
aravindhp, line 41: Does this need to be a global variable? If there are performance implications that made you add this as a global variable, please call that out here.

pkg/daemon/manager/manager.go
```diff
@@ -98,6 +115,26 @@ func (m *manager) EnsureServiceState(service winsvc.Service, state svc.State) er
...
+		dependentServices, err := m.listDependentServices(winSvc.Handle)
+		if err != nil {
+			return errors.Wrap(err, "error finding dependent services")
+		}
+		for _, dependentServiceName := range dependentServices {
+			dependentSvc, err := m.OpenService(dependentServiceName)
+			if err != nil {
+				return fmt.Errorf("error opening dependent service %s", dependentServiceName)
+			}
+			err = m.EnsureServiceState(dependentSvc, svc.Stopped)
```
aravindhp, line 133: Is it possible to have interdependent services? While it will not happen with services we configuring but I am wondering about the BYOH case where the customer could set some service of theirs dependent on one of the WICD managed services and vice versa.

pkg/daemon/manager/manager.go
aravindhp: +1

test/e2e/upgrade_test.go
aravindhp: I have the same concern with the length of the upgrade test increasing. It is the prime candidate for overrunning the 4hr time out.

aravindhp (approved):
I have approved with a few questions. Nothing blocking though.

test/e2e/create_test.go
```diff
@@ -61,6 +61,7 @@ func creationTestSuite(t *testing.T) {
 	t.Run("Certificates", testCertificates)
 	t.Run("Node Logs", testNodeLogs)
 	t.Run("Metrics validation", testMetrics)
+	t.Run("Dependent service reconcilation", testDependentServiceChanges)
```
saifshaikh48, line 64: Won't this still run in the upgrade job, extending its runtime? Perhaps we can add a platform dependent skip for AWS

pkg/daemon/manager/manager.go
```diff
@@ -82,6 +83,32 @@ func (m *manager) DeleteService(name string) error {
...
+		if err != nil {
+			return err
+		}
+	case svc.Stopped:
+		_, err = service.Control(svc.Stop)
+		if err != nil {
+			return err
+		}
+	default:
+		return errors.New("unexpected state request")
```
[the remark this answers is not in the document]
saifshaikh48: This can still be addressed

saifshaikh48 (commented):
/lgtm overall, but the time concern may still be open

saifshaikh48 (commented):
/lgtm

Great stuff with this

saifshaikh48 (commented):
/lgtm

sebsoto:
All tests passed but this isnt merging. not sure whats going on with prow

PR #1316, opened 2022-11-07
OCPBUGS-1513: Fix check for Windows Defender

### LINKED ISSUE #1240: [Bug]:  Check if Defender is running is wrong
### Attention

- [X] I am using the community version of WMCO

### Version

OCP

### Cluster Version

4.11

### Platform

vSphere

### Proxy

Yes, without HTTPS

### WMCO Version

6.0.0

### Windows version

2022

### What happened?

The check, whether or not Defender is running is wrong. [SourceCode](https://github.com/openshift/windows-machine-config-operator/blob/5928b620ee7fdf3d1328e7d393ea8c6f467616c3/pkg/windows/windows.go#L1025)
The Defender Services are always there, so that check returns always "true".
As [mentioned in the Microsoft documentation](https://docs.microsoft.com/en-us/microsoft-365/security/defender-endpoint/microsoft-defender-antivirus-compatibility?view=o365-worldwide#use-windows-powershell-to-confirm-that-microsoft-defender-antivirus-is-running), it must be checked whether the process is running. 

That line needs to be changed to 
```powershell
Get-Process | where-object Name -eq 'MsMpEng.exe'
```

This is the result from the current command on a server where Defender DOES NOT run:
```powershell
PS > (Get-Service | where {$_.DisplayName -Like 'Windows Defender*'}) | ft  -AutoSize

Status  Name   DisplayName
------  ----   -----------
Running mpssvc Windows Defender Firewall
Stopped Sense  Windows Defender Advanced Threat Protection Service
```

And the logs from the operator:

```
{"level":"error","ts":1662711753.9641004,"logger":"wc __IP__","msg":"error running","cmd":"powershell.exe -NonInteractive -ExecutionPolicy Bypass \"Add-MpPreference -ExclusionProcess C:\\k\\containerd\\containerd.exe\"","out":"Add-MpPreference : The term 'Add-MpPreference' is not recognized as the name of a cmdlet, function, script file, or \r\noperable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try \r\nagain.\r\nAt line:1 char:1\r\n+ Add-MpPreference -ExclusionProcess C:\\k\\containerd\\containerd.exe\r\n+ ~~~~~~~~~~~~~~~~\r\n    + CategoryInfo          : ObjectNotFound: (Add-MpPreference:String) [], CommandNotFoundException\r\n    + FullyQualifiedErrorId : CommandNotFoundException\r\n \r\n","error":"Process exited with status 1","stacktrace":"github.com/openshift/windows-machine-config-operator/pkg/windows.(*windows).createWindowsDefenderExclusion\n\t/remote-source/build/windows-machine-config-operator/pkg/windows/windows.go:1056\ngithub.com/openshift/windows-machine-config-operator/pkg/windows.(*windows).configureContainerd\n\t/remote-source/build/windows-machine-config-operator/pkg/windows/windows.go:455\ngithub.com/openshift/windows-machine-config-operator/pkg/windows.(*windows).Configure\n\t/remote-source/build/windows-machine-config-operator/pkg/windows/windows.go:436\ngithub.com/openshift/windows-machine-config-operator/pkg/nodeconfig.(*nodeConfig).Configure\n\t/remote-source/build/windows-machine-config-operator/pkg/nodeconfig/nodeconfig.go:197\ngithub.com/openshift/windows-machine-config-operator/controllers.(*instanceReconciler).ensureInstanceIsUpToDate\n\t/remote-source/build/windows-machine-config-operator/controllers/controllers.go:91\ngithub.com/openshift/windows-machine-config-operator/controllers.(*ConfigMapReconciler).ensureInstancesAreUpToDate\n\t/remote-source/build/windows-machine-config-operator/controllers/configmap_controller.go:293\ngithub.com/openshift/windows-machine-config-operator/controllers.(*ConfigMapReconciler).reconcileNodes\n\t/remote-source/build/windows-machine-config-operator/controllers/configmap_controller.go:258\ngithub.com/openshift/windows-machine-config-operator/controllers.(*ConfigMapReconciler).Reconcile\n\t/remote-source/build/windows-machine-config-operator/controllers/configmap_controller.go:153\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile\n\t/remote-source/build/windows-machine-config-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:121\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/remote-source/build/windows-machine-config-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:320\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/remote-source/build/windows-machine-config-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:273\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/remote-source/build/windows-machine-config-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:234"}
```

As you can see the defender module is not present at all on our servers. 

We do not use community version of the operator, but I believe this issue should be visible here..

### What did you expect to happen?

The check whether or not the defender is installed should not be based on the number of services.

### Steps to reproduce the issue

Install a server and disable Defender. 

### Do you have a workaround for this issue?

Created a "fake" powershell module with an empty defender pscmdlet...
But it can not be a solution for our production environment.
It's an absolute show stopper for our migration to 2022.

saifshaikh48:
Before, the check to see if Windows Defender antivirus was running was wrongly
checking for any services whose name starts with Windows Defender regardless of
state, leading to an error creating an exclusion for containerd on instances
without Defender installed. 

This PR fixes this by checking for the presence of the specific running process 
associated with the Defender antivirus service.
It also adds an e2e test to validate instances are able to be configured into
Nodes regardless of the Windows Defender antivirus status.

Fixes OCPBUGS-1513
Fixes #1240

### DIFF
--- a/pkg/windows/windows.go
+++ b/pkg/windows/windows.go
@@ -487,8 +487,8 @@ func (vm *windows) Bootstrap(desiredVer, apiServerURL string, credentials *Authe
 // configureContainerd configures the Windows defender exclusion and starts the
 // Windows containerd service
 func (vm *windows) configureContainerd() error {
-	// if Windows Defender is installed on the instance, create an exclusion for containerd
-	exclusionNeeded, err := vm.isWindowsDefenderEnabled()
+	// if Windows Defender is running on the instance, create an exclusion for containerd
+	exclusionNeeded, err := vm.isWindowsDefenderRunning()
 	if err != nil {
 		return err
 	}
@@ -930,9 +930,11 @@ func (vm *windows) isContainersFeatureEnabled() (bool, error) {
 	return strings.Contains(out, "Enabled"), nil
 }
 
-// isWindowsDefenderEnabled returns true if the Windows Defender antivirus/firewall is installed on the Windows instance
-func (vm *windows) isWindowsDefenderEnabled() (bool, error) {
-	command := "(Get-Service | where {$_.DisplayName -Like 'Windows Defender*'}).Count -gt 0"
+// isWindowsDefenderRunning returns true if the Windows Defender antivirus/firewall is running on the Windows instance
+func (vm *windows) isWindowsDefenderRunning() (bool, error) {
+	// Check if the process associated with Windows Defender exists. Reference:
+	// https://learn.microsoft.com/en-us/microsoft-365/security/defender-endpoint/microsoft-defender-antivirus-compatibility?view=o365-worldwide#use-windows-powershell-to-confirm-that-microsoft-defender-antivirus-is-running
+	command := "(Get-Process -Name MsMpEng -ErrorAction SilentlyContinue) -ne $null"
 	out, err := vm.Run(command, true)
 	if err != nil {
 		return false, errors.Wrap(err, "error checking if Windows Defender is enabled")
--- a/test/e2e/create_test.go
+++ b/test/e2e/create_test.go
@@ -245,6 +245,14 @@ func (tc *testContext) provisionBYOHConfigMapWithMachineSet() error {
 	if err != nil {
 		return errors.Wrapf(err, "unable to change default shell of machine %s", machines.Items[0].GetName())
 	}
+	if tc.CloudProvider.GetType() == config.VSpherePlatformType {
+		// By default the Windows Defender antivirus is running on VMs in the test suite. Disable it for a BYOH node to
+		// validate instances are able to be configured into Nodes regardless of the Windows Defender antivirus status.
+		err = tc.uninstallWindowsDefender(&machines.Items[0])
+		if err != nil {
+			return errors.Wrapf(err, "unable to uninstall Windows Defender on machine %s", machines.Items[0].GetName())
+		}
+	}
 	return tc.createWindowsInstanceConfigMap(machines)
 }
 
@@ -302,6 +310,20 @@ func (tc *testContext) setPowerShellDefaultShell(machine *mapi.Machine) error {
 		}
 		return true, nil
 	})
+}
+
+// uninstallWindowsDefender uninstalls Windows Defender from the given machine. Reboots the underlying VM.
+func (tc *testContext) uninstallWindowsDefender(machine *mapi.Machine) error {
+	addr, err := controllers.GetAddress(machine.Status.Addresses)
+	if err != nil {
+		return err
+	}
+	command := "Uninstall-WindowsFeature -Name Windows-Defender; Restart-Computer -Force"
+	_, err = tc.runPowerShellSSHJob("uninstall-windows-defender", command, addr)
+	if err != nil {
+		return errors.Wrap(err, "failed to uninstall Windows Defender antivirus")
+	}
+	return nil
 }
 
 // createWindowsMachineSet creates given number of Windows Machines.


pkg/windows/windows.go
```diff
@@ -932,7 +932,10 @@ func (vm *windows) isContainersFeatureEnabled() (bool, error) {
 
 // isWindowsDefenderEnabled returns true if the Windows Defender antivirus/firewall is installed on the Windows instance
 func (vm *windows) isWindowsDefenderEnabled() (bool, error) {
-	command := "(Get-Service | where {$_.DisplayName -Like 'Windows Defender*'}).Count -gt 0"
+	// Check if process associated with Windows Defender does not exist. Reference:
+	// https://learn.microsoft.com/en-us/microsoft-365/security/defender-endpoint/microsoft-defender-antivirus-compatibility?view=o365-worldwide#use-windows-powershell-to-confirm-that-microsoft-defender-antivirus-is-running
+	winDefProcess := "MsMpEng.exe"
+	command := "(Get-Process -Name " + winDefProcess + " -ErrorAction SilentlyContinue) -eq $null"
 	out, err := vm.Run(command, true)
```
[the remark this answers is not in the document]
saifshaikh48: If there is no process with a name equal to MsMpEng.exe that means the Microsoft defender antivirus is not running (and likewise, if there is a MsMpEng.exe process then Defender is running). I don't follow the ask here.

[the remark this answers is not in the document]
jrvaldes: Correct. Manually debugging this code to expose my concern.

`out` is equal to `"True"` if there is no process with a name equal to `MsMpEng.exe`, which means Microsoft defender antivirus **is not** running and the function returns `strings.TrimSpace(out) == "True"` that evaluate to `true`. 

So, `isWindowsDefenderRunning()` returning `true`  means the Microsoft defender antivirus **is** running


IMO, `isWindowsDefenderRunning()` should returns `false` if there is no process with a name equal to `MsMpEng.exe`.


```go
        ...
	out, err := vm.Run(command, true)
	if err != nil {
		return false, errors.Wrap(err, "error checking if Windows Defender is enabled")
	}
	return strings.TrimSpace(out) == "True", nil
```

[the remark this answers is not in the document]
saifshaikh48: Thanks for the explanation, I see now. I need to flip the condition i.e. `return strings.TrimSpace(out) == "False"`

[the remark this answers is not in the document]
jrvaldes: You may not need a conditional statement at all in the PowerShell command. You can compare with the output of `Get-Process -Name MsMpEng.exe` directly.

pkg/windows/windows.go
```diff
@@ -930,9 +930,11 @@ func (vm *windows) isContainersFeatureEnabled() (bool, error) {
 	return strings.Contains(out, "Enabled"), nil
 }
 
-// isWindowsDefenderEnabled returns true if the Windows Defender antivirus/firewall is installed on the Windows instance
-func (vm *windows) isWindowsDefenderEnabled() (bool, error) {
-	command := "(Get-Service | where {$_.DisplayName -Like 'Windows Defender*'}).Count -gt 0"
+// isWindowsDefenderRunning returns true if the Windows Defender antivirus/firewall is enabled on the Windows instance
```
jrvaldes, line 933:
```suggestion
// isWindowsDefenderRunning returns true if the Windows Defender antivirus/firewall is running on the Windows instance
```

saifshaikh48: Enabled == running according to [MS docs](https://learn.microsoft.com/en-us/microsoft-365/security/defender-endpoint/microsoft-defender-antivirus-compatibility?view=o365-worldwide#use-windows-powershell-to-confirm-that-microsoft-defender-antivirus-is-running)

> Use Windows PowerShell to confirm that Microsoft Defender Antivirus is running
> - You should see MsMpEng.exe if Microsoft Defender Antivirus is enabled.

saifshaikh48:
I've found that [turning off Windows Defender](https://learn.microsoft.com/en-us/microsoft-365/security/defender-endpoint/microsoft-defender-antivirus-on-windows-server?view=o365-worldwide#passive-mode-and-windows-server) is tricky. Just disabling it does not seem to actually stop it, and manually trying to stop the associated service gives a "permission denied" error. The best way I've found to work is uninstalling the module completely.

jrvaldes:
> I've found that [turning off Windows Defender](https://learn.microsoft.com/en-us/microsoft-365/security/defender-endpoint/microsoft-defender-antivirus-on-windows-server?view=o365-worldwide#passive-mode-and-windows-server) is tricky. Just disabling it does not seem to actually stop it, and manually trying to stop the associated service gives a "permission denied" error. The best way I've found to work is uninstalling the module completely.

Are you thinking about the e2e test?

saifshaikh48:
> Are you thinking about the e2e test?

@jrvaldes Yup. Interestingly I'm seeing that uninstalling Windows Defender hits [permission issues on AWS](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/1316/pull-ci-openshift-windows-machine-config-operator-master-aws-e2e-upgrade/1590109180977483776/artifacts/aws-e2e-upgrade/e2e-aws-upgrade/artifacts/pods/job-name=uninstall-windows-defender-job-jsccq/job-name=uninstall-windows-defender-job-jsccq.log). Works fine on Azure and vSphere

mansikulkarni96:
Fixes #1240

test/e2e/create_test.go
```diff
@@ -304,6 +312,20 @@ func (tc *testContext) setPowerShellDefaultShell(machine *mapi.Machine) error {
 	})
 }
 
+// disableWindowsDefender uninstalls Windows Defender from the given machine. Reboots the underlying VM.
```
aravindhp, line 315: You are uninstalling the feature and naming the function disable which is misleading. So how about `uninstallWindowsDefender`?

pkg/windows/windows.go
aravindhp: I can see Windows Defender being enabled i.e. the service is set to start but not auto-restart and the process has crashed and hence is not running. 

PS: every pattern that you see in "upstream" or other documentation need not be a good pattern and we should use our judgement.

test/e2e/create_test.go
```diff
@@ -245,6 +245,14 @@ func (tc *testContext) provisionBYOHConfigMapWithMachineSet() error {
...
 		return errors.Wrapf(err, "unable to change default shell of machine %s", machines.Items[0].GetName())
 	}
+	if tc.CloudProvider.GetType() == config.VSpherePlatformType {
+		// By default the Windows Defender antivirus is running on VMs in the test suite. Disable it for a BYOH node to
+		// validate instances are able to be configured into Nodes regardless of the Windows Defender antivirus status.
+		err = tc.disableWindowsDefender(&machines.Items[0])
+		if err != nil {
+			return errors.Wrapf(err, "unable to disable Windows Defender on machine %s", machines.Items[0].GetName())
+		}
+	}
```
aravindhp, line 255: Do you need a delay or ideally a check if Machine is `running`  here before creating the ConfigMap to avoid potential flakes? I would think not but just checking. My worry is SSH starts too early and we start seeing red herrings in the logs that we will only notice debugging an orthogonal issue.

aravindhp (changes requested):
Thanks for the quick turn around with this, @saifshaikh48

pkg/windows/windows.go
saifshaikh48: I'll re-word to consistently use `running`

test/e2e/create_test.go
saifshaikh48: I had considered this and as far as I can see, the wait+check is not needed. After we create the ConfigMap, the operator will try to establish SSH connection to the machine, [retrying for up to 10 mins](https://github.com/openshift/windows-machine-config-operator/blob/master/pkg/windows/connectivity.go#L87) which is far greater than the time it takes for a machine to reboot. During all this, the test suite itself is [waiting for the BYOH instance to be configured as a node (for up to 20 mins)](https://github.com/openshift/windows-machine-config-operator/blob/master/test/e2e/create_test.go#L198-L203) so the risk of flaking is minimal.

test/e2e/create_test.go
```diff
@@ -304,6 +312,20 @@ func (tc *testContext) setPowerShellDefaultShell(machine *mapi.Machine) error {
...
+// disableWindowsDefender uninstalls Windows Defender from the given machine. Reboots the underlying VM.
+func (tc *testContext) disableWindowsDefender(machine *mapi.Machine) error {
+	addr, err := controllers.GetAddress(machine.Status.Addresses)
+	if err != nil {
+		return errors.Wrapf(err, "Machine %s does not have a valid address", machine.GetName())
+	}
+	command := "Uninstall-WindowsFeature -Name Windows-Defender; Restart-Computer -Force"
+	_, err = tc.runPowerShellSSHJob("uninstall-windows-defender", command, addr)
+	if err != nil {
+		return errors.Wrapf(err, "failed to uninstall Windows Defender antivirus on machine %s", machine.GetName())
```
sebsoto, line 324: The Machine name is being logged redundantly as it is already done at the callsite. Please remove it from these error messages

sebsoto (changes requested):
Thanks @saifshaikh48 mainly LGTM

pkg/windows/windows.go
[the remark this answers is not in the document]
saifshaikh48: The condition seems to be the easiest way. Comparing with the output would require Golang-based Powershell error discernment, which I'd like to avoid.

test/e2e/create_test.go
```diff
@@ -304,6 +312,20 @@ func (tc *testContext) setPowerShellDefaultShell(machine *mapi.Machine) error {
...
+// uninstallWindowsDefender uninstalls Windows Defender from the given machine. Reboots the underlying VM.
+func (tc *testContext) uninstallWindowsDefender(machine *mapi.Machine) error {
+	addr, err := controllers.GetAddress(machine.Status.Addresses)
+	if err != nil {
+		return err
+	}
+	command := "Uninstall-WindowsFeature -Name Windows-Defender; Restart-Computer -Force"
+	_, err = tc.runPowerShellSSHJob("uninstall-windows-defender", command, addr)
+	if err != nil {
+		return errors.Wrapf(err, "failed to uninstall Windows Defender antivirus", machine.GetName())
```
sebsoto, line 324: missed variable removal here
could do errors.Wrap as well if you want

sebsoto (changes requested):

test/e2e/create_test.go
saifshaikh48: Yep thanks, fixed

saifshaikh48:
/test vsphere-e2e-operator

cluster spinup failure

test/e2e/create_test.go
```diff
@@ -245,6 +245,14 @@ func (tc *testContext) provisionBYOHConfigMapWithMachineSet() error {
 	if err != nil {
 		return errors.Wrapf(err, "unable to change default shell of machine %s", machines.Items[0].GetName())
 	}
+	if tc.CloudProvider.GetType() == config.VSpherePlatformType {
+		// By default the Windows Defender antivirus is running on VMs in the test suite. Disable it for a BYOH node to
```
aravindhp, line 249: s\Disable\Uninstall

aravindhp (approved):

mtnbikenc:
Cluster provisioning failed.
/test vsphere-e2e-operator

saifshaikh48:
Cluster spinup failure
/test vsphere-e2e-operator

PR #1323, opened 2022-11-10
WINC-941: [services] Use IMDSv1 to get hostname in AWS

jrvaldes:
Before, the AWS Tools for PowerShell library (https://github.com/aws/aws-tools-for-powershell)
was used to retrieve the local hostname of the Windows instances in AWS.

This change propose the instance metadata service version 1 (IMDSv1) endpoint
since the library may not be available, especially in the BYOH use case.

### DIFF
--- a/pkg/services/services.go
+++ b/pkg/services/services.go
@@ -223,7 +223,9 @@ func klogVerbosityArg(debug bool) string {
 func getHostnameCmd(platformType config.PlatformType) string {
 	switch platformType {
 	case config.AWSPlatformType:
-		return "Get-EC2InstanceMetadata -Category LocalHostname"
+		// Use the Instance Metadata Service Version 1 (IMDSv1) to fetch the hostname. IMDSv1 will continue to be
+		// supported indefinitely as per AWS docs. https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html
+		return "Invoke-RestMethod -UseBasicParsing -Uri http://169.254.169.254/latest/meta-data/local-hostname"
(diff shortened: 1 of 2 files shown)


pkg/services/services.go
```diff
@@ -223,7 +223,9 @@ func klogVerbosityArg(debug bool) string {
 func getHostnameCmd(platformType config.PlatformType) string {
 	switch platformType {
 	case config.AWSPlatformType:
-		return "Get-EC2InstanceMetadata -Category LocalHostname"
+		// For the instance metadata service to work on BYOH Windows instances using a custom Windows AMI, the AMI must
+		// be a standardized image created using the AWS's sysprep workflow. https://docs.aws.amazon.com/AWSEC2/latest/WindowsGuide/Creating_EBSbacked_WinAMI.html#ami-create-standard
```
aravindhp, line 227: What does the doc have to do with the metadata endpoint?

aravindhp: Please add the doc for the metadata endpoint here.

aravindhp (changes requested):

pkg/services/services.go
aravindhp: In addition are using IMDS v1 or v2?

aravindhp (approved):

pkg/services/services.go
jrvaldes: > What does the doc have to do with the metadata endpoint?

Nothing. Not sure where that came from 👽

saifshaikh48 (commented):
/lgtm

PR #1330, opened 2022-11-21
OCPBUGS-3203: [metrics] Add check for endpoints subsets address

jrvaldes:
This change includes checks to validate the addresses in the metric endpoint subsets before looking for a matching node name, so that prevents a NPE if the field is not present.

Fixes OCPBUGS-3203

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -194,6 +194,11 @@ func isEndpointsValid(nodes *v1.NodeList, endpoints *v1.Endpoints) bool {
 	for _, node := range nodes.Items {
 		nodeFound := false
 		for _, address := range endpoints.Subsets[0].Addresses {
+			// check TargetRef is present and has the expected kind
+			if address.TargetRef == nil || address.TargetRef.Kind != "Node" {
+				// otherwise, skip the invalid address
+				continue
+			}
 			if address.TargetRef.Name == node.Name {
 				nodeFound = true
 				break
--- a/pkg/metrics/metrics_test.go
+++ b/pkg/metrics/metrics_test.go
@@ -1,0 +1,201 @@
+package metrics
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"k8s.io/api/core/v1"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestIsEndpointsValid(t *testing.T) {
+	tests := []struct {
+		name      string
+		nodes     *v1.NodeList
+		endpoints *v1.Endpoints
+		want      bool
+	}{
+		{
+			name: "EndpointAddresses match nodes count and name",
+			nodes: &v1.NodeList{
+				Items: []v1.Node{
+					{
+						ObjectMeta: meta.ObjectMeta{Name: "the-node-name"},
+					},
+				},
+			},
+			endpoints: &v1.Endpoints{
+				Subsets: []v1.EndpointSubset{
+					{Addresses: []v1.EndpointAddress{
+						{
+							TargetRef: &v1.ObjectReference{
+								Kind: "Node",
+								Name: "the-node-name",
+							},
+						},
+					}},
+				},
+			},
+			want: true,
+		},
+		{
+			name: "EndpointAddresses match two nodes",
+			nodes: &v1.NodeList{
+				Items: []v1.Node{
+					{
+						ObjectMeta: meta.ObjectMeta{Name: "the-node-name-one"},
+					},
+					{
+						ObjectMeta: meta.ObjectMeta{Name: "the-node-name-two"},
+					},
+				},
+			},
+			endpoints: &v1.Endpoints{
+				Subsets: []v1.EndpointSubset{
+					{
+						Addresses: []v1.EndpointAddress{
+							{
+								TargetRef: &v1.ObjectReference{
+									Kind: "Node",
+									Name: "the-node-name-two",
+								},
+							},
+							{
+								TargetRef: &v1.ObjectReference{
+									Kind: "Node",
+									Name: "the-node-name-one",
+								},
+							},
+						},
+					},
+				},
+			},
+			want: true,
+		},
+		{
+			name: "No Endpoint Subsets",
+			nodes: &v1.NodeList{
+				Items: []v1.Node{
+					{
+						ObjectMeta: meta.ObjectMeta{Name: "the-node-name"},
+					},
+				},
+			},
+			endpoints: &v1.Endpoints{
+				Subsets: []v1.EndpointSubset{},
+			},
+			want: false,
+		},
+		{
+			name: "No nodes",
+			nodes: &v1.NodeList{
+				Items: []v1.Node{},
+			},
+			endpoints: &v1.Endpoints{
+				Subsets: []v1.EndpointSubset{
+					{Addresses: []v1.EndpointAddress{
+						{
+							TargetRef: &v1.ObjectReference{
+								Kind: "Node",
+								Name: "the-node-name",
+							},
+						},
+					}},
+				},
+			},
+			want: false,
+		},
+		{
+			name: "EndpointAddress does not match node name",
+			nodes: &v1.NodeList{
+				Items: []v1.Node{
+					{
+						ObjectMeta: meta.ObjectMeta{Name: "the-node-name"},
+					},
+				},
+			},
+			endpoints: &v1.Endpoints{
+				Subsets: []v1.EndpointSubset{
+					{Addresses: []v1.EndpointAddress{
+						{
+							TargetRef: &v1.ObjectReference{
+								Kind: "Node",
+								Name: "wrong-node-name",
+							},
+						},
+					}},
+				},
+			},
+			want: false,
+		},
+		{
+			name: "EndpointAddress without targetRef",
+			nodes: &v1.NodeList{
+				Items: []v1.Node{
+					{
+						ObjectMeta: meta.ObjectMeta{Name: "the-node-name"},
+					},
+				},
+			},
+			endpoints: &v1.Endpoints{
+				Subsets: []v1.EndpointSubset{
+					{Addresses: []v1.EndpointAddress{
+						{
+							IP: "1.2.3.4",
+						},
+					}},
+				},
+			},
+			want: false,
+		},
+		{
+			name: "EndpointAddress with targetRef without name",
+			nodes: &v1.NodeList{
+				Items: []v1.Node{
+					{
+						ObjectMeta: meta.ObjectMeta{Name: "the-node-name"},
+					},
+				},
+			},
+			endpoints: &v1.Endpoints{
+				Subsets: []v1.EndpointSubset{
+					{Addresses: []v1.EndpointAddress{
+						{
+							TargetRef: &v1.ObjectReference{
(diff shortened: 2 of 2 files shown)


pkg/metrics/metrics.go
```diff
@@ -194,6 +194,14 @@ func isEndpointsValid(nodes *v1.NodeList, endpoints *v1.Endpoints) bool {
 	for _, node := range nodes.Items {
 		nodeFound := false
 		for _, address := range endpoints.Subsets[0].Addresses {
+			if address.TargetRef == nil {
+				// TargetRef not found
+				continue
+			}
+			if address.TargetRef.Kind != "Node" {
```
saifshaikh48, line 201: nit: consider collapsing the two if statements 
```golang
if address.TargetRef == nil || address.TargetRef.Kind != "Node" {
    // TargetRef not found or invalid
    continue
}
```

pkg/metrics/metrics.go
```diff
@@ -194,6 +194,14 @@ func isEndpointsValid(nodes *v1.NodeList, endpoints *v1.Endpoints) bool {
 	for _, node := range nodes.Items {
 		nodeFound := false
 		for _, address := range endpoints.Subsets[0].Addresses {
+			if address.TargetRef == nil {
```
saifshaikh48, line 197: I'm curious why we use the `TargetRef` to check the node name rather than the [`NodeName` field](https://pkg.go.dev/k8s.io/api/core/v1#EndpointAddress.NodeName)?

saifshaikh48 (commented):
@jrvaldes Thanks for fixing this

pkg/metrics/metrics.go
mansikulkarni96: Not 100% sure but probably that can be nil in some cases, could be verified and used if that's not the case.

aravindhp: Do we know why we ended up in a scenario where the `TargetRef` was nil?

>I'm curious why we use the TargetRef to check the node name rather than the [NodeName field](https://pkg.go.dev/k8s.io/api/core/v1#EndpointAddress.NodeName)?

I have the same question. I would think if `TargetRef` is nil, `NodeName` also would be `nil` but we should confirm that. Either way we need to have `nil` checks.

jrvaldes: > Do we know why we ended up in a scenario where the TargetRef was nil?

IDK.

Below the object extracted from the case, where only the `ip` is present in the list of addresses.

```yaml
---
apiVersion: v1
kind: EndpointsList
items:
- apiVersion: v1
  kind: Endpoints
  metadata:
    creationTimestamp: "2022-10-27T06:53:02Z"
    labels:
      name: windows-exporter
    name: windows-exporter
    namespace: openshift-windows-machine-config-operator
    resourceVersion: "136231886"
    uid: e4a3e21d-cdcb-4de9-907d-5a9e908aa27c
  subsets:
  - addresses:
    - ip: 10.0.38.87
    - ip: 10.0.38.88
    - ip: 10.0.38.85
    - ip: 10.0.38.84
    - ip: 10.0.38.86
    ports:
    - name: metrics
      port: 9182
      protocol: TCP

```

jrvaldes: > I would think if TargetRef is nil, NodeName also would be nil but we should confirm that.

In the [docs](https://docs.openshift.com/container-platform/4.11/rest_api/network_apis/endpoints-v1.html#subsets-addresses-2), looks like `TargetRef`  and `NodeName` are independent. 

`NodeName` is [not in use ](https://github.com/openshift/windows-machine-config-operator/blob/c8f92e7244d5a41af51302e715a846efdd7468d3/pkg/metrics/metrics.go#L173) by WMCO as it is set to `nil`.  Hence, only `TargetRef` requires validation.

pkg/metrics/metrics.go
jrvaldes: Addressed.

saifshaikh48 (commented):
/lgtm

pkg/metrics/metrics.go
```diff
@@ -194,6 +194,11 @@ func isEndpointsValid(nodes *v1.NodeList, endpoints *v1.Endpoints) bool {
 	for _, node := range nodes.Items {
 		nodeFound := false
 		for _, address := range endpoints.Subsets[0].Addresses {
+			// check TargetRef is present and have the expected kind
```
sebsoto, line 197: have -> has

pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,203 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"k8s.io/api/core/v1"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const nodeName = "NodeName"
```
sebsoto, line 11: I don't see where this is used

sebsoto (changes requested):

pkg/metrics/metrics.go
jrvaldes: Fixed

pkg/metrics/metrics_test.go
jrvaldes: Good catch, forgot to remove it. Fixed.

pkg/metrics/metrics.go
aravindhp: >Below the object extracted from the case, where only the ip is present in the list of addresses.

Please investigate the controller that manages `EndpointList` and ensure this is not pointing at bug in another component. This can be done independently of fixing this bug. We need to figure why the TargetRef` is being removed.

>NodeName is [not in use ](https://github.com/openshift/windows-machine-config-operator/blob/c8f92e7244d5a41af51302e715a846efdd7468d3/pkg/metrics/metrics.go#L173)by WMCO as it is set to nil. Hence, only TargetRef requires validation.

Yes I know that and my question if we should be setting that. PS: if yes, then it needs to be done independently in a story under the monitoring epic.

sebsoto (approved):

jfrancoa:
@jrvaldes Could you help me with some tips to validate this issue? the customer was using a UPI cluster with vSphere, however I see from the fix, that the source of the problem was in the metrics code...how could I reproduce it so I could validate that the fix works as expected?

saifshaikh48:
@jfrancoa You could edit the [`Endpoints` resource](https://pkg.go.dev/k8s.io/api/core/v1#Endpoints) named `windows-exporter` in the WMCO namespace, adding [a Subset entry with an Address that lacks a TargetRef](https://github.com/openshift/windows-machine-config-operator/pull/1330/files#diff-c3ade6c0a09b6258d971210f5a41c2ae5beb77873cee44a9ccb78055caf271ddR140-R148). Then trigger a reconfiguration (e.g. change version annotation on any Windows node). This will trigger metrics re-configuration as well, reproducing the error

aravindhp:
I wonder if we should have added an e2e test that did this.

jfrancoa:
@saifshaikh48 I was performing the test in two clusters, one with GCP using OCP 4.13 (and WMCO which includes the fix) and another on AWS using 4.12 (with WCMO without the fix). So my approach was, I will try to reproduce it first in the cluster which does not include the fix, however I couldn't reproduce it. As soon as I modify the windows-exporter, WMCO restores it back during the reconciliation whithout spitting any error:

$ oc get endpoints windows-exporter -n openshift-windows-machine-config-operator  -o yaml
apiVersion: v1
kind: Endpoints
metadata:
  creationTimestamp: "2022-12-01T06:48:00Z"
  labels:
    name: windows-exporter
  name: windows-exporter
  namespace: openshift-windows-machine-config-operator
  resourceVersion: "147708"
  uid: 4aa2cb8f-c17c-4c48-aa1a-d8ddedd9c66b
subsets:
- addresses:
  - ip: 10.0.153.109
    targetRef:
      kind: Node
      name: ip-10-0-153-109.us-east-2.compute.internal
  - ip: 10.0.155.99
    targetRef:
      kind: Node
      name: ip-10-0-155-99.us-east-2.compute.internal
  - ip: 10.0.155.101
  ports:
  - name: metrics
    port: 9182
    protocol: TCP

$ oc edit node ip-10-0-153-109.us-east-2.compute.internal
node/ip-10-0-153-109.us-east-2.compute.internal edited

$ oc get endpoints windows-exporter -n openshift-windows-machine-config-operator  -o yaml
apiVersion: v1
kind: Endpoints
metadata:
  creationTimestamp: "2022-12-01T06:48:00Z"
  labels:
    name: windows-exporter
  name: windows-exporter
  namespace: openshift-windows-machine-config-operator
  resourceVersion: "148287"
  uid: 4aa2cb8f-c17c-4c48-aa1a-d8ddedd9c66b
subsets:
- addresses:
  - ip: 10.0.155.99
    targetRef:
      kind: Node
      name: ip-10-0-155-99.us-east-2.compute.internal
  ports:
  - name: metrics
    port: 9182
    protocol: TCP

This occurs when this is logged:

{"level":"info","ts":"2022-12-01T11:43:32Z","logger":"metrics","msg":"Prometheus configured","endpoints":"windows-exporter","port":9182,"name":"metrics"}
{"level":"info","ts":"2022-12-01T11:44:02Z","logger":"metrics","msg":"Prometheus configured","endpoints":"windows-exporter","port":9182,"name":"metrics"}
{"level":"info","ts":"2022-12-01T11:45:51Z","logger":"controller.windowsmachine","msg":"deleting machine","windowsmachine":"openshift-machine-api/jfrancoa-0112-92cjm-windows-worker-us-east-2a-jg7nk"}
{"level":"info","ts":"2022-12-01T11:45:51Z","logger":"controller.windowsmachine","msg":"unhealthy machine count for machineset","name":"jfrancoa-0112-92cjm-windows-worker-us-east-2a","total":2,"unhealthy":0}
{"level":"info","ts":"2022-12-01T11:45:51Z","logger":"controller.windowsmachine","msg":"machine has been remediated by deletion","name":"jfrancoa-0112-92cjm-windows-worker-us-east-2a-jg7nk"}
{"level":"info","ts":"2022-12-01T11:46:21Z","logger":"metrics","msg":"Prometheus configured","endpoints":"windows-exporter","port":9182,"name":"metrics"}
{"level":"info","ts":"2022-12-01T11:46:51Z","logger":"metrics","msg":"Prometheus configured","endpoints":"windows-exporter","port":9182,"name":"metrics"}

I tried adding the ip field in the existing subsets.addresses and also creating a new -addresses section with its own ip field (without targetRef), but no sign of the panic. Am I doing it wrong?

jrvaldes:
> Am I doing it wrong?

@jfrancoa  I suggest you scale down WMCO deployment so that there are no WMCO pods running. Perform the modifications in the endpoint object and scale back up the WMCO deployment. [Here](https://github.com/openshift/windows-machine-config-operator/pull/1330#discussion_r1028806590) a sample endpoint object where WMCO panic while starting.

jfrancoa:
@jrvaldes thanks a lot! that's very helpful, I will get to it.

jfrancoa:
Nice! that worked. I had to delete the windows-exporter after scaling down WMCO and then create another CSR with a new Endpoint kind CSR with addresses without targetRef. This way it didn't override it (probably the CSR was still in etcd and it was just rewritting it) and the panic appeared. I will go ahead and validate the bug. Thanks both for your help.

PR #1332, opened 2022-11-22
WINC-733: Instance cleanup is done by WICD cleanup command

mansikulkarni96:
With this PR, WMCO starts invoking WICD cleanup command to stop and delete services and delete the node object.
Responsibility of removing directories and HNS networks is mainatined by WMCO while excluding deletion of log directories for future reference.

### DIFF
--- a/pkg/daemon/cleanup/cleanup.go
+++ b/pkg/daemon/cleanup/cleanup.go
@@ -33,10 +33,10 @@ import (
 	"github.com/openshift/windows-machine-config-operator/pkg/servicescm"
 )
 
-// Deconfigure removes an instance from the cluster. If preserveNode is true, the node object is not deleted.
+// Deconfigure removes an instance from the cluster.
 // If we are able to get the services ConfigMap tied to the desired version, all services defined in it are cleaned up.
 // TODO: Otherwise, perform cleanup based on a combination of the OpenShift managed tag and latest services ConfigMap.
-func Deconfigure(cfg *rest.Config, ctx context.Context, preserveNode bool, configMapNamespace string) error {
+func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace string) error {
 	// Cannot use a cached client as no manager will be started to populate cache
 	directClient, err := controller.NewDirectClient(cfg)
 	if err != nil {
@@ -81,13 +81,7 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, preserveNode bool, confi
 	if err = removeServices(svcMgr, cmData.Services); err != nil {
 		return err
 	}
-
-	// Delete node if needed
-	if preserveNode || node == nil {
-		return nil
-	}
-	klog.Infof("deleting node %s", node.GetName())
-	return directClient.Delete(ctx, node)
+	return nil
 }
 
 // removeServices uses the given manager to remove all the given Windows services from this instance.
--- a/pkg/daemon/cleanup/cleanup.go
+++ b/pkg/daemon/cleanup/cleanup.go
@@ -69,7 +69,7 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, preserveNode bool, confi
 	}()
 	if err != nil {
 		// TODO: best effort cleanup of all OpenShift managed services and those in the latest services ConfigMap
-		// https://issues.redhat.com/browse/WINC-733
+		// https://issues.redhat.com/browse/WINC-853
 		return err
 	}
 
--- a/pkg/daemon/cleanup/cleanup.go
+++ b/pkg/daemon/cleanup/cleanup.go
@@ -77,6 +77,7 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, preserveNode bool, confi
 	if err != nil {
 		klog.Exitf("could not create service manager: %s", err.Error())
 	}
+	defer svcMgr.Disconnect()
 	if err = removeServices(svcMgr, cmData.Services); err != nil {
 		return err
 	}
--- a/pkg/daemon/controller/controller.go
+++ b/pkg/daemon/controller/controller.go
@@ -265,6 +265,7 @@ func (sc *ServiceController) reconcileServices(services []servicescm.Service) er
 			if err != nil {
 				return err
 			}
+			defer winSvcObj.Close()
 			klog.Infof("created service %s", service.Name)
 		} else {
 			// open the service
@@ -272,6 +273,7 @@ func (sc *ServiceController) reconcileServices(services []servicescm.Service) er
 			if err != nil {
 				return err
 			}
+			defer winSvcObj.Close()
 			klog.Infof("reconciling existing service %s", service.Name)
 		}
 		if err := sc.reconcileService(winSvcObj, service); err != nil {
--- a/pkg/nodeconfig/nodeconfig.go
+++ b/pkg/nodeconfig/nodeconfig.go
@@ -242,7 +242,7 @@ func (nc *nodeConfig) Configure() error {
 	// Stop the kubelet so that the node is marked NotReady in case of an error in configuration. We are stopping all
 	// the required services as they are interdependent and is safer to do so given the node is going to be NotReady.
 	if err != nil {
-		if err := nc.Windows.EnsureRequiredServicesStopped(); err != nil {
+		if err := nc.Windows.RunWICDCleanup(nodeConfigCache.apiServerEndpoint, nc.wmcoNamespace); err != nil {
 			nc.log.Info("Unable to mark node as NotReady", "error", err)
 		}
 	}
@@ -511,10 +511,9 @@ func (nc *nodeConfig) Deconfigure() error {
 	}
 
 	// Revert the changes we've made to the instance by removing services and deleting all installed files
-	if err := nc.Windows.Deconfigure(); err != nil {
+	if err := nc.Windows.Deconfigure(nc.wmcoNamespace, nodeConfigCache.apiServerEndpoint); err != nil {
 		return errors.Wrap(err, "error deconfiguring instance")
 	}
-
 	// Clear the version annotation from the node object to indicate the node is not configured
 	patchData, err := metadata.GenerateRemovePatch([]string{}, []string{metadata.VersionAnnotation})
 	if err != nil {
--- a/pkg/windows/windows.go
+++ b/pkg/windows/windows.go
@@ -144,7 +144,6 @@ var (
 		KubeletServiceName}
 	// RequiredDirectories is a list of directories to be created by WMCO
 	RequiredDirectories = []string{
-		K8sDir,
 		remoteDir,
 		cniDir,
 		CniConfDir,
@@ -156,6 +155,7 @@ var (
 		ContainerdDir,
 		containerdLogDir,
 		podManifestDirectory,
+		K8sDir,
 	}
 )
 
@@ -235,10 +235,10 @@ type Windows interface {
 	ConfigureWICD(string, string, *Authentication) error
 	// ConfigureKubeProxy ensures that the kube-proxy service is running
 	ConfigureKubeProxy() error
-	// EnsureRequiredServicesStopped ensures that all services that are needed to configure a VM are stopped
-	EnsureRequiredServicesStopped() error
-	// Deconfigure removes all files and services created as part of the configuration process
-	Deconfigure() error
+	// Deconfigure removes all files and networks created by WMCO and runs the WICD cleanup command.
+	Deconfigure(string, string) error
+	// RunWICDCleanup runs the WICD cleanup command that ensures all services configured by WICD are stopped.
+	RunWICDCleanup(string, string) error
 }
 
 // windows implements the Windows interface
@@ -400,60 +400,24 @@ func (vm *windows) Reinitialize() error {
 	return nil
 }
 
-func (vm *windows) EnsureRequiredServicesStopped() error {
-	// TODO: In this transitional period of migrating services to WICD's responsibility, WICD must be stopped first to
-	//       ensure it does not start any of these stopped services. This code should be removed as part of
-	//       https://issues.redhat.com/browse/WINC-733
-	svc := &service{name: wicdServiceName}
-	if err := vm.ensureServiceNotRunning(svc); err != nil {
-		return errors.Wrapf(err, "could not stop service %s", wicdServiceName)
-	}
-	for _, svcName := range append(RequiredServices, AzureCloudNodeManagerServiceName) {
-		svc := &service{name: svcName}
-		if err := vm.ensureServiceNotRunning(svc); err != nil {
-			return errors.Wrapf(err, "could not stop service %s", svcName)
-		}
+func (vm *windows) RunWICDCleanup(apiServerURL string, watchNamespace string) error {
+	wicdCleanupCmd := fmt.Sprintf("%s cleanup --api-server %s --sa-ca %s --sa-token %s --namespace %s",
+		wicdPath, apiServerURL, wicdCAFile, wicdTokenFile, watchNamespace)
+	if out, err := vm.Run(wicdCleanupCmd, true); err != nil {
+		vm.log.Info("failed to cleanup node", "command", wicdCleanupCmd, "output", out)
+		return err
 	}
 	return nil
 }
 
-// ensureServicesAreRemoved ensures that all services installed by WMCO are removed from the instance
-func (vm *windows) ensureServicesAreRemoved() error {
-	// TODO: In this transitional period of migrating services to WICD's responsibility, WICD must be stopped first to
-	//       ensure it does not start any of these stopped services. This code should be removed as part of
-	//       https://issues.redhat.com/browse/WINC-733
-	svc := &service{name: wicdServiceName}
-	if err := vm.ensureServiceNotRunning(svc); err != nil {
-		return errors.Wrapf(err, "could not stop service %s", wicdServiceName)
-	}
-	for _, svcName := range append(RequiredServices, AzureCloudNodeManagerServiceName) {
-		svc := &service{name: svcName}
-
-		// If the service is not installed, do nothing
-		exists, err := vm.serviceExists(svc.name)
-		if err != nil {
-			return errors.Wrapf(err, "unable to check if %s service exists", svc.name)
-		}
-		if !exists {
-			continue
-		}
-
-		// Make sure the service is stopped before we attempt to delete it
-		if err := vm.ensureServiceNotRunning(svc); err != nil {
-			return errors.Wrapf(err, "could not stop service %s", svc.name)
-		}
-		if err := vm.deleteService(svc); err != nil {
-			return errors.Wrapf(err, "could not delete service %s", svcName)
-		}
-		vm.log.Info("deconfigured", "service", svc.name)
-	}
-	return nil
-}
-
-func (vm *windows) Deconfigure() error {
+func (vm *windows) Deconfigure(watchNamespace string, apiServerURL string) error {
 	vm.log.Info("deconfiguring")
-	if err := vm.ensureServicesAreRemoved(); err != nil {
-		return errors.Wrap(err, "unable to remove Windows services")
+
+	if err := vm.deconfigureWICD(); err != nil {
+		return err
+	}
+	if err := vm.RunWICDCleanup(apiServerURL, watchNamespace); err != nil {
+		return errors.Wrap(err, "Unable to cleanup the Windows instance")
 	}
 	if err := vm.removeDirectories(); err != nil {
 		return errors.Wrap(err, "unable to remove created directories")
@@ -470,7 +434,7 @@ func (vm *windows) Bootstrap(desiredVer, apiServerURL, watchNamespace string, cr
 	if err := vm.deconfigureWICD(); err != nil {
 		return err
 	}
-	if err := vm.EnsureRequiredServicesStopped(); err != nil {
+	if err := vm.ensureRequiredServicesStopped(); err != nil {
 		return errors.Wrap(err, "unable to stop all services")
 	}
 	if err := vm.ensureHostNameAndContainersFeature(); err != nil {
@@ -526,6 +490,18 @@ func (vm *windows) ConfigureKubeProxy() error {
 }
 
 // Interface helper methods
+// ensureRequiredServicesStopped ensures that all services that are needed to configure a VM are stopped
+func (vm *windows) ensureRequiredServicesStopped() error {
+	// TODO: WMCO no longer stops services started by WICD before calling WICD bootstrap command
+	//		 This code should be removed as part of: https://issues.redhat.com/browse/WINC-741
+	for _, svcName := range append(RequiredServices, AzureCloudNodeManagerServiceName) {
+		svc := &service{name: svcName}
+		if err := vm.ensureServiceNotRunning(svc); err != nil {
+			return errors.Wrapf(err, "could not stop service %s", svcName)
+		}
+	}
+	return nil
+}
 
 // ensureHostNameAndContainersFeature ensures hostname of the Windows VM matches the expected name
 // and the required Windows feature is enabled.
@@ -883,15 +859,21 @@ func (vm *windows) ensureHNSNetworksAreRemoved() error {
 	// VIP HNS endpoint created by the operator is also deleted when the HNS networks are deleted.
 	for _, network := range []string{BaseOVNKubeOverlayNetwork, OVNKubeOverlayNetwork} {
 		err = wait.Poll(retry.Interval, retry.Timeout, func() (bool, error) {
+			// reinitialize and retry on failure to avoid connection reset SSH errors
 			if err := vm.removeHNSNetwork(network); err != nil {
-				return false, errors.Wrapf(err, "error removing %s HNS network", network)
+				vm.log.V(1).Error(err, "error removing %s HNS network", "network", network)
+				if err := vm.Reinitialize(); err != nil {
+					return false, errors.Wrapf(err, "error reinitializing VM after removing %s HNS network", network)
+				}
+				return false, nil
 			}
 			if err := vm.Reinitialize(); err != nil {
 				return false, errors.Wrapf(err, "error reinitializing VM after removing %s HNS network", network)
 			}
 			out, err := vm.Run(getHNSNetworkCmd(network), true)
 			if err != nil {
-				return false, errors.Wrapf(err, "error waiting for %s HNS network", network)
+				vm.log.V(1).Error(err, "error waiting for HNS network", "network", network)
+				return false, nil
 			}
 			return !strings.Contains(out, network), nil
 		})
@@ -905,8 +887,8 @@ func (vm *windows) ensureHNSNetworksAreRemoved() error {
 // removeHNSNetwork removes the given HNS network.
 func (vm *windows) removeHNSNetwork(networkName string) error {
 	cmd := getHNSNetworkCmd(networkName) + " | Remove-HnsNetwork;"
-	// PowerShell returns error waiting without exit status or signal error when the OVNKubeOverlayNetwork is removed.
-	if out, err := vm.Run(cmd, true); err != nil && !(networkName == OVNKubeOverlayNetwork && strings.Contains(err.Error(), cmdExitNoStatus)) {
+	// PowerShell returns error waiting without exit status or signal error when the networks are removed.
+	if out, err := vm.Run(cmd, true); err != nil && !strings.Contains(err.Error(), cmdExitNoStatus) {
 		return errors.Wrapf(err, "failed to remove %s HNS network with output: %s", networkName, out)
 	}
 	return nil
--- a/pkg/windows/windows.go
+++ b/pkg/windows/windows.go
@@ -466,6 +466,10 @@ func (vm *windows) Deconfigure() error {
 
 func (vm *windows) Bootstrap(desiredVer, apiServerURL, watchNamespace string, credentials *Authentication) error {
 	vm.log.Info("configuring")
+	// Make sure WICD service is not running before calling node bootstrap
+	if err := vm.deconfigureWICD(); err != nil {
+		return err
+	}
 	if err := vm.EnsureRequiredServicesStopped(); err != nil {
 		return errors.Wrap(err, "unable to stop all services")
 	}
@@ -698,6 +702,28 @@ func (vm *windows) ensureServiceNotRunning(svc *service) error {
 
 }
 
+// ensureServiceIsRemoved ensures that the given service installed by WMCO is removed from the instance
+func (vm *windows) ensureServiceIsRemoved(svcName string) error {
+	svc := &service{name: svcName}
+	// If the service is not installed, do nothing
+	exists, err := vm.serviceExists(svcName)
+	if err != nil {
+		return errors.Wrapf(err, "error checking if %s Windows service exists", svc.name)
+	}
+	if !exists {
+		return nil
+	}
+	// Make sure the service is stopped before we attempt to delete it
+	if err := vm.ensureServiceNotRunning(svc); err != nil {
+		return errors.Wrapf(err, "error stopping %s Windows service", svc.name)
+	}
+	if err := vm.deleteService(svc); err != nil {
+		return errors.Wrapf(err, "error deleting %s Windows service", svc.name)
+	}
+	vm.log.Info("deconfigured", "service", svc.name)
+	return nil
+}
+
 // stopService stops the service that was already running
 func (vm *windows) stopService(svc *service) error {
 	if svc == nil {
@@ -928,6 +954,14 @@ func (vm *windows) ensureWICDSecretContent(credentials *Authentication) error {
 	}
 	tokenDir, tokenFileName := SplitPath(wicdTokenFile)
 	return vm.EnsureFileContent(credentials.Token, tokenFileName, tokenDir)
+}
+
+// deconfigureWICD ensures the WICD service running on the Windows instance is removed
+func (vm *windows) deconfigureWICD() error {
+	if err := vm.ensureServiceIsRemoved(wicdServiceName); err != nil {
+		return errors.Wrapf(err, "error ensuring %s Windows service is removed", wicdServiceName)
+	}
+	return nil
 }
 
 // Generic helper methods
--- a/cmd/daemon/cleanup.go
+++ b/cmd/daemon/cleanup.go
@@ -35,15 +35,10 @@ var (
 			"according to information given by Windows Service ConfigMaps present within the cluster",
 		Run: runCleanupCmd,
 	}
-	// preserveNode is an optional flag that instructs WICD to deconfigure an instance without deleting the Node object.
-	// This is useful in the node upgrade scenario.
-	preserveNode bool
 )
 
 func init() {
 	rootCmd.AddCommand(cleanupCmd)
-	cleanupCmd.PersistentFlags().BoolVar(&preserveNode, "preserveNode", false,
-		"If set to true, preserves the Node associated with this instance. Defaults to false.")
 }
 
 func runCleanupCmd(cmd *cobra.Command, args []string) {
@@ -52,7 +47,7 @@ func runCleanupCmd(cmd *cobra.Command, args []string) {
 		klog.Exitf("error using service account to build config: %s", err.Error())
 	}
 	ctx := ctrl.SetupSignalHandler()
-	if err := cleanup.Deconfigure(cfg, ctx, preserveNode, namespace); err != nil {
+	if err := cleanup.Deconfigure(cfg, ctx, namespace); err != nil {
 		klog.Exitf(err.Error())
 	}
 }
--- a/pkg/daemon/fake/manager.go
+++ b/pkg/daemon/fake/manager.go
@@ -68,6 +68,10 @@ type testMgr struct {
 	svcList *fakeServiceList
 }
 
+func (t *testMgr) Disconnect() error {
+	return nil
+}
+
 // CreateService installs new service name on the system.
 // The service will be executed by running exepath binary.
 // Use config c to specify service parameters.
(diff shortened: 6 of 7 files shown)


pkg/daemon/controller/controller.go
```diff
@@ -252,6 +252,7 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
 		klog.Error(err)
 		return ctrl.Result{}, err
 	}
+	defer sc.Disconnect()
```
mansikulkarni96, line 255: added the manager disconnect here

pkg/daemon/cleanup/cleanup.go
```diff
@@ -92,6 +93,8 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, preserveNode bool) error
 
 // removeServices uses the given manager to remove all the given Windows services from this instance.
 func removeServices(svcMgr manager.Manager, services []servicescm.Service) error {
+	wicdSvc := &servicescm.Service{Name: windows.WicdServiceName}
```
aravindhp, line 96: So WICD is removing itself? If yes, that would be an anti-pattern. WMCO should be the one removing WICD given it is the entity that installed it.

pkg/windows/windows.go
```diff
@@ -417,50 +419,27 @@ func (vm *windows) EnsureRequiredServicesStopped() error {
...
-	return nil
-}
-
-func (vm *windows) Deconfigure() error {
-	vm.log.Info("deconfiguring")
-	if err := vm.ensureServicesAreRemoved(); err != nil {
-		return errors.Wrap(err, "unable to remove Windows services")
+	if err := vm.ensureHNSNetworksAreRemoved(); err != nil {
+		return errors.Wrap(err, "unable to ensure HNS networks are removed")
+	}
```
aravindhp, line 430: Why is this being done before the services are deleted?

pkg/daemon/cleanup/cleanup.go
```diff
@@ -77,6 +77,7 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, preserveNode bool) error
 	if err != nil {
 		klog.Exitf("could not create service manager: %s", err.Error())
 	}
+	defer svcMgr.Disconnect()
```
aravindhp, line 81: I assume you ran into an issue because handles were not being closed. Please update the commit message with the issue you ran into.

pkg/windows/windows.go
```diff
@@ -417,50 +419,27 @@ func (vm *windows) EnsureRequiredServicesStopped() error {
...
 
-		// Make sure the service is stopped before we attempt to delete it
-		if err := vm.ensureServiceNotRunning(svc); err != nil {
-			return errors.Wrapf(err, "could not stop service %s", svc.name)
-		}
-		if err := vm.deleteService(svc); err != nil {
-			return errors.Wrapf(err, "could not delete service %s", svcName)
-		}
-		vm.log.Info("deconfigured", "service", svc.name)
+	if err := vm.ensureWICDSecretContent(credentials); err != nil {
```
aravindhp, line 425: Why is this needed?

pkg/windows/windows.go
```diff
@@ -417,50 +419,27 @@ func (vm *windows) EnsureRequiredServicesStopped() error {
...
-}
-
-func (vm *windows) Deconfigure() error {
-	vm.log.Info("deconfiguring")
-	if err := vm.ensureServicesAreRemoved(); err != nil {
-		return errors.Wrap(err, "unable to remove Windows services")
+	if err := vm.ensureHNSNetworksAreRemoved(); err != nil {
+		return errors.Wrap(err, "unable to ensure HNS networks are removed")
+	}
+	wicdCleanupCmd := fmt.Sprintf("%s cleanup --api-server %s --sa-ca %s --sa-token %s",
```
aravindhp, line 431: There should be a `DeconfigureWICD()` call equivalent to the `ConfigureWICD()` call before calling clean up. WICD shouldn't be cleaning up itself. In fact the `cleanup` and `bootstrap` commands should fail if the WICD service is running.

aravindhp (changes requested):
Thanks for working on this @mansikulkarni96. Please address my comments.

pkg/daemon/cleanup/cleanup.go
mansikulkarni96: I agree with this. I had it that way before, I think I just made this change as it was cleaner. Thanks for pointing out, will make the change.

pkg/windows/windows.go
mansikulkarni96: Attempt was to deconfigure it in a reverse order of configuration:
create dirs
create services
create networks

aravindhp: I am surprised we are creating the network after the services are running. Don't hybrid-overlay or kube-proxy needs these networks?

saifshaikh48 (commented):
Left a few comments, nice work @mansikulkarni96

pkg/windows/windows.go
saifshaikh48: I thought network creation occurs before kube-proxy service is started via the network-conf script?

pkg/windows/windows.go
aravindhp: That's a good point @saifshaikh48

pkg/windows/windows.go
```diff
@@ -137,26 +137,22 @@ var (
...
-		KubeProxyLogDir,
-		wicdLogDir,
-		HybridOverlayLogDir,
 		ContainerdDir,
-		containerdLogDir,
 		podManifestDirectory,
+		K8sDir,
 	}
+	// RequiredDirectories is a list of directories to be created by WMCO
+	RequiredDirectories = append(RequiredConfigDirectories, logDir, KubeletLogDir, KubeProxyLogDir, wicdLogDir, HybridOverlayLogDir, containerdLogDir)
```
saifshaikh48, line 155: nit: over 120 col

pkg/windows/windows.go
```diff
@@ -61,8 +61,8 @@ const (
 	ContainerdLogPath = containerdLogDir + "containerd.log"
 	// ContainerdServiceName is containerd Windows service name
 	ContainerdServiceName = "containerd"
-	// wicdServiceName is the Windows service name for WICD
-	wicdServiceName = "windows-instance-config-daemon"
+	// WicdServiceName is the Windows service name for WICD
+	WicdServiceName = "windows-instance-config-daemon"
```
saifshaikh48, line 65: Don't think this needs to be exported

pkg/windows/windows.go
```diff
@@ -417,50 +408,44 @@ func (vm *windows) EnsureRequiredServicesStopped() error {
...
 }
 
-// ensureServicesAreRemoved ensures that all services installed by WMCO are removed from the instance
-func (vm *windows) ensureServicesAreRemoved() error {
-	// TODO: In this transitional period of migrating services to WICD's responsibility, WICD must be stopped first to
-	//       ensure it does not start any of these stopped services. This code should be removed as part of
-	//       https://issues.redhat.com/browse/WINC-733
-	svc := &service{name: wicdServiceName}
+// ensureServiceIsRemoved ensures that the given service installed by WMCO is removed from the instance
+func (vm *windows) ensureServiceIsRemoved(svc *service) error {
```
saifshaikh48, line 412: IMO have this take in a name for ease of use at call sites and create `service` obj within the method as before

pkg/windows/windows.go
```diff
@@ -498,14 +483,22 @@ func (vm *windows) ConfigureWICD(apiServerURL string, credentials *Authenticatio
...
+		return errors.Wrapf(err, "error ensuring %s Windows service has started running", WicdServiceName)
+	}
+	vm.log.Info("configured", "service", WicdServiceName, "args", wicdServiceArgs)
+	return nil
+}
+
+func (vm *windows) DeconfigureWICD() error {
+	wicdSvc := &service{name: WicdServiceName}
+	if err := vm.ensureServiceIsRemoved(wicdSvc); err != nil {
+		return errors.Wrapf(err, "could not remove service %s", WicdServiceName)
```
saifshaikh48, line 500: We aren't always removing the WICD service (if it doesn't exist already for example) so the error msg should reflect that
e.g. `unable to ensure WICD service is removed`

pkg/nodeconfig/nodeconfig.go
```diff
@@ -149,6 +149,10 @@ func (nc *nodeConfig) Configure() error {
 		return err
 	}
 
+	if err := nc.Windows.DeconfigureWICD(); err != nil {
+		return errors.Wrap(err, "deconfiguring WICD failed")
```
saifshaikh48, line 153: IMO `return err` is sufficient, since DeconfigureWICD itself is a wrapper func that should give a descriptive enough err

pkg/windows/windows.go
```diff
@@ -401,13 +399,6 @@ func (vm *windows) Reinitialize() error {
...
 
 func (vm *windows) EnsureRequiredServicesStopped() error {
-	// TODO: In this transitional period of migrating services to WICD's responsibility, WICD must be stopped first to
-	//       ensure it does not start any of these stopped services. This code should be removed as part of
-	//       https://issues.redhat.com/browse/WINC-733
-	svc := &service{name: wicdServiceName}
-	if err := vm.ensureServiceNotRunning(svc); err != nil {
-		return errors.Wrapf(err, "could not stop service %s", wicdServiceName)
-	}
 	for _, svcName := range append(RequiredServices, AzureCloudNodeManagerServiceName) {
```
sebsoto, line 402: WICD should be stopping these services, not WMCO.

pkg/windows/windows.go
sebsoto: This error is stuttering, if you look at the errors returned by ensureServiceIsRemoved

pkg/windows/windows.go
```diff
@@ -417,50 +408,44 @@ func (vm *windows) EnsureRequiredServicesStopped() error {
...
-		vm.log.Info("deconfigured", "service", svc.name)
+	if err := vm.deleteService(svc); err != nil {
+		return errors.Wrapf(err, "could not delete service %s", svc.name)
 	}
+	vm.log.Info("deconfigured", "service", svc.name)
 	return nil
 }
 
-func (vm *windows) Deconfigure() error {
+func (vm *windows) Deconfigure(preserveNode bool, apiServerURL string, credentials *Authentication) error {
```
sebsoto, line 432: I don't see credentials being used here. Don't you need to get the wicdCAFile and wicdTokenFile values from it?

pkg/windows/windows.go
sebsoto: Shouldnt this whole method should be removed from the Windows interface?

pkg/nodeconfig/nodeconfig.go
sebsoto, line 153: Why does this need to be done here? Please add a comment

sebsoto: Additionally it should be done before createBootstrapFiles if it is necessary, it is odd to start bootstrapping and then clean something up in the middle.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -505,11 +509,19 @@ func (nc *nodeConfig) Deconfigure() error {
...
+	}
+
 	// Revert the changes we've made to the instance by removing services and deleting all installed files
-	if err := nc.Windows.Deconfigure(); err != nil {
+	if err := nc.Windows.Deconfigure(preserveNode, nodeConfigCache.apiServerEndpoint, nodeConfigCache.credentials); err != nil {
 		return errors.Wrap(err, "error deconfiguring instance")
 	}
+	nc.log.Info("instance has been deconfigured", "node", nc.node.GetName())
```
sebsoto, line 520: This has the possiblity of printing multiple times for a single instance if preserveNode is set and the RemovePatch fails and this is re-reconciled.

sebsoto (changes requested):
Good work @mansikulkarni96 please see my comments

pkg/windows/windows.go
mansikulkarni96: @sebsoto This method is now only being used in the WICD [bootstrap call ](https://github.com/openshift/windows-machine-config-operator/blob/be61290e3f49cf006b95f003b48a1e559eb348cf/pkg/windows/windows.go#L469) and configure not during the cleanup process. Are you saying this needs to be replaced?

pkg/nodeconfig/nodeconfig.go
mansikulkarni96: Please see this [conversation](https://github.com/openshift/windows-machine-config-operator/pull/1332/commits/0486bd02e166e0319d35cf11ed501849851c2b8c#r1038511228). I can move the call site.

pkg/windows/windows.go
mansikulkarni96: I remoed the credentials param now. wicdCAFile and wicdTokenFile  are present in the k8sdir and their location is defined by constants, I am using them from there. Are you suggesting using ensureWICDSecretContent to verify their contents?

pkg/nodeconfig/nodeconfig.go
sebsoto: That link doesn't work for me, but regardless a comment on the PR doesn't help someone reading the code, please add a comment.

pkg/windows/windows.go
sebsoto: WMCO should not be stopping any services besides the WICD controller service. You should invoke WICD cleanup if the services started by WICD need to be stopped.

pkg/windows/windows.go
sebsoto: No, I was just commenting on the unused parameter and wondering if you needed to do something with it.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -505,22 +509,26 @@ func (nc *nodeConfig) Deconfigure() error {
...
+		// Clear the version annotation from the node object to indicate the node is not configured
+		patchData, err := metadata.GenerateRemovePatch([]string{}, []string{metadata.VersionAnnotation})
+		if err != nil {
+			return errors.Wrapf(err, "error creating version annotation remove request")
+		}
+		_, err = nc.k8sclientset.CoreV1().Nodes().Patch(context.TODO(), nc.node.GetName(), kubeTypes.JSONPatchType,
+			patchData, meta.PatchOptions{})
+		if err != nil {
+			return errors.Wrapf(err, "error removing version annotation from node %s", nc.node.GetName())
+		}
```
aravindhp, line 530: Shouldn't this also be done by WICD given it is the entity that decides whether to delete the node or node based on the `preservedNode` option?

saifshaikh48: While this should be done, it might be out of scope of the PR. See [WINC-741](https://issues.redhat.com/browse/WINC-741)

aravindhp: Ah yes, I keep forgetting about that story.

pkg/windows/windows.go
```diff
@@ -879,8 +885,9 @@ func (vm *windows) ensureHNSNetworksAreRemoved() error {
 // removeHNSNetwork removes the given HNS network.
 func (vm *windows) removeHNSNetwork(networkName string) error {
 	cmd := getHNSNetworkCmd(networkName) + " | Remove-HnsNetwork;"
-	// PowerShell returns error waiting without exit status or signal error when the OVNKubeOverlayNetwork is removed.
-	if out, err := vm.Run(cmd, true); err != nil && !(networkName == OVNKubeOverlayNetwork && strings.Contains(err.Error(), cmdExitNoStatus)) {
+	// PowerShell returns error waiting without exit status or signal error when the networks are removed.
+	out, err := vm.Run(cmd, true)
+	if err != nil && !strings.Contains(err.Error(), cmdExitNoStatus) {
```
mansikulkarni96, line 890: This change needs to be added to commit msg, will update in next iteration.
Removing both networks has been leading to error waiting without exit status, in my opinion we can ignore error for both network types as we have a poll with a check for hnsNetwork call following this method in ensureHNSNetworksAreRemoved().

aravindhp (changes requested):

pkg/windows/windows.go
```diff
@@ -594,7 +599,8 @@ func (vm *windows) createDirectories() error {
 // removeDirectories removes all directories created as part of the configuration process
 func (vm *windows) removeDirectories() error {
 	vm.log.Info("removing directories")
-	for _, dir := range RequiredDirectories {
+	// Remove all config directories, exclude logs
```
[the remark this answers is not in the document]
mansikulkarni96: We had a discussion in design meeting about preserving logs for future reference, could be helpful in debugging.

pkg/windows/windows.go
```diff
@@ -235,10 +232,13 @@ type Windows interface {
 	ConfigureWICD(string, *Authentication) error
 	// ConfigureKubeProxy ensures that the kube-proxy service is running
 	ConfigureKubeProxy() error
-	// EnsureRequiredServicesStopped ensures that all services that are needed to configure a VM are stopped
-	EnsureRequiredServicesStopped() error
-	// Deconfigure removes all files and services created as part of the configuration process
-	Deconfigure() error
+	// Cleanup runs the WICD cleanup command that ensures all services configured by WICD are stopped. If the bool is
+	// not set, this function will clean up the node object as well.
+	Cleanup(string, bool) error
```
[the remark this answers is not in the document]
mansikulkarni96: runWICDCleanup() is run  on the Windows instance, this method is also used in Configure() from nodeconfig to cleanup all services in case of error configuring the node.

pkg/windows/windows.go
```diff
@@ -235,10 +232,13 @@ type Windows interface {
...
+	// Deconfigure removes all files and networks created by WMCO and runs the WICD cleanup command
+	Deconfigure(bool, string) error
+	// DeconfigureWICD ensures that the Windows Instance Config Daemon is not running on the node
```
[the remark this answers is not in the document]
mansikulkarni96: same comment applies here, this method includes commands that run on the Windows instance. Please let me know if I should add a comment for this.

pkg/windows/windows.go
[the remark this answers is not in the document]
aravindhp: - Aren't you now changing behavior across versions? 
- Please mention why in the comment
- Shouldn't this be documented?

pkg/windows/windows.go
```diff
@@ -509,6 +512,13 @@ func (vm *windows) ConfigureWICD(apiServerURL string, credentials *Authenticatio
 	return nil
 }
 
+func (vm *windows) deconfigureWICD() error {
```
aravindhp, line 515: - Please add a comment
- Please move this along with all the other non-struct methods.

aravindhp (changes requested):

pkg/windows/windows.go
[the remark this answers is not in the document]
mansikulkarni96: Decided to skip adding this change as it is out of scope of this ticket, will start a new discussion about why we need this and create a ticket for it.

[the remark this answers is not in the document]
aravindhp: Please remove the comment.

saifshaikh48 (commented):
/lgtm

test/e2e/delete_test.go
```diff
@@ -67,13 +67,13 @@ func (tc *testContext) testBYOHRemoval(t *testing.T) {
...
-			require.NoError(t, err, "error determining if created directories exist")
-			assert.True(t, dirsCleanedUp, "directories not removed")
-
 			networksRemoved, err := tc.checkNetworksRemoved(addr)
 			require.NoError(t, err, "error determining if HNS networks are removed")
 			assert.True(t, networksRemoved, "HNS networks not removed")
+
+			dirsCleanedUp, err := tc.checkDirsDoNotExist(addr)
+			require.NoError(t, err, "error determining if created directories exist")
+			assert.True(t, dirsCleanedUp, "directories not removed")
```
mansikulkarni96, line 76: test order

test/e2e/delete_test.go
```diff
@@ -74,6 +74,7 @@ func (tc *testContext) testBYOHRemoval(t *testing.T) {
 			networksRemoved, err := tc.checkNetworksRemoved(addr)
 			require.NoError(t, err, "error determining if HNS networks are removed")
 			assert.True(t, networksRemoved, "HNS networks not removed")
+
```
aravindhp, line 77: Please revert this whenever you get the chance

pkg/windows/windows.go
```diff
@@ -417,44 +413,29 @@ func (vm *windows) EnsureRequiredServicesStopped() error {
...
-	if err := vm.ensureServicesAreRemoved(); err != nil {
-		return errors.Wrap(err, "unable to remove Windows services")
+
+	if err := vm.deconfigureWICD(); err != nil {
+		return err
+	}
+	if err := vm.RunWICDCleanup(apiServerURL, preserveNode); err != nil {
+		return errors.Wrap(err, "Unable to cleanup the Windows instance")
 	}
+	time.Sleep(time.Minute)
```
aravindhp, line 438: Is this for debugging purposes or actually required?

pkg/windows/windows.go
```diff
@@ -857,15 +864,19 @@ func (vm *windows) ensureHNSNetworksAreRemoved() error {
 	// VIP HNS endpoint created by the operator is also deleted when the HNS networks are deleted.
 	for _, network := range []string{BaseOVNKubeOverlayNetwork, OVNKubeOverlayNetwork} {
 		err = wait.Poll(retry.Interval, retry.Timeout, func() (bool, error) {
+			// retry removing network on failure
```
aravindhp, line 867: If we keep seeing `read: connection reset by peer` on the retry we might have to call `vm.Reinitialize()`.

pkg/windows/windows.go
```diff
@@ -857,15 +864,19 @@ func (vm *windows) ensureHNSNetworksAreRemoved() error {
...
 			if err := vm.removeHNSNetwork(network); err != nil {
-				return false, errors.Wrapf(err, "error removing %s HNS network", network)
+				vm.log.V(1).Error(err, "error removing %s HNS network", "network", network)
+				return false, nil
 			}
 			if err := vm.Reinitialize(); err != nil {
-				return false, errors.Wrapf(err, "error reinitializing VM after removing %s HNS network", network)
+				vm.log.V(1).Error(err, "error reinitializing VM after removing HNS network", "network", network)
+				return false, nil
```
aravindhp, line 874: If `vm.Reinitialize()` failed how would any commands over ssh work? So you would have to retry the `vm.Reinitialize()` call or fail on the error.

aravindhp (changes requested):

pkg/windows/windows.go
mansikulkarni96: debugging only

pkg/windows/windows.go
mansikulkarni96: No I don't think this change is required I will remove it in my next iteration.
Apologies for some unrelated changes left over for debugging.

pkg/windows/windows.go
```diff
@@ -417,44 +413,30 @@ func (vm *windows) EnsureRequiredServicesStopped() error {
...
-		return errors.Wrap(err, "unable to remove Windows services")
+
+	if err := vm.deconfigureWICD(); err != nil {
+		return err
+	}
+	if err := vm.RunWICDCleanup(apiServerURL, watchNamespace, preserveNode); err != nil {
+		return errors.Wrap(err, "Unable to cleanup the Windows instance")
 	}
+	// wait for log files to be closed after cleanup
+	time.Sleep(time.Second * 30)
```
aravindhp, line 439: This is risky. There is potential for `removeDirectories()` to still fail on busy instances where it could take more than 30 seconds for log files to be closed. Safer would be to retry `removeDirectories()` similar to `ensureHNSNetworksAreRemoved()`.

But what is still puzzling is why is the following not returning an error?
```
2023-01-12T16:25:28Z	DEBUG	wc 10.0.228.20	run	{"cmd": "cmd /c if exist C:\\var\\log\\ rmdir C:\\var\\log\\ /s /q", "out": "C:\\var\\log\\HYBRID~1\\hybrid-overlay.log - The process cannot access the file because it is being used by another process.\r\n"}
```
Is it because of the `cmd /c` prefix not propagating the error?

In addition, this is the general problem @sebsoto is also facing with hcs-shim binary.

aravindhp: I missed:
```
2023-01-12T16:25:27Z	DEBUG	wc 10.0.228.20	run	{"cmd": "cmd /c if exist C:\\k\\ rmdir C:\\k\\ /s /q", "out": "C:\\k\\hybrid-overlay-node.exe - Access is denied.\r\n"}
```
This means that hybrid-overlay is still running. So even though the service is stopped and deleted the process is still running??? This is very strange.

mansikulkarni96: The sleep is working, it seems like it takes longer to close the log files as I mentioned. I can add the retry change as per your suggestion and see if it is helpful.

sebsoto: PTAL at https://github.com/openshift/windows-machine-config-operator/pull/1372/commits/9a25d79b582df3395a093edd854dd3690125ebf9

mansikulkarni96:
/test aws-e2e-operator
Error bringing up cluster with 
`level=error msg=Error: creating EC2 Instance: InvalidNetworkInterface.InUse: Interface: [eni-05a8b4a2ac07d2eb5] in use.`

pkg/windows/windows.go
mansikulkarni96: I cherry-picked @sebsoto 's commit and another [commit](https://github.com/openshift/windows-machine-config-operator/pull/1332/commits/5f3fae1a3aad3ddbda8c8db21921efefeb89dcc9) for a retry, hitting a connection reset now:
```
2023-01-17T10:53:43Z	ERROR	wc 10.0.251.204	error running	{"cmd": "cmd /c if exist C:\\var\\log\\kubelet\\ rmdir C:\\var\\log\\kubelet\\ /s /q", "out": "", "error": "read tcp 10.0.213.9:36156->10.0.251.204:22: read: connection reset by peer"}
github.com/openshift/windows-machine-config-operator/pkg/windows.(*windows).removeDirectories.func1
	/build/windows-machine-config-operator/pkg/windows/windows.go:584
```
Kicking off the reconciliation after node has been deleted is not useful, hence the retry is added.

pkg/daemon/controller/controller.go
```diff
@@ -282,6 +282,7 @@ func (sc *ServiceController) reconcileServices(services []servicescm.Service) er
 		if err := sc.reconcileService(winSvcObj, service); err != nil {
 			return err
 		}
+		defer winSvcObj.Close()
```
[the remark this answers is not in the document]
sebsoto: This should be called immediately after the service object is opened. 
```
			winSvcObj, err = sc.OpenService(service.Name)
			if err != nil {
				return err
			}
			defer winSvcObj.Close()
```
and
```
			winSvcObj, err = sc.CreateService(service.Name, "", mgr.Config{})
			if err != nil {
				return err
			}
			defer winSvcObj.Close()
```

as it is there is the opportunity for the handle to be leaked if reconcileService returns error.

pkg/windows/windows.go
```diff
@@ -235,10 +235,11 @@ type Windows interface {
 	ConfigureWICD(string, string, *Authentication) error
 	// ConfigureKubeProxy ensures that the kube-proxy service is running
 	ConfigureKubeProxy() error
-	// EnsureRequiredServicesStopped ensures that all services that are needed to configure a VM are stopped
-	EnsureRequiredServicesStopped() error
-	// Deconfigure removes all files and services created as part of the configuration process
-	Deconfigure() error
+	// Deconfigure removes all files and networks created by WMCO and runs the WICD cleanup command
+	Deconfigure(bool, string, string) error
```
sebsoto, line 239: Please doc the parameters to this method

pkg/windows/windows.go
```diff
@@ -857,15 +865,21 @@ func (vm *windows) ensureHNSNetworksAreRemoved() error {
...
 			if err := vm.removeHNSNetwork(network); err != nil {
-				return false, errors.Wrapf(err, "error removing %s HNS network", network)
+				vm.log.V(1).Error(err, "error removing %s HNS network", "network", network)
+				if err := vm.Reinitialize(); err != nil {
+					return false, errors.Wrapf(err, "error reinitializing VM after removing %s HNS network", network)
+				}
+				return false, nil
 			}
 			if err := vm.Reinitialize(); err != nil {
 				return false, errors.Wrapf(err, "error reinitializing VM after removing %s HNS network", network)
```
sebsoto, line 877: by returning errors here on the two reinitialize calls the choice is being made not to retry. Is that on purpose? I would think it would be better to return false, nil and retry.

pkg/windows/windows.go
```diff
@@ -698,6 +684,28 @@ func (vm *windows) ensureServiceNotRunning(svc *service) error {
...
+	// Make sure the service is stopped before we attempt to delete it
+	if err := vm.ensureServiceNotRunning(svc); err != nil {
+		return errors.Wrapf(err, "error stopping %s Windows service", svc.name)
+	}
+	if err := vm.deleteService(svc); err != nil {
+		return errors.Wrapf(err, "error deleting %s Windows service", svc.name)
+	}
+	vm.log.Info("deconfigured", "service", svc.name)
+	return nil
+}
```
sebsoto, line 707: nit: This is not new code, but because it was moved within the file it is presenting as if it is, making the PR longer to review.

sebsoto (changes requested):
Good work @mansikulkarni96,  a few comments

pkg/windows/windows.go
aravindhp: Please add a story to remove this sleep as we cannot cut a release with this.

aravindhp (changes requested):

bundle/manifests/windows-instance-config-daemon_rbac.authorization.k8s.io_v1_clusterrole.yaml
```diff
@@ -12,3 +12,4 @@ rules:
   - list
   - watch
   - get
+  - delete
```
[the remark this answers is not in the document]
mansikulkarni96: No we can't do this, the comment was valid for way we had RBAC in the past. With the new RBAC changes, WMCO grants permissions to WICD so it needs to have them.

aravindhp (approved):

pkg/windows/windows.go
```diff
@@ -880,7 +896,7 @@ func (vm *windows) ensureHNSNetworksAreRemoved() error {
 func (vm *windows) removeHNSNetwork(networkName string) error {
 	cmd := getHNSNetworkCmd(networkName) + " | Remove-HnsNetwork;"
 	// PowerShell returns error waiting without exit status or signal error when the OVNKubeOverlayNetwork is removed.
```
saifshaikh48, line 898: nit: comment should be updated to drop the mention of OVNKubeOverlayNetwork

saifshaikh48 (commented):
/lgtm

Good work getting this to the finish line. One nit (but since it's a nit, I'm fine with this merging without it being addressed)

sebsoto:
/hold
`instance has been deconfigured` is missing from the wmco logs.
https://github.com/openshift/windows-machine-config-operator/blob/a560bfe44aa85083938383c795c515e93c7ee1ae/pkg/nodeconfig/nodeconfig.go#L529

holding until cause is figured out

sebsoto (changes requested):

pkg/windows/windows.go
```diff
@@ -156,6 +155,7 @@ var (
 		ContainerdDir,
 		containerdLogDir,
 		podManifestDirectory,
+		K8sDir,
```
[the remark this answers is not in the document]
mansikulkarni96: This change is necessary and mentioned in the commit msg.

controllers/controllers.go
```diff
@@ -192,9 +192,6 @@ func (r *instanceReconciler) deconfigureInstance(node *core.Node) error {
 	if err = nc.Deconfigure(); err != nil {
 		return err
 	}
-	if err = r.client.Delete(context.TODO(), instance.Node); err != nil {
-		return errors.Wrapf(err, "error deleting node %s", instance.Node.GetName())
-	}
 	return nil
```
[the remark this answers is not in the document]
mansikulkarni96: thanks for the catch. Looking at previous commit history, it looks like this got pushed during an incorrect squash. Updated.

pkg/windows/windows.go
```diff
@@ -401,13 +401,8 @@ func (vm *windows) Reinitialize() error {
 }
 
 func (vm *windows) EnsureRequiredServicesStopped() error {
-	// TODO: In this transitional period of migrating services to WICD's responsibility, WICD must be stopped first to
-	//       ensure it does not start any of these stopped services. This code should be removed as part of
-	//       https://issues.redhat.com/browse/WINC-733
-	svc := &service{name: wicdServiceName}
-	if err := vm.ensureServiceNotRunning(svc); err != nil {
-		return errors.Wrapf(err, "could not stop service %s", wicdServiceName)
-	}
```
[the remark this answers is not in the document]
mansikulkarni96: It's being used with deconfigureWICD, this method is called from both bootstrap and decofigure methods.

saifshaikh48 (commented):
/lgtm

pkg/windows/windows.go
```diff
@@ -400,60 +400,24 @@ func (vm *windows) Reinitialize() error {
...
+	wicdCleanupCmd := fmt.Sprintf("%s cleanup --api-server %s --sa-ca %s --sa-token %s --namespace %s",
+		wicdPath, apiServerURL, wicdCAFile, wicdTokenFile, watchNamespace)
+	if out, err := vm.Run(wicdCleanupCmd, true); err != nil {
+		vm.log.Info("failed to cleanup node", "command", wicdCleanupCmd, "output", out)
+		return err
 	}
 	return nil
 }
 
-// ensureServicesAreRemoved ensures that all services installed by WMCO are removed from the instance
```
aravindhp, line 420: This is getting hard to review. You are introducing this function in the previous commit and then removing it in this. I see no point in having such history.

pkg/windows/windows.go
```diff
@@ -879,15 +859,21 @@ func (vm *windows) ensureHNSNetworksAreRemoved() error {
 	// VIP HNS endpoint created by the operator is also deleted when the HNS networks are deleted.
 	for _, network := range []string{BaseOVNKubeOverlayNetwork, OVNKubeOverlayNetwork} {
 		err = wait.Poll(retry.Interval, retry.Timeout, func() (bool, error) {
+			// reinitialize and retry on failure
```
aravindhp, line 862: Please mention why we need to reinitialize.

pkg/daemon/cleanup/cleanup.go
```diff
@@ -81,13 +81,7 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, preserveNode bool, confi
 	if err = removeServices(svcMgr, cmData.Services); err != nil {
 		return err
 	}
-
-	// Delete node if needed
-	if preserveNode || node == nil {
-		return nil
-	}
-	klog.Infof("deleting node %s", node.GetName())
-	return directClient.Delete(ctx, node)
```
aravindhp, line 89: So is the node object no longer getting deleted? I don't see it being deleted by WMCO

pkg/windows/windows.go
```diff
@@ -400,60 +400,24 @@ func (vm *windows) Reinitialize() error {
...
-	for _, svcName := range append(RequiredServices, AzureCloudNodeManagerServiceName) {
-		svc := &service{name: svcName}
-		if err := vm.ensureServiceNotRunning(svc); err != nil {
-			return errors.Wrapf(err, "could not stop service %s", svcName)
-		}
+func (vm *windows) RunWICDCleanup(apiServerURL string, watchNamespace string) error {
+	wicdCleanupCmd := fmt.Sprintf("%s cleanup --api-server %s --sa-ca %s --sa-token %s --namespace %s",
+		wicdPath, apiServerURL, wicdCAFile, wicdTokenFile, watchNamespace)
+	if out, err := vm.Run(wicdCleanupCmd, true); err != nil {
+		vm.log.Info("failed to cleanup node", "command", wicdCleanupCmd, "output", out)
```
aravindhp, line 407: Shouldn't be log.Error()?

aravindhp (changes requested):

pkg/windows/windows.go
mansikulkarni96: Probably the name is confusing, I am just removing this function. The one introduced in previous commit is ensureServiceIsRemoved

pkg/daemon/cleanup/cleanup.go
aravindhp: OK, I see it being deleted [here](https://github.com/openshift/windows-machine-config-operator/blob/014992656747dde271921e814ffd9ef5f03159e0/controllers/controllers.go#L195). Please call this out in your commit message.

mansikulkarni96: [Done](https://github.com/openshift/windows-machine-config-operator/pull/1332/commits/76280c74da1ed3ee132fa0b9b9b71211b14ad442) in this commit.
Removing directories, HNS networks and
deleting the node object will be done by WMCO.

aravindhp (commented):
Thanks for the hard work on this, @mansikulkarni96. LGTM from me. I will let @sebsoto do the official approval.

saifshaikh48 (commented):
/lgtm

aravindhp:
/retest-required

Cluster install [failure](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/1332/pull-ci-openshift-windows-machine-config-operator-master-vsphere-e2e-operator/1617675261782790144):
```
level=error msg=Error: error attaching tags to object ID "vm-518760": POST https://ibmvcenter.vmc-ci.devcluster.openshift.com/rest/com/vmware/cis/tagging/tag-association/id:urn:vmomi:InventoryServiceTag:24386d58-75d9-4a01-b7d6-038a3be2ac3c:GLOBAL?~action=attach: 404 Not Found
level=error
level=error msg=  with vsphere_virtual_machine.vm_master[1],
level=error msg=  on main.tf line 12, in resource "vsphere_virtual_machine" "vm_master":
level=error msg=  12: resource "vsphere_virtual_machine" "vm_master" {
level=error
level=error
```

PR #1331, opened 2022-11-21
Add merge blocker issue template

aravindhp:
This template will be used to created an issue that will result in blocking of PRs merges to a release branch. This is useful when code freeze is in effect. Meant to be used only by @openshift/openshift-team-windows-containers

PR #1335, opened 2022-11-23
Fix merge blocker template

aravindhp:
Use `value` instead of `placeholder` so that the text is actually populated in the issue.


jrvaldes:
/lgtm

Follow-up to:
- https://github.com/openshift/windows-machine-config-operator/pull/1331

sebsoto (approved):

PR #1345, opened 2022-11-30
Adjust annotation wait debug log message

sebsoto:
Change from error level to info level, to stop stack traces from being logged. This is a debug log and will not effect logging in production.

### DIFF
--- a/pkg/windows/windows.go
+++ b/pkg/windows/windows.go
@@ -768,7 +768,7 @@ func (vm *windows) deleteService(svc *service) error {
 	err = wait.Poll(retry.Interval, retry.Timeout, func() (bool, error) {
 		exists, err := vm.serviceExists(svc.name)
 		if err != nil {
-			vm.log.V(1).Error(err, "unable to check if Windows service exists", "service", svc.name)
+			vm.log.V(1).Info("unable to check if Windows service exists", "service", svc.name, "err", err)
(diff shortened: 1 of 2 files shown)


saifshaikh48:
@sebsoto thanks for opening this, can we also change these to debug logs:
- https://github.com/openshift/windows-machine-config-operator/blob/master/pkg/nodeconfig/nodeconfig.go#L440
- https://github.com/openshift/windows-machine-config-operator/blob/master/pkg/windows/windows.go#L743
- https://github.com/openshift/windows-machine-config-operator/blob/master/pkg/windows/windows.go#L771

pkg/windows/windows.go
```diff
@@ -768,7 +768,7 @@ func (vm *windows) deleteService(svc *service) error {
 	err = wait.Poll(retry.Interval, retry.Timeout, func() (bool, error) {
 		exists, err := vm.serviceExists(svc.name)
 		if err != nil {
-			vm.log.V(1).Error(err, "unable to check if Windows service exists", "service", svc.name)
+			vm.log.V(1).Info("unable to check if Windows service exists", "service", svc.name, "err", err)
```
saifshaikh48, line 771: one more place :) https://github.com/openshift/windows-machine-config-operator/blob/master/pkg/windows/windows.go#L743

PR #1352, opened 2022-12-05
WINC-949: Rename powershellVariablesInCommand

sebsoto:
Renames powershellVariablesInCommand to powershellPreScripts to remove the implicit requirement of a variable to replace, which is not always necessary. Also renames powershellVaraiblesInCommand.Name, to VariableName, to be more clear about the purpose of the field with the new naming.

### DIFF
--- a/pkg/daemon/controller/controller.go
+++ b/pkg/daemon/controller/controller.go
@@ -340,7 +340,7 @@ func (sc *ServiceController) expectedServiceCommand(expected servicescm.Service)
 			return "", err
 		}
 	}
-	if len(expected.PowershellVariablesInCommand) > 0 {
+	if len(expected.PowershellPreScripts) > 0 {
 		psVars, err = sc.resolvePowershellVariables(expected)
 		if err != nil {
 			return "", err
@@ -422,13 +422,13 @@ func (sc *ServiceController) resolveNodeVariables(svc servicescm.Service) (map[s
 // replace the variable with. Variables with blank names will not result in a map entry, but their script will be run.
 func (sc *ServiceController) resolvePowershellVariables(svc servicescm.Service) (map[string]string, error) {
 	vars := make(map[string]string)
-	for _, psVar := range svc.PowershellVariablesInCommand {
-		out, err := sc.psCmdRunner.Run(psVar.Path)
+	for _, script := range svc.PowershellPreScripts {
+		out, err := sc.psCmdRunner.Run(script.Path)
 		if err != nil {
-			return nil, errors.Wrapf(err, "could not resolve PowerShell variable %s", psVar.Name)
+			return nil, errors.Wrapf(err, "could not resolve PowerShell variable %s", script.VariableName)
 		}
-		if psVar.Name != "" {
-			vars[psVar.Name] = strings.TrimSpace(out)
+		if script.VariableName != "" {
+			vars[script.VariableName] = strings.TrimSpace(out)
 		}
 	}
 	return vars, nil
--- a/pkg/services/services.go
+++ b/pkg/services/services.go
@@ -33,13 +33,13 @@ func GenerateManifest(kubeletArgsFromIgnition map[string]string, vxlanPort strin
 		return nil, errors.Wrap(err, "could not determine kubelet service configuration spec")
 	}
 	services := &[]servicescm.Service{{
-		Name:                         windows.WindowsExporterServiceName,
-		Command:                      windows.WindowsExporterServiceCommand,
-		NodeVariablesInCommand:       nil,
-		PowershellVariablesInCommand: nil,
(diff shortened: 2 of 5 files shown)


saifshaikh48 (commented):
/lgtm

thanks for doing this :zap:

pkg/services/services.go
```diff
@@ -189,13 +187,13 @@ func getKubeletServiceConfiguration(argsFromIginition map[string]string, debug b
...
-		Bootstrap:                    true,
-		Dependencies:                 []string{windows.ContainerdServiceName},
-		PowershellVariablesInCommand: powershellVars,
-		NodeVariablesInCommand:       nil,
+		Name:                   windows.KubeletServiceName,
+		Command:                kubeletServiceCmd,
+		Priority:               1,
+		Bootstrap:              true,
+		Dependencies:           []string{windows.ContainerdServiceName},
+		PowershellPreScripts:   powershellVars,
```
aravindhp, line 195: Now this looks a bit odd. Why would `PowershellPreScripts` be referred to as `powershellVars`

pkg/daemon/controller/controller.go
```diff
@@ -422,13 +422,13 @@ func (sc *ServiceController) resolveNodeVariables(svc servicescm.Service) (map[s
...
-	for _, psVar := range svc.PowershellVariablesInCommand {
+	for _, psVar := range svc.PowershellPreScripts {
 		out, err := sc.psCmdRunner.Run(psVar.Path)
 		if err != nil {
-			return nil, errors.Wrapf(err, "could not resolve PowerShell variable %s", psVar.Name)
+			return nil, errors.Wrapf(err, "could not resolve PowerShell variable %s", psVar.VariableName)
 		}
-		if psVar.Name != "" {
-			vars[psVar.Name] = strings.TrimSpace(out)
+		if psVar.VariableName != "" {
```
aravindhp, line 430: Now this variable is stuttering because this is no longer a `psVar`.

pkg/servicescm/servicescm.go
```diff
@@ -47,11 +47,12 @@ type NodeCmdArg struct {
 	NodeObjectJsonPath string `json:"nodeObjectJsonPath"`
 }
 
-// PowershellCmdArg describes a PowerShell variable and how its value can be populated
-type PowershellCmdArg struct {
-	// Name is the variable name as it appears in commands
-	Name string `json:"name"`
-	// Path is the location of a PowerShell script whose output is the value of the variable
+// PowershellPreScript describes a PowerShell script to be ran and an optional variable to be populated
+type PowershellPreScript struct {
```
aravindhp, line 51: If you are thinking about `PowerShellPostScript` shouldn't this be generic?

aravindhp (changes requested):

pkg/servicescm/servicescm.go
sebsoto: PowerShellPostScript would just need to be a []string, as there wouldn't be any variables to replace, given it runs after the service.

aravindhp: Would this also be the case for arbitrary scripts that we were going to introduce instead of always having scripts associated with a service? While there might not be variables to replace there either do you think the struct could be reused with other members added to it?

pkg/servicescm/servicescm.go
```diff
@@ -47,11 +47,12 @@ type NodeCmdArg struct {
...
+	// VariableName is the name of a variable which should be replaced by the output of the script. An empty value will
+	// cause no variable replacement to occur, but the script will still be ran.
+	VariableName string `json:"variableName"`
```
aravindhp, line 54: Do we need to add `omitempty`?

aravindhp (approved):

pkg/services/services.go
```diff
@@ -67,9 +67,7 @@ func containerdConfiguration(debug bool) servicescm.Service {
 		Name:                   windows.ContainerdServiceName,
 		Command:                containerdServiceCmd,
 		NodeVariablesInCommand: nil,
-		PowershellVariablesInCommand: []servicescm.PowershellCmdArg{{
-			// Name is left blank as we just want this script to execute, without any service command replacements
```
jrvaldes, line 71: It may be helpful to keep the comment.

pkg/servicescm/servicescm.go
jrvaldes: Yes, I think so. In addition, would be nice to add a unit test to cover this scenario, where a `PowershellPreScript`  gets created without `VariableName`.

jrvaldes (commented):
LGMT, just a few suggestions.

saifshaikh48 (commented):
/lgtm 

LGTM from my end

PR #1364, opened 2022-12-13
Remove EnforceNodeAllocatable kubelet param hack

saifshaikh48:
This PR removes the hacky solution to supply a Windows compatible
kubelet parameter value via the KubeletConfiguration API.
Follow-up to https://github.com/openshift/windows-machine-config-operator/pull/1279#discussion_r998769079


pkg/nodeconfig/nodeconfig.go
```diff
@@ -24,7 +24,7 @@ import (
 	cloudproviderapi "k8s.io/cloud-provider/api"
 	cloudnodeutil "k8s.io/cloud-provider/node/helpers"
 	"k8s.io/kubectl/pkg/drain"
-	kubeletconfig "k8s.io/kubelet/config/v1beta1"
+	kubeletconfig "k8s.io/kubernetes/pkg/kubelet/apis/config"
```
aravindhp, line 27: Please call this out in your commit message i.e. what you are doing and why.

aravindhp (approved):
Thanks for doing, @saifshaikh48. Approved with a minor request.

aravindhp:
@saifshaikh48 you can move the URL to individual lines.

saifshaikh48:
The issue I'm seeing is a strange one. When the kubeletConfiguration object is marshalled to a string, the EnforceNodeAllocatable and CGroupsPerQOS options are capitalized -- kubelet refuses to start because it cannot recognize these options when they start with a capital letter. Exploring if there is another way around this rather than just forcing the config data ToLower()

pkg/nodeconfig/nodeconfig.go
```diff
@@ -346,19 +346,15 @@ func createKubeletConf(clusterServiceCIDR string) (string, error) {
...
-	// the empty value will be omitted, so it would end up being incorrectly populated at service start time.
-	// Can be moved to kubelet configuration object with https://issues.redhat.com/browse/WINC-926
-	enforceNodeAllocatable := []byte("\"enforceNodeAllocatable\":[]}")
-	kubeletConfigData = append(kubeletConfigData, enforceNodeAllocatable...)
-
-	return string(kubeletConfigData), nil
+	kubeletConfigStrData := string(kubeletConfigBytes)
+	// Windows kubelet does not recognize the uppercase versions of these arguments
+	kubeletConfigStrData = strings.ReplaceAll(kubeletConfigStrData, "CgroupsPerQOS", "cgroupsPerQOS")
+	kubeletConfigStrData = strings.ReplaceAll(kubeletConfigStrData, "EnforceNodeAllocatable", "enforceNodeAllocatable")
```
saifshaikh48, line 356: After doing these replacements the node is able to be configured/join the cluster but things like `oc adm node-logs` breaks. I suspect every JSON key in the config file needs to be replaced like this unless we fix the marshalling upstream.

saifshaikh48:
/close

Not actionable until https://github.com/kubernetes/kubernetes/issues/114644 is fixed

PR #1366, opened 2022-12-20
OCPBUGS-5065: Add namespace parameter to WICD commands

sebsoto:
This commit fixes an issue preventing WMCO from being deployed to non-default namespaces. If WMCO was deployed to a namespace other than openshift-windows-machine-config-operator, WICD would be unable to retrieve the services ConfigMap. This is fixed by WMCO informing WICD of the namespace to look in, by passing it in as a parameter in WICD commands.

Fixes OCPBUGS-5065


saifshaikh48 (commented):
/lgtm

saifshaikh48 (commented):
/lgtm

Re-applied after reviewing test approach

PR #1367, opened 2022-12-20
WINC-677: [e2e] Modularize linux curler job creation

saifshaikh48:
This PR consolidates the creation of the linux curl commands in the e2e tests so that there is a standard way of calling the curler jobs.

### DIFF
--- a/test/e2e/network_test.go
+++ b/test/e2e/network_test.go
@@ -131,8 +131,8 @@ func testEastWestNetworking(t *testing.T) {
 
 					// create the curler job based on the specified curlerOS
 					if tt.curlerOS == linux {
-						curlerCommand := []string{"bash", "-c", "curl " + endpointIP}
-						curlerJob, err = testCtx.createLinuxJob("linux-curler-"+strings.ToLower(node.Status.NodeInfo.MachineID), curlerCommand)
+						curlerJob, err = testCtx.createLinuxCurlerJob(strings.ToLower(node.Status.NodeInfo.MachineID),
+							endpointIP, false)
 						require.NoError(t, err, "could not create Linux job")
 					} else if tt.curlerOS == windowsOS {
 						// Always deploy the Windows curler pod on the first node. Because we test scaling multiple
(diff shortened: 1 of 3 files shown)


saifshaikh48:
/test aws-e2e-upgrade

cluster install failure

alinaryan:
Thanks for working on this, 
/lgtm

test/e2e/network_test.go
```diff
@@ -574,6 +575,30 @@ func (tc *testContext) getPodEvents(name string) ([]v1.Event, error) {
 	return podEvents, nil
 }
 
+// createLinuxCurlerJob creates a linux job to curl a specific endpoint. curl must be present in the container image.
+func (testCtx *testContext) createLinuxCurlerJob(jobSuffix, endpoint, portNum, portName string) (*batch.Job, error) {
```
aravindhp, line 579: I suggest making the signature much simpler i.e. combine endpoint, portNum and portName into a URL and construct that at the call site. WDYT?

aravindhp (changes requested):

test/e2e/network_test.go
```diff
@@ -574,6 +575,30 @@ func (tc *testContext) getPodEvents(name string) ([]v1.Event, error) {
...
+	}
+	// curl endpoint every 5 seconds. If the webserver is not accessible for a few minutes, the pod exits with an error.
+	command := []string{
+		"bash",
+		"-c",
+		" while true;" +
+			" do curl " + endpoint + ";" +
+			" if [ $? != 0 ]; then" +
+			" sleep 60;" +
+			" curl " + endpoint + " || exit 1;" +
```
aravindhp, line 594: Why do we need to do curl twice here? I think this can be collapsed into one.

saifshaikh48: Not sure, let me try to rework it

aravindhp: This code is pretty convoluted and can be simplified 😄

test/e2e/network_test.go
```diff
@@ -574,6 +574,28 @@ func (tc *testContext) getPodEvents(name string) ([]v1.Event, error) {
...
 }
 
+// createLinuxCurlerJob creates a linux job to curl a specific endpoint. curl must be present in the container image.
+func (testCtx *testContext) createLinuxCurlerJob(jobSuffix, endpoint string, continuous bool) (*batchv1.Job, error) {
+	// Retries a failed curl attempt once to avoid flakes
+	curlCommand := fmt.Sprintf(
+		"curl %s;"+
+			" if [ $? != 0 ]; then"+
+			" sleep 60;"+
+			" curl %s || exit 1;"+
```
aravindhp, line 584: You are still calling `curl` twice here. Can't you use `break` or `continue`?

test/e2e/network_test.go
saifshaikh48: @aravindhp After a bit I was able to see the purpose of it. I added some docs around it

test/e2e/network_test.go
aravindhp: In other words, always have the [while loop](https://linuxize.com/post/bash-while-loop/).

test/e2e/network_test.go
aravindhp: I still don't see why it needs to be this way. If we are trying to avoid flakes, we just retry the curl command N number of times and exit with an error.

test/e2e/network_test.go
saifshaikh48: the original intent of the 2 curls was to error on the second failure. I'll see if a do-while loop will allow me to remove the second one

aravindhp: You can use [this](https://www.networkworld.com/article/3541298/how-to-repeat-a-linux-command-until-it-succeeds.html) for inspiration.

saifshaikh48: After exploring this further I lean towards the current approach. To make it only 1 curl call we'd have to introduce a counter and it looks more complex than the current code.

aravindhp: Yeah I was thinking the counter would be more clear but I will defer to your judgement on this. Do we really need the continuous option? Won't doing the loop work in all cases?

aravindhp (approved):

mansikulkarni96:
LGTM

saifshaikh48:
/hold

depends on https://github.com/openshift/windows-machine-config-operator/pull/1372 to unblock CI

saifshaikh48:
/retest-required

AWS cluster install failures

PR #1368, opened 2022-12-20
OCPBUGS-5065: Explictly support only the recommended namespace

sebsoto:


README.md
```diff
@@ -17,6 +17,10 @@ More design details can be explored in the [WMCO enhancement](https://github.com
 The operator can be installed from the *community-operators* catalog on OperatorHub.
 It can also be build and installed from source manually, see the [development instructions](docs/HACKING.md).
 
+The operator must be deployed in the `openshift-windows-machine-config-operator` namespace for it to function properly.
```
aravindhp, line 20: @mburke5678 in the OpenShift docs in addition to above, I suggest adding a known upgrade issue. Something along the lines of:
```
If the operator has been installed in a namespace other than `openshift-windows-machine-config-operator`, upgrading to 7.0.0 will result in failure when adding a node or bringing up the operator (TBD). Workaround is to reinstall the operator. Workloads will not be affected.
```

We should then backport the installation update to other releases.

aravindhp (approved):

cmd/operator/main.go
```diff
@@ -87,6 +87,31 @@ func main() {
...
+	// Check that the OperatorGroup + CSV were not deployed with a cluster scope (namespace = ""), and that the
+	// OperatorGroup does not target multiple namespaces. This should not be able to happen as both `AllNamespaces` and
+	// `MultiNamespaces` are not supported InstallModes.
+	if watchNamespace == "" || strings.Contains(watchNamespace, ",") {
+		setupLog.Error(err, "WMCO has an invalid target namespace. "+
+			"OperatorGroup target namespace must be a single, non-cluster-scoped value", "target namespace",
+			watchNamespace)
+		os.Exit(1)
+	}
+
```
aravindhp, line 114: Please wrap this into a ` namespaceCheck()` helper. This `main()` function is growing.

aravindhp (changes requested):

cmd/operator/main.go
```diff
@@ -274,5 +264,11 @@ func getWatchNamespace() (string, error) {
 	if !found {
 		return "", fmt.Errorf("%s must be set", watchNamespaceEnvVar)
 	}
+	// WMCO can only function properly in this namespace, as OLM is unable to adjust the WICD RoleBinding to target a
+	// ServiceAccount in any namespace but this one.
+	supportedNamespace := "openshift-windows-machine-config-operator"
+	if ns != supportedNamespace {
+		return "", fmt.Errorf("the operator must be deployed in the %s namespace", supportedNamespace)
+	}
```
aravindhp, line 272: You cannot just add this to `getWatchNamespace()`. You have now changed the meaning of the function. Like I mentioned before add an explicit `checkNamespace()` method to cleary call this out. We don't want this inlined (as simple as it might be) or buried in another method.

aravindhp (changes requested):

cmd/operator/main.go
aravindhp: I can even live with inlining if the comment is retained.

aravindhp (approved):

aravindhp:
/hold

The only open question I have is if WMCO can apply the required manifests instead of having OLM do it if a non standard namespace is used?

aravindhp:
/close

in favor of openshift/windows-machine-config-operator#1369

PR #1369, opened 2022-12-21
OCPBUGS-5065: Add namespace parameter and RBAC to WICD

saifshaikh48:
This PR fixes an issue preventing WMCO from being deployed
to non-default namespaces. If WMCO was deployed to a namespace
other than openshift-windows-machine-config-operator, WICD would
be unable to retrieve the services ConfigMap. This is fixed by WMCO
informing WICD of the namespace to look in, by passing it in as a
parameter in WICD commands.

It also shifts the responsibility to ensure the proper RBAC resources are
created for the WICD component from OLM to WMCO. OLM was
unable to point the WICD RoleBinding to the right ServiceAccount
namespace since it treats bundle resources as static manifests.

Fixes OCPBUGS-5065


cmd/operator/main.go
```diff
@@ -239,6 +239,14 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Update WICD RBAC with proper namespace
+	wicdResourceName := "windows-instance-config-daemon"
+	if err := configMapReconciler.UpdateWICDRoleBindingNamespace(wicdResourceName); err != nil {
```
aravindhp, line 244: I suggest not updating this always but rather only do this if a non-default namespace is being used.
```golang
if watchNamespace != controller.WMCONamespace {
    // Update WICD RoleBinding
}
```

controllers/configmap_controller.go
```diff
@@ -477,6 +478,28 @@ func (r *ConfigMapReconciler) EnsureServicesConfigMapExists() error {
 	return r.createServicesConfigMapOnBootup()
 }
 
+// UpdateWICDRoleBindingNamespace updates the WICD RoleBinding to point to the right ServiceAccount namespace
+func (r *ConfigMapReconciler) UpdateWICDRoleBindingNamespace(wicdResourceName string) error {
```
aravindhp, line 482: `wicdResource` should not be an input parameter given it is something that will never change.

controllers/configmap_controller.go
```diff
@@ -477,6 +478,28 @@ func (r *ConfigMapReconciler) EnsureServicesConfigMapExists() error {
...
+	for _, serviceAccount := range roleBinding.Subjects {
+		if serviceAccount.Name == wicdResourceName {
+			found = true
+			serviceAccount.Namespace = r.watchNamespace
+		}
+	}
+	if !found {
+		return errors.Errorf("WICD %s missing reference to WICD ServiceAccount", roleBinding.Kind)
+	}
+	_, err = r.k8sclientset.RbacV1().RoleBindings(r.watchNamespace).Update(context.TODO(), roleBinding,
```
aravindhp, line 498: If you are doing this only with a non-default namespace, can this be patch or would it be too complicated?

aravindhp (changes requested):
Thanks for the quick turn around on this @saifshaikh48.  Just a couple of comments.

test/e2e/main_test.go
```diff
@@ -28,6 +28,8 @@ var (
 	privateKeyPath string
 	// wmcoPath is the path to the WMCO binary that was used within the operator image
 	wmcoPath string
+	// wmcoPath is the namespace WMCO is deployed to
```
aravindhp, line 31: copy pasta error 😄

aravindhp (changes requested):

controllers/configmap_controller.go
```diff
@@ -68,6 +71,8 @@ const (
 	UsernameAnnotation = "windowsmachineconfig.openshift.io/username"
 	// ConfigMapController is the name of this controller in logs and other outputs.
 	ConfigMapController = "configmap"
+	// defaultWMCONamespace is the namespace WMCO is installed unless otherwise specified
+	defaultWMCONamespace = "openshift-windows-machine-config-operator"
```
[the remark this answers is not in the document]
saifshaikh48: It is not an exported value so no need to capitalize (I agree, golang is strange in that capitalization affects scope)

aravindhp (approved):

controllers/configmap_controller.go
```diff
@@ -477,6 +482,39 @@ func (r *ConfigMapReconciler) EnsureServicesConfigMapExists() error {
...
+	}
+
+	patchData, err := json.Marshal([]*patch.JSONPatch{patch.NewJSONPatch("replace", "/subjects", roleBinding.Subjects)})
+	if err != nil {
+		return errors.Wrapf(err, "unable to generate patch request body for RoleBinding %s", wicdResourceName)
+	}
+	roleBinding, err = r.k8sclientset.RbacV1().RoleBindings(r.watchNamespace).Patch(context.TODO(), roleBinding.Name,
+		kubeTypes.JSONPatchType, patchData, meta.PatchOptions{})
+	r.log.Info("Patched resource", "RoleBinding",
+		kubeTypes.NamespacedName{Namespace: r.watchNamespace, Name: roleBinding.Name})
```
aravindhp, line 514: This should only be printed in the case there is no error.

aravindhp (changes requested):

controllers/configmap_controller.go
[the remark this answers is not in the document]
jfrancoa: Thanks for explanation.

saifshaikh48:
/test gcp-e2e-operator
/test platform-none-vsphere-e2e-operator

kicking off 1 test with custom namespace, 1 with default

saifshaikh48:
/test gcp-e2e-operator
/test platform-none-vsphere-e2e-operator

retesting with extra commit to resolve permission issues when deconfiguring

saifshaikh48:
dropped the daemonset RBAC commit as it is being done in https://github.com/openshift/windows-machine-config-operator/pull/1372

controllers/configmap_controller.go
```diff
@@ -477,6 +480,66 @@ func (r *ConfigMapReconciler) EnsureServicesConfigMapExists() error {
 	return r.createServicesConfigMapOnBootup()
 }
 
+// EnsureWICDRBAC ensures the WICD RBAC resources exist, creates them if not
+func (r *ConfigMapReconciler) EnsureWICDRBAC() error {
+	wicdResourceName := "windows-instance-config-daemon"
+	_, err := r.k8sclientset.RbacV1().RoleBindings(r.watchNamespace).Get(context.TODO(), wicdResourceName,
+		meta.GetOptions{})
+	if err != nil {
```
aravindhp, line 488: Don't we need to ensure that the `RoleBinding` is what we expect and overwrite it if it is not?

controllers/configmap_controller.go
```diff
@@ -477,6 +480,66 @@ func (r *ConfigMapReconciler) EnsureServicesConfigMapExists() error {
...
+		_, err = r.k8sclientset.RbacV1().RoleBindings(r.watchNamespace).Create(context.TODO(), rb, meta.CreateOptions{})
+		if err != nil {
+			return errors.Wrapf(err, "unable to create RoleBinding %s/%s", r.watchNamespace, wicdResourceName)
+		}
+		r.log.Info("Created resource", "RoleBinding",
+			kubeTypes.NamespacedName{Namespace: r.watchNamespace, Name: rb.Name})
+	}
+
+	_, err = r.k8sclientset.RbacV1().ClusterRoleBindings().Get(context.TODO(), wicdResourceName, meta.GetOptions{})
+	if err == nil || !k8sapierrors.IsNotFound(err) {
```
aravindhp, line 517: Same question here.

controllers/configmap_controller.go
saifshaikh48: addressed with new changes

controllers/configmap_controller.go
```diff
@@ -477,6 +482,98 @@ func (r *ConfigMapReconciler) EnsureServicesConfigMapExists() error {
 	return r.createServicesConfigMapOnBootup()
 }
 
+// EnsureWICDRBAC ensures the WICD RBAC resources exist as expected, creates them if not
+func (r *ConfigMapReconciler) EnsureWICDRBAC() error {
```
aravindhp, line 486: I suggest you have a function `ensureWICDRoleBinding()` that you call from here. Only if that succeeds will you call `ensureWICDClusterRoleBinding()`. This will make the flow much easier to follow.

aravindhp (changes requested):

controllers/configmap_controller.go
saifshaikh48: Makes sense, updated

aravindhp (approved):

controllers/configmap_controller.go
```diff
@@ -477,6 +482,108 @@ func (r *ConfigMapReconciler) EnsureServicesConfigMapExists() error {
...
+		}
+		err = r.k8sclientset.RbacV1().RoleBindings(r.watchNamespace).Delete(context.TODO(), wicdRBACResourceName,
+			meta.DeleteOptions{})
+		if err != nil {
+			return errors.Wrapf(err, "unable to delete RoleBinding %s/%s", r.watchNamespace, wicdRBACResourceName)
+		}
+		r.log.Info("Deleted malformed resource", "RoleBinding",
+			kubeTypes.NamespacedName{Namespace: r.watchNamespace, Name: existingRB.Name})
+	}
+	// create proper resource does if it does not exist
```
mansikulkarni96, line 531: * create proper resource if it does not exist

controllers/configmap_controller.go
```diff
@@ -477,6 +482,108 @@ func (r *ConfigMapReconciler) EnsureServicesConfigMapExists() error {
...
+			return nil
+		}
+		err = r.k8sclientset.RbacV1().ClusterRoleBindings().Delete(context.TODO(), wicdRBACResourceName,
+			meta.DeleteOptions{})
+		if err != nil {
+			return err
+		}
+		r.log.Info("Deleted malformed resource", "ClusterRoleBinding", existingCRB.Name)
+	}
+	// create proper resource does if it does not exist
```
mansikulkarni96, line 579: * // create proper resource if it does not exist

mansikulkarni96 (changes requested):
LGTM minor changes needed.

sebsoto:
adding /hold for @mansikulkarni96 comment

saifshaikh48:
/test aws-e2e-upgrade

Cluster install failure

aravindhp:
/hold
All PRs are seeing:
```
Timed out waiting for the catalog source oo-8lc9h to become ready after 10 minutes.
193
Catalogsource state at timeout is ""
194
Catalogsource image used is "registry.build03.ci.openshift.org/ci-op-zn2hbc63/pipeline@sha256:76a7735bbb7d30c53753d0f5d25a414872dbb6c7129ee7919cf6205557a95316"
```
in the upgrade job and needs to be investigated.

saifshaikh48:
/retest-required

Cluster installed flake (GCP + azure)

PR #1372, opened 2023-01-04
OCPBUGS-5354: Enable draining DaemonSet workloads

sebsoto:
Draining a workload created with a DaemonSet was causing an error due
to lacking DaemonSet `get` permissions. This change adds the RBAC
permissions to resolve said error.

Additionally, the drain flags had to be adjusted by adding the `force`
and `IgnoreAllDaemonSets` flags.

Because draining ignores Daemonset pods, the pod's containers would be
leaked, continuing to run in the background even with kubelet and
containerd stopped. Stopping containerd does not stop containerd
containers running on the instance.

This behavior is preventing WMCO from removing the containerd shim.
To address this, WICD cleanup was made to kill all processes using
the shim, which will stop the leaked containers.

An issue with how the deletion tests were being invoked needed to be
corrected in order to properly test this issue. The tests were
incorrectly using the default expected Node count of 1 machine node
and 1 byoh node, instead of being passed the actual correct value.

Fixes OCPBUGS-5354

### DIFF
--- a/hack/run-ci-e2e-test.sh
+++ b/hack/run-ci-e2e-test.sh
@@ -175,7 +175,7 @@ fi
 # Run the deletion tests while testing operator restart functionality. This will clean up VMs created
 # in the previous step
 if ! $SKIP_NODE_DELETION; then
-  go test ./test/e2e/... -run=TestWMCO/destroy -v -timeout=60m -args --private-key-path=$KUBE_SSH_KEY_PATH --wmco-namespace=$WMCO_DEPLOY_NAMESPACE
+  go test ./test/e2e/... -run=TestWMCO/destroy -v -timeout=60m -args $BYOH_NODE_COUNT_OPTION $MACHINE_NODE_COUNT_OPTION --private-key-path=$KUBE_SSH_KEY_PATH --wmco-namespace=$WMCO_DEPLOY_NAMESPACE
   # Get logs on success before cleanup
   PRINT_UPGRADE=""
   if [[ "$TEST" = "upgrade" ]]; then
--- a/pkg/daemon/cleanup/cleanup.go
+++ b/pkg/daemon/cleanup/cleanup.go
@@ -29,6 +29,7 @@ import (
 
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/controller"
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/manager"
+	"github.com/openshift/windows-machine-config-operator/pkg/daemon/powershell"
 	"github.com/openshift/windows-machine-config-operator/pkg/nodeconfig"
 	"github.com/openshift/windows-machine-config-operator/pkg/servicescm"
 )
@@ -81,6 +82,7 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
 	if err = removeServices(svcMgr, cmData.Services); err != nil {
 		return err
 	}
+	cleanupContainers()
 	return nil
 }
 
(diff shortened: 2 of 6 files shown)


aravindhp (approved):

aravindhp:
/hold

until we figure out:
```
 === CONT  TestWMCO/destroy/Deletion/BYOH_node_removal/ip-10-0-243-24.us-east-2.compute.internal
    delete_test.go:72: 
        	Error Trace:	/go/src/github.com/openshift/windows-machine-config-operator/test/e2e/delete_test.go:72
        	Error:      	Should be true
        	Test:       	TestWMCO/destroy/Deletion/BYOH_node_removal/ip-10-0-243-24.us-east-2.compute.internal
        	Messages:   	directories not removed 
```

sebsoto:
This should be left held until a daemonset is added to our e2e tests

saifshaikh48 (commented):
/lgtm

hack/run-ci-e2e-test.sh
```diff
@@ -175,7 +175,7 @@ fi
 # Run the deletion tests while testing operator restart functionality. This will clean up VMs created
 # in the previous step
 if ! $SKIP_NODE_DELETION; then
-  go test ./test/e2e/... -run=TestWMCO/destroy -v -timeout=60m -args --private-key-path=$KUBE_SSH_KEY_PATH --wmco-namespace=$WMCO_DEPLOY_NAMESPACE
+  go test ./test/e2e/... -run=TestWMCO/destroy -v -timeout=60m -args $BYOH_NODE_COUNT_OPTION $MACHINE_NODE_COUNT_OPTION --private-key-path=$KUBE_SSH_KEY_PATH --wmco-namespace=$WMCO_DEPLOY_NAMESPACE
```
aravindhp, line 178: Why is this change needed?

pkg/daemon/cleanup/cleanup.go
```diff
@@ -104,3 +106,10 @@ func removeServices(svcMgr manager.Manager, services []servicescm.Service) error
 	}
 	return nil
 }
+
+// killContainerdContainers makes a best effort to stop all processes with the name containerd-shim-runhcs-v1
+func killContainerdContainers() {
```
aravindhp, line 111: I suggest a gentler name like `deleteContainerdContainers()` or just `cleanupContainers()`. You are not deleting containers per se in some sense.

test/e2e/delete_test.go
```diff
@@ -112,6 +115,16 @@ func testWindowsNodeDeletion(t *testing.T) {
 	testCtx, err := NewTestContext()
 	require.NoError(t, err)
 
+	// Deploy a DaemonSet and wait until its pods have been made ready on each Windows node. DaemonSet pods cannot be
+	// drained from a Node.
+	// Doing this ensures that WMCO is able to handle containers being present on the instance when deconfiguring.
+	ds, err := testCtx.deployNOOPDaemonSet()
+	require.NoError(t, err, "error creating DaemonSet")
+	defer testCtx.client.K8s.AppsV1().DaemonSets(ds.GetNamespace()).Delete(context.TODO(), ds.GetName(),
```
aravindhp, line 123: Why are you deleting the `DaemonSet`? In the real world case nothing is going to delete the `DaemonSet`. Are you just doing this to do cleanup after all the tests have finished running?

aravindhp (changes requested):
Thanks for doing this, @sebsoto. Please take a look at my comments. 

>Because draining ignores Daemonset pods, the pod's containers would be leaked,

Don't you also have to mention the issue of stopping containerd not deleting the containers?

test/e2e/delete_test.go
sebsoto: Yes this is just cleaning up after test completion.

hack/run-ci-e2e-test.sh
sebsoto: Without this change, the deletion suite is incorrectly using the default expected Node count of 1 machine node and 1 byoh node, instead of being passed the actual correct value.

aravindhp (approved):

saifshaikh48:
/lgtm

:+1:

aravindhp (approved):

PR #1378, opened 2023-01-11
OCPBUGS-5732: [wm_controller] Fix Machine node upgrades

saifshaikh48:
This PR fixes an issue where Machine nodes were not being identified as
out-of-date. This made them skip the deconfigure process, preventing draining
nodes, which would have resulted in a user workload disruptions during upgrade.

Fixes [OCPBUGS-5732](https://issues.redhat.com/browse/OCPBUGS-5732)


aravindhp (approved):
Thanks for the quick turnaround on this, @saifshaikh48

PR #1386, opened 2023-01-16
OCPBUGS-5887: Check for error when deleting directories

sebsoto:
Directory deletion is not being caught because err is nil, while
the actual reported error is present in out. This is being fixed by
using the PowerShell alternative, which will properly report an error.

In order to make use of Remove-Dir, trailing backslashes were removed
from constants, as they were causing unintentional escaping of quotes.

Fixes OCPBUGS-5887

### DIFF
--- a/pkg/windows/windows.go
+++ b/pkg/windows/windows.go
@@ -22,70 +22,70 @@ import (
 
 const (
 	// remoteDir is the remote temporary directory created on the Windows VM
-	remoteDir = "C:\\Temp\\"
+	remoteDir = "C:\\Temp"
 	// GcpGetHostnameScriptRemotePath is the remote location of the PowerShell script that resolves the hostname
 	// for GCP instances
-	GcpGetHostnameScriptRemotePath = remoteDir + payload.GcpGetHostnameScriptName
+	GcpGetHostnameScriptRemotePath = remoteDir + "\\" + payload.GcpGetHostnameScriptName
 	// WinDefenderExclusionScriptRemotePath is the remote location of the PowerShell script that creates an exclusion
 	// for containerd if the Windows Defender Antivirus is active
-	WinDefenderExclusionScriptRemotePath = remoteDir + payload.WinDefenderExclusionScriptName
+	WinDefenderExclusionScriptRemotePath = remoteDir + "\\" + payload.WinDefenderExclusionScriptName
 	// HNSPSModule is the remote location of the hns.psm1 module
-	HNSPSModule = remoteDir + "hns.psm1"
+	HNSPSModule = remoteDir + "\\hns.psm1"
 	// K8sDir is the remote kubernetes executable directory
-	K8sDir = "C:\\k\\"
+	K8sDir = "C:\\k"
 	// KubeconfigPath is the remote location of the kubelet's kubeconfig
-	KubeconfigPath = K8sDir + "kubeconfig"
+	KubeconfigPath = K8sDir + "\\kubeconfig"
 	// logDir is the remote kubernetes log directory
-	logDir = "C:\\var\\log\\"
+	logDir = "C:\\var\\log"
 	// KubeletLogDir is the remote kubelet log directory
-	KubeletLogDir = logDir + "kubelet\\"
+	KubeletLogDir = logDir + "\\kubelet"
 	// KubeProxyLogDir is the remote kube-proxy log directory
-	KubeProxyLogDir = logDir + "kube-proxy\\"
+	KubeProxyLogDir = logDir + "\\kube-proxy"
 	// HybridOverlayLogDir is the remote hybrid-overlay log directory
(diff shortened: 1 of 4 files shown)


pkg/windows/windows.go
```diff
@@ -595,8 +596,8 @@ func (vm *windows) createDirectories() error {
 func (vm *windows) removeDirectories() error {
 	vm.log.Info("removing directories")
 	for _, dir := range RequiredDirectories {
-		if _, err := vm.Run(rmDirCmd(dir), false); err != nil {
-			return errors.Wrapf(err, "unable to remove directory %s", dir)
+		if out, err := vm.Run(rmDirCmd(dir), true); err != nil {
+			return fmt.Errorf("unable to remove directory %s, out: %s, err:%s", dir, out, err)
```
mansikulkarni96, line 600: nit: add a space after err:

pkg/windows/windows.go
```diff
@@ -120,13 +61,73 @@ const (
...
 	// containersFeatureName is the name of the Windows feature that is required to be enabled on the Windows instance.
 	containersFeatureName = "Containers"
-	// wicdCAFile is the name of the file that holds the WICD service account CA certificate
-	wicdCAFile = K8sDir + "sa-ca.crt"
-	// wicdTokenFile is the name of the file that holds the WICD service account secret token
-	wicdTokenFile = K8sDir + "sa-token"
 )
 
 var (
+	// GcpGetHostnameScriptRemotePath is the remote location of the PowerShell script that resolves the hostname
```
aravindhp, line 67: If we are moving away from these being constants, we should look to see if all of the below need to be global variables.

pkg/windows/windows.go
```diff
@@ -961,3 +962,8 @@ func SplitPath(filepath string) (dir string, fileName string) {
 	splitIndex := strings.LastIndexByte(filepath, '\\') + 1
 	return filepath[:splitIndex], filepath[splitIndex:]
 }
+
+// JoinPath joins path elements into a single path separating them with the Windows file separator.
+func JoinPath(elem1, elem2 string) string {
```
aravindhp, line 967: What about using [filepath.Join()](https://pkg.go.dev/path/filepath#Join)?

aravindhp (changes requested):

pkg/windows/windows.go
sebsoto: Using that will use a linux path separator `/` as the WMCO is compiled for linux.

pkg/windows/windows.go
sebsoto: Some definitely don't, such as `WindowsExporterServiceCommand`. I was worried about including too much within this bug fix, let me know if you think thats within scope here.

aravindhp: I would then suggest to fix them in place in the constants without introducing these global variables. i.e. just remove the trailing back slashes where needed and not change all the constants. Alternatively you can do this in the command by removing a suffix. You can then open a follow up PR to fix it up correctly.

pkg/windows/windows.go
aravindhp: If you do have to retain this, please add a comment.

aravindhp (approved):

saifshaikh48 (commented):
/lgtm

sebsoto:
Cluster install failure, retesting azure
```
 level=info msg=Destroying the bootstrap resources...
level=error
level=error msg=Error: waiting for the deletion of Security Rule: (Name "bootstrap_ssh_in" / Network Security Group Name "ci-op-38nswfqk-0e6bb-kgpbl-nsg" / Resource Group "ci-op-38nswfqk-0e6bb-kgpbl-rg"): Code="Canceled" Message="Operation was canceled." Details=[{"code":"CanceledAndSupersededDueToAnotherOperation","message":"Operation DeleteSecurityRuleOperation (1554f237-980e-4a85-9c65-1a5604eba27e) was canceled and superseded by operation PutNetworkSecurityGroupOperation (0573bae3-a287-46de-983d-73bcdb1d9674)."}]
level=error
level=fatal msg=terraform destroy: failed doing terraform destroy: exit status 1
level=fatal
level=fatal msg=Error: waiting for the deletion of Security Rule: (Name "bootstrap_ssh_in" / Network Security Group Name "ci-op-38nswfqk-0e6bb-kgpbl-nsg" / Resource Group "ci-op-38nswfqk-0e6bb-kgpbl-rg"): Code="Canceled" Message="Operation was canceled." Details=[{"code":"CanceledAndSupersededDueToAnotherOperation","message":"Operation DeleteSecurityRuleOperation (1554f237-980e-4a85-9c65-1a5604eba27e) was canceled and superseded by operation PutNetworkSecurityGroupOperation (0573bae3-a287-46de-983d-73bcdb1d9674)."}]
level=fatal
level=fatal 
```

PR #1390, opened 2023-01-18
Remove gcloud dependency in hack/machineset.sh

sebsoto:
Using the image family will use the latest released image from the
family (Windows server 2022) without having to exec gcloud.

The projectID is available within the cluster infrastructure object
and should be retrieved that way.

### DIFF
--- a/hack/machineset.sh
+++ b/hack/machineset.sh
@@ -186,8 +186,7 @@ get_gcp_ms() {
   # For GCP the zone field returns the region + zone, like: `us-central1-a`.
   # Installer created MachineSets only append the `-a` portion, so we should do the same.
   local az_suffix=$(echo $az |awk -F "-" '{print $NF}')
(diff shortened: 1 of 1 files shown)


hack/machineset.sh
```diff
@@ -187,7 +187,6 @@ get_gcp_ms() {
   # Installer created MachineSets only append the `-a` portion, so we should do the same.
   local az_suffix=$(echo $az |awk -F "-" '{print $NF}')
   local projectID=$(gcloud config get-value project)
```
aravindhp, line 189: If the intent is remove the dependency on using `glcoud`, you still need it to find the project ID. Can we hard code this? Or would that be leaking some info?

aravindhp (approved):

saifshaikh48 (commented):
/lgtm

mtnbikenc:
Failover from build04 complete.
/retest-required

PR #1391, opened 2023-01-20
[community-4.12][submodule] update 01/20

alinaryan:
[submodule][ovn-kubernetes] Update to f51b0c8d9


mtnbikenc:
/hold
If we open a direct PR instead of a backport it will break fast forwarding.

PR #1395, opened 2023-01-30
WINC-962: Pick up openshift/kubernetes 1.26 rebase updates

mansikulkarni96:
- [services] Replace deprecated klog flags with kube-log-runner 
- [submodule][kubelet] Update to 149fe52 
- [vendor] Use 1.26 k8s and OpenShift libraries
- [manifests] Set v1.26 as minimum k8s version


pkg/services/services.go
```diff
@@ -178,7 +178,10 @@ func getKubeletServiceConfiguration(argsFromIginition map[string]string, debug b
 		preScripts = append(preScripts, hostnameOverridePowershellVar)
 	}
 
-	kubeletServiceCmd := windows.KubeletPath
+	kubeLogCmd := fmt.Sprintf("%s -log-file=%s -redirect-stderr=false ",
+		windows.KubeLogRunnerPath, windows.KubeletLog)
+
+	kubeletServiceCmd := kubeLogCmd + windows.KubeletPath
```
sebsoto, line 184: Please add windows.KubeletPath as part of the Sprintf, no reason for it to be separate

sebsoto: Does this need to be done for kube-proxy as well?

sebsoto (changes requested):

pkg/services/services.go
mansikulkarni96: yes it needs to be, I am contemplating on whether to add it in this PR or add it as part of kube-proxy bump when the new sdn branch is created. [kube-log-runner](https://github.com/kubernetes/kubernetes/tree/62d9f8ba80f4cc660a88dcc34f56c5c6f7df17ea/staging/src/k8s.io/component-base/logs/kube-log-runner) only has the functionality to direct the logs to a file. For kube-proxy, we have been redirecting to a log-dir for a long time now. Is it acceptable to move them only to info and error files?

pkg/windows/windows.go
```diff
@@ -177,6 +179,7 @@ func getFilesToTransfer() (map[*payload.FileInfo]string, error) {
 		payload.KubeProxyPath:                  K8sDir,
 		payload.KubeletPath:                    K8sDir,
 		payload.AzureCloudNodeManagerPath:      K8sDir,
+		payload.KubeLogRunnerPath:              K8sDir,
```
saifshaikh48, line 182: Should we also add this to [requiredFiles](https://github.com/openshift/windows-machine-config-operator/blob/master/cmd/operator/main.go#L111)?
Although I'm seeing that var is quite outdated so maybe it warrants a task of its own to update properly.

pkg/services/services.go
saifshaikh48: Would doing this for kube-proxy have effect on how we collect log files (e.g. must-gather changes)?

saifshaikh48 (commented):
LGTM aside from a comment

pkg/services/services.go
aravindhp: If `kube-proxy` does not require this change we can leave it as is.

mansikulkarni96: @aravindhp It will once we update to the 1.26 SDN branch. 
@saifshaikh48 yes we will need to update the file info [here](https://github.com/openshift/must-gather/blob/3860800ff98445620e40d474a9f6dd2f36027fca/collection-scripts/gather_windows_node_logs#L6)

pkg/cluster/config_test.go
```diff
@@ -125,9 +125,9 @@ func TestIsValidKubernetesVersion(t *testing.T) {
 		error   bool
 	}{
 		{"cluster version lower than supported version ", "v1.17.1", true},
-		{"cluster version equals supported version", "v1.25.0", false},
-		{"cluster version equals supported version", "v1.26.4", false},
-		{"cluster version greater than supported version ", "v1.27.2", true},
+		{"cluster version equals supported version", "v1.26.0", false},
+		{"cluster version equals supported version", "v1.27.4", false},
+		{"cluster version greater than supported version ", "v1.28.2", true},
```
aravindhp, line 130: Do we need to keep updating this function given it is just a unit test? Or does the version actually change even with the fake client?  If it does, won't this break the e2e tests on kube version bumps in the SDN repos?

aravindhp (commented):
LGTM. Just a few questions. I will leave official approval to @sebsoto.

pkg/cluster/config_test.go
mansikulkarni96: No it doesn't change with the fake client, we are not required to do this change. We have already removed the e2e test that tested with the updated version.

aravindhp: I suggest leaving it as is then

saifshaikh48 (commented):
/lgtm

PR #1397, opened 2023-01-30
WINC-741: WICD takes more responsibility of Node configuration

saifshaikh48:
This PR moves the version annotation application and removal to WICD,
instead of WMCO. This allows us to reduce WMCO's responsibility when reconciling
nodes -- it no longer watches or stops any WICD managed Windows services. It also
shifts the responsibility of stopping services before bootstrapping from WMCO to WICD.
To do this, the WICD cleanup command was with partial best effort cleanup, using 
most recently created services ConfigMap  if no associated node object exists.

The PR also includes miscellaneous changes:
Adds containerd to the services ConfigMap validation e2e test.
Removes hardcoded values and consolidates Machine API namespace & user data constants.
Rehomes the desired version annotation constant to the metadata pkg.

### DIFF
--- a/bundle/manifests/windows-instance-config-daemon_rbac.authorization.k8s.io_v1_clusterrole.yaml
+++ b/bundle/manifests/windows-instance-config-daemon_rbac.authorization.k8s.io_v1_clusterrole.yaml
@@ -12,3 +12,4 @@ rules:
   - list
   - watch
   - get
+  - patch
--- a/controllers/controllers.go
+++ b/controllers/controllers.go
@@ -27,11 +27,6 @@ import (
 	"github.com/openshift/windows-machine-config-operator/version"
 )
 
-const (
-	// machineAPINamespace is the name of the namespace in which machine objects and userData secret is created.
-	machineAPINamespace = "openshift-machine-api"
-)
-
 // instanceReconciler contains everything needed to perform actions on a Windows instance
 type instanceReconciler struct {
 	// Client is the cache client
--- a/pkg/daemon/cleanup/cleanup.go
+++ b/pkg/daemon/cleanup/cleanup.go
@@ -30,7 +30,7 @@ import (
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/controller"
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/manager"
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/powershell"
-	"github.com/openshift/windows-machine-config-operator/pkg/nodeconfig"
+	"github.com/openshift/windows-machine-config-operator/pkg/metadata"
 	"github.com/openshift/windows-machine-config-operator/pkg/servicescm"
 )
 
@@ -54,7 +54,7 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
 		if err != nil {
 			return err
 		}
-		desiredVersion, present := node.Annotations[nodeconfig.DesiredVersionAnnotation]
+		desiredVersion, present := node.Annotations[metadata.DesiredVersionAnnotation]
 		if !present {
 			return errors.Wrapf(err, "node missing desired version annotation")
 		}
--- a/pkg/daemon/cleanup/cleanup.go
+++ b/pkg/daemon/cleanup/cleanup.go
@@ -34,7 +34,7 @@ import (
 	"github.com/openshift/windows-machine-config-operator/pkg/servicescm"
 )
 
-// Deconfigure removes an instance from the cluster.
+// Deconfigure removes all managed services from the instance and the version annotation
 // If we are able to get the services ConfigMap tied to the desired version, all services defined in it are cleaned up.
 // TODO: Otherwise, perform cleanup based on a combination of the OpenShift managed tag and latest services ConfigMap.
 func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace string) error {
@@ -83,6 +83,10 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
 		return err
 	}
 	cleanupContainers()
+
+	if node != nil {
+		return metadata.RemoveVersionAnnotation(ctx, directClient, *node)
+	}
 	return nil
 }
 
--- a/pkg/daemon/cleanup/cleanup.go
+++ b/pkg/daemon/cleanup/cleanup.go
@@ -34,9 +34,10 @@ import (
 	"github.com/openshift/windows-machine-config-operator/pkg/servicescm"
 )
 
-// Deconfigure removes all managed services from the instance and the version annotation
+// Deconfigure removes all managed services from the instance and the version annotation, if it has an associated node.
 // If we are able to get the services ConfigMap tied to the desired version, all services defined in it are cleaned up.
-// TODO: Otherwise, perform cleanup based on a combination of the OpenShift managed tag and latest services ConfigMap.
+// Otherwise, cleanup is based on the latest services ConfigMap.
+// TODO: remove services with the OpenShift managed tag in best effort cleanup https://issues.redhat.com/browse/WINC-853
 func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace string) error {
 	// Cannot use a cached client as no manager will be started to populate cache
 	directClient, err := controller.NewDirectClient(cfg)
@@ -50,16 +51,22 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
 	var node *core.Node
 	var cmData *servicescm.Data
 	err = func() error {
-		node, err = controller.GetAssociatedNode(directClient, addrs)
+		cm := &core.ConfigMap{}
+		node, err := controller.GetAssociatedNode(directClient, addrs)
 		if err != nil {
+			// If no node is found, fetch the most recently created services ConfigMap for best effort cleanup
+			cm, err = servicescm.GetLatest(directClient, ctx, configMapNamespace)
+			if err != nil {
+				return errors.Wrapf(err, "cannot get latest services ConfigMap from namespace %s", configMapNamespace)
+			}
+			cmData, err = servicescm.Parse(cm.Data)
 			return err
 		}
+		// Otherwise, fetch the ConfigMap tied to the node's desired version annotation
 		desiredVersion, present := node.Annotations[metadata.DesiredVersionAnnotation]
 		if !present {
-			return errors.Wrapf(err, "node missing desired version annotation")
+			return errors.Wrapf(err, "node %s missing desired version annotation", node.Name)
 		}
-		// Fetch the CM of the desired version
-		cm := &core.ConfigMap{}
 		err = directClient.Get(ctx,
 			client.ObjectKey{Namespace: configMapNamespace, Name: servicescm.NamePrefix + desiredVersion}, cm)
 		if err != nil {
@@ -69,8 +76,6 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
 		return err
 	}()
 	if err != nil {
-		// TODO: best effort cleanup of all OpenShift managed services and those in the latest services ConfigMap
-		// https://issues.redhat.com/browse/WINC-853
 		return err
 	}
 
--- a/pkg/daemon/controller/controller.go
+++ b/pkg/daemon/controller/controller.go
@@ -205,7 +205,7 @@ func (sc *ServiceController) SetupWithManager(mgr ctrl.Manager) error {
 
 // mapToCurrentNode maps all events to the node associated with this Windows instance
 func (sc *ServiceController) mapToCurrentNode(_ client.Object) []reconcile.Request {
-	return []reconcile.Request{{types.NamespacedName{Name: sc.nodeName}}}
+	return []reconcile.Request{{NamespacedName: types.NamespacedName{Name: sc.nodeName}}}
 }
 
 // Reconcile fulfills the Reconciler interface
@@ -435,8 +435,7 @@ func (sc *ServiceController) resolvePowershellVariables(svc servicescm.Service) 
 // It also returns the config and scheme used to created the client.
 func NewDirectClient(cfg *rest.Config) (client.Client, error) {
 	clientScheme := runtime.NewScheme()
-	err := clientgoscheme.AddToScheme(clientScheme)
-	if err = clientgoscheme.AddToScheme(clientScheme); err != nil {
+	if err := clientgoscheme.AddToScheme(clientScheme); err != nil {
 		return nil, err
 	}
 
--- a/pkg/daemon/controller/controller.go
+++ b/pkg/daemon/controller/controller.go
@@ -247,6 +247,10 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
 		klog.Error(err)
 		return ctrl.Result{}, err
 	}
+	// Version annotation is the indicator that the node was fully configured by this version of the services ConfigMap
+	if err = metadata.ApplyVersionAnnotation(sc.ctx, sc.client, node, desiredVersion); err != nil {
+		return ctrl.Result{}, errors.Wrapf(err, "error updating version annotation on node %s", sc.nodeName)
+	}
 	return ctrl.Result{}, nil
 }
 
--- a/pkg/daemon/controller/controller.go
+++ b/pkg/daemon/controller/controller.go
@@ -49,7 +49,7 @@ import (
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/manager"
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/powershell"
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/winsvc"
-	"github.com/openshift/windows-machine-config-operator/pkg/nodeconfig"
+	"github.com/openshift/windows-machine-config-operator/pkg/metadata"
 	"github.com/openshift/windows-machine-config-operator/pkg/nodeutil"
 	"github.com/openshift/windows-machine-config-operator/pkg/services"
 	"github.com/openshift/windows-machine-config-operator/pkg/servicescm"
@@ -179,15 +179,15 @@ func (sc *ServiceController) SetupWithManager(mgr ctrl.Manager) error {
 		// A node's name will never change, so it is fine to use the name for node identification
 		// The node must have a desired-version annotation for it to be reconcilable
 		CreateFunc: func(e event.CreateEvent) bool {
-			return sc.nodeName == e.Object.GetName() && e.Object.GetAnnotations()[nodeconfig.DesiredVersionAnnotation] != ""
+			return sc.nodeName == e.Object.GetName() && e.Object.GetAnnotations()[metadata.DesiredVersionAnnotation] != ""
 		},
 		UpdateFunc: func(e event.UpdateEvent) bool {
 			// Only process update events if the desired version has changed
 			return sc.nodeName == e.ObjectNew.GetName() &&
-				e.ObjectOld.GetAnnotations()[nodeconfig.DesiredVersionAnnotation] != e.ObjectNew.GetAnnotations()[nodeconfig.DesiredVersionAnnotation]
+				e.ObjectOld.GetAnnotations()[metadata.DesiredVersionAnnotation] != e.ObjectNew.GetAnnotations()[metadata.DesiredVersionAnnotation]
 		},
 		GenericFunc: func(e event.GenericEvent) bool {
-			return sc.nodeName == e.Object.GetName() && e.Object.GetAnnotations()[nodeconfig.DesiredVersionAnnotation] != ""
+			return sc.nodeName == e.Object.GetName() && e.Object.GetAnnotations()[metadata.DesiredVersionAnnotation] != ""
 		},
 		DeleteFunc: func(e event.DeleteEvent) bool {
 			return false
@@ -223,7 +223,7 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
 		return ctrl.Result{}, err
 	}
 
-	desiredVersion, present := node.Annotations[nodeconfig.DesiredVersionAnnotation]
+	desiredVersion, present := node.Annotations[metadata.DesiredVersionAnnotation]
 	if !present {
 		// node missing desired version annotation, don't requeue
 		return ctrl.Result{}, nil
--- a/pkg/metadata/metadata.go
+++ b/pkg/metadata/metadata.go
@@ -9,9 +9,11 @@ import (
 	"github.com/pkg/errors"
 	core "k8s.io/api/core/v1"
 	kubeTypes "k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/openshift/windows-machine-config-operator/pkg/patch"
+	"github.com/openshift/windows-machine-config-operator/pkg/retry"
 )
 
 const (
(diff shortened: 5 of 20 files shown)


saifshaikh48:
/test aws-e2e-upgrade
/test gcp-e2e-operator

fixed a validation e2e test

saifshaikh48:
/test aws-e2e-upgrade
/test gcp-e2e-operator

Everything passed last run except the deletion test. I've added a check before removing the version annotation, this should allow node deletion to properly complete

saifshaikh48:
No issues in local testing. Seeing [OCPBUGS-6611 - Hybrid Overlay logfile is in use and cannot be deleted](https://issues.redhat.com/browse/OCPBUGS-6611) in CI though

test/e2e/validation_test.go
```diff
@@ -481,9 +481,13 @@ func testServicesConfigMap(t *testing.T) {
 		require.NoError(t, err, "unable to parse ConfigMap data")
 
 		// Check that the expected services are defined within the CM data
```
aravindhp, line 483: So this is doing more than just what the comment says. I suggest updating the comment.

bundle/manifests/windows-instance-config-daemon_rbac.authorization.k8s.io_v1_clusterrole.yaml
```diff
@@ -12,3 +12,4 @@ rules:
   - list
   - watch
   - get
+  - patch
```
aravindhp, line 15: Do we need to remove this permission from WMCO?

pkg/daemon/cleanup/cleanup.go
```diff
@@ -83,6 +85,17 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
...
+		// Clear the version annotation from the node object to indicate the node is not configured
+		patchData, err := metadata.GenerateRemovePatch([]string{}, []string{metadata.VersionAnnotation})
+		if err != nil {
+			return errors.Wrapf(err, "error creating version annotation remove request")
+		}
+		err = directClient.Patch(ctx, node, client.RawPatch(types.JSONPatchType, patchData))
+		if err != nil {
+			return errors.Wrapf(err, "error removing version annotation from node %s", node.GetName())
+		}
+	}
```
aravindhp, line 98: Please make this a separate helper in the metadata package.

pkg/daemon/controller/controller.go
```diff
@@ -247,6 +248,11 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
 		klog.Error(err)
 		return ctrl.Result{}, err
 	}
+	// Version annotation is the indicator that the node was fully configured by this version of the services ConfigMap
+	err = metadata.ApplyAnnotations(sc.client, sc.ctx, node, map[string]string{metadata.VersionAnnotation: desiredVersion})
```
aravindhp, line 252: I would rather make `ApplyAnnotations` an internal helper and expose `metadata.ApplyDesiredVersion()` and so on.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -460,9 +427,9 @@ func (nc *nodeConfig) setNode(quickCheck bool) error {
 	return errors.Wrapf(err, "unable to find node with address %s", instanceAddress)
 }
 
-// waitForNodeAnnotation checks if the node object has the given annotation and waits for retry.Interval seconds and
-// returns an error if the annotation does not appear in that time frame.
-func (nc *nodeConfig) waitForNodeAnnotation(annotation string) error {
+// waitForNodeAnnotation checks if the node object has the given annotation with the expected value.
+// Waits for retry.Interval seconds and returns an error if it does not appear in that time frame.
+func (nc *nodeConfig) waitForNodeAnnotation(annotation, expectedVal string) error {
```
aravindhp, line 432: This function can be moved to the metadata package and be made specific to the version annotation.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -193,13 +194,14 @@ func (nc *nodeConfig) Configure() error {
 			return errors.Wrap(err, "configuring WICD failed")
 		}
 		// Set the desired version annotation, communicating to WICD which Windows services configmap to use
-		if err := nc.applyLabelsAndAnnotations(nil, map[string]string{DesiredVersionAnnotation: version.Get()}); err != nil {
+		if err := nc.applyLabelsAndAnnotations(nil, map[string]string{DesiredVersionAnnotation: wmcoVersion}); err != nil {
```
aravindhp, line 197: Do we use `applyLabelsAndAnnotations()` only for the version now? If yes, then we should move that to metadata package and make it specific to the version.

aravindhp (changes requested):

bundle/manifests/windows-instance-config-daemon_rbac.authorization.k8s.io_v1_clusterrole.yaml
saifshaikh48: Nope, WMCO still patches the node with labels (worker label, BYOH label) and other annotations (public key hash, username anno)

pkg/nodeconfig/nodeconfig.go
saifshaikh48: Yep, we also use it for [other labels/annotations WMCO applies](https://github.com/openshift/windows-machine-config-operator/blob/master/pkg/nodeconfig/nodeconfig.go#L181). But let me see if I can reuse what we have in the metadata pkg to get rid of this method

pkg/nodeconfig/nodeconfig.go
```diff
@@ -193,13 +194,15 @@ func (nc *nodeConfig) Configure() error {
 			return errors.Wrap(err, "configuring WICD failed")
 		}
 		// Set the desired version annotation, communicating to WICD which Windows services configmap to use
-		if err := nc.applyLabelsAndAnnotations(nil, map[string]string{DesiredVersionAnnotation: version.Get()}); err != nil {
+		if err := metadata.ApplyLabelsAndAnnotations(nc.client, context.TODO(), *nc.node, nil,
```
aravindhp, line 197: Adding a wrapper `metadata.ApplyVersionAnnotation()` will make the code readable when paired with `metadata.WaitForVersionAnnotation()`.

pkg/windows/windows.go
```diff
@@ -419,15 +419,27 @@ func (vm *windows) Deconfigure(watchNamespace string, apiServerURL string) error
...
+	// stopped and their files closed before modifying them.
+	if err := vm.ensureWICDExists(); err != nil {
 		return err
 	}
-	if err := vm.ensureRequiredServicesStopped(); err != nil {
-		return errors.Wrap(err, "unable to stop all services")
+	// Stop any services that may be running. This prevents the node being shown as Ready after a failed configuration.
+	// TODO: WICD cleanup can only be called if a node object exists since it relies on the desired version annotation.
+	// The nodeExists flag should be removed once best-effort cleanup is done, https://issues.redhat.com/browse/WINC-853
+	if nodeExists {
```
aravindhp, line 437: Shouldn't you stopping services always even if the node object does not exist like we were doing before? Consider the case where we error out during kubelet setup and should start again. At that point it is best we stop all the services kubelet depends on.

aravindhp (changes requested):

pkg/secrets/secrets.go
```diff
@@ -16,8 +16,8 @@ import (
 const (
 	// UserDataSecret is the name of the userData secret that WMCO creates
 	UserDataSecret = "windows-user-data"
-	// UserDataNamespace is the namespace of the userData secret that WMCO creates
-	UserDataNamespace = "openshift-machine-api"
+	// MachineAPINamespace is the name of the namespace in which machine objects and userData secret is created.
+	MachineAPINamespace = "openshift-machine-api"
```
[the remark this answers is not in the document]
saifshaikh48: Are you suggesting just moving the consolidated `secrets.MachineAPINamespace` const to the `cluster` pkg? Or not consolidating at all and moving the existing constant `controller.machineAPINamespace` to `cluster`?

pkg/windows/windows.go
```diff
@@ -875,14 +890,6 @@ func (vm *windows) ensureWICDSecretContent(credentials *Authentication) error {
 	return vm.EnsureFileContent(credentials.Token, tokenFileName, tokenDir)
 }
 
-// deconfigureWICD ensures the WICD service running on the Windows instance is removed
```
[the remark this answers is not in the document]
saifshaikh48: I didn't feet it was a super useful wrapper honestly. The analog ConfigureWICD() does more than just call ensureServiceIsRunning().

I'm still a proponent of removing it given the function it wraps takes in a simple parameter, but if more folks want to keep it that's fine with me.

pkg/windows/windows.go
saifshaikh48: This is the issue that almost made me drop this from the story's acceptance criteria. WICD [does not support cleanup if a node object doesn't exist](https://github.com/openshift/windows-machine-config-operator/blob/master/pkg/daemon/cleanup/cleanup.go#L53) - this is part of the best-effort cleanup work that was punted from the epic.

saifshaikh48: One way around this is to pass the desiredVersion to the cleanup command instead of relying on a node annotation, as we do to the bootstrap command. @sebsoto likely has thoughts on this, so looping him in

pkg/metadata/metadata.go
```diff
@@ -82,3 +84,40 @@ func ApplyAnnotations(c client.Client, ctx context.Context, node core.Node, anno
...
+		if err != nil {
+			return errors.Wrapf(err, "error removing version annotation from node %s", node.GetName())
+		}
+	}
+	return nil
+}
+
+// WaitForVersionAnnotation checks if the node object has the version annotation with the given value.
+// Waits for retry.Interval seconds and returns an error if it does not appear in that time frame.
+func WaitForVersionAnnotation(c client.Client, node *core.Node, expectedVal string) error {
```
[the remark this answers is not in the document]
saifshaikh48: I'll be consistent and pass it in. We pass the context to these metadata functions at many other call sites, just not from the nodeconfig pkg.

test/e2e/validation_test.go
```diff
@@ -480,13 +480,14 @@ func testServicesConfigMap(t *testing.T) {
...
 		require.NoError(t, err, "unable to parse ConfigMap data")
 
-		// Check that the expected services are defined within the CM data
-		expectedSvcs, err := tc.expectedWindowsServices(windows.RequiredServicesOwnedByWICD)
+		// Check that only the expected services are defined within the CM data. WICD itself should not be defined in it
+		expectedSvcs, err := tc.expectedWindowsServices(windows.RequiredServices)
+		expectedSvcs[windows.WicdServiceName] = false
 		require.NoError(t, err)
-		for svcName, shouldBeRunning := range expectedSvcs {
+		for svcName, shouldBeInConfigMap := range expectedSvcs {
```
mansikulkarni96, line 487: nit: If this is a boolean check, a name like inConfigMap could be sufficient

pkg/windows/windows.go
```diff
@@ -419,15 +419,27 @@ func (vm *windows) Deconfigure(watchNamespace string, apiServerURL string) error
...
+func (vm *windows) Bootstrap(desiredVer, apiServerURL, watchNamespace string, credentials *Authentication, nodeExists bool) error {
 	vm.log.Info("configuring")
-	// Make sure WICD service is not running before calling node bootstrap
-	if err := vm.deconfigureWICD(); err != nil {
+
+	// Make sure WICD service is not running before calling node cleanup and/or bootstrap
+	if err := vm.ensureServiceIsRemoved(WicdServiceName); err != nil {
+		return errors.Wrapf(err, "error ensuring %s Windows service is removed", WicdServiceName)
+	}
+	// We have to do ensure WICD exists separately before the rest of the files because we must ensure services are
```
mansikulkarni96, line 429: nit: WICD binary exists

mansikulkarni96 (commented):

test/e2e/validation_test.go
saifshaikh48: I think I'll leave this as is for readability, since it is the expected value in the assert.

pkg/metadata/metadata.go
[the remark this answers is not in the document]
aravindhp: I am fine with that too.

pkg/windows/windows.go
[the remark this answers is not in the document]
aravindhp: Having such wrappers is a good pattern and adopted in other projects like the Linux kernel. It makes the code easier to read and understandable without having to parse what the helper is actually doing.

pkg/windows/windows.go
aravindhp: It does seem weird that WICD cannot perform cleanup if the node object is not present.

>One way around this is to pass the desiredVersion to the cleanup command instead of relying on a node annotation, as we do to the bootstrap command.

I am fine with that. Why do we even need the `desiredVersion` annotation if we are passing it to WICD?

controllers/controllers.go
```diff
@@ -27,11 +27,6 @@ import (
 	"github.com/openshift/windows-machine-config-operator/version"
 )
 
-const (
```
aravindhp, line 30: Please prefix the commit header with `[cluster]`

pkg/daemon/controller/controller.go
```diff
@@ -247,6 +247,11 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
 		klog.Error(err)
 		return ctrl.Result{}, err
 	}
+	// Version annotation is the indicator that the node was fully configured by this version of the services ConfigMap
+	if err = metadata.ApplyLabelsAndAnnotations(sc.ctx, sc.client, node, nil,
+		map[string]string{metadata.VersionAnnotation: desiredVersion}); err != nil {
+		return ctrl.Result{}, errors.Wrapf(err, "error updating version annotation on node %s", sc.nodeName)
+	}
```
aravindhp, line 254: This can be `ApplyVersionAnnotation`

aravindhp (changes requested):

pkg/daemon/cleanup/cleanup.go
```diff
@@ -83,7 +83,7 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
 		return err
 	}
 	cleanupContainers()
-	return nil
+	return metadata.RemoveVersionAnnotation(ctx, directClient, *node)
```
sebsoto, line 86: Its not clear why this is being done, please document what the full effects of the `Deconfigure`   function are in the function doc. All the high level changes to the node object and vm should be captured.

pkg/daemon/cleanup/cleanup.go
```diff
@@ -54,7 +54,7 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
 		if err != nil {
 			return err
 		}
-		desiredVersion, present := node.Annotations[nodeconfig.DesiredVersionAnnotation]
+		desiredVersion, present := node.Annotations[metadata.DesiredVersionAnnotation]
```
sebsoto, line 57: In the future, please separate a "rename" change like this into a different commit, it doesn't have any impact on functionality but creates noise for the real functionality changes.

pkg/metadata/metadata.go
```diff
@@ -82,3 +88,43 @@ func ApplyAnnotations(c client.Client, ctx context.Context, node core.Node, anno
...
+		if err != nil {
+			return errors.Wrapf(err, "error removing version annotation from node %s", node.GetName())
+		}
+	}
+	return nil
+}
+
+// WaitForVersionAnnotation checks if the node object has the version annotation with the given value.
+// Waits for retry.Interval seconds and returns an error if it does not appear in that time frame.
+func WaitForVersionAnnotation(ctx context.Context, c client.Client, node core.Node, expectedVal string) error {
```
sebsoto, line 113: Any reason this can't just wait until desiredVersion annotation == version annotation?

pkg/metadata/metadata.go
```diff
@@ -82,3 +88,43 @@ func ApplyAnnotations(c client.Client, ctx context.Context, node core.Node, anno
 	}
 	return nil
 }
+
+func ApplyDesiredVersionAnnotation(ctx context.Context, c client.Client, node core.Node, value string) error {
```
sebsoto, line 92: does this need the value arg when it can just use version.Get() internally?
Or is there a dependency issue there?

pkg/metadata/metadata.go
sebsoto: Also, why is node being passed in when `Get` is being called to retrieve it? The node name should be the argument here.

pkg/metadata/metadata.go
```diff
@@ -82,3 +88,43 @@ func ApplyAnnotations(c client.Client, ctx context.Context, node core.Node, anno
...
+	}
+	return nil
+}
+
+// WaitForVersionAnnotation checks if the node object has the version annotation with the given value.
+// Waits for retry.Interval seconds and returns an error if it does not appear in that time frame.
+func WaitForVersionAnnotation(ctx context.Context, c client.Client, node core.Node, expectedVal string) error {
+	var found bool
+	err := wait.Poll(retry.Interval, retry.Timeout, func() (bool, error) {
+		err := c.Get(ctx, kubeTypes.NamespacedName{Namespace: node.Namespace, Name: node.Name}, &node)
```
sebsoto, line 116: Nodes don't have a namespace, only NamespacedName.Name needs to be set

pkg/nodeconfig/nodeconfig.go
```diff
@@ -193,13 +192,14 @@ func (nc *nodeConfig) Configure() error {
...
-		// Now that basic kubelet configuration is complete, configure networking in the node
-		if err := nc.configureNetwork(); err != nil {
-			return errors.Wrap(err, "configuring node network failed")
+		// Wait for version annotation. This prevents uncordoning the node until all node services and networks are up
+		if err := metadata.WaitForVersionAnnotation(context.TODO(), nc.client, *nc.node, wmcoVersion); err != nil {
+			return errors.Wrapf(err, "error waiting for proper %s annotation for node %s", metadata.VersionAnnotation,
+				nc.node.GetName())
 		}
 
 		// Now that the node has been fully configured, add the version annotation to signify that the node
```
sebsoto, line 205: Outdated comment and potentially outdated logic

pkg/windows/windows.go
sebsoto: > Why do we even need the desiredVersion annotation if we are passing it to WICD

The original thought process there is that by changing the desiredVersion annotation we can trigger an upgrade without stopping and changing the WICD service, in the case where there is no change to the WICD exe.

>  One way around this is to pass the desiredVersion to the cleanup command instead of relying on a node annotation, as we do to the bootstrap command.

This works for me. Another way is to just use the latest windows-services configmap in the absence of a node object

sebsoto (changes requested):

pkg/daemon/cleanup/cleanup.go
saifshaikh48: I've split it out

pkg/metadata/metadata.go
saifshaikh48: No strong reason for either. I've taken both suggestions, thanks

pkg/metadata/metadata.go
saifshaikh48: No dependency issue, updated

pkg/nodeconfig/nodeconfig.go
saifshaikh48: Adjusted the comment. The logic is still fine as we want to update the nc.node object after applying all these annotations to prevent stale resource

pkg/windows/windows.go
saifshaikh48: > Another way is to just use the latest windows-services configmap in the absence of a node object

I went ahead and implemented this (which is half of https://issues.redhat.com/browse/WINC-853)

pkg/metadata/metadata.go
saifshaikh48: `ApplyDesiredVersionAnnotation()` can rely on `version.Get()`, but note that we do have to pass in the value for `ApplyVersionAnnotation()` since it is used from WICD. Calling functions that do `version.Get()` from WICD results in an empty version value.

pkg/metadata/metadata.go
```diff
@@ -86,3 +89,51 @@ func ApplyAnnotations(c client.Client, ctx context.Context, node core.Node, anno
...
+	}
+	return nil
+}
+
+// WaitForVersionAnnotation checks if the node object has equivalent version and desiredVersion annotations.
+// Waits for retry.Interval seconds and returns an error if the version annotation does not appear in that time frame.
+func WaitForVersionAnnotation(ctx context.Context, c client.Client, nodeName string) error {
+	var found bool
+	node := &core.Node{}
+	err := wait.Poll(retry.Interval, retry.Timeout, func() (bool, error) {
```
aravindhp, line 123: Checking  every 15 seconds for the version annotation seems very aggressive. And is a 10 minute timeout sufficient? I suggest we introduce specific constants for this case.

pkg/windows/windows.go
```diff
@@ -469,17 +473,20 @@ func (vm *windows) ConfigureWICD(apiServerURL, watchNamespace string, credential
...
-func (vm *windows) ensureRequiredServicesStopped() error {
-	// TODO: WMCO no longer stops services started by WICD before calling WICD bootstrap command
-	//		 This code should be removed as part of: https://issues.redhat.com/browse/WINC-741
-	for _, svcName := range append(RequiredServices, AzureCloudNodeManagerServiceName) {
-		svc := &service{name: svcName}
-		if err := vm.ensureServiceNotRunning(svc); err != nil {
-			return errors.Wrapf(err, "could not stop service %s", svcName)
-		}
+
+// ensureWICDExists ensures the WICD binary and SA credential files exist. Creates the directories and files, if needed.
```
aravindhp, line 477: You are overloading this function. Either rename it to `ensureWICDArtifactsExists()` and reflect what it is doing (but overloading in this manner is not good practice) or split the calls at the call site.

aravindhp (changes requested):

pkg/windows/windows.go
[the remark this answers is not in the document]
saifshaikh48: Understood, I reverted the func removal :+1:

pkg/windows/windows.go
saifshaikh48: I'll split the calls at the call site, thanks

pkg/metadata/metadata.go
saifshaikh48: In my testing I saw successful version annotation application within 2 minutes. The 10 min timeout should be fine IMO. Should I change this to poll every 30 seconds instead?

aravindhp: Given what you are seeing 10 minutes is good. And if it is 2 minutes total time I am fine with 15 min intervals also.

saifshaikh48: Looks like the WICD work has sped up configuration times significantly. Machine-backed node took less than 2 min :rocket: 
```
20:34:19 - init SSH connection 
20:35:58 - start WICD controller
20:36:36 - kubelet posts ready status
20:37:13 - instance has been fully configured as a worker node
```

pkg/windows/windows.go
```diff
@@ -469,15 +476,18 @@ func (vm *windows) ConfigureWICD(apiServerURL, watchNamespace string, credential
...
-		if err := vm.ensureServiceNotRunning(svc); err != nil {
-			return errors.Wrapf(err, "could not stop service %s", svcName)
-		}
+
+// ensureWICDExists ensures the WICD executable exists. Creates the destination directory and binary file, if needed.
+func (vm *windows) ensureWICDExists(credentials *Authentication) error {
+	if _, err := vm.Run(mkdirCmd(K8sDir), false); err != nil {
+		return errors.Wrapf(err, "unable to create remote directory %s", K8sDir)
+	}
+	f, err := payload.NewFileInfo(payload.WICDPath)
```
sebsoto, line 485: nit: A more descriptive variable name could be helpful

sebsoto (changes requested):
LGTM besides a nit, will let @aravindhp approve if his comments have been addressed

mansikulkarni96:
LGTM

pkg/metadata/metadata.go
aravindhp: I expect we can make this even faster by moving the HNS network removal to WICD as part of it's clean up or a separate sub-command.

aravindhp (approved):

saifshaikh48:
/test gcp-e2e-operator

Cluster spin up failure, timeout 
```
INFO Updating openshift-install configmap map with the start and end times. 
…
error: failed to get configmap
```

saifshaikh48:
/retest-required

couple of cluster install failures and a flake “waiting for no-op daemonset to scale”

saifshaikh48:
/retest-required

Operator install failure, retesting since there have been no changes since everything was green...

aravindhp:
@saifshaikh48 operator did not fail to install. Please see the full [log](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/1397/pull-ci-openshift-windows-machine-config-operator-master-azure-e2e-operator/1623652132034973696) for the real failure. The operator eventually installs. 
```
2023/02/09 15:02:36 failed to collect logs from e2e-wm-hnp22: error waiting for job to succeed
<snip>
2023/02/09 15:04:02 failed to change e2e-cmg6s default shell to Powershell
```
Not sure if it is a flake.

The other `aws-e2c-ccm-install` failed on cluster bring up:
```
error: unable to parse image registry.build03.ci.openshift.org/ci-op-3b73d7ws/release@sha256:4493a270dade7ccd5b08876f19001942fb1682529e69e04825c6a328c5e5b734: cannot retrieve image configuration for manifest sha256:4493a270dade7ccd5b08876f19001942fb1682529e69e04825c6a328c5e5b734: Get "https://d29zazo61hdry4.cloudfront.net/docker/registry/v2/blobs/sha256/3f/3fa721b4c403cee355f28f4dec4e544715c7b6500cde8db9183a76f3bfa24d6a/data?Expires=1675946989&Signature=T2Hdn7BCu4WSfeyw89Y3JxgwxCK2Gbk2rezDHZVzIk-c2ACtgnGHW4SHOWBwGfZrZ4p5w4i7QU0P1IK51vq3fJEqulnpvYtbM1uERAvGSs6ANYrgj-Kd9VRHGBBa0DsC4RrTRyjhiQKglaz9NyxYoIVjskexRyhSPfGQ-2DEiM4nE3PY1EqO4VnYZDYbksV11NcXeVry9sY3PoAuO6MkDenABZf4~x05ACDz88GYe9RRyfruFz1MvR9gz2Onyhy69FQenrMHaOCRiN5AX4rOtmo7CRaDufLbDGwY35ukwk~ckmcgoedDXyxtboXbvW5xdG4CiHDNR2mWH67guJBrtw__&Key-Pair-Id=K35IN2XBI3W7KZ": read tcp 10.129.28.196:36408->108.138.82.27:443: read: connection reset by peer
```
We should report this if we are seeing it consistently.

aravindhp:
/hold

I am again [seeing](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/1397/pull-ci-openshift-windows-machine-config-operator-master-azure-e2e-operator/1623716281704255488):
```
 2023/02/09 17:50:40 35m0.22467367s time is required to configure 1 nodes
    create_test.go:161: 
        	Error Trace:	/go/src/github.com/openshift/windows-machine-config-operator/test/e2e/create_test.go:161
        	            	timed out waiting for the condition
        	Test:       	TestWMCO/create/Creation/Machine_controller
        	Messages:   	Windows node creation failed
W0209 17:50:40.706877   14646 warnings.go:70] would violate PodSecurity "restricted:latest": host namespaces (hostNetwork=true), allowPrivilegeEscalation != false (container "print-logs-e2e-wm-jrjwn" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container "print-logs-e2e-wm-jrjwn" must set securityContext.capabilities.drop=["ALL"]), seccompProfile (pod or container "print-logs-e2e-wm-jrjwn" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
2023/02/09 17:50:56 failed to collect logs from e2e-wm-jrjwn: error waiting for job to succeed: 
```
@saifshaikh48 please investigate

saifshaikh48:
/retest-required

There seems to be an infrastructure issue in azure, machine-backed nodes are taking uncharacteristically long to get configured (across PRs). Retesting to see if the behavior persists

saifshaikh48:
/unhold

Back to normal configuration times
`16m0.159533856s time is required to configure 1 nodes`

PR #1399, opened 2023-02-02
OCPBUGS-6611: Wait until processes exit when stopping services

sebsoto:
This commit waits for the process associated with a service to exit
after stopping a service to ensure that WICD is correctly returning that
cleanup was successful, only after all the services have fully stopped.

If the process does not exit before the timeout period, the process will
be terminated by WICD.

### DIFF
--- a/pkg/daemon/manager/manager.go
+++ b/pkg/daemon/manager/manager.go
@@ -11,8 +11,10 @@ import (
 	"golang.org/x/sys/windows"
 	"golang.org/x/sys/windows/svc"
 	"golang.org/x/sys/windows/svc/mgr"
+	"k8s.io/apimachinery/pkg/util/wait"
 
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/winsvc"
+	"github.com/openshift/windows-machine-config-operator/pkg/retry"
 )
 
 type Manager interface {
@@ -139,8 +141,7 @@ func (m *manager) EnsureServiceState(service winsvc.Service, state svc.State) er
 				return errors.Wrapf(err, "unable to stop dependent service %s", dependentServiceName)
 			}
 		}
-		_, err = service.Control(svc.Stop)
-		if err != nil {
+		if err := m.stopServiceAndProcess(winSvc); err != nil {
 			return err
 		}
 	default:
@@ -148,6 +149,37 @@ func (m *manager) EnsureServiceState(service winsvc.Service, state svc.State) er
 	}
 	// Wait for the state change to actually take place
 	return winsvc.WaitForState(service, state)
+}
+
+// Stop the service, and wait for the process associated with the service to stop
+func (m *manager) stopServiceAndProcess(winSvc *mgr.Service) error {
+	status, err := winSvc.Query()
+	if err != nil {
+		return errors.Wrap(err, "error querying service")
+	}
+	var pHandle windows.Handle
+	// A value of 0 indicates that no process is running
+	if status.ProcessId != 0 {
+		pHandle, err = windows.OpenProcess(windows.PROCESS_TERMINATE|windows.PROCESS_QUERY_INFORMATION, false,
+			status.ProcessId)
+		if err != nil {
+			return errors.Wrap(err, "unable to open service's associated process")
+		}
+		defer windows.CloseHandle(pHandle)
+	}
+	_, err = winSvc.Control(svc.Stop)
+	if err != nil {
+		return err
+	}
+	if status.ProcessId != 0 {
+		if err = waitForProcessToStop(pHandle); err != nil {
+			// Terminate the process if it does not exit on its own
+			if err = windows.TerminateProcess(pHandle, uint32(1)); err != nil {
+				return errors.Wrap(err, "error terminating stalled process")
+			}
+		}
+	}
+	return nil
 }
 
 // listDependentServices returns a list of names of all services dependent on the given service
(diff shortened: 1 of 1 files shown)


pkg/daemon/cleanup/cleanup.go
```diff
@@ -107,6 +120,35 @@ func removeServices(svcMgr manager.Manager, services []servicescm.Service) error
...
+	if len(exeSplit) < 2 {
+		return "", fmt.Errorf("command does not include '.exe'")
+	}
+	return filepath.Base(exeSplit[0]), nil
+}
+
+// processRunning returns true if the given process is running
+func processRunning(name string) bool {
+	// Assume that an error means the process doesn't exist.
+	// A better solution for this would be to call CreateToolhelp32Snapshot and iterate through the processes
```
aravindhp, line 143: The better solution seems like a heavy hammer. What is the advantage of doing that? Can't you get the [StatusHandle()](https://cs.opensource.google/go/x/sys/+/v0.4.0:windows/svc/service.go;l=297) of the service which contains the pid and the use that to wait for the process to stop?

pkg/daemon/cleanup/cleanup.go
```diff
@@ -98,6 +103,14 @@ func removeServices(svcMgr manager.Manager, services []servicescm.Service) error
 			failedRemovals = append(failedRemovals, err)
 		} else {
 			servicesRemoved = append(servicesRemoved, service.Name)
+			backingProcess, err := processName(service.Command)
+			if err != nil {
+				klog.Warningf("could not determine process associated with the %s service", service.Name)
+				continue
+			}
+			if err = waitForProcessToStop(backingProcess); err != nil {
```
aravindhp, line 111: Should you try killing the process if it doesn't stop?

sebsoto: I was on the fence, but I've added that now.

pkg/daemon/manager/manager.go
```diff
@@ -150,6 +151,37 @@ func (m *manager) EnsureServiceState(service winsvc.Service, state svc.State) er
...
+	status, err := winSvc.Query()
+	if err != nil {
+		return errors.Wrap(err, "error querying service")
+	}
+	var pHandle windows.Handle
+	// A value of 0 indicates that no process is running
+	if status.ProcessId != 0 {
+		pHandle, err = windows.OpenProcess(windows.PROCESS_TERMINATE|windows.PROCESS_QUERY_INFORMATION, false,
+			status.ProcessId)
+		if err != nil {
```
saifshaikh48, line 165: Could it ever be the case that the PID is non-zero but the process doesn't exist (timing issue)? If so, consider checking
``` golang
errors.Is(err, windows.ERROR_PROC_NOT_FOUND)
```

pkg/daemon/manager/manager.go
```diff
@@ -150,6 +151,37 @@ func (m *manager) EnsureServiceState(service winsvc.Service, state svc.State) er
...
+			return errors.Wrap(err, "unable to open service's associated process")
+		}
+		defer windows.CloseHandle(pHandle)
+	}
```
saifshaikh48, line 169: Could the above suggestion help merge this block with the other `if status.ProcessId != 0 {` below?

pkg/daemon/manager/manager.go
```diff
@@ -222,3 +254,19 @@ func enumDependentServicesSyscall(hService windows.Handle, dwServiceState uint32
...
+// waitForProcessToStop waits until the process has exited
+func waitForProcessToStop(process windows.Handle) error {
+	return wait.PollImmediate(retry.WindowsAPIInterval, retry.ResourceChangeTimeout, func() (done bool, err error) {
+		var exitCode uint32
+		if err := windows.GetExitCodeProcess(process, &exitCode); err != nil {
+			// unexpected error, most likely related to permissions
+			return false, errors.Wrap(err, "error getting process exit code")
+		}
+		// STATUS_PENDING indicates the process has not exited, keep retrying.
+		if exitCode == uint32(windows.STATUS_PENDING) {
```
saifshaikh48, line 267: Should/can we also check for `STILL_ACTIVE`? [Microsoft docs](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getexitcodeprocess):

> If the process has not terminated and the function succeeds, the status returned is STILL_ACTIVE (a macro for STATUS_PENDING (minwinbase.h))

pkg/daemon/manager/manager.go
sebsoto: My understanding is that ERROR_PROC_NOT_FOUND refers to a procedure, not a process, I wasn't able to find a specific error to handle here.

pkg/daemon/manager/manager.go
sebsoto: > a macro for STATUS_PENDING

This implies they are the same value, `0x00000103`

pkg/daemon/manager/manager.go
sebsoto: As for if the case is possible, it would only be possible if the process crashed, or someone killed the process in the very small time between querying the service and trying to open the process.

saifshaikh48: > I wasn't able to find a specific error to handle here

Me neither. I am OK with how things are written currently

pkg/daemon/manager/manager.go
```diff
@@ -150,6 +151,37 @@ func (m *manager) EnsureServiceState(service winsvc.Service, state svc.State) er
 	return winsvc.WaitForState(service, state)
 }
 
+// Stop the service, and wait for the process associated with the service to stop
+func (m *manager) stopServiceAndProcess(winSvc *mgr.Service) error {
+	status, err := winSvc.Query()
+	if err != nil {
+		return errors.Wrap(err, "error querying service")
+	}
```
aravindhp, line 159: Can't you just return here `if status.ProcessId == 0`. The code will read much easier then.

aravindhp (changes requested):

pkg/daemon/manager/manager.go
sebsoto: Unfortunately no, the service itself needs to be stopped.
The stopping of the service needs to be sandwiched between opening the process and waiting/stopping the process

aravindhp: So there could be a running service without a process associated with. That does not sound right. What happens if you kill the process? The service stays at running?

sebsoto: > So there could be a running service without a process associated with

I don't know if its possible but that edge case needs to be handled.

> What happens if you kill the process

Killing the process without sending a stop signal for the service will be handled dependent on the service configuration, ideally the service will restart the process.

aravindhp: >ideally the service will restart the process.

I know that does not happen with our current service configuration. The services stay stopped when the process crashes.

aravindhp (approved):

PR #1405, opened 2023-02-09
OCPBUGS-7287: Fix Kublet logs are not written to kubelet.log

mansikulkarni96:
This PR fixes an issue where the kubelet logs were not being written to the kubelet.log file.
The kubelet.exe invocation was not ordered correctly with the arguments being passed before it. 
Moreover, we do not want to set the `-redirect-stderr` flag to false and want to default to true which means we want our error logs to be logged to the same kubelet log file instead of stderr.

Fixes: OCPBUGS-7287

### DIFF
--- a/pkg/services/services.go
+++ b/pkg/services/services.go
@@ -178,7 +178,7 @@ func getKubeletServiceConfiguration(argsFromIginition map[string]string, debug b
 		preScripts = append(preScripts, hostnameOverridePowershellVar)
 	}
 
-	kubeletServiceCmd := fmt.Sprintf("%s -log-file=%s -redirect-stderr=false %s",
+	kubeletServiceCmd := fmt.Sprintf("%s -log-file=%s %s",
 		windows.KubeLogRunnerPath, windows.KubeletLog, windows.KubeletPath)
 
 	for _, arg := range kubeletArgs {


mansikulkarni96:
Latest run [kubelet.log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/1405/pull-ci-openshift-windows-machine-config-operator-master-aws-e2e-operator/1623901741399412736/artifacts/aws-e2e-operator/windows-e2e-operator-test/artifacts/nodes/ip-10-0-243-98.us-west-2.compute.internal/kubelet/kubelet.log) file from artifacts dir.

pkg/services/services.go
```diff
@@ -178,7 +178,7 @@ func getKubeletServiceConfiguration(argsFromIginition map[string]string, debug b
 		preScripts = append(preScripts, hostnameOverridePowershellVar)
 	}
 
-	kubeletServiceCmd := fmt.Sprintf("%s -log-file=%s -redirect-stderr=false %s",
+	kubeletServiceCmd := fmt.Sprintf("%s -log-file=%s %s",
```
sebsoto, line 181: `--` should be used instead of `-` for arguments that are longer than one character

sebsoto (changes requested):

pkg/services/services.go
mansikulkarni96: @sebsoto [kube-log-runner docs](https://github.com/kubernetes/kubernetes/tree/62d9f8ba80f4cc660a88dcc34f56c5c6f7df17ea/staging/src/k8s.io/component-base/logs/kube-log-runner#readme) only mention using -, I remember trying to use it and that didn't work. I can check again.
```
# Merge stderr and stdout, write to stdout (same as 2>&1).
kube-log-runner echo "hello world"

# Redirect both into log file (same as 1>>/tmp/log 2>&1).
kube-log-runner -log-file=/tmp/log echo "hello world"
```

sebsoto:
> Any idea if we we have to explicitly set --also-stdout=false? I tried to search but couldn't find what the flag's default value is

The default is set here
https://github.com/kubernetes/kubernetes/blob/62d9f8ba80f4cc660a88dcc34f56c5c6f7df17ea/staging/src/k8s.io/component-base/logs/kube-log-runner/kube-log-runner.go#L33

pkg/services/services.go
mansikulkarni96: Updated to use --log, I could verify it locally. Not sure if the issue I ran into earlier was side-effect of other flags. Thanks for noticing it @sebsoto

aravindhp (approved):
/lgtm

pkg/services/services.go
aravindhp: This is what the kube-log-runner help shows:
```
kubelet   detached:remotes/origin/HEAD  _output/bin/kube-log-runner --help
Usage of _output/bin/kube-log-runner:
  -also-stdout
    	useful with log-file, log to standard output as well as the log file
  -log-file string
    	If non-empty, save stdout to this file
  -redirect-stderr
    	treat stderr same as stdout (default true)
```
Given that we should stick to using `-log-file`.

aravindhp (approved):
/lgtm

mtnbikenc:
@mansikulkarni96 The PR description mentions `Fixes: OCPBUGS-6635`.  Should that be updated to `OCPBUGS-7287`?

aravindhp (approved):
/lgtm

PR #1411, opened 2023-02-14
WINC-977: Update kube-proxy submodule to sdn-4.13

mansikulkarni96:
- Replace the usage of deprecated klog flags like --log-dir and --logstostderr with the kube-log-runner invocation to direct logs to a single log file.
- [submodule][kube-proxy] Update to 9500d080756 
- [build] Update kube-proxy version to v1.26.0+9500d08

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -19,7 +19,7 @@
 [submodule "kube-proxy"]
 	path = kube-proxy
 	url = https://github.com/openshift/kubernetes
-	branch = sdn-4.12-kubernetes-1.25.1
+	branch = sdn-4.13-kubernetes-1.26.0
 [submodule "containerd"]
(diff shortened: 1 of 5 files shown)


aravindhp (approved):
Thanks for working on this, @mansikulkarni96. 

Shouldn't this be associated with OCPBUGS-4133?

saifshaikh48 (commented):
/lgtm

I'm wondering if it would be beneficial to also check that the log files aren't empty in the logs tests?

aravindhp:
> I'm wondering if it would be beneficial to also check that the log files aren't empty in the logs tests?

Good point. Yes, we should but in a separate PR so that we can easily backport it to all branches.

PR #1412, opened 2023-02-14
WINC-981: [release-4.12] Prep for 7.1.0 release

saifshaikh48:
* Updates version in Makefile to 7.1.0.
* Removes `skipRange` from CSV as it has no effect on 7.1.0 upgrade path.
* `make bundle` ran to update the `bundle/` directory

### DIFF
--- a/bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
@@ -8,7 +8,6 @@ metadata:
     certified: "false"
     createdAt: REPLACE_DATE
     description: An operator that enables Windows container workloads on OCP
-    olm.skipRange: '>=6.0.0 <7.0.1'
     operatorframework.io/cluster-monitoring: "true"
     operatorframework.io/suggested-namespace: openshift-windows-machine-config-operator
     operators.openshift.io/valid-subscription: '["Red Hat OpenShift support for Windows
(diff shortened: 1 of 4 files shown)


aravindhp (approved):

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
```diff
@@ -17,7 +17,7 @@ metadata:
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v3
     repository: https://github.com/openshift/windows-machine-config-operator
     support: Red Hat
-  name: windows-machine-config-operator.v7.0.1
+  name: windows-machine-config-operator.v7.1.0
```
sebsoto, line 20: Shouldn't the skip range be removed for this version?

sebsoto (changes requested):

saifshaikh48:
/hold

for skipVersion convo

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
saifshaikh48: I thought we want to keep 7.0.0 in the skip range so that OLM will update away from it if automatic upgrades are enabled? That's my understanding from [these docs](https://access.redhat.com/documentation/en-us/openshift_container_platform/4.4/html/operators/understanding-the-operator-lifecycle-manager-olm#olm-upgrades-replacing-multiple_olm-understanding-olm) but I could be wrong

```
When searching catalogs for updates, if the head of a channel has a skipRange 
annotation and the currently installed Operator has a version field that falls
in the range, OLM updates to the latest entry in the channel.
```

sebsoto: > we want to keep 7.0.0 in the skip range so that OLM will update away from it

Yes to 7.0.1, but for 7.1.0 the upgrade path should go from 7.0.1 -> 7.1.0.

aravindhp:

e2e tests passed but failed in deprovisioning the cluster.

PR #1421, opened 2023-02-18
WINC-983: [e2e] Ensure required log files are non-empty

mansikulkarni96:
This commit  adds a check to the logs tests to check that log files have non-zero size, ensuring they are actually being written to by the services.

### DIFF
--- a/test/e2e/logs_test.go
+++ b/test/e2e/logs_test.go
@@ -40,6 +40,15 @@ func testNodeLogs(t *testing.T) {
 						log.Printf("unable to retrieve log %s from node %s: %s", file, node.GetName(), err)
 						return false, nil
 					}
+					logInfo, err := os.Stat(filepath.Join(nodeDir, file))
+					if err != nil {
+						log.Printf("unable to get info for retrieved log %s from node %s: %s",
+							file, node.GetName(), err)
+						return false, nil
+					}
+					if logInfo.Size() == 0 {
+						return true, errors.Wrap(err, "log file should not be empty")
+					}
 					return true, nil
 				})
 				assert.NoError(t, err)


test/e2e/logs_test.go
```diff
@@ -66,5 +66,10 @@ func retrieveLog(nodeName, srcPath, destDir string) error {
 		return errors.Wrap(err, "failed to create log directory")
 	}
 	outputFile := filepath.Join(destDir, splitPath[0], filepath.Base(srcPath))
-	return ioutil.WriteFile(outputFile, out, os.ModePerm)
+	ioutil.WriteFile(outputFile, out, os.ModePerm)
+	logInfo, err := os.Stat(outputFile)
+	if err != nil || logInfo.Size() == 0 {
+		return errors.Wrap(err, "log file should not be empty")
+	}
```
saifshaikh48, line 73: I feel like this should be done at the retrieveLogs() call site to avoid giving too much responsibility to this function. Also we're retrying on error at the call site -- when we get a log size of zero, we don't need to retry

test/e2e/logs_test.go
```diff
@@ -40,6 +40,15 @@ func testNodeLogs(t *testing.T) {
 						log.Printf("unable to retrieve log %s from node %s: %s", file, node.GetName(), err)
 						return false, nil
 					}
+					logInfo, err := os.Stat(file)
+					if err != nil {
+						log.Printf("unable to retrieve info for log %s from node %s: %s",
```
aravindhp, line 45: This makes it seem you are running stat on the node which is not the case. `unable to get info for retrieved log %s from node %s: %s` seems more appropriate.

aravindhp (changes requested):

aravindhp (approved):

saifshaikh48 (commented):
/lgtm

aravindhp:
/retest-required. azure-e2e-ccm-install failed with:
```
 ERRO[2023-02-27T16:23:11Z] Some steps failed:                           
ERRO[2023-02-27T16:23:11Z] 
  * could not run steps: step [release-inputs:latest] failed: could not create snapshot imagestream ocp/4.13 for release latest: Timeout: request did not complete within requested timeout - context deadline exceeded 
```

mansikulkarni96:
/retest-required
There was an outage reported in prow, status is fixed now retrying.

PR #1422, opened 2023-02-20
WINC-838: Add OS field to pod spec in Windows Server deployment

saifshaikh48:
This PR adds the OS field in the pod specification for the
Windows Server deployment used in the network test. This ensures that
Windows pods can be identified authoritatively at admission time. As of
now, the SCC Admission Plugin does not perform any validation for Windows
pods, it ignores them. See to https://github.com/openshift/apiserver-library-go/pull/82

Fixes a CI issue introduced by https://github.com/openshift/kubernetes/pull/1290
where SCC Admission Plugin defaults the `runAsNonRoot` with value `true`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,6 +256,11 @@ It is recommended to use images from public registries or pre-pull the images in
 ### Trunk port
 WMCO does not support adding Windows nodes to a cluster through a trunk port. The only supported networking setup for adding Windows nodes is through an access port carrying the VLAN traffic.
 
+## Running Windows workloads
+Be sure to set the [OS field in the Pod spec](https://kubernetes.io/docs/concepts/workloads/pods/#pod-os) to Windows
+when deploying Windows workloads. This field is used to authoritatively identify the pod OS for validation. 
+In OpenShift, it is used when enforcing OS-specific pod security standards.
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -256,6 +256,11 @@ It is recommended to use images from public registries or pre-pull the images in
 ### Trunk port
 WMCO does not support adding Windows nodes to a cluster through a trunk port. The only supported networking setup for adding Windows nodes is through an access port carrying the VLAN traffic.
 
+## Running Windows workloads
+Be sure to set the [OS field in the Pod spec](https://kubernetes.io/docs/concepts/workloads/pods/#pod-os) to Windows
+when deploying Windows workloads. This field is used by kubelet to validate a node's operating system compatibility
+before a pod is run, as well as when enforcing OS-specific pod security standards.
```
aravindhp, line 262: I suggest being more generic and saying `This fields is used to authoritatively identify the pod OS and for validation`. In addition you can saying `In OpenShift, it is used when when enforcing OS-specific pod security standards.`

aravindhp (changes requested):

aravindhp (approved):

PR #1423, opened 2023-02-21
OCPBUGS-5378: [build] Fix containerd version reporting

mansikulkarni96:
This is the root cause of containerd Makefile not setting the version and we see containerd 1.19  as the misreported version in 4.12 and 1.18 in 4.11 cluster.

Fixes:  OCPBUGS-5378

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ WMCO_VERSION ?= 8.0.0
 # command in generate_k8s_version_commit() in hack/update_submodules.sh
 KUBELET_GIT_VERSION=v1.26.0+919a59b
 KUBE-PROXY_GIT_VERSION=v1.26.0+9500d08
+CONTAINERD_GIT_VERSION=v1.6.16-3+feedbee
 
 # CHANNELS define the bundle channels used in the bundle.
 # Add a new line here if you would like to change its default config. (E.g CHANNELS = "preview,fast,stable")
@@ -224,3 +225,7 @@ kubelet:
 .PHONY: kube-proxy
 kube-proxy:
 	KUBE_GIT_VERSION=$(KUBE-PROXY_GIT_VERSION) KUBE_BUILD_PLATFORMS=windows/amd64 make -C kube-proxy WHAT=cmd/kube-proxy
+
+.PHONY : containerd
+containerd:
+	GOOS=windows VERSION=$(CONTAINERD_GIT_VERSION) make -C containerd bin/containerd.exe
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ WMCO_VERSION ?= 8.0.0
 # command in generate_k8s_version_commit() in hack/update_submodules.sh
 KUBELET_GIT_VERSION=v1.26.0+919a59b
 KUBE-PROXY_GIT_VERSION=v1.26.0+9500d08
-CONTAINERD_GIT_VERSION=v1.6.16-3+feedbee
+CONTAINERD_GIT_VERSION=v1.6.16-3-g2d3127cbb
 
 # CHANNELS define the bundle channels used in the bundle.
 # Add a new line here if you would like to change its default config. (E.g CHANNELS = "preview,fast,stable")
--- a/build/Dockerfile.ci
+++ b/build/Dockerfile.ci
@@ -33,9 +33,10 @@ COPY windows_exporter/ .
 RUN GOOS=windows ./promu build -v
 
 # Build containerd
-WORKDIR /build/windows-machine-config-operator/containerd/
-COPY containerd/ .
-RUN GOOS=windows make
+WORKDIR /build/windows-machine-config-operator/
+COPY containerd/ containerd/
+COPY Makefile Makefile
+RUN make containerd
 
 # Build containerd shim
 WORKDIR /build/windows-machine-config-operator/hcsshim/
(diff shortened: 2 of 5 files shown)


mansikulkarni96:
Verified on node
```
administrator@EC2AMAZ-78MFG33 C:\Users\Administrator>powershell
Windows PowerShell
Copyright (C) Microsoft Corporation. All rights reserved.

PS C:\Users\Administrator> C:\k\containerd\containerd.exe --version
containerd github.com/containerd/containerd v1.6.16-3-g2d3127cbb 2d3127cbb487b46e65bb4329c6f7836d12cec9c8
```

Makefile
```diff
@@ -224,3 +224,7 @@ kubelet:
 .PHONY: kube-proxy
 kube-proxy:
 	KUBE_GIT_VERSION=$(KUBE-PROXY_GIT_VERSION) KUBE_BUILD_PLATFORMS=windows/amd64 make -C kube-proxy WHAT=cmd/kube-proxy
+
+.PHONY : containerd
+containerd:
+	GOOS=windows VERSION=$(shell cd containerd && git describe --match 'v[0-9]*' --dirty='.m' --always) make -C containerd WHAT=bin/containerd.exe
```
saifshaikh48, line 230: For my own education, what does the `WHAT` directive do?
Is this the equivalent of what we do in the [midstream dockerfile](https://gitlab.cee.redhat.com/openshift-winc-midstream/openshift-winc-midstream/-/blob/rhaos-4.13-rhel-8/distgit/containers/windows-machine-config-operator/Dockerfile#L36) i.e. just building container.exe (not building `ctr.exe` and `containerd-stress.exe` anymore)?

build/Dockerfile.ci
```diff
@@ -42,6 +42,12 @@ WORKDIR /build/windows-machine-config-operator/hcsshim/
 COPY hcsshim/ .
 RUN GOOS=windows go build ./cmd/containerd-shim-runhcs-v1
 
+# Build containerd
```
saifshaikh48, line 45: There is a `# Build containerd` stanza a few lines lines above that should be modified instead of adding this one

saifshaikh48 (commented):
Nice work with this!

build/Dockerfile.ci
mansikulkarni96: Thanks for the catch, looks like this came from my local testing

Makefile
saifshaikh48, line 230: Do we need to set the [REVISION](https://github.com/containerd/containerd/blob/main/Makefile#L34) as well?

mansikulkarni96: Yes that's the intention WHAT directive builds from the directory it is pointed to, I realize it's not needed in this case, updating.

mansikulkarni96: It gets set from the containerd Makefile, since it is not getting pre-set in our Dockerfile before the containerd make target runs we are good.
PTAL at the [output comment](https://github.com/openshift/windows-machine-config-operator/pull/1423#issuecomment-1437820355), revision comes after the hyphen.

Makefile
```diff
@@ -224,3 +224,7 @@ kubelet:
 .PHONY: kube-proxy
 kube-proxy:
 	KUBE_GIT_VERSION=$(KUBE-PROXY_GIT_VERSION) KUBE_BUILD_PLATFORMS=windows/amd64 make -C kube-proxy WHAT=cmd/kube-proxy
+
+.PHONY : containerd ## Ref: Containerd Version https://github.com/containerd/containerd/blob/8cb00f45c973b017f403d05d4d208ab2971ff335/Makefile#L33
+containerd:
+	GOOS=windows VERSION=$(shell cd containerd && git describe --match 'v[0-9]*' --dirty='.m' --always) make -C containerd bin/containerd.exe
```
aravindhp, line 230: Does this still require tags to be manually applied to the downstream branch?

aravindhp (commented):

Makefile
mansikulkarni96: yes I think we still need to have tags downstream which are in sync with upstream tags, since this command does a describe on the pattern matching the current tag in the submodule directory.

aravindhp: Could we then do something similar like we did for the k8s submodules? i.e. pull the tags when updating the submodules and hard code the version here.

hack/update_submodules.sh
```diff
@@ -25,25 +25,34 @@ function help() {
...
+    # git describe --abbrev=7 returns vx.y.z-number_of_commits-hash. Remove the -number_of_commits-g and replace it
+    # with + to match the Linux kubelet version pattern
+    version=$(git describe --abbrev=7  | sed --expression="s|-.*-g|+|g")
+  fi
   cd ..
-  # Construct the Makefile variable KUBE(LET/-PROXY)_GIT_VERSION
+  # Construct the Makefile variable $submodule_GIT_VERSION
   local makefile_var="$(echo "$submodule" | tr '[:lower:]' '[:upper:]')_GIT_VERSION"
-  # Makefile has "KUBE(LET/-PROXY)_GIT_VERSION=vx.y.z+hash. Replace the KUBE(LET/-PROXY)_GIT_VERSION= with the new k8s
+  # Makefile has "$submodule_GIT_VERSION=vx.y.z+hash. Replace the $submodule_GIT_VERSION= with the new k8s
```
mansikulkarni96, line 50: will update this comment to remove k8s version specific context

saifshaikh48 (commented):
/lgtm

aravindhp (approved):

alinaryan:
Thanks for working on this, 
/lgtm

PR #1407, opened 2023-02-10
[release-4.12] OCPBUGS-7336: [build] Fix k8s version reporting

aravindhp:
The openshift/kubernetes repo does not propagate all tags from upstream. Only the X.Y.0 tag is propagated. This causes the Windows kubelet and kube-proxy Z version to lag behind its Linux counterparts. The Linux [build](https://github.com/openshift/doozer/blob/a34e303682ca180e4f071f85cc4c34e6d333787e/doozerlib/metadata.py#L788) of kubelet and kube-proxy figures out the Z version at build time. To do something similar for Windows, we can leverage the hack/update_submodules.sh script which is always used to update the kubelet and kube-proxy submodules. To achieve this, the following changes are made:
- Add kubelet and kube-proxy version variables to the Makefile along with make endpoints
- Call these make endpoints in the Dockerfile
- Update hack/update_submodules.sh to pull upstream tags and generate a commit that updates the Makefile version variables

Now the built binaries will have the correct version encoded in them.

(manually cherry picked from https://github.com/openshift/windows-machine-config-operator/pull/1401)

PR #1424, opened 2023-02-22
[release-4.12] OCPBUGS-7843:  [build] Fix containerd version reporting

mansikulkarni96:
This is the root cause of containerd Makefile not setting the version and we see containerd 1.19 as the misreported version in 4.12 and 1.18 in 4.11 cluster.

(manually cherry-pick from #1423)


mtnbikenc:
Let's not manually backport this one yet.  We are going to start working through getting WICD work backported first.  Then this one will probably cherry-pick cleanly.
/hold

mtnbikenc:
This manual backport is required because #1407 was manually backported.  This will also not conflict with future 4.12 backport work for WICD.
/hold cancel

aravindhp:
@mansikulkarni96 hopefully https://github.com/openshift/windows-machine-config-operator/pull/1434 will fix the above merge conflict

PR #1427, opened 2023-02-22
WINC-736: Periodically correct service state

sebsoto:
Ensures that services return to the correct state if tampered with

### DIFF
--- a/pkg/daemon/controller/controller.go
+++ b/pkg/daemon/controller/controller.go
@@ -24,6 +24,7 @@ import (
 	"net"
 	"reflect"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 	"golang.org/x/sys/windows/svc"
@@ -153,7 +154,7 @@ func RunController(ctx context.Context, apiServerURL, saCA, saToken, watchNamesp
 	if err != nil {
 		return err
 	}
-	if err = sc.SetupWithManager(ctrlMgr); err != nil {
+	if err = sc.SetupWithManager(ctx, ctrlMgr); err != nil {
 		return err
 	}
 	klog.Info("Starting manager, awaiting events")
@@ -174,7 +175,7 @@ func NewServiceController(ctx context.Context, nodeName, watchNamespace string, 
 }
 
 // SetupWithManager sets up the controller with the Manager.
-func (sc *ServiceController) SetupWithManager(mgr ctrl.Manager) error {
+func (sc *ServiceController) SetupWithManager(ctx context.Context, mgr ctrl.Manager) error {
 	nodePredicate := predicate.Funcs{
 		// A node's name will never change, so it is fine to use the name for node identification
 		// The node must have a desired-version annotation for it to be reconcilable
@@ -196,10 +197,17 @@ func (sc *ServiceController) SetupWithManager(mgr ctrl.Manager) error {
 	cmPredicate := predicate.NewPredicateFuncs(func(object client.Object) bool {
 		return strings.HasPrefix(object.GetName(), servicescm.NamePrefix)
 	})
+
+	// Keeping this on the longer side, as each reconciliation requires running each service's powershell scripts
+	// This value is based on CVO's resync period
+	reconcilePeriod := 2 * time.Minute
+	eventChan := newPeriodicEventGenerator(ctx, reconcilePeriod)
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&core.Node{}, builder.WithPredicates(nodePredicate)).
 		Watches(&source.Kind{Type: &core.ConfigMap{}}, handler.EnqueueRequestsFromMapFunc(sc.mapToCurrentNode),
 			builder.WithPredicates(cmPredicate)).
+		Watches(&source.Channel{Source: eventChan}, handler.EnqueueRequestsFromMapFunc(sc.mapToCurrentNode)).
 		Complete(sc)
 }
 
@@ -433,6 +441,26 @@ func (sc *ServiceController) resolvePowershellVariables(svc servicescm.Service) 
 		}
 	}
 	return vars, nil
+}
+
+// newPeriodicEventGenerator returns a channel which will have an empty event sent on it at an interval specified by the
+// given period
+func newPeriodicEventGenerator(ctx context.Context, period time.Duration) <-chan event.GenericEvent {
+	eventChan := make(chan event.GenericEvent)
+	go func() {
+		ticker := time.NewTicker(period)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-ctx.Done():
+				close(eventChan)
+				return
+			case <-ticker.C:
+				eventChan <- event.GenericEvent{}
+			}
+		}
+	}()
+	return eventChan
 }
 
 // NewDirectClient creates and returns an authenticated client that reads directly from the API server.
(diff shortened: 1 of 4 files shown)


pkg/daemon/controller/controller.go
```diff
@@ -316,6 +318,11 @@ func (sc *ServiceController) reconcileService(service winsvc.Service, expected s
 		updateRequired = true
 	}
 
+	// Recovery options are not part of the normal configuration and must be updated separately.
+	if err = sc.ensureRecoveryActions(service); err != nil {
+		return errors.Wrap(err, "error configuring service recovery")
+
```
saifshaikh48, line 331: nit: extra line

pkg/daemon/controller/controller.go
```diff
@@ -331,6 +338,22 @@ func (sc *ServiceController) reconcileService(service winsvc.Service, expected s
...
 
+// ensureRecoveryActions sets the recovery actions for the service if they are configured incorrectly
+func (sc *ServiceController) ensureRecoveryActions(service winsvc.Service) error {
+	actions, err := service.RecoveryActions()
+	if err != nil {
+		return errors.Wrap(err, "error getting current recovery actions")
+	}
+
+	// Ensure the recovery action for the service is to restart 10 seconds after a crash.
+	expectedAction := mgr.RecoveryAction{Type: mgr.ServiceRestart, Delay: 10 * time.Second}
```
saifshaikh48, line 356: Consider having a constant for recovery delay

pkg/daemon/controller/controller.go
```diff
@@ -458,6 +465,25 @@ func (sc *ServiceController) resolvePowershellVariables(svc servicescm.Service)
 	return vars, nil
 }
 
+// startPeriodicServiceReconciliation initiates periodic polling to ensure that the instance is properly reconciled.
+func (sc *ServiceController) startPeriodicServiceReconciliation() {
+	go func() {
+		// Keeping this on the longer side, as each reconciliation requires running each service's powershell scripts
+		wait.PollInfinite(60*time.Second, func() (done bool, err error) {
```
saifshaikh48, line 472: Consider a constant for this too

saifshaikh48 (commented):
LGTM overall, satisfies story acceptance criteria

pkg/daemon/controller/controller.go
aravindhp: What happens if a service is constantly crashing and it keeps getting restarted? Is there a an upper limit for retries that we can set? Ideally we should retry N number of times and then WICD should generate an event.

pkg/daemon/winsvc/winsvc.go
```diff
@@ -18,6 +18,8 @@ type Service interface {
 	Control(svc.Cmd) (svc.Status, error)
 	Query() (svc.Status, error)
 	UpdateConfig(mgr.Config) error
+	RecoveryActions() ([]mgr.RecoveryAction, error)
+	SetRecoveryActions([]mgr.RecoveryAction, uint32) error
```
aravindhp, line 22: I am not following  this change. Where is the implementation?

pkg/daemon/controller/controller.go
```diff
@@ -104,6 +106,7 @@ type ServiceController struct {
 	nodeName       string
 	watchNamespace string
 	psCmdRunner    powershell.CommandRunner
+	mu             sync.Mutex
```
aravindhp, line 109: Why do we need this lock? Ideally whenever introducing a lock, the reason for doing so should be mentioned in the commit and in the code.

pkg/daemon/controller/controller.go
```diff
@@ -458,6 +465,25 @@ func (sc *ServiceController) resolvePowershellVariables(svc servicescm.Service)
 	return vars, nil
 }
 
+// startPeriodicServiceReconciliation initiates periodic polling to ensure that the instance is properly reconciled.
+func (sc *ServiceController) startPeriodicServiceReconciliation() {
+	go func() {
+		// Keeping this on the longer side, as each reconciliation requires running each service's powershell scripts
+		wait.PollInfinite(60*time.Second, func() (done bool, err error) {
+			select {
+			case <-sc.ctx.Done():
```
aravindhp, line 474: I am noticing this just now. It is bad practice to add context to a struct. This is clearly called out in the [doc](https://pkg.go.dev/context):
>Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx

Why are we doing this?

aravindhp (commented):
Thanks for working on this, @sebsoto. I have left a few comments.

pkg/daemon/winsvc/winsvc.go
sebsoto: https://github.com/golang/sys/blob/master/windows/svc/mgr/recovery.go#L63

pkg/daemon/controller/controller.go
aravindhp: I am fine doing it as a follow up PR.

pkg/daemon/controller/controller.go
sebsoto: We can specify a sequence of actions to occur within a given time period, for example
Crash 1: Restart service after 1 min
Crash 2: Restart service after 2 min 
Crash 3: Restart service after 5 min
Crash 4+: Do nothing

And then 30 min after the first failure, reset the crash counter.

Generating a kubernetes event would have to be separate from this, as these counts are all managed within the Windows Service Manager

pkg/windows/windows.go
```diff
@@ -467,7 +467,24 @@ func (vm *windows) ConfigureWICD(apiServerURL, watchNamespace string, credential
...
+		{
+			actionType: serviceRestart,
+			delay:      60,
+		},
+		{
+			actionType: serviceRestart,
+			delay:      120,
+		},
+	}
+	wicdService, err := newService(wicdPath, WicdServiceName, wicdServiceArgs, nil, recoveryActions, 600)
```
saifshaikh48, line 487: the 600 magic number is a good candidate to be stored as a constant in the `retry` package

pkg/windows/windows.go
```diff
@@ -467,7 +467,24 @@ func (vm *windows) ConfigureWICD(apiServerURL, watchNamespace string, credential
 	}
 	wicdServiceArgs := fmt.Sprintf("controller --windows-service --log-dir %s --api-server %s --sa-ca %s --sa-token %s --namespace %s",
 		wicdLogDir, apiServerURL, wicdCAFile, wicdTokenFile, watchNamespace)
-	wicdService, err := newService(wicdPath, WicdServiceName, wicdServiceArgs, nil)
+
+	// if WICD crashes, attempt to restart WICD after 30 seconds, 60 seconds, and then every 2 minutes after that.
```
saifshaikh48, line 471: nit: does starting at 30s for WICD seem like a slow reaction? For comparison, k8s exp backoff for containers starts at 10s

mansikulkarni96:
LGTM

alinaryan:
LGTM

pkg/windows/windows.go
sebsoto: I've updated the doc for the recoveryPeriod, I'm not sure if that this will fit in the retry package, given this is a rolling window, with not much potential to be used elsewhere.

pkg/windows/windows.go
```diff
@@ -467,7 +467,30 @@ func (vm *windows) ConfigureWICD(apiServerURL, watchNamespace string, credential
...
+	// reset this counter 10 min after the first crash
```
saifshaikh48, line 472: comment needs updating

saifshaikh48 (commented):
LGTM, nice work with the generated events

aravindhp (approved):
Nice work, @sebsoto. Looking forward to seeing a live Windows service tampering demo 😄

PR #1439, opened 2023-02-28
OCPBUGS-8055: [release-4.11]  [build] Fix containerd version reporting

mansikulkarni96:
This is the root cause of containerd Makefile not setting the version and we see containerd 1.19 as the misreported version in 4.12 and 1.18 in 4.11 cluster.

(manually cherry-pick from https://github.com/openshift/windows-machine-config-operator/pull/1424)


aravindhp (approved):
/lgtm

aravindhp:
/retest-required

[vsphere-e2e-operator](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/1439/pull-ci-openshift-windows-machine-config-operator-release-4.11-vsphere-e2e-operator/1633139902403055616) failed to get log files which is a known flake and is being investigated.

PR #1462, opened 2023-03-07
[hack] Add script to check PR merge status in errata build

mtnbikenc:
The script will query Errata Tool for the active errata based on the branch specified.  It will then look up the build attached and determine the WMCO commit the build was built from.  Finally, it will list the merge commits since the last WMCO release for that version.  The commit associated with the build will be tagged with a shortened build NVR, 'wmco-container-X.Y.Z-n'.


hack/errata-check.sh
```diff
@@ -0,0 +1,262 @@
...
+  local text=$1
+  printf "\n${BRIGHT}%s${NORMAL}\n" "$text"
+}
+
+function print-error {
+  local text=$1
+    printf "\n${RED}ERROR: %s${NORMAL}\n" "$text" 1>&2
+}
+
+function set-branch-vars {
```
sebsoto, line 88: Instead of using a lookup table approach, I suggest determining the hash and version dynamically.

You can take in a branch as an argument, for example, `set-branch-vars release-4.13`.
```
local release_branch=$1
branch_prefix=$(semver=$(cut -d '-' -f 1 <<< "$input"))
semver=$(cut -d '-' -f 2 <<< "$input")
major_version=$(cut -d '.' -f 1 <<< "$semver")
minor_version=$(cut -d '.' -f 2 <<< "$semver")
previous_release_branch="4.$((minor_version-1))"
echo $previous_release_branch
```

You can then use the method you stated above to get the branching hash
```
# diff -u <(git rev-list --first-parent $NEXT_RELEASE) \
#         <(git rev-list --first-parent $LAST_RELEASE) \
#         | sed -ne 's/^ //p' | head -1 
```

The LAST_WMCO_VERSION can be found by looking at the release branch in question and searching `Makefile`:
```
$ git show release-4.12:Makefile |grep "WMCO_VERSION ?=
WMCO_VERSION ?= 7.0.0
```
The version can be captured directly using some bash regex magic, but I'm not sure what the best tool for that is .

sebsoto (changes requested):

hack/errata-check.sh
mtnbikenc: I had considered dynamically determining the last release, however my initial thought was to do that through another errata query for 'SHIPPED LIVE' errata.  That was going to be more expensive requiring more calls.  I like the suggestion of using git locally, however I'll need to think that through further.  When I do the `git show` mentioned above, I'm getting the "next" version for that branch.

```
$ git show release-4.12:Makefile |grep "WMCO_VERSION ?="
WMCO_VERSION ?= 7.1.0
```
Aravindh and I also discussed pushing tags to the repo for released versions, 7.0.0, 7.0.1, etc.  That would simplify the task of finding the commit used for the release without having to go through the errata.

`LAST_WMCO_VERSION` is only used in printing a description and `LAST_WMCO_HASH` is only used to limit the number of commits returned.  The vars are not really necessary to do most of the work in the script.

mtnbikenc: If we do push tags I could do something like this:
```
$ git tag --merged release-4.12 | grep -E "^v" | sort --version-sort | tail -1
v7.0.1

$ git show-ref --tags "refs/tags/v7.0.1" | awk '{ print $1 }'                 
bc9473bf8c4200b0a016608e808122ef783570bc
```

If nothing is returned for the branch in the first step (no release for the branch), then it could drop back to running the diff to find the branching point.

aravindhp (changes requested):
My only nit is to use 50 columns for the commit header.

mtnbikenc:
> My only nit is to use 50 columns for the commit header.

Could you explain the reasoning?  I went with 40 because that is the length of a SHA-1 hash produced by `git`.

aravindhp:
> Could you explain the reasoning? I went with 40 because that is the length of a SHA-1 hash produced by `git`.

I meant for the git commit header `[hack] Add script to check PR merge status in errata build`. Ref: [CONTRIBUTION.md#format-of-the-commit-message](https://github.com/openshift/windows-machine-config-operator/blob/master/CONTRIBUTION.md#format-of-the-commit-message). Maybe `[hack] Add PR to errata build correlation script`.

mtnbikenc:
Ah, _that_ commit header.  Done.

hack/errata-check.sh
mansikulkarni96: I am in favor of pushing tags, we have been thinking about it for a while now.

hack/errata-check.sh
```diff
@@ -0,0 +1,262 @@
...
+#
+# Install Required Tools
+# ----------------------
+# dnf install curl jq gh
+#
+# Usage
+# -----
+# * VPN connectivity and valid kerberos ticket required
+# * Update your local clone and branches
+# * Run this script ./errata-check.sh <branch_name>
```
alinaryan, line 26: nit: assume that this is ran from repo root
`./hack/errata-check.sh <branch_name>`

hack/errata-check.sh
alinaryan: +1

alinaryan (commented):
Thanks for working on this!

hack/errata-check.sh
```diff
@@ -0,0 +1,262 @@
...
+
+TARGET_BRANCH=$1
+set-branch-vars "$TARGET_BRANCH"
+check-upstream-status "$TARGET_BRANCH"
+
+# Query Errata Tool for the active errata for the desired release version
+ERRATA_SEARCH=$(curl-json "$ERRATA_API_URL/search?product%5B%5D=79&show_state_IN_PUSH=1&show_state_NEW_FILES=1&show_state_PUSH_READY=1&show_state_QE=1&show_state_REL_PREP=1&synopsis_text=Windows+Containers+$WMCO_VERSION")
+
+# Parse errata search result for errata ID
+if ! ERRATA_ID=$(echo -E "$ERRATA_SEARCH" | jq '.data | .[] | .id') ; then
```
aravindhp, line 211: I see a parse error with an invalid kerberos ticket:
```
windows-machine-config-operator   release-4.10 → upstream {4}  1  errata-check.sh release-4.10
WMCO Version        = v5
Branch              = release-4.10
parse error: Invalid numeric literal at line 1, column 10

ERROR: Unable to parse errata search results.  Valid kerberos ticket?
```
Not a big deal but I think you can error check `ERRATA_ID` before passing it to `jq`. 

PS: this not a blocker for merging this.

mtnbikenc: Added a [check](https://github.com/openshift/windows-machine-config-operator/pull/1462/files#diff-860bfe2660f784c0e166c943676def7106a2c50fcfe353ee6c6c6f48a2c4c95aR150-R153) for `unauthorized` in the response before continuing.

hack/errata-check.sh
mtnbikenc: Updated.

hack/errata-check.sh
```diff
@@ -0,0 +1,262 @@
...
+}
+
+function curl-json {
+  local url=$1
+  local result
+  if ! result=$(curl --silent --negotiate --user : "$url"); then
+    print-error "Unable to curl server.  Connected to VPN?"
+    exit 1
+  fi
+  if echo -E "$result" | grep "Errata Tool - Unauthorized"; then
```
aravindhp, line 150: Nit: I suggest just checking for `Unauthorized` given `curl-json` is generic and could be used to curl URLs other than the errata website.

mtnbikenc:
I've pushed some updates as separate commits to make it easier to review what is new.  I'll squash before we merge this.
Additions include branch error checking, removing the need to have hard coded hashs/versions, retrieving the `FixVersion` from Jira for each PR/issue.

hack/errata-check.sh
mtnbikenc: Updated.  Hashes are no longer stored in the script now that release tags have been pushed.

hack/errata-check.sh
mtnbikenc: Updated.  I now check the result code of curl as well as HTTP headers for errors before pulling content.

hack/errata-check.sh
```diff
@@ -0,0 +1,306 @@
...
+    local commit_hash=""
+    local pr_num=""
+    local pr_title=""
+    local jira_id=""
+    local jira_fix_version=""
+    local attached=""
+    commit_hash=$(echo "$commit" | awk '{ print $1 }')
+    pr_num=$(echo "$commit" | awk '{ print substr($5,2) }')
+    pr_title=$(gh pr view "$pr_num" --json title | jq -r .title)
+    jira_id=$(echo "$pr_title" | grep -Eo '(WINC|OCPBUGS)-[0-9]*') || true
```
jrvaldes, line 208: Considering moving the PR title prefixes (`WINC|OCPBUGS`) as a constant at the beginning of the script so that is easier to maintain and reuse.

jrvaldes (commented):
@mtnbikenc  Thanks for working on this hackenstein script. 

/lgtm

aravindhp:
@mtnbikenc please squash merge your commits and lets get this merged. For initial implementations of a new component we don't require commits describing how you arrived at the destination.

aravindhp (approved):
Thanks for doing this, @mtnbikenc 

@wgahnagl Should we add a note about using this script in the contribution guidelines as to when we can move a story to done and a bug to ON_QA?

saifshaikh48 (commented):
/lgtm

Updated my local tags (`git fetch --tags`) and this worked very well. Awesome stuff, @mtnbikenc :rocket:

saifshaikh48:
> Should we add a note about using this script in the contribution guidelines as to when we can move a story to done and a bug to ON_QA?

@aravindhp I think this is something better suited for the Definition of Ready and Done doc

aravindhp:
> @aravindhp I think this is something better suited for the Definition of Ready and Done doc

Good point. Will you take an action item to update it?

mtnbikenc:
/retest-required
Cluster spin up failures and unrelated test flake.
```
=== RUN   TestWMCO/destroy/Deletion
W0320 16:23:46.054921   15114 warnings.go:70] would violate PodSecurity "restricted:latest": runAsNonRoot != true (pod or container "sleep" must set securityContext.runAsNonRoot=true)
    delete_test.go:126: 
        	Error Trace:	/go/src/github.com/openshift/windows-machine-config-operator/test/e2e/delete_test.go:126
        	            				/go/src/github.com/openshift/windows-machine-config-operator/test/e2e/delete_test.go:31
        	            	timed out waiting for daemonset noop-ds to scale, current status: {CurrentNumberScheduled:2 NumberMisscheduled:0 DesiredNumberScheduled:2 NumberReady:1 ObservedGeneration:1 UpdatedNumberScheduled:2 NumberAvailable:1 NumberUnavailable:1 CollisionCount:<nil> Conditions:[]}
```

saifshaikh48:
> Good point. Will you take an action item to update it?

Sure I updated it, but there is still a lot of outdated info on the doc (BZ specific info, QE verification, etc). It might need a proper re-write so I added it as an item to discuss in the retro notes

aravindhp:
/override ci/prow/gcp-e2e-operator

The script does not touch e2e scripts

aravindhp:

The script does not touch e2e scripts

PR #1464, opened 2023-03-07
[test] Capture stderr on oc adm node-logs failures

sebsoto:
Fixes only stdout being captured.

### DIFF
--- a/test/e2e/logs_test.go
+++ b/test/e2e/logs_test.go
@@ -63,7 +63,12 @@ func retrieveLog(nodeName, srcPath, destDir string) error {
 	cmd := exec.Command("oc", "adm", "node-logs", "--path="+srcPath, nodeName)
 	out, err := cmd.Output()
 	if err != nil {
-		return errors.Wrapf(err, "oc adm node-logs failed with output: %s", string(out))
+		var exitError *exec.ExitError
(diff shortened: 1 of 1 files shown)


aravindhp (approved):

test/e2e/logs_test.go
```diff
@@ -63,7 +63,12 @@ func retrieveLog(nodeName, srcPath, destDir string) error {
 	cmd := exec.Command("oc", "adm", "node-logs", "--path="+srcPath, nodeName)
 	out, err := cmd.Output()
 	if err != nil {
-		return errors.Wrapf(err, "oc adm node-logs failed with output: %s", string(out))
+		var exitError *exec.ExitError
+		stderr := ""
+		if errors.As(err, exitError) {
+			stderr = string(exitError.Stderr)
+		}
+		return fmt.Errorf("oc adm node-logs failed with exit code %s and output: %s: %s", err, string(out), stderr)
```
mansikulkarni96, line 71: can we also mention the path this failed on here, instead of saying oc adm node-logs failed just use the cmd var? or is that part of the out anyways?

sebsoto: @mansikulkarni96 that is included as part of error returned by the calling function

PR #1479, opened 2023-03-14
OCPBUGS-10222: [metrics] Ignore case of cluster monitoring label

### LINKED ISSUE #1478: [Bug]: Case sensitivity issue when label "openshift.io/cluster-monitoring" set to 'True' on openshift-windows-machine-config-operator namespace
### Attention

- [X] I am using the community version of WMCO

### Version

OCP

### Cluster Version

4.12

### Platform

AWS

### Proxy

Yes, without HTTPS

### WMCO Version

7.1

### Windows version

2019

### What happened?

WMCO seems to be ignoring cluster monitoring label when the value is set to 'True'. Can see the prometheus windows exporter service running on windows nodes added to cluster but no metrics are available on cluster monitoring prometheus server. Couldn't edit the label from web console, did not try to update via command line to see what the experience is. Ended up having to delete the machineset, uninstall WMCO, delete namespace. Basically re-install configure. Tested this in OCP 4.12 version on AWS, See the ansible playbook here in this repo WMCO. https://github.com/rprakashg-redhat/windemos

This seems to be the offending line of code -> https://github.com/openshift/windows-machine-config-operator/blame/d0035fd0b3224bc90b87d3b436c66bd484047db4/pkg/metrics/metrics.go#L256 Thanks to @mansikulkarni96 for tracking this down

### What did you expect to happen?

Metrics from windows nodes available on Cluster monitoring prometheus server. Label value should be case insensitive. Should allow both 'True' and 'true'

### Steps to reproduce the issue

Yes. Please see this repo https://github.com/rprakashg-redhat/windemos

### Do you have a workaround for this issue?

No

saifshaikh48:
This PR makes the cluster monitoring label for the WMCO namespace case insensitive to provide better user experience. Previously, not adhering to an all lowercase `true` value would result in WMCO ignoring the label and not configuring/exposing Prometheus metrics for Windows nodes.

Fixes OCPBUGS-10222

Fixes #1478

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -3,6 +3,7 @@ package metrics
 import (
 	"context"
 	"encoding/json"
+	"strconv"
 
 	"github.com/pkg/errors"
 	monclient "github.com/prometheus-operator/prometheus-operator/pkg/client/versioned/typed/monitoring/v1"
@@ -245,21 +246,29 @@ func (c *Config) Configure(ctx context.Context) error {
 	return nil
 }
 
-// validate will verify if cluster monitoring is enabled in the operator namespace.
-// If the label is not present, it will log and send warning events to the user.
+// validate will verify if cluster monitoring is enabled in the operator namespace. If the label is set to false or not
+// present, it will log and send warning events to the user. If the label holds a non-boolean value, returns an error.
 func (c *Config) validate(ctx context.Context) (bool, error) {
 	// validate if metrics label is added to namespace
 	wmcoNamespace, err := c.CoreV1().Namespaces().Get(ctx, c.namespace, metav1.GetOptions{})
 	if err != nil {
 		return false, errors.Wrap(err, "error getting operator namespace")
 	}
-	if wmcoNamespace.Labels["openshift.io/cluster-monitoring"] != "true" {
-		metricsEnabled = false
+
+	labelValue := false
+	// if the label exists, update value from default of false
+	if value, ok := wmcoNamespace.Labels["openshift.io/cluster-monitoring"]; ok {
+		labelValue, err = strconv.ParseBool(value)
+		if err != nil {
+			return false, errors.Wrap(err, "monitoring label must have a boolean value")
+		}
+	}
+	if !labelValue {
 		c.recorder.Eventf(wmcoNamespace, v1.EventTypeWarning, "labelValidationFailed",
 			"Cluster monitoring openshift.io/cluster-monitoring=true label is not enabled in %s namespace", c.namespace)
-		return false, nil
 	}
-	return true, nil
+	metricsEnabled = labelValue
+	return metricsEnabled, nil
 }
 
 // createEndpoint creates an endpoint object in the operator namespace.


pkg/metrics/metrics.go
```diff
@@ -253,7 +254,7 @@ func (c *Config) validate(ctx context.Context) (bool, error) {
 	if err != nil {
 		return false, errors.Wrap(err, "error getting operator namespace")
 	}
-	if wmcoNamespace.Labels["openshift.io/cluster-monitoring"] != "true" {
+	if strings.ToLower(wmcoNamespace.Labels["openshift.io/cluster-monitoring"]) != "true" {
```
aravindhp, line 257: So this will accept strings like `tRue` etc. Do we want to do that or throw an error?

saifshaikh48: Should be fine IMO. I don't see a reason accept `True` but not `tRue`

alinaryan:
Thanks for taking this on!
/lgtm

pkg/metrics/metrics.go
aravindhp: Thinking more about this, I suggest we rather do [strconv.ParseBool](https://cs.opensource.google/go/go/+/go1.20.2:src/strconv/atob.go;l=10).

aravindhp (changes requested):

pkg/metrics/metrics.go
```diff
@@ -245,21 +246,30 @@ func (c *Config) Configure(ctx context.Context) error {
...
 	if err != nil {
 		return false, errors.Wrap(err, "error getting operator namespace")
 	}
-	if wmcoNamespace.Labels["openshift.io/cluster-monitoring"] != "true" {
-		metricsEnabled = false
+
+	boolVal := false
+	value, ok := wmcoNamespace.Labels["openshift.io/cluster-monitoring"]
+	// if the label exists, update value from default of false
+	if ok {
```
aravindhp, line 261: This can be combined:
```suggestion
if value, ok := wmcoNamespace.Labels["openshift.io/cluster-monitoring"]; ok {
```

aravindhp (changes requested):
>It also makes WMCO able to react to a change in the namespace's cluster monitoring label. 

How does that happen? Don't we have to restart the operator?

pkg/metrics/metrics.go
```diff
@@ -245,21 +246,29 @@ func (c *Config) Configure(ctx context.Context) error {
...
 func (c *Config) validate(ctx context.Context) (bool, error) {
 	// validate if metrics label is added to namespace
 	wmcoNamespace, err := c.CoreV1().Namespaces().Get(ctx, c.namespace, metav1.GetOptions{})
 	if err != nil {
 		return false, errors.Wrap(err, "error getting operator namespace")
 	}
-	if wmcoNamespace.Labels["openshift.io/cluster-monitoring"] != "true" {
-		metricsEnabled = false
+
+	boolVal := false
```
mansikulkarni96, line 258: Instead of bool val, can we give it a monitoring specific name? maybe monitoringEnabled?

saifshaikh48: I opted for the non-specific name because there's already a global variable by the name of `metricsEnabled` (which gets assigned to the `boolVal` later in this function). I can make this `labelValue` instead?

aravindhp (approved):

mansikulkarni96:
Fixes #1478

pkg/metrics/metrics.go
```diff
@@ -245,21 +246,29 @@ func (c *Config) Configure(ctx context.Context) error {
 	return nil
 }
 
-// validate will verify if cluster monitoring is enabled in the operator namespace.
-// If the label is not present, it will log and send warning events to the user.
+// validate will verify if cluster monitoring is enabled in the operator namespace. If the label is set to false or not
+// present, it will log and send warning events to the user. If the label holds a non-boolean value, returns an error.
 func (c *Config) validate(ctx context.Context) (bool, error) {
```
jrvaldes, line 251: Does it make sense to add a unit test to cover the [mentioned](https://github.com/openshift/windows-machine-config-operator/pull/1479#discussion_r1136100529) cases?

https://github.com/openshift/windows-machine-config-operator/blob/767f93c6eac744c84d0741989de9b358fa161159/pkg/metrics/metrics_test.go#L4

saifshaikh48: I don't think it's necessary to add one since there's e2e tests covering this functionality

saifshaikh48:
/retest-required

cluster spinup flakes

aravindhp:
@saifshaikh48 for bugs please do not cherry-pick to all releases from the master PR. Just to the release-4.n-1 branch. Then do the release-4.n-2 cherry pick from the release-4.n-1PR. If you don't do this then you have to manually fix up all the Jira links. So please close all the other cherry picks.

PR #1480, opened 2023-03-14
[hack] MachineSet script platform discovery fix

sebsoto:
The hack/machineset.sh script was relying on an incorrect way to determine the platform of the cluster. The expected result of the query has changed, and broken the script. This commit changes the script to use use the infrastructure/cluster object to correctly determine the platform.

### DIFF
--- a/hack/machineset.sh
+++ b/hack/machineset.sh
@@ -287,7 +287,7 @@ EOF
 }
 
 # Retrieves the Cloud Provider for the OpenShift Cluster
-provider="$(oc -n openshift-kube-apiserver get configmap config -o json | jq -r '.data."config.yaml"' | jq '.apiServerArguments."cloud-provider"' | jq -r '.[]')"
+platform="$(oc get infrastructure cluster -ojsonpath={.spec.platformSpec.type})"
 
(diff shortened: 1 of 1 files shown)


hack/machineset.sh
```diff
@@ -287,7 +287,7 @@ EOF
 }
 
 # Retrieves the Cloud Provider for the OpenShift Cluster
-provider="$(oc -n openshift-kube-apiserver get configmap config -o json | jq -r '.data."config.yaml"' | jq '.apiServerArguments."cloud-provider"' | jq -r '.[]')"
+platform="$(oc get infrastructure cluster -ojsonpath={.spec.platformSpec.type})"
```
aravindhp, line 290: Will this work for older clusters?

sebsoto: It should, its the same method we use in the e2e tests: https://github.com/openshift/windows-machine-config-operator/blob/master/test/e2e/providers/cloudprovider.go#L34

aravindhp (approved):

alinaryan:
/lgtm

saifshaikh48 (commented):
/lgtm

aravindhp:
/override ci/prow/vsphere-e2e-operator

The MachineSet script does not affect any code or e2e tests

PR #1482, opened 2023-03-15
OCPBUGS-10572: Fix cleanup failure on WMCO upgrade

mansikulkarni96:
This PR rearranges the order of WMCO instance deconfiguration operations. It also updates the directory removal operation to delete the ksDir files excluding the WICD service related files. This allows us to retry
reconciliation in case of a deletion failure, resulting in WICD cleanup cmd being called. k8sDir which only contains the WICD files is deleted as the last step in the directory cleanup process with best effort, logging any errors. At this point we don't want to retry reconciliation as the WICD files would have been deleted.

### DIFF
--- a/pkg/windows/windows.go
+++ b/pkg/windows/windows.go
@@ -414,11 +414,11 @@ func (vm *windows) Deconfigure(watchNamespace string, apiServerURL string) error
 	if err := vm.RunWICDCleanup(apiServerURL, watchNamespace); err != nil {
 		return errors.Wrap(err, "Unable to cleanup the Windows instance")
 	}
-	if err := vm.removeDirectories(); err != nil {
-		return errors.Wrap(err, "unable to remove created directories")
-	}
 	if err := vm.ensureHNSNetworksAreRemoved(); err != nil {
 		return errors.Wrap(err, "unable to ensure HNS networks are removed")
+	}
+	if err := vm.removeDirectories(); err != nil {
+		return errors.Wrap(err, "unable to remove created directories")
 	}
 	return nil
 }
@@ -592,9 +592,23 @@ func (vm *windows) createDirectories() error {
 func (vm *windows) removeDirectories() error {
 	vm.log.Info("removing directories")
 	for _, dir := range RequiredDirectories {
+		if dir == K8sDir {
+			// Exclude WICD binary and credential files, used in the WICD cleanup cmd
+			// This allows us to retry reconciliation again in case of a failure here
+			// which will result in the WICD cleanup cmd being called.
+			if out, err := vm.Run(rmK8sFilesCmd(), true); err != nil {
+				return fmt.Errorf("unable to remove directory %s, out: %s, err: %s", dir, out, err)
+			}
+			continue
+		}
 		if out, err := vm.Run(rmDirCmd(dir), true); err != nil {
 			return fmt.Errorf("unable to remove directory %s, out: %s, err: %s", dir, out, err)
 		}
+	}
+	// Do a best effort deletion of k8sDir which now includes only the WICD related files
+	// At this point, we don't want to retry reconciliation again as the WICD files could have been deleted.
+	if out, err := vm.Run(rmDirCmd(K8sDir), true); err != nil {
+		vm.log.V(1).Error(err, "unable to remove directory", "dir", K8sDir, "out", out)
 	}
 	return nil
 }
@@ -958,6 +972,12 @@ func mkdirCmd(dirName string) string {
 // rmDirCmd returns the PowerShell command to recursively remove a directory if it exists
 func rmDirCmd(dirName string) string {
 	return fmt.Sprintf("if(Test-Path %s) {Remove-Item -Recurse -Force %s}", dirName, dirName)
+}
+
+// rmK8sFilesCmd() returns the PowerShell command to remove the k8sDir files excluding WICD files
+func rmK8sFilesCmd() string {
+	return fmt.Sprintf("if(Test-Path %s) {Get-ChildItem %s -Recurse -Exclude %s,%s,%s | Remove-Item -Force -Recurse}",
+		K8sDir, K8sDir, wicdPath, wicdTokenFile, wicdCAFile)
 }
 
 // getHNSNetworkCmd returns the Windows command to get HNS network by name


pkg/windows/windows.go
```diff
@@ -591,6 +597,21 @@ func (vm *windows) createDirectories() error {
 // removeDirectories removes all directories created as part of the configuration process
 func (vm *windows) removeDirectories() error {
 	vm.log.Info("removing directories")
+	existingFiles, err := getFilesToTransfer()
+	if err != nil {
+		return errors.Wrapf(err, "error getting list of files transferred to the instance")
+	}
+	for src, dest := range existingFiles {
+		filePath := dest + "\\" + filepath.Base(src.Path)
+		if out, err := vm.Run(rmFileCmd(filePath), true); err != nil {
```
aravindhp, line 606: Should we consider retrying a few times? I am worried that given the amount network issues we see executing commands over SSH we are going to have to reconcile a few times before all files are deleted. Or should we introduce a script that does the file deletion and HSN network removal?

pkg/windows/windows.go
```diff
@@ -591,6 +597,21 @@ func (vm *windows) createDirectories() error {
...
+		}
+	}
+	for _, file := range wicdFiles {
+		if out, err := vm.Run(rmFileCmd(file), true); err != nil {
+			vm.log.V(1).Error(err, "unable to remove file", "file", file, "out", out)
+		}
+	}
 	for _, dir := range RequiredDirectories {
 		if out, err := vm.Run(rmDirCmd(dir), true); err != nil {
 			return fmt.Errorf("unable to remove directory %s, out: %s, err: %s", dir, out, err)
```
aravindhp, line 617: This actually does not fix the issue. If there is an error at this stage, we will attempt to reconcile again and fail as the WICD files have been removed. You don't need to delete the WICD files individually but just make the directory deletion best effort.

aravindhp (changes requested):

pkg/windows/windows.go
mansikulkarni96: Agreed directory deletion should be best effort. I am deleting WICD files separately as a best effort as the cleanup command requires all the three files, if any one of them is deleted but the other one fails then we would end up in the same spot.

aravindhp: If you are going to make WICD file deletion and directory deletion all best effort, what is the point of deleting WICD files individually? We should be trying to reduce the number of SSH calls we make.

mansikulkarni96: ack, they will get removed as part of the directory deletion.

pkg/windows/windows.go
```diff
@@ -157,6 +157,12 @@ var (
 		podManifestDirectory,
 		K8sDir,
 	}
+	// wicdFiles is a list of files associated with the WICD service
```
mansikulkarni96, line 160: will remove this var in the next update.

pkg/windows/windows.go
```diff
@@ -591,9 +597,32 @@ func (vm *windows) createDirectories() error {
...
+			return !(strings.TrimSpace(out) == "True"), nil
+		})
+		if err != nil {
+			return errors.Wrapf(err, "unable to remove file: %s", filePath)
+		}
+	}
 	for _, dir := range RequiredDirectories {
 		if out, err := vm.Run(rmDirCmd(dir), true); err != nil {
-			return fmt.Errorf("unable to remove directory %s, out: %s, err: %s", dir, out, err)
+			vm.log.V(1).Error(err, "unable to remove directory", "dir", dir, "out", out)
```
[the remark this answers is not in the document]
mansikulkarni96: @saifshaikh48 we want to make directory removals best effort so we are deleting files one by one. There is no motivation to remove files sequentially otherwise.

pkg/windows/windows.go
```diff
@@ -591,9 +597,32 @@ func (vm *windows) createDirectories() error {
...
+				vm.log.V(1).Error(err, "error removing file", "file", filePath, "out", out)
+				return false, nil
+			}
+			out, err := vm.Run("Test-Path "+filePath, true)
+			if err != nil {
+				vm.log.V(1).Error(err, "error checking if file exists", "file", filePath,
+					"out", out)
+				return false, nil
+			}
+			return !(strings.TrimSpace(out) == "True"), nil
```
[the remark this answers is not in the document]
mansikulkarni96: I see your point, I had added a verification just to be sure the file is removed. But it is not necessary.

pkg/windows/windows.go
```diff
@@ -591,9 +597,32 @@ func (vm *windows) createDirectories() error {
 // removeDirectories removes all directories created as part of the configuration process
 func (vm *windows) removeDirectories() error {
 	vm.log.Info("removing directories")
+	existingFiles, err := getFilesToTransfer()
+	if err != nil {
+		return errors.Wrapf(err, "error getting list of files transferred to the instance")
+	}
+	for src, dest := range existingFiles {
+		filePath := dest + "\\" + filepath.Base(src.Path)
```
aravindhp, line 605: Shouldn't you be using `path.Join` or `filepath.Join` or does that not work given these are Windows filepaths?

pkg/windows/windows.go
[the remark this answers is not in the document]
aravindhp: In addition we should remove lines 611-615 i.e. the 2nd `Test-Path` call.

aravindhp (changes requested):

pkg/windows/windows.go
```diff
@@ -608,11 +608,15 @@ func (vm *windows) removeDirectories() error {
 			return errors.Wrapf(err, "unable to remove file: %s", filePath)
 		}
 	}
-	for _, dir := range RequiredDirectories {
+	dirsToDelete := RequiredDirectories[:len(RequiredDirectories)-1]
```
aravindhp, line 611: You are assuming that the `k8sDir` will always be the last entry in the slice. That is a risky approach and could potentially introduce a bug in the future. We should explicitly be checking for `k8sDir`.

aravindhp (changes requested):

pkg/windows/windows.go
mansikulkarni96: This is the pattern used in this file. I looked into it, "path/filepath" can be used as it is OS aware.

aravindhp: >This is the pattern used in this file.

That does not make it correct though and we shouldn't be adding to it.

>I looked into it, "path/filepath" can be used as it is OS aware.

Aren't you constructing a Windows path on a Linux OS? So maybe that is why filepath.Join() was not used.

mansikulkarni96: I agree, we should not follow the pattern. I was stating the fact, I am testing an alternative, will update.

mansikulkarni96: I tested it out, current way is the only way to proceed. As you mentioned we are creating a path for Windows on linux using an inbuilt function is not feasible here.

aravindhp (approved):

pkg/windows/windows.go
```diff
@@ -591,11 +591,36 @@ func (vm *windows) createDirectories() error {
 // removeDirectories removes all directories created as part of the configuration process
 func (vm *windows) removeDirectories() error {
 	vm.log.Info("removing directories")
+	existingFiles, err := getFilesToTransfer()
+	if err != nil {
+		return errors.Wrapf(err, "error getting list of files transferred to the instance")
+	}
+	for src, dest := range existingFiles {
+		filePath := dest + "\\" + filepath.Base(src.Path)
+		err = wait.Poll(retry.Interval, retry.Timeout, func() (bool, error) {
```
sebsoto, line 600: Why does this polling need to be introduced?
If an error occurs the instance deconfigure wll be re-queued by the controller-runtime manager.
The node object will still exist at this point, so that won't be a problem.

pkg/windows/windows.go
```diff
@@ -591,11 +591,36 @@ func (vm *windows) createDirectories() error {
...
+		if dir == K8sDir {
+			continue
+		}
 		if out, err := vm.Run(rmDirCmd(dir), true); err != nil {
 			return fmt.Errorf("unable to remove directory %s, out: %s, err: %s", dir, out, err)
 		}
 	}
+	if out, err := vm.Run(rmDirCmd(K8sDir), true); err != nil {
+		vm.log.V(1).Error(err, "unable to remove directory", "dir", K8sDir, "out", out)
+	}
```
sebsoto, line 623: I'm trying to understand the reasoning behind why this is the only path being made optional.
```
		// Skip cleanup of k8sDir as it contains the WICD binary and credentials file, required for
		// running the WICD cleanup command.
```
WICD cleanup is ran before this. I'm assuming this is in case this is reconciled again after this point and WICD cleanup cannot be ran. Considering that case is still possible even with this change, have you tried instead checking if the required files for WICD exist [here](https://github.com/openshift/windows-machine-config-operator/blob/master/pkg/windows/windows.go#L414) before calling `RunWICDCleanup`?

As you have already made it so that this function only runs after `RunWICDCleanup`, checking for the WICD files, will make the order of file deletion in this function not matter.

sebsoto (changes requested):

pkg/windows/windows.go
aravindhp: I had suggested this in an earlier [comment](https://github.com/openshift/windows-machine-config-operator/pull/1482#discussion_r1140443886) but I am not set on it. I understand you are suggesting the regular k8s pattern but we need to be pragmatic. These reconcile roundtrips are becoming expensive because of SSH failures. But I am not set on it and can be convinced otherwise.

pkg/windows/windows.go
aravindhp: What about the corner case where we end up in this path with an out of band WICD binary removal. We will go ahead and attempt to delete rest of the directories without rest of the services being properly cleaned. I would rather we fail on WICD clean up execution.

pkg/windows/windows.go
sebsoto: If we are worried about time here, I think the best thing to do would be to condense this into as little powershell commands as possible, `rmDirCmd()` returns the PS equivalent of `rm -rd -f`.

Depending on the outcome of [this disscusion](https://github.com/openshift/windows-machine-config-operator/pull/1482#discussion_r1142802352) this could be simplified by deleting `remoteDir` (C:\Temp) and `k8sdir` (C:\k). All the other files will be removed as part of that.


As a side note, if polling needs to be used, `PollImmediate` should be used instead of `Poll`, the code as is introduces `len(existingFiles) * retry.Interval` unnecessary waiting time.

pkg/windows/windows.go
sebsoto: In the Deconfigure case, if an out of band WICD binary removal occurred how would that situation be remediated? WMCO would continuously fail as it is not going to copy those files over anymore, as it is trying to remove the node from the cluster.

Manual intervention would be required, for the user to re-add the instance to the windows-instances ConfigMap, wait for it to be configured, and then remove it from the ConfigMap.

To be clear, I'm suggesting this be approached by making the change within `func (vm *windows) Deconfigure`:

```
	if err := vm.deconfigureWICD(); err != nil {
		return err
	}
	if wicdBinaryExistsOnNode && requiredServiceAccountCredsExistOnNode{
		if err := vm.RunWICDCleanup(apiServerURL, watchNamespace); err != nil {
			return errors.Wrap(err, "Unable to cleanup the Windows instance")
		}
		else {
			vm.log.Warning("Can't clean up service for node...")
		}
	}
```

saifshaikh48 (commented):
LGTM from me, but there's some open discussion to go another route

pkg/windows/windows.go
mansikulkarni96: I see your point, we are doing everything here just to ensure we don't end up in a situation where WICD files don't exist on cleanup. If we are able to avoid getting into it without increasing the time required to deconfigure. I am up for it.

mansikulkarni96: @sebsoto I still feel maintaining a retry for directory removals would be beneficial, I can use pollImmediate for polling.

aravindhp: >In the Deconfigure case, if an out of band WICD binary removal occurred how would that situation be remediated? WMCO would continuously fail as it is not going to copy those files over anymore, as it is trying to remove the node from the cluster.

The situation should not be remediated. 

>To be clear, I'm suggesting this be approached by making the change within func (vm *windows) Deconfigure:

Please don't do this. Lets discuss this as a parking lot item.

pkg/windows/windows.go
```diff
@@ -592,10 +592,21 @@ func (vm *windows) createDirectories() error {
 func (vm *windows) removeDirectories() error {
 	vm.log.Info("removing directories")
 	for _, dir := range RequiredDirectories {
+		// exclude WICD binary and credential files removal used in the WICD cleanup cmd
```
aravindhp, line 595: You can remove `removal` from the comment, however the comment is misplaced. It should come inside the if statement.

pkg/windows/windows.go
```diff
@@ -960,6 +971,12 @@ func rmDirCmd(dirName string) string {
 	return fmt.Sprintf("if(Test-Path %s) {Remove-Item -Recurse -Force %s}", dirName, dirName)
 }
 
+// rmK8sDirCmd returns the PowerShell command to remove the k8sDir excluding WICD files
+func rmK8sDirCmd() string {
```
aravindhp, line 975: The name is misleading as you are not removing the directory but just the non WICD files. I suggest naming this `rmK8sFilesCmd()` or `rmNonWICDFilesCmd()` something similar.

pkg/windows/windows.go
```diff
@@ -592,10 +592,21 @@ func (vm *windows) createDirectories() error {
...
+			if out, err := vm.Run(rmK8sDirCmd(), true); err != nil {
+				return fmt.Errorf("unable to remove directory %s, out: %s, err: %s", dir, out, err)
+			}
+			continue
+		}
 		if out, err := vm.Run(rmDirCmd(dir), true); err != nil {
 			return fmt.Errorf("unable to remove directory %s, out: %s, err: %s", dir, out, err)
 		}
 	}
+	// cleanup k8sDir, uses best-effort cleanup for WICD service related files
```
aravindhp, line 606: `Do a best effort deletion of k8sDir which now includes just the WICD related files`

aravindhp (changes requested):

pkg/windows/windows.go
mansikulkarni96: updated to rmK8sFilesCmd()

pkg/windows/windows.go
```diff
@@ -592,10 +592,21 @@ func (vm *windows) createDirectories() error {
...
+				return fmt.Errorf("unable to remove directory %s, out: %s, err: %s", dir, out, err)
+			}
+			continue
+		}
 		if out, err := vm.Run(rmDirCmd(dir), true); err != nil {
 			return fmt.Errorf("unable to remove directory %s, out: %s, err: %s", dir, out, err)
 		}
 	}
+	// Do a best effort deletion of k8sDir which now includes only the WICD related files
+	if out, err := vm.Run(rmDirCmd(K8sDir), true); err != nil {
```
aravindhp, line 607: Please add:
```
At this point we don't want to retry reconciliation again as the WICD files could have been deleted.
```

aravindhp (approved):
Nit:
```
This commit rearranges the order of WMCO instance
deconfiguration operations.
It also updates the directory removal operation to
delete the ksDir files excluding the WICD service
related files. This allows us to retry
reconciliation in case of a deletion failure, resulting in WICD
cleanup cmd being called.
k8sDir which only contains the WICD files is deleted
as the last step in the directory cleanup process with
best effort, logging any errors. At this point we don't 
want to retry reconciliation as the WICD files would
have been deleted.
```
Sorry about being pedantic here but I would like to capture what we are doing clearly for the future.

PR #1349, opened 2022-12-01
WINC-942: [bundle] Prep for 8.0.0 release

alinaryan:
This change bumps the operator and OCP version
according to the WMCO 8.0.0 release.

Ran 'make bundle' after changing the Makefile and 'config/' files.

PR #1491, opened 2023-03-16
[bundle] Fix metadata version annotation

saifshaikh48:
This commit sets the proper OCP version in the bundle metadata so the
operator is only included in catalogs targeting the compatible OCP version.

Follow-up to https://github.com/openshift/windows-machine-config-operator/pull/1349/


aravindhp (approved):

mtnbikenc:
/hold
This PR targets `master` but the title says `release-4.13`.

aravindhp:
/approve cancel

Good catch @mtnbikenc. @saifshaikh48 you need to open this against the release-4.13 branch.

saifshaikh48:
@mtnbikenc @aravindhp Thanks for catching that. I updated the target branch to release-4.13

aravindhp (approved):
/unhold

PR #1492, opened 2023-03-16
[bundle] Fix metadata version annotation

saifshaikh48:
This PR sets the proper OCP version in the bundle metadata so the
operator is only included in catalogs targeting the compatible OCP version.

Follow-up to https://github.com/openshift/windows-machine-config-operator/pull/1456


aravindhp (approved):

saifshaikh48:
/retest-required

vsphere infra flakes seem to be resolved

PR #1498, opened 2023-03-16
[wmco] Remove netutil package

sebsoto:
netutil.ResolveToIPv4Address was reimplementing a standard library function net.ResolveIPAddr. The package has been removed and all uses of the reimplemented function have been switched to ResolveIpAddr().


aravindhp (approved):

PR #1499, opened 2023-03-17
Use ClusterFirstWithHostNet DNS policy

sebsoto:
As WMCO is ran with host network privileges, its DNS policy is currently
using the DNS servers of the Node it is running on.

Instead, the policy should be set to `ClusterFirstWithHostNet`. Without
this it will not be able to resolve any addresses defined within the
cluster's DNS, such as internal services. With this policy, cluster
defined services can be resolved, while also using the cluster
configured upstream name server for any names that are not within the
cluster.


saifshaikh48 (commented):
/lgtm

aravindhp (approved):

PR #1510, opened 2023-03-21
[windows] Remove vxlanPort parameter for New()

sebsoto:
This is no longer used within the windows package now that the hybrid
overlay service configuration is defined within the services package.

This commit removes all unnecessary instances where this is being
passed down through the code.


saifshaikh48 (commented):
/lgtm

PR #1516, opened 2023-03-22
[WIP] WICD kubeconfig

sebsoto:


aravindhp (changes requested):
Thanks for working on this, @sebsoto. Mostly LGTM.

cmd/daemon/main.go
```diff
@@ -32,19 +32,14 @@ var (
 		Long: "The Windows Instance Config Daemon performs multiple functions related to maintaining the expected " +
 			"state of a Windows Node.",
 	}
-	saToken      string
-	saCA         string
+	kubeconfig   string
 	apiServerURL string
```
aravindhp, line 36: Do we still need this here?

aravindhp (approved):
/lgtm

pkg/nodeconfig/init.go
```diff
@@ -43,7 +35,6 @@ func init() {
 	}
 	// populate the cache
 	nodeConfigCache.apiServerEndpoint = kubeAPIServerEndpoint
-	nodeConfigCache.credentials, err = getWICDCredentials()
 	if err != nil {
 		log.Error(err, "unable to get WICD service account credentials")
 	}
```
saifshaikh48, line 40: we can remove this as well

saifshaikh48 (commented):
LGTM, clear and useful refactor. just one small comment

pkg/nodeconfig/init.go
sebsoto: Good catch, updated

PR #1512, opened 2023-03-22
WINC-1001: [wmco] Use standard library errors package

alinaryan:
Uses the standard library errors package given the package, github.com/pkg/errors, has been archived.

PR #1520, opened 2023-03-23
[release-4.12] OCPBUGS-10784: Correct CSI feature gates

sebsoto:
The AzureFile and vSphere CSIMigration feature flags should be false. This puts them in line with Linux nodes, and allows for the in tree storage drivers to be used, enabling storage on Windows nodes.


pkg/nodeconfig/nodeconfig.go
```diff
@@ -526,6 +526,8 @@ func generateKubeletConfiguration(clusterDNS string) kubeletconfig.KubeletConfig
 			"SCTPSupport":                    true,
 			"ServiceNodeExclusion":           true,
 			"SupportPodPidsLimit":            true,
+			"CSIMigrationAzureFile":          false,
+			"CSIMigrationvSphere":            false,
```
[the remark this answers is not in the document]
aravindhp: I see this is targeted at release-4.12 so we should be good as long as we have picked up the feature gate unlock in o/k.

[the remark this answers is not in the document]
sebsoto: @aravindhp Yeah this is targeted specifically at 4.12 to match the feature flags there.
I'm not completely sure what the final state of CSI for 4.13  is going to be on the Linux side, but I don't think the in-tree azure-file drivers will be used, as the plugin hit GA in kube 1.26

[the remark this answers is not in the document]
sebsoto: Looking into it with the azure-file flag being GA in 1.26, it cannot be disabled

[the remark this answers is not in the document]
aravindhp: The unlock is only needed for 1.26 and release-4.12 is on 1.25.

aravindhp (approved):

aravindhp:
This applies only to release-4.12

/label jira/valid-bug

pkg/nodeconfig/nodeconfig.go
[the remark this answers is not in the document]
gnufied: The migration for vSphere is already false on 4.12 though. So there is no need to disable it explicitly.

sebsoto:
/hold
to add azure e2e test

test/e2e/providers/azure/azure.go
```diff
@@ -4,14 +4,13 @@ import (
...
 	config "github.com/openshift/api/config/v1"
 	mapi "github.com/openshift/api/machine/v1beta1"
-	core "k8s.io/api/core/v1"
-	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
-
 	"github.com/openshift/windows-machine-config-operator/test/e2e/clusterinfo"
 	"github.com/openshift/windows-machine-config-operator/test/e2e/providers/machineset"
+	core "k8s.io/api/core/v1"
+	storage "k8s.io/api/storage/v1"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
```
saifshaikh48, line 13: nit: move these to above import group

test/e2e/providers/azure/azure.go
```diff
@@ -33,19 +32,27 @@ type Provider struct {
...
 	*config.InfrastructureStatus
 	vmSize string
+	// vmLocation is the Azure "location" in which VMs should be created. This should not be read directly, instead
+	// it should be accessed through getVMLocation().
+	vmLocation string
 }
 
 // New returns a new Azure provider struct with the give client set and ssh key pair
-func New(clientset *clusterinfo.OpenShift, infraStatus *config.InfrastructureStatus) *Provider {
+func New(clientset *clusterinfo.OpenShift, infraStatus *config.InfrastructureStatus) (*Provider, error) {
```
saifshaikh48, line 41: Is there a case where the constructor can return error? IMO no need for this.

test/e2e/providers/azure/azure.go
```diff
@@ -86,32 +93,55 @@ func (p *Provider) newAzureMachineProviderSpec(location, zone string) (*mapi.Azu
...
-
 	return machineset.New(rawProviderSpec, p.InfrastructureName, replicas, withIgnoreLabel, ""), nil
 }
 
 func (p *Provider) GetType() config.PlatformType {
 	return config.AzurePlatformType
 }
+
+func (p *Provider) getVMLocation() (string, error) {
+	if p.vmLocation != "" {
```
saifshaikh48, line 113: Will this condition ever be true? I don't see `vmLocation` being set anywhere

test/e2e/storage_test.go
```diff
@@ -0,0 +1,57 @@
...
+// TestStorage tests that persistent storage can be used with Windows workloads on supported providers
+func TestStorage(t *testing.T) {
+	tc, err := NewTestContext()
+	require.NoError(t, err)
+	// If more cloud provider storage tests are to be added, the code which provisions StorageClasses and PVCs should
+	// be moved into the CloudProvider interface
+	provider, isAzureProvider := tc.CloudProvider.(*azureProvider.Provider)
+	if !isAzureProvider {
+		t.Skip("storage not supported on this platform")
+	}
```
saifshaikh48, line 25: With this, if there is an issue with the azure provider that causes the cast to fail, we ignore it and skip storage testing (even if tested on azure). Consider
```
	if tc.CloudProvider.GetType() != config.AzurePlatformType {
		t.Skip("storage not supported on this platform")
	}
	provider, isAzureProvider := tc.CloudProvider.(*azureProvider.Provider)
	require.True(t, isAzureProvider)
```

saifshaikh48 (commented):
Nice work

test/e2e/providers/azure/azure.go
sebsoto: You're right, was left over from a previous method I was trying

test/e2e/network_test.go
```diff
@@ -461,14 +463,32 @@ func (tc *testContext) getWindowsServerContainerImage() string {
...
+	if pvcToMount != nil {
+		volumes = append(volumes, v1.Volume{
+			Name: pvcToMount.ClaimName,
+			VolumeSource: v1.VolumeSource{
+				PersistentVolumeClaim: pvcToMount,
+			},
+		})
+		volumeMounts = append(volumeMounts, v1.VolumeMount{
+			Name:      pvcToMount.ClaimName,
+			MountPath: "C:\\mnt\\storage",
```
mansikulkarni96, line 488: nit: path can be made a constant

test/e2e/providers/azure/azure.go
```diff
@@ -86,32 +95,56 @@ func (p *Provider) newAzureMachineProviderSpec(location, zone string) (*mapi.Azu
 
 // GenerateMachineSet generates the machineset object which is aws provider specific
```
mansikulkarni96, line 96: Azure provider specific?

test/e2e/providers/azure/azure.go
```diff
@@ -86,32 +95,56 @@ func (p *Provider) newAzureMachineProviderSpec(location, zone string) (*mapi.Azu
...
-	if err != nil {
-		return nil, fmt.Errorf("failed to get master-0 machine resource: %v", err)
-	}
-	masterProviderSpec := new(mapi.AzureMachineProviderSpec)
-	err = json.Unmarshal(machines.Spec.ProviderSpec.Value.Raw, masterProviderSpec)
-	if err != nil {
-		return nil, fmt.Errorf("failed to unmarshal master-0 azure machine provider spec: %v", err)
-	}
-
 	// create new machine provider spec for deploying Windows node in the same Location and Zone as master-0
```
mansikulkarni96, line 98: comment needs to be updated, we are not getting zone from master-0

test/e2e/providers/azure/azure.go
```diff
@@ -86,32 +95,56 @@ func (p *Provider) newAzureMachineProviderSpec(location, zone string) (*mapi.Azu
...
 	}
-
 	return machineset.New(rawProviderSpec, p.InfrastructureName, replicas, withIgnoreLabel, ""), nil
 }
 
 func (p *Provider) GetType() config.PlatformType {
 	return config.AzurePlatformType
 }
+
+func (p *Provider) getVMLocation() (string, error) {
```
mansikulkarni96, line 114: Needs doc

test/e2e/providers/azure/azure.go
```diff
@@ -86,32 +95,56 @@ func (p *Provider) newAzureMachineProviderSpec(location, zone string) (*mapi.Azu
...
+	if p.vmLocation != "" {
+		return p.vmLocation, nil
+	}
+	// Inspect master-0 to get the Azure Location and Zone
```
mansikulkarni96, line 118: only location here?

test/e2e/storage_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/stretchr/testify/require"
+	core "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	azureProvider "github.com/openshift/windows-machine-config-operator/test/e2e/providers/azure"
+)
+
+// TestStorage tests that persistent storage can be used with Windows workloads on supported providers
+func TestStorage(t *testing.T) {
```
mansikulkarni96, line 18: Do we need to export this?

mansikulkarni96 (changes requested):
Thanks for working on this @sebsoto  PTAL at my comments.

test/e2e/storage_test.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	pvc, err := tc.client.K8s.CoreV1().PersistentVolumeClaims(tc.workloadNamespace).Create(context.TODO(), &pvcSpec,
+		meta.CreateOptions{})
+	require.NoError(t, err)
+	defer tc.client.K8s.CoreV1().PersistentVolumeClaims(tc.workloadNamespace).Delete(context.TODO(), pvc.GetName(),
+		meta.DeleteOptions{})
+
+	// Testing that the pod goes to ready is sufficient as the pod will not become ready without successfully mounting
+	// the volume
+	ws, err := tc.deployWindowsWebServer("storage-test", nil,
```
aravindhp, line 54: So we are deploying yet another webserver to test storage. Can't we deploy the webserver before we run the networking and storage tests and use them in both? I am worried about the increase in time this is going to take. I realize this might take a refactor but please see if it is easily doable. We are going to have to do this anyhow as we keep adding more storage support for more providers.

aravindhp (changes requested):

sebsoto:
/hold cancel
e2e test added

test/e2e/providers/cloudprovider.go
```diff
@@ -23,8 +23,9 @@ type CloudProvider interface {
 	GetType() config.PlatformType
 	// StorageSupport indicates if we support Windows storage on this provider
 	StorageSupport() bool
-	// CreatePVC creates a new PersistentVolumeClaim that can be used by a workload
-	CreatePVC(p client.Interface) (*core.PersistentVolumeClaim, error)
+	// CreatePVC creates a new PersistentVolumeClaim that can be used by a workload. The PVC will be created with
+	// the given client, in the given namespace.
+	CreatePVC(client.Interface, string) (*core.PersistentVolumeClaim, error)
```
aravindhp, line 28: Don't we need to do this in master also?

test/e2e/providers/azure/azure.go
```diff
@@ -113,14 +110,76 @@ func (p *Provider) GenerateMachineSet(withIgnoreLabel bool, replicas int32) (*ma
...
 func (p *Provider) GetType() config.PlatformType {
 	return config.AzurePlatformType
 }
 
 func (p *Provider) StorageSupport() bool {
-	return false
+	return true
+}
+
+func (p *Provider) CreatePVC(client client.Interface, namespace string) (*core.PersistentVolumeClaim, error) {
```
aravindhp, line 138: Please add a doc

sebsoto:
@aravindhp
>@sebsoto don't you need to create a Pod that is using the PVC? In addition don't you need a Persistent Volume (PV)? I was thinking we would need to do something similar to [Configure a Pod to Use a PersistentVolume for Storage](https://kubernetes.io/docs/tasks/configure-pod-container/configure-persistent-volume-storage/)

That is being done by making `azureProvider.SupportsStorage()` return true.
https://github.com/openshift/windows-machine-config-operator/blob/e50773d460b9b7821ac58eff09d2dbefafc16104/test/e2e/network_test.go#L249-L255

That ends up causing the PVC to be mounted in the pod https://github.com/openshift/windows-machine-config-operator/blob/e50773d460b9b7821ac58eff09d2dbefafc16104/test/e2e/network_test.go#L485-L496

I've not found creating a PV is necessary, the backing storage [is dynamically provisioned](https://learn.microsoft.com/en-us/azure/aks/azure-csi-files-storage-provision#dynamically-provision-a-volume) if you don't associate the PVC with a PV.

sebsoto:
@aravindhp I was wrong about this

Its actually specifically when the PVC is associated with a StorageClass and not just when its not linked to a PV.
I'll add the [OCP docs](https://docs.openshift.com/container-platform/4.12/storage/dynamic-provisioning.html#about_dynamic-provisioning) link to the code

test/e2e/providers/azure/azure.go
```diff
@@ -113,14 +110,79 @@ func (p *Provider) GenerateMachineSet(withIgnoreLabel bool, replicas int32) (*ma
...
 
-func (p *Provider) CreatePVC(_ client.Interface, _ string) (*core.PersistentVolumeClaim, error) {
-	return nil, fmt.Errorf("storage not supported on azure")
+// ensureStorageClass ensures an azure-file storage class exists for use with in-tree storage
+func (p *Provider) ensureStorageClass(client client.Interface) (*storage.StorageClass, error) {
+	sc, err := client.StorageV1().StorageClasses().Get(context.TODO(), storageClassName, meta.GetOptions{})
+	if err == nil {
+		return sc, nil
+	} else if !k8sapierrors.IsNotFound(err) {
+		return nil, fmt.Errorf("error getting storage class 'e2e':%w", err)
```
alinaryan, line 167: space after `:` to follow the err formatting I'm adding in #1512

sebsoto:
/hold
Cherry-pick commit from https://github.com/openshift/windows-machine-config-operator/pull/1525
This must merge after that.

test/e2e/providers/vsphere/vsphere.go
```diff
@@ -134,9 +140,50 @@ func (p *Provider) GetType() config.PlatformType {
...
+	return client.CoreV1().PersistentVolumeClaims(namespace).Create(context.TODO(), &pvcSpec, meta.CreateOptions{})
+}
+
+// ensureStorageClass ensures a vsphere-volume NTFS storage class exists for use with in-tree storage
+func (p *Provider) ensureStorageClass(client client.Interface) (*storage.StorageClass, error) {
+	sc, err := client.StorageV1().StorageClasses().Get(context.TODO(), storageClassName, meta.GetOptions{})
+	if err == nil {
+		return sc, nil
+	} else if !k8sapierrors.IsNotFound(err) {
+		return nil, fmt.Errorf("error getting storage class 'e2e': %w", err)
```
saifshaikh48, line 175: nit: use the storageClassName constant in the error msg

test/e2e/providers/vsphere/vsphere.go
```diff
@@ -11,14 +11,20 @@ import (
...
 	client "k8s.io/client-go/kubernetes"
 
 	"github.com/openshift/windows-machine-config-operator/test/e2e/clusterinfo"
 	"github.com/openshift/windows-machine-config-operator/test/e2e/providers/machineset"
 )
 
-const defaultCredentialsSecretName = "vsphere-cloud-credentials"
+const (
+	defaultCredentialsSecretName = "vsphere-cloud-credentials"
+	storageClassName             = "e2e"
```
saifshaikh48, line 26: The same constant was introduced in the previous commit. Is it possible to consolidate?

saifshaikh48 (commented):
/lgtm
Nice work, just a few nits

test/e2e/providers/vsphere/vsphere.go
sebsoto: Unfortunately not with the way the packages are structured

pkg/nodeconfig/nodeconfig.go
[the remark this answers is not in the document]
aravindhp: @jsafrane please chime in here.

test/e2e/network_test.go
aravindhp, line 1: Please add a story to extract the workload creation independent of networking tests.

test/e2e/providers/vsphere/vsphere.go
aravindhp: Can't you introduce a constant in a higher level package?

pkg/nodeconfig/nodeconfig.go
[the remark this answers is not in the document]
jsafrane: We handle CSIMigrationvSphere & AzureFile via FeatureGates object. [Default 4.12 OCP FeatureGates disable them](https://github.com/openshift/api/blob/release-4.12/config/v1/types_feature.go#L133-L143), TechPreviewNoUpgrade enables them.

In 4.13, we don't say anything in openshift/api, i.e. both are be enabled today, incl. CSIMigrationvSphere. We will disable its behavior in a hackish way using a carry patch + env. var. https://github.com/openshift/kubernetes/pull/1514

test/e2e/providers/vsphere/vsphere.go
sebsoto: I could put it in the clusterinfo package or create a new one.
That will make cherry-picking the in-tree vsphere test to release-4.13 more difficult.

pkg/nodeconfig/nodeconfig.go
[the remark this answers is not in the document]
aravindhp: @sebsoto looks like you don't have to explicitly disable this here. Or did you discover something else otherwise in your tests? It is possible that the change in the OpenShift API is not reflected in Windows nodes.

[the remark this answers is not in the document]
sebsoto: > looks like you don't have to explicitly disable this here

I don't see how thats the case, Windows nodes dont use the FeatureGates defined there, we use our own set. The Linux worker node feature gates are defined in the worker MachineConfig, which we don't use.

aravindhp (approved):

test/e2e/network_test.go
sebsoto: https://issues.redhat.com/browse/WINC-1021

sebsoto:
Keep getting node-logs errors, but had [one non-node log](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/1520/pull-ci-openshift-windows-machine-config-operator-release-4.12-vsphere-e2e-operator/1641449548108795904) failure, which I'm trying to debug with the final commit

saifshaikh48 (commented):
/lgtm

nice work getting this all working together

sebsoto:
/test vsphere-e2e-operator

node logs flake

PR #1536, opened 2023-03-31
WINC-637: Set cluster-wide proxy environment variables on Windows instances

saifshaikh48:
This PR expands the services ConfigMap data schema and the expected services ConfigMap to hold a new field, a list of name/value pairs representing environment variables to be set on Windows instances.  If the cluster-wide proxy variables change at any point, the ConfigMap is re-created to hold the new values. It also introduces a functionality to the WICD controller to ensure all the environment variables defined in the services ConfigMap are set as expected on the Windows nodes. If there is a discrepancy, WICD updates the values and signals to WMCO to restarts the instance to allow all OpenShift managed Windows services to pick up changes.

summary of proxy vars flow:
- WICD controller is running
- WICD reconciles proxy vars. if it updates any of them, WICD applies a reboot-required annotation to the node
  - WICD now waits for reboot to occur (polls Windows processes to see if they have inherited the correct env vars values)
- WMCO's node controller reacts to setting of reboot annotation by restarting the instance
  - node controller now waits for WICD to remove the reboot annotation
  - after this, WICD is can see the env vars are set properly at the Windows service level
- WICD continues as it currently does, reconciling Windows services and applying version annotation when node is ready


mansikulkarni96:
/retitle WINC-637: Set cluster-wide proxy environment variables on Windows instances

test/e2e/proxy_test.go
```diff
@@ -15,25 +16,63 @@ import (
...
+			require.NoError(t, err, "unable to get node address")
 
-	t.Run("Trusted CA ConfigMap validation", testTrustedCAConfigMap)
+			for _, proxyVar := range cluster.AcceptableProxyVars {
+				value, err := tc.getSystemEnvVar(addr, proxyVar)
+				require.NoError(t, err, "unable to get value of %s from instance", proxyVar)
+				assert.Equalf(t, expectedEnvVars[proxyVar], value, "incorrect value for %s", proxyVar)
+			}
+
+			for _, svcName := range windows.RequiredServices {
```
aravindhp, line 65: Are we adding the proxy variables to the configuration all services? Where are we doing that?

test/e2e/proxy_test.go
```diff
@@ -99,3 +138,75 @@ func (tc *testContext) waitForValidTrustedCAConfigMap() error {
...
+		}
+		if len(valueLines) > 0 {
+			value = strings.Join(valueLines, "")
+			value = strings.ReplaceAll(value, ";", ",")
+		}
+	}
+	return value, nil
+}
+
+// getServiceProxyEnvVars returns a map of proxy environment variables present in a service's config
```
aravindhp, line 173: I don't follow this. Where are we adding the proxy variables to the service's configuration?

aravindhp (changes requested):
Thanks for working on this, @saifshaikh48 @mansikulkarni96. Please address my comments.

test/e2e/proxy_test.go
mansikulkarni96: We set the ENV vars at machine level and they get propogated to the services. We are just verifying they are set correctly in the e2e test as having them at machine and service level is important.
PTAL https://github.com/openshift/enhancements/blob/4f18064f4ad9124efec465c59346e37171c0fcc9/enhancements/windows-containers/windows-node-egress-proxy.md?plain=1#L177

aravindhp: I completely forgot about this. Sorry about that. I suggest adding this as a comment somewhere in `reconcileEnvironment()`

pkg/daemon/controller/controller.go
```diff
@@ -253,13 +254,55 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
...
+		if actualVal != expectedVal {
+			err := registryKey.SetStringValue(name, expectedVal)
+			if err != nil {
+				// Do not log value as proxy information is sensitive
+				return fmt.Errorf("unable to set environment variable %s: %w", name, err)
+			}
+			restartRequired = true
+		}
+	}
+	if restartRequired {
```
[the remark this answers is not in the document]
mansikulkarni96: @aravindhp while I am working on resolving other comments. I wanted to get a clear understanding on how to move forward with this. Will draining and cordoning the node if this operation occurs after a node has been configured be a good remediation for this issue?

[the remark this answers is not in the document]
aravindhp: Correct. We would need to drain and cordon and the node before restarting. And then uncordon after the restart is complete.

pkg/servicescm/servicescm.go
```diff
@@ -91,11 +93,16 @@ type Data struct {
 	Services []Service `json:"services"`
 	// Files contains the path and checksum of all the files copied to a Windows VM by WMCO
 	Files []FileInfo `json:"files"`
+	// EnvironmentVars will be set as system-level environment variables on each Windows instance
```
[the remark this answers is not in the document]
saifshaikh48: It was part of the proxy enhancement

pkg/cluster/config.go
```diff
@@ -34,14 +34,14 @@ const (
...
+	// SupportedProxyVars is a list of the supported proxy variables
+	SupportedProxyVars = []string{"HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY"}
 	// ClusterWideProxyVars is a map of the global egress proxy variables and values from WMCO's environment
 	ClusterWideProxyVars map[string]string
 )
 
 // init runs once, initializing global variables
 func init() {
-	// OLM restarts the operator when the global cluster proxy config changes so these values will always be up-to-date
-	// at start time. We never expect a user to update these values via editing the operator pod spec
```
saifshaikh48, line 44: this is a useful comment IMO

pkg/daemon/controller/controller.go
```diff
@@ -253,13 +258,72 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
...
+			}
+			restartRequired = true
+		}
+		if err = registryKey.Close(); err != nil {
+			return fmt.Errorf("unable to close Windows system registry: %w", err)
+		}
+	}
+	if restartRequired {
+		// Cordon and drain the Node before we restart the instance
+		drainHelper := nodeconfig.NewDrainHelper(sc.k8sclientset, klog.NewKlogr())
```
saifshaikh48, line 306: where does `sc.k8sclientset` get initialized? we likely have to edit the `Options` struct and `setDefaults()` func

pkg/servicescm/servicescm.go
```diff
@@ -97,19 +97,17 @@ type Data struct {
 	EnvironmentVars map[string]string
```
saifshaikh48, line 97:
```suggestion
	EnvironmentVars map[string]string `json:"environmentVars,omitempty"`
```

saifshaikh48 (commented):
could you also fix up the commit message for 
`[wicd] Move cordon/drain responsiblities to WICD`

pkg/servicescm/servicescm.go
saifshaikh48, line 1: should these changes be part of the first commit?

pkg/services/services.go
```diff
@@ -52,7 +52,10 @@ func GenerateManifest(kubeletArgsFromIgnition map[string]string, vxlanPort strin
 	}
 	// TODO: All payload filenames and checksums must be added here https://issues.redhat.com/browse/WINC-847
 	files := &[]servicescm.FileInfo{}
-	return servicescm.NewData(services, files, cluster.ClusterWideProxyVars)
+	if proxyEnabled {
+		return servicescm.NewData(services, files, cluster.ClusterWideProxyVars)
+	}
+	return servicescm.NewData(services, files, nil)
```
saifshaikh48, line 58: Is it possible to remove the need for the `proxyEnabled` parameter?
When proxy is not enabled, cluster.ClusterWideProxyVars should be empty (len=0). Is there a difference in how empty and nil values are treated? I don't think there should be

mansikulkarni96: I agree but if that's the case do we even need the proxyenabled parameter in the configmapReconciler struct?

pkg/cluster/config.go
```diff
@@ -34,6 +34,8 @@ const (
 )
 
 var (
+	// SupportedProxyVars is a list of the supported proxy variables
+	SupportedProxyVars = []string{"HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY"}
 	// ClusterWideProxyVars is a map of the global egress proxy variables and values from WMCO's environment
 	ClusterWideProxyVars map[string]string
```
sebsoto, line 40: Neither of these need to be public variables, see comment lower in the file.

pkg/cluster/config.go
```diff
@@ -354,24 +356,22 @@ func IsCloudControllerOwnedByCCM(oclient configclient.Interface) (bool, error) {
...
+func IsProxyEnabled() bool {
+	return len(ClusterWideProxyVars) > 0
+}
 
-	proxyVars := make(map[string]string, len(proxyEnvVars))
-	for _, envVar := range proxyEnvVars {
-		// read settings from the WMCO container's environment
+// getProxyVars returns a map of the proxy variables and values from the WMCO container's environment. The presence of
+// any implies a proxy is enabled, as OLM would have injected them into the operator spec. Returns an empty map otherwise.
+func getProxyVars() map[string]string {
```
sebsoto, line 366: Please make this a getter, which will return `ClusterWideProxyVars` instead of having the user access the variable directly. If isn't already cached the variable can be populated before returning it, and the init function can be removed.

Using a getter with a documented function provides a better interface for a user of this package.

Init functions cannot be error handled, and I believe we should avoid them where possible.

pkg/servicescm/servicescm.go
```diff
@@ -91,15 +93,21 @@ type Data struct {
...
-// NewData returns a new 'Data' object with the given services and files. Validates given object contents on creation.
-func NewData(services *[]Service, files *[]FileInfo) (*Data, error) {
-	cmData := &Data{*services, *files}
+// NewData returns a new 'Data' object with the given services, files and ENV vars if they exist.
+// Validates given object contents on creation.
+func NewData(services *[]Service, files *[]FileInfo, envVars map[string]string) (*Data, error) {
+	cmData := Data{Services: *services, Files: *files}
+	if envVars != nil {
+		cmData.EnvironmentVars = envVars
+	}
```
sebsoto, line 106: Is this nil check necessary?
it will be set to nil either way will it not?

Both by being uninitialized with a default of nil or being set to the value of envVars, which is nil.

pkg/servicescm/servicescm.go
```diff
@@ -158,14 +166,21 @@ func Generate(name, namespace string, data *Data) (*core.ConfigMap, error) {
...
 	return servicesConfigMap, nil
 }
 
 // Parse converts ConfigMap data into the objects representing a Windows services ConfigMap schema
 // Returns error if the given data is invalid in structure
 func Parse(dataFromCM map[string]string) (*Data, error) {
-	if len(dataFromCM) != 2 {
-		return nil, fmt.Errorf("services ConfigMap should have exactly 2 keys")
+	if len(dataFromCM) != 2 && len(dataFromCM) != 3 {
+		return nil, fmt.Errorf("services ConfigMap can only have 2 or 3 number of keys")
```
sebsoto, line 183: Please add a comment explaining this.
Is this to provide compatibility when upgrading from a ConfigMap from a previous version? We need to make sure thats captured and removed when necessary. Making this less restrictive should also be considered.

pkg/servicescm/servicescm.go
```diff
@@ -186,7 +201,14 @@ func Parse(dataFromCM map[string]string) (*Data, error) {
 		return nil, err
 	}
 
-	return NewData(services, files)
+	envVars := &map[string]string{}
```
sebsoto, line 204: is a map not a pointer already?
The `&` is confusing me here

pkg/daemon/controller/controller.go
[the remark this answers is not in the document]
sebsoto: This has the potential to take down all windows instances on the cluster at once. This needs to be addressed in some way.

test/e2e/proxy_test.go
```diff
@@ -99,3 +140,63 @@ func (tc *testContext) waitForValidTrustedCAConfigMap() error {
...
+		return nil, fmt.Errorf("error running SSH job: %w", err)
+	}
+	return parseEnvVars(strings.Split(out, "\n")), nil
+}
+
+// parseEnvVars parses a given slice of string containing all environment variables with their name, value pairs
+// and returns a map of ENV vars to their corresponding values.
+// Sample input: ["Name  : HTTP_PROXY", "Value : http://dev:d3436c0b817f7ca8e23f7b47be49945d@10.0.1.10:3128/",
+// "Name  : SHELL", "Value : c:\windows\system32\cmd.exe"]
+func parseEnvVars(lines []string) map[string]string {
```
sebsoto, line 174: rename this to parseWindowsEnvVars please.

Why is this made to take a slice of strings and not the direct output of  the powershell command?

test/e2e/proxy_test.go
```diff
@@ -99,3 +140,63 @@ func (tc *testContext) waitForValidTrustedCAConfigMap() error {
 	}
 	return nil
 }
+
+// getSystemEnvVar returns the value corresponding to the input proxy ENV var as set in the registry
+func (tc *testContext) getSystemEnvVar(addr, variableName string) (map[string]string, error) {
+	command := fmt.Sprintf("Get-ChildItem -Path Env: | Where-Object -Property Name -eq '%s' | Format-List ",
+		variableName)
+	out, err := tc.runPowerShellSSHJob(fmt.Sprintf("get-%s-system-env-var",
```
sebsoto, line 148: Can you share the documentation for this powershell command?

test/e2e/proxy_test.go
```diff
@@ -99,3 +140,63 @@ func (tc *testContext) waitForValidTrustedCAConfigMap() error {
...
+	value := parseEnvVars(strings.Split(out, "\n"))
+	return value, nil
+}
+
+// getServiceProxyEnvVars returns a map of all environment variables present in a service's config
+func (tc *testContext) getServiceProxyEnvVars(addr, svcName string) (map[string]string, error) {
+	command := fmt.Sprintf("Get-Process %s | ForEach-Object { $_.StartInfo.EnvironmentVariables.GetEnumerator() "+
+		"| Format-List }",
+		svcName)
+	out, err := tc.runPowerShellSSHJob(fmt.Sprintf("get-%s-service-env-vars",
```
sebsoto, line 162: Same question as above

test/e2e/proxy_test.go
```diff
@@ -15,25 +16,65 @@ import (
...
 		require.NoError(t, err)
 	}
-	if clusterProxy.Status.HTTPProxy == "" && clusterProxy.Status.HTTPSProxy == "" {
-		t.Skip("cluster-wide proxy is not enabled in this environment")
-	}
+	expectedEnvVars := map[string]string{}
+	expectedEnvVars["HTTP_PROXY"] = clusterProxy.Status.HTTPProxy
+	expectedEnvVars["HTTPS_PROXY"] = clusterProxy.Status.HTTPSProxy
+	expectedEnvVars["NO_PROXY"] = clusterProxy.Status.NoProxy
+	err = tc.waitForWindowsNodes(gc.numberOfMachineNodes, false, false, false)
```
sebsoto, line 51: I suggest you use this instead https://github.com/openshift/windows-machine-config-operator/blob/f079f3ddbcbae368756f535ed78229295388924c/test/e2e/create_test.go#L564

Additionally, why is this only tested on Machine nodes?

sebsoto (changes requested):
Thanks for the PR, PTAL at my comments

pkg/cluster/config.go
saifshaikh48: SupportedProxyVars is exported for usage in the e2e tests. let me know if this a bad practice

pkg/servicescm/servicescm.go
saifshaikh48: The environmentVars key is optional in the services CM (present when a proxy is enabled). I'll add a comment explaining this

pkg/daemon/controller/controller.go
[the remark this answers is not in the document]
saifshaikh48: Here's how we plan to handle this:

- if restart is required, check if at least 1 windows node is ready+schedulable
- if yes, then cordon/drain/restart/uncordon
- if no, retry the check until a timeout 
  - if the timeout is reached, output a warning

pkg/servicescm/servicescm.go
saifshaikh48: changed+added a comment for clarity

test/e2e/proxy_test.go
mansikulkarni96: The powershell command is the one above that line it's a generic `Get-ChildIteam -Path` command
https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.management/get-childitem?view=powershell-7.3
the output is then filtered to get value of the specific ENV var
eg: Get-ChildItem -Path Env: | Where-Object -Property Name -eq 'HTTP_PROXY'

test/e2e/proxy_test.go
mansikulkarni96: same here, generic get-process command for service and then we get all env vars
eg: Get-Process kubelet | ForEach-Object { $_.StartInfo.EnvironmentVariables.GetEnumerator()}

pkg/daemon/controller/controller.go
```diff
@@ -275,6 +285,88 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
...
+	for name, expectedVal := range envVars {
+		// We cannot use the os.Setenv since it calls syscall.Setenv which, calls syscall.SetEnvironmentVariable
+		// As per Microsoft docs: "Calling SetEnvironmentVariable has no effect on the system environment variables"
+		registryKey, err := registry.OpenKey(registry.LOCAL_MACHINE, systemEnvVarRegistryPath, registry.ALL_ACCESS)
+		if err != nil {
+			return fmt.Errorf("unable to open Windows system registry key %s: %w", systemEnvVarRegistryPath, err)
+		}
+		defer func() {
+			if registryErr := registryKey.Close(); err != nil {
+				err = fmt.Errorf("unable to close Windows system registry key: %w", registryErr)
```
aravindhp, line 303: Won't you overwrite if another error occurred before the drain function ran?

pkg/daemon/controller/controller.go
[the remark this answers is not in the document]
aravindhp: Your plan just reduces the window where all instances could be down at once. More I think about this I am not sure if WICD can make the decision to restart. You need an external entity to ensure only one node is restarted at a time.

test/e2e/proxy_test.go
```diff
@@ -99,3 +139,67 @@ func (tc *testContext) waitForValidTrustedCAConfigMap() error {
...
+		strings.ToLower(strings.ReplaceAll(variableName, "_", "-"))), command, addr)
+	if err != nil {
+		return nil, fmt.Errorf("error running SSH job: %w", err)
+	}
+	value := parseWindowsEnvVars(out)
+	return value, nil
+}
+
+// getServiceProxyEnvVars returns a map of all environment variables present in a service's config
+func (tc *testContext) getServiceProxyEnvVars(addr, svcName string) (map[string]string, error) {
```
aravindhp, line 157: `getServiceProxyEnvVars` and `getSystemEnvVar` are identical except for the `command`. Please modularize with a helper that is called from each of these functions. Also this function should be `getProxyEnvVarsFromService`. 

```golang
func (tc *testContext) getProxyEnvVarsFromService(addr, svcName string) (map[string]string, error) {
        command := fmt.Sprintf("Get-Process %s | ForEach-Object ...")
        envVars, err  := getEnvVar(addr, svcName, command)
```

test/e2e/proxy_test.go
```diff
@@ -99,3 +139,67 @@ func (tc *testContext) waitForValidTrustedCAConfigMap() error {
...
+		strings.ReplaceAll(svcName, "_", "-")), command, addr)
+	if err != nil {
+		return nil, fmt.Errorf("error running SSH job: %w", err)
+	}
+	return parseWindowsEnvVars(out), nil
+}
+
+// parseWindowsEnvVars parses the Powershell output listing all environment variables with their name, value pairs
+// and returns a map of ENV vars to their corresponding values.
+// Sample input:
```
aravindhp, line 171: Thank you for adding this

aravindhp (changes requested):
I think the main point of contention is how are we going to handle reboots.

pkg/daemon/controller/controller.go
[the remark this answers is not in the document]
aravindhp: To expand on this, say there are 3 Windows nodes in the cluster. WICD running on all of them hits this point. It checks if there are other Windows nodes in `Ready` state. The answer comes back as yes, and all of them get restarted.

pkg/cluster/config.go
```diff
@@ -354,24 +349,27 @@ func IsCloudControllerOwnedByCCM(oclient configclient.Interface) (bool, error) {
...
+		return clusterWideProxyVars
+	}
+	// if clusterWideProxyVars is not already cached, initialize it. We never expect these values to change during
+	// runtime as OLM restarts the operator when the global cluster proxy config changes
+	clusterWideProxyVars = make(map[string]string, 3)
+	for _, envVar := range SupportedProxyVars {
 		value, found := os.LookupEnv(envVar)
 		if found {
-			proxyVars[envVar] = value
+			// on Windows, hostname lists are separated by semicolons rather than the Linux default of commas
```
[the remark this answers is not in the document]
saifshaikh48: It is typically a single host. Some custom applications may allow multiple (if routing through first proxy URL fails, second one is used as fallback, etc) but this is definitely not the standard

pkg/servicescm/servicescm.go
```diff
@@ -158,14 +167,23 @@ func Generate(name, namespace string, data *Data) (*core.ConfigMap, error) {
...
 	return servicesConfigMap, nil
 }
 
 // Parse converts ConfigMap data into the objects representing a Windows services ConfigMap schema
 // Returns error if the given data is invalid in structure
 func Parse(dataFromCM map[string]string) (*Data, error) {
-	if len(dataFromCM) != 2 {
-		return nil, fmt.Errorf("services ConfigMap should have exactly 2 keys")
+	// 2 required keys: services and files
+	// 1 optional key: environmentVars, which won't be present in the services CM if nil or empty
```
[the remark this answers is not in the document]
saifshaikh48: This check may pass but the later validation in the function will catch that, there's a unit test to ensure this.

If we remove this check, we would allow extra keys in the CM data. Although they would be ignored, I prefer stricter validation for more predictable results, especially since WMCO generates the services CM itself.

pkg/daemon/controller/controller.go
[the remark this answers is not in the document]
sebsoto: That approach in general is fine with me
however that doesn't need to be shoehorned into `ensureInstanceIsUpToDate`.
Instead, I feel you should introduce a node controller which reacts to changes in the node object.

controllers/controllers.go
```diff
@@ -56,6 +57,39 @@ func (r *instanceReconciler) ensureInstanceIsUpToDate(instanceInfo *instance.Inf
...
+			}
+
+			if err := drain.RunCordonOrUncordon(drainer, instanceInfo.Node, false); err != nil {
+				return fmt.Errorf("unable to uncordon node %s: %w", instanceInfo.Node.Name, err)
+			}
+			if err := metadata.RemoveRebootAnnotation(context.TODO(), r.client, *instanceInfo.Node); err != nil {
+				return err
+			}
+		}
+	}
```
aravindhp, line 91: At a minimum please make this a helper method. 

I am a bit confused about when this is being called. Has WICD finished running at this point? I also don't understand how this would work if WICD updated the environment variables outside of a configuration cycle.

pkg/daemon/controller/controller.go
```diff
@@ -271,9 +270,15 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
 	if err = sc.reconcileServices(cmData.Services); err != nil {
 		return ctrl.Result{}, err
 	}
-	if err = sc.reconcileEnvironmentVariables(cmData.EnvironmentVars, &node); err != nil {
+	restartRequired, err := sc.reconcileEnvironmentVariables(cmData.EnvironmentVars)
```
aravindhp, line 273: Do we really need a bool here? Don't we always need to restart if no error is returned?

controllers/controllers.go
```diff
@@ -56,6 +57,39 @@ func (r *instanceReconciler) ensureInstanceIsUpToDate(instanceInfo *instance.Inf
...
+			restartCmd := "Restart-Computer -Force"
+			out, err := nc.Windows.Run(restartCmd, true)
+			if err != nil {
+				return fmt.Errorf("'%s' failed with out: %s, err: %w", restartCmd, out, err)
+			}
+
+			if err := drain.RunCordonOrUncordon(drainer, instanceInfo.Node, false); err != nil {
+				return fmt.Errorf("unable to uncordon node %s: %w", instanceInfo.Node.Name, err)
+			}
+			if err := metadata.RemoveRebootAnnotation(context.TODO(), r.client, *instanceInfo.Node); err != nil {
```
aravindhp, line 87: Isn't it possible that WICD could be adding the annotation at the same time given it is async?

pkg/daemon/controller/controller.go
mansikulkarni96: No we don't always have to restart, we only restart when we change a value in the OS registry.
On each reconcile, we do a comparison between the proxy variable value currently set in the Windows OS registry with the value we have specified in the configmap.

controllers/node_controller.go
```diff
@@ -0,0 +1,177 @@
...
+		// Error reading the object - return error to requeue the request.
+		return ctrl.Result{}, err
+	}
+
+	// Create a new signer using the private key that the instances will be configured with
+	r.signer, err = signer.Create(types.NamespacedName{Namespace: r.watchNamespace,
+		Name: secrets.PrivateKeySecret}, r.client)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("unable to create signer from private key secret: %w", err)
+	}
```
sebsoto, line 102: Setting the signer like this is a bit weird.
Instead you can just assign it to a new variable and use it directly in the NewNodeConfig() call

controllers/node_controller.go
```diff
@@ -0,0 +1,177 @@
...
+	if err := drain.RunCordonOrUncordon(drainer, node, true); err != nil {
+		return fmt.Errorf("unable to cordon node %s: %w", node.Name, err)
+	}
+	if err := drain.RunNodeDrain(drainer, node.Name); err != nil {
+		return fmt.Errorf("unable to drain node %s: %w", node.Name, err)
+	}
+
+	if out, err := nc.Windows.Run("Start-Job { Restart-Computer -Force }", true); err != nil {
+		return fmt.Errorf("error restarting the Windows instance with output %s: %w", out, err)
+	}
```
sebsoto, line 135: We try not to use powershell directly outside of the windows package. Please abstract this by creating a `Reboot` method for Windows.

controllers/node_controller.go
```diff
@@ -0,0 +1,177 @@
...
+		return fmt.Errorf("failed to create new nodeconfig: %w", err)
+	}
+
+	drainer := nc.NewDrainHelper()
+	if err := drain.RunCordonOrUncordon(drainer, node, true); err != nil {
+		return fmt.Errorf("unable to cordon node %s: %w", node.Name, err)
+	}
+	if err := drain.RunNodeDrain(drainer, node.Name); err != nil {
+		return fmt.Errorf("unable to drain node %s: %w", node.Name, err)
+	}
```
sebsoto, line 131: Instead of exporting NewDrainHelper, what do you think about keeping the drain logic contained within the Nodeconfig package, and moving `SafeReboot` to be a method of Nodeconfig?

pkg/daemon/controller/controller_test.go
```diff
@@ -654,9 +654,15 @@ func TestReconcile(t *testing.T) {
 
 			winSvcMgr := fake.NewTestMgr(test.existingServices)
 			c, err := NewServiceController(context.TODO(), "node", wmcoNamespace, Options{
-				Client:    clientfake.NewClientBuilder().WithObjects(clusterObjs...).Build(),
-				Mgr:       winSvcMgr,
-				cmdRunner: &fakePSCmdRunner{},
+				Client: clientfake.NewClientBuilder().WithObjects(clusterObjs...).Build(),
```
sebsoto, line 657: Please test that the reboot-annotation is being applied to the node.

sebsoto (changes requested):
Thanks for adding the node controller, PR is looking mostly good, just have a few comments.

pkg/daemon/controller/controller.go
```diff
@@ -265,6 +271,75 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
...
+					return false, fmt.Errorf("error running PowerShell command %s with output %s: %w", cmd, out, err)
+				}
+				if strings.TrimSpace(out) != envVars[varName] {
+					stillNeedsReboot = true
+				}
+			}
+			return !stillNeedsReboot, nil
+		})
+	if err != nil {
+		return fmt.Errorf("error waiting for environment vars to get picked up by processes: %w", err)
```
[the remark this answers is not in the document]
saifshaikh48: All processes forked from the default shell. For clarity the poll above checks the env vars in the shell process. Since child processes inherit from the parent processes, and all the processes we care about (containerd, kubelet, etc) are all spawned from the shell, this is a sufficient check

pkg/metadata/metadata.go
```diff
@@ -135,3 +157,20 @@ func WaitForVersionAnnotation(ctx context.Context, c client.Client, nodeName str
...
 	return nil
 }
+
+// WaitForVersionAnnotation waits for the reboot annotation to be cleared from the node
+func WaitForRebootAnnotationRemoval(ctx context.Context, c client.Client, nodeName string) error {
+	node := &core.Node{}
+	err := wait.Poll(retry.Interval, retry.Timeout, func() (bool, error) {
+		err := c.Get(ctx, kubeTypes.NamespacedName{Name: nodeName}, node)
+		if err != nil {
+			return false, err
```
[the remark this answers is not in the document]
saifshaikh48: This is intended. If we can't get the node object, won't we want to fail? We should only retry if we can read the annotation and it is not as expected

pkg/daemon/controller/controller.go
```diff
@@ -265,6 +271,75 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
...
+	}
+	if restartRequired {
+		// Applying the reboot annotation results in an event picked up by WMCO's node controller to reboot the instance
+		if err = metadata.ApplyRebootAnnotation(sc.ctx, sc.client, node); err != nil {
+			return fmt.Errorf("error setting reboot annotation on node %s: %w", sc.nodeName, err)
+		}
+	}
+	// Wait until the environment variables at the process level are set as expected.
+	// This will only be picked up after WMCO reboots the instance
+	err = wait.PollUntilContextTimeout(sc.ctx, 15*time.Second, 2*time.Minute, true,
```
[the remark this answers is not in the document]
saifshaikh48: I see what you're saying, if a restart is required, I will return successfully at the top level Reconcile func.

aravindhp (changes requested):

controllers/node_controller.go
```diff
@@ -0,0 +1,177 @@
...
+// Reconcile is part of the main kubernetes reconciliation loop which reads that state of the cluster for a
+// CertificateSigningRequests object and aims to move the current state of the cluster closer to the desired state.
+func (r *nodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, err error) {
+	_ = r.log.WithValues(NodeController, req.NamespacedName)
+	// Prevent WMCO upgrades while Node objects are being processed
+	if err := condition.MarkAsBusy(r.client, r.watchNamespace, r.recorder, NodeController); err != nil {
+		return ctrl.Result{}, err
+	}
+	defer func() {
+		err = markAsFreeOnSuccess(r.client, r.watchNamespace, r.recorder, NodeController, result.Requeue, err)
```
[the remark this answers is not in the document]
saifshaikh48: markAsFreeOnSuccess does not overwrite the error. please view the function doc and body

controllers/node_controller.go
```diff
@@ -0,0 +1,177 @@
...
+		return fmt.Errorf("error restarting the Windows instance with output %s: %w", out, err)
+	}
+
+	// wait for WICD to remove Reboot annotation
+	if err := metadata.WaitForRebootAnnotationRemoval(ctx, r.client, node.Name); err != nil {
+		return err
+	}
+
+	if err := drain.RunCordonOrUncordon(drainer, node, false); err != nil {
+		return fmt.Errorf("unable to uncordon node %s: %w", node.Name, err)
```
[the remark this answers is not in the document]
saifshaikh48: @sebsoto Could you suggest a way to mitigate? I was not able to find a good way to address your concern.

pkg/daemon/controller/controller_test.go
saifshaikh48: How can i test this in the WICD unit tests? the fake client always returns the node obj as defined in the test, even if the reboot-annotation is "applied" using the fake client during reconcile

sebsoto: are you sure it is being applied? I would think if the reconciler is being created by the same fake client it should be seen in the test, this can be tested by making a change in the unit test which uses the client to apply the annotation, and then checking immediately after if it is being applied

controllers/node_controller.go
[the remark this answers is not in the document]
sebsoto: Oops, yes, `Note that` is what i meant

pkg/metadata/metadata.go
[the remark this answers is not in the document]
sebsoto: No, because there are various reasons why a node object cannot be gotten and in some cases a retry should be done.

pkg/servicescm/servicescm.go
```diff
@@ -158,14 +167,23 @@ func Generate(name, namespace string, data *Data) (*core.ConfigMap, error) {
...
+	if len(dataFromCM) < 2 || len(dataFromCM) > 3 {
+		return nil, fmt.Errorf("services ConfigMap can only have 2 or 3 keys")
```
aravindhp, line 186:
```suggestion
		return nil, fmt.Errorf("services ConfigMap can only have the required services and files key in addition to an optional environmentVars key")
```

pkg/servicescm/servicescm.go
```diff
@@ -158,14 +167,23 @@ func Generate(name, namespace string, data *Data) (*core.ConfigMap, error) {
 	}
 	servicesConfigMap.Data[filesKey] = string(jsonFiles)
 
+	if len(data.EnvironmentVars) > 0 {
+		jsonEnvVars, err := json.Marshal(data.EnvironmentVars)
```
aravindhp, line 171: Do we need to sanity check the `EnvironmentVars` data? Is it possible for the user try and sneak in a custom variable that WICD will end up setting on the host? I realize WMCO will revert the ConfigMap to the expected state but there could be a Window  where WICD ends up adding the environment variable. It is not as if we will end up removing the added variable in that case.

pkg/services/services.go
aravindhp, line 1: How do we handle removing of an environment variable? i.e. say proxy has been enabled on a cluster and now the user disables it. How are the proxy variables removed?

controllers/node_controller.go
aravindhp, line 1: The introduction of the node controller needs to be a separate commit. We need to clearly call that out.

controllers/controllers.go
aravindhp, line 1: Please introduce `isWindowsNode()` in this commit.

aravindhp (changes requested):

pkg/servicescm/servicescm.go
saifshaikh48: Hm, the services CM is immutable so sneaking this in would be tough

aravindhp: Yes I am aware it is immutable but the user could delete and create one with the same name but different content.

pkg/services/services.go
mansikulkarni96: After the user disables proxy which means removing the proxy vars from the cluster proxy object, OLM will reflect those to WMCO container environment. The controller will pick up the changes and set the values to empty on the instance.

aravindhp: Will the environment variables will be removed from the WMCO pod? Or will it be replaced with blanks values?

saifshaikh48: They are removed from the WMCO pod. There is a followup story to delete the env vars entries from the windows registry rather than setting them to empty on the instances.

pkg/servicescm/servicescm.go
saifshaikh48: Anyway, the scenario you described will never occur as WICD only sets variables that are in the SupportedProxyVars set. It does not iterate over all env vars present in the services CM

pkg/daemon/controller/controller_test.go
saifshaikh48: turns out it was being applied I just had a bogus test... fixed now

pkg/services/services.go
aravindhp: That is [WINC-998](https://issues.redhat.com/browse/WINC-998). We need to ensure that it is not just specific to deconfiguration though. Please update that story to make it more generic i.e. WICD is able to remove the variables it added.

controllers/controllers.go
```diff
@@ -197,24 +197,21 @@ func windowsNodeVersionChangePredicate() predicate.Funcs {
 	return predicate.Funcs{
 		CreateFunc: func(e event.CreateEvent) bool {
 			// catch Machine-backed Windows node upgrades as they are re-created
-			return e.Object.GetLabels()[core.LabelOSStable] == "windows" &&
-				e.Object.GetAnnotations()[metadata.VersionAnnotation] == version.Get()
+			return isWindowsNode(e.Object) && e.Object.GetAnnotations()[metadata.VersionAnnotation] == version.Get()
```
aravindhp, line 200: While it is fine to introduce `isWindowsNode()` in this method, updating all the places where it can be used should be a separate commit. I would have suggested a separate PR but I know @sebsoto requested the change be made here and in my opinion it is out of scope but can live with it if it is a separate commit.

pkg/servicescm/servicescm.go
aravindhp: That is incorrect now that you have stated it explicitly. WICD should be blindly following what is set in the ConfigMap. WMCO should control what goes into the ConfigMap. IOW, WMCO sets the "policy" and WICD implements it and that boundary should be maintained.

pkg/daemon/controller/controller.go
```diff
@@ -265,6 +275,78 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
...
+			return !stillNeedsReboot, nil
+		})
+	if err != nil {
+		return true, fmt.Errorf("error waiting for environment vars to get picked up by processes: %w", err)
+	}
+	// Remove the reboot annotation after we know the reboot occurred successfully. No-op if already not present
+	return false, metadata.RemoveRebootAnnotation(sc.ctx, sc.client, node)
+}
+
+// ensureVarsAreUpToDate ensures that the proxy environment variables are set as expected on the instance
```
aravindhp, line 316: `ensureVarsAreUpToDate` should be generic and not specific to proxy environment variables.

aravindhp (changes requested):

pkg/services/services.go
sebsoto: In the meantime this probably should be documented, if a user wants to stop using proxy they need to manually change the env variables on the VM and restart it.

sebsoto: Alternatively WMCO should always set the proxy env variables to an empty value in the CM if they are not present. Unless thats already happening.

controllers/controllers.go
sebsoto: separate commit is my preference

pkg/daemon/controller/controller.go
saifshaikh48: created [WINC-1061](https://issues.redhat.com/browse/WINC-1061) for this

pkg/daemon/controller/controller.go
```diff
@@ -250,6 +252,14 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
 		return ctrl.Result{}, err
 	}
 
+	awaitingRestart, err := sc.reconcileEnvironmentVariables(cmData.EnvironmentVars, node)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if awaitingRestart {
+		// if a restart is required, there is nothing more the controller can do until the instance is rebooted
```
sebsoto, line 260: A `waiting for reboot` log message might be useful here

sebsoto (changes requested):

pkg/services/services.go
saifshaikh48: > if a user wants to stop using proxy they need to manually change the env variables on the VM and restart it

To be clear, the proxy values are cleared from the instance with current implementation, meaning if a user or application polled for any of the variables, they would get an _empty value_ back. The story above is to delete the key entirely, meaning if someone polled for a proxy variable, they'd get a _ErrNotExist_ from the registry. There is no need to document anything IMO

controllers/node_controller.go
aravindhp: Was this addressed?

aravindhp (approved):
/lgtm

saifshaikh48:
/retest-required

cluster spinup fail in GCP
csr flake in vSphere
ran e2e with proxy locally and it passed

saifshaikh48:
/retest-required

Trying again as it passed 2x locally with proxy enabled in cluster. Strange node log collection failures in CI for only some services, other services are collected with no issue.

`unable to retrieve log hybrid-overlay/hybrid-overlay.log from node winhost-9wzn3: oc adm node-logs failed with exit code exit status 1 and output: : error: error trying to reach service: dial tcp 192.168.99.22:10250: i/o timeout`

saifshaikh48:
/hold

to investigate node log failures

controllers/node_controller.go
```diff
@@ -0,0 +1,177 @@
...
+			watchNamespace:     watchNamespace,
+			recorder:           mgr.GetEventRecorderFor(NodeController),
+		},
+	}, nil
+}
+
+// Reconcile is part of the main kubernetes reconciliation loop which reads that state of the cluster for a
+// CertificateSigningRequests object and aims to move the current state of the cluster closer to the desired state.
+func (r *nodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, err error) {
+	_ = r.log.WithValues(NodeController, req.NamespacedName)
```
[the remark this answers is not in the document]
aravindhp: I assume you will revert this when squashing.

controllers/node_controller.go
```diff
@@ -111,9 +110,11 @@ func (r *nodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 			return ctrl.Result{}, fmt.Errorf("failed to create new nodeconfig: %w", err)
 		}
 
+		r.log.Info("rebooting instance")
```
aravindhp, line 113: This shows up as:
```
INFO	controllers.node	rebooting instance
```
I suggest moving the log into `RebootAndReinitialize()` so that more context is added plus we know for sure that the reboot will be attempted.

controllers/node_controller.go
```diff
@@ -111,9 +110,11 @@ func (r *nodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
 		if err := nc.SafeReboot(ctx, node); err != nil {
 			return ctrl.Result{}, fmt.Errorf("full instance reboot failed: %w", err)
 		}
+		r.log.V(1).Info("successful reboot", "instance", instanceInfo.Address)
```
aravindhp, line 117: You can move this log also to `RebootAndReinitialize()` right after we reinitialize SSH.

aravindhp (changes requested):

pkg/daemon/controller/controller.go
```diff
@@ -265,6 +276,79 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
...
+			// we cannot use os.Setenv, which is a wrapper for syscall.SetEnvironmentVariable
+			// As per Microsoft docs: "Calling SetEnvironmentVariable has no effect on the system environment variables"
+			err := registryKey.SetStringValue(key, expectedVal)
+			if err != nil {
+				// Do not log value as proxy information is sensitive
+				return false, fmt.Errorf("unable to set environment variable %s: %w", key, err)
+			}
+			restartRequired = true
+		}
+		if err := registryKey.Close(); err != nil {
```
aravindhp, line 345: Don't you need to always close this key in case of errors above?

saifshaikh48: Yes, it should always be closed. We had changed it based on [this comment](https://github.com/openshift/windows-machine-config-operator/pull/1536#discussion_r1223540487), but now I am leaning towards doing back to `defer registryKey.Close()`

aravindhp: You can go back to `defer registryKey.Close()` but you need to catch the error and not just ignore it. I thought the PR was updated to do that correctly but moved back.

saifshaikh48: Sorry I don't understand the ask. How should we catch/return the error in a defer call? Whatever is returned from a defer func will not affect the function it is called from, ex: https://go.dev/play/p/bXAnQkdk517

saifshaikh48: The [sys/windows/registry package](https://pkg.go.dev/golang.org/x/sys/windows/registry) itself recommends just deferring the close
```
key, err := registry.OpenKey(...)
if err != nil {
    return err
}
defer key.Close()
```

pkg/daemon/controller/controller.go
```diff
@@ -265,6 +276,86 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
...
+		registryKey, err := registry.OpenKey(registry.LOCAL_MACHINE, systemEnvVarRegistryPath, registry.ALL_ACCESS)
+		if err != nil {
+			return false, fmt.Errorf("unable to open Windows system registry key %s: %w", systemEnvVarRegistryPath, err)
+		}
+		defer func() { // always close the registry key, without swallowing any error returned before the defer call
+			closeErr := registryKey.Close()
+			if closeErr != nil {
+				if err != nil {
+					err = fmt.Errorf("multiple errors: %w, %w", err, closeErr)
+				}
```
aravindhp, line 333: Don't you need an `else` here?

pkg/daemon/controller/controller.go
```diff
@@ -265,6 +276,86 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
...
+			// As per Microsoft docs: "Calling SetEnvironmentVariable has no effect on the system environment variables"
+			err := registryKey.SetStringValue(key, expectedVal)
+			if err != nil {
+				// Do not log value as proxy information is sensitive
+				return false, fmt.Errorf("unable to set environment variable %s: %w", key, err)
+			}
+			restartRequired = true
+		}
+	}
+	return restartRequired, nil
```
aravindhp, line 356: Shouldn't this be `return restartRequired, err`?

saifshaikh48: The err value should be `nil` as this is the happy path. By the time we get here, if there was a non-nil error of note, we'd have returned already

mansikulkarni96: @saifshaikh48 we only set the error in the defer function but where do we return it, if we don't return it here?

saifshaikh48: When the overarching function (ensureVarsAreUpToDate) returns, its second return value is stored in the variable `err`. Then the defer function can access and modify it. And finally, after the defer function runs, whatever value is in `err` is returned

aravindhp: Did you confirm this in a go playground?

saifshaikh48: yes, https://go.dev/play/p/JTCvFeEsYrs

aravindhp (approved):

controllers/node_controller.go
```diff
@@ -88,6 +93,27 @@ func (r *nodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		nc, err := nodeconfig.NewNodeConfig(r.client, r.k8sclientset, r.clusterServiceCIDR, r.watchNamespace,
+			instanceInfo, signer, nil, nil, r.platform)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to create new nodeconfig: %w", err)
+		}
+
+		if err := nc.SafeReboot(ctx, node); err != nil {
```
sebsoto, line 113: I'm realizing that passing in node here doesn't make much sense. Considering `Nodeconfig` has node as a part of its struct. Not asking for a change here, this isn't a pattern you're introducing, just something i'm realizing now.

pkg/daemon/controller/controller.go
```diff
@@ -265,6 +276,86 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
...
+			for _, varName := range cluster.SupportedProxyVars {
+				cmd := fmt.Sprintf("[Environment]::GetEnvironmentVariable('%s', 'Process')", varName)
+				out, err := sc.psCmdRunner.Run(cmd)
+				if err != nil {
+					return false, fmt.Errorf("error running PowerShell command %s with output %s: %w", cmd, out, err)
+				}
+				if strings.TrimSpace(out) != envVars[varName] {
+					stillNeedsReboot = true
+				}
+			}
```
sebsoto, line 307: nit: this logic could be contained within its own function which would help readability of this larger function.

pkg/daemon/controller/controller.go
```diff
@@ -265,6 +276,86 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
...
+		defer func() { // always close the registry key, without swallowing any error returned before the defer call
+			closeErr := registryKey.Close()
+			if closeErr != nil {
+				if err != nil {
+					err = fmt.Errorf("multiple errors: %w, %w", err, closeErr)
+				} else {
+					err = closeErr
+				}
+			}
+		}()
```
sebsoto, line 336: Why does this need to be done this way and not just logged?
In case of this error, the handle is being leaked no matter what, triggering another reconcile is not going to fix that.

sebsoto (changes requested):

pkg/daemon/controller/controller.go
sebsoto: Taking the straightforward approach and just logging it shouldn't contradict @aravindhp's comment. The error is still being noticed.
https://github.com/openshift/windows-machine-config-operator/pull/1536#discussion_r1258583602

saifshaikh48: > another reconcile is not going to fix that

Do we know this for sure? It could be that next time we reconcile the handle can successfully closes

controllers/node_controller.go
aravindhp: @sebsoto thanks for catching this bad pattern. If we are doing this in other places, we should fix it sooner otherwise we will keep copying the same bad pattern.

aravindhp: To be clear. It doesn't need to be fixed as part of this PR.

pkg/daemon/controller/controller.go
aravindhp: Good point @sebsoto. Ideally there would be `waitForEnvironmentVariablesToReconcile()` function that calls the helper you are talking about.

PR #1541, opened 2023-04-04
[test] Move storage testing to its own test

sebsoto:
Instead of using the network tests to indirectly test storage, move it to a separate test. This ensures that neither test is compromised by the requirements of the other. As storage testing uses the same container image, this should not add much to the testing time, since no images need to be pulled.

### DIFF
--- a/test/e2e/storage_test.go
+++ b/test/e2e/storage_test.go
@@ -1,0 +1,45 @@
+package e2e
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	core "k8s.io/api/core/v1"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// testStorage tests that persistent volumes can be accessed by Windows pods
+func testStorage(t *testing.T) {
+	tc, err := NewTestContext()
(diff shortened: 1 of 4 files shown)


saifshaikh48 (commented):
/lgtm

test/e2e/storage_test.go
```diff
@@ -0,0 +1,44 @@
...
+		t.Skip("storage is not supported on this platform")
+	}
+	err = tc.waitForWindowsNodes(gc.numberOfMachineNodes, false, false, false)
+	require.NoError(t, err, "timed out waiting for Windows Machine nodes")
+	err = tc.waitForWindowsNodes(gc.numberOfBYOHNodes, false, false, true)
+	require.NoError(t, err, "timed out waiting for BYOH Windows nodes")
+	require.Greater(t, len(gc.allNodes()), 0, "test requires at least one Windows node to run")
+
+	// Create the PVC and choose the node the deployment will be scheduled on. This is necessary as ReadWriteOnly
+	// volumes can only be bound to a single node.
```
aravindhp, line 27: Please link to the documentation that states this

aravindhp (approved):
Thanks for reworking this @sebsoto. This is much cleaner.

aravindhp (approved):
/lgtm

PR #1547, opened 2023-04-10
[release-4.13] WINC-1014: Enable vSphere in-tree storage

sebsoto:
Enables the use of in-tree storage for Windows nodes on vSphere.

Includes:
* A kubelet update to capture https://github.com/openshift/kubernetes/pull/1514, which allows the kubelet feature vSphereCSIMigration to be disabled.
* Changes to WMCO to disable the feature flag on Windows nodes, allowing the use of in-tree storage.
* e2e testing, cherry-picked from release-4.12


pkg/ignition/ignition.go
```diff
@@ -84,6 +84,15 @@ func (ign *Ignition) GetKubeletArgs() (map[string]string, error) {
 	if err != nil {
 		return nil, fmt.Errorf("error parsing kubelet systemd unit args: %w", err)
 	}
+	// Special casing for vSphere to disable CSI Migration. These args have the potential to be either set to this, or
+	// not, depending on if the cluster was upgraded from 4.12 or if it is a fresh 4.13 install.
+	// See: https://github.com/openshift/machine-config-operator/pull/3655
+	// Because in-tree storage is the only option for Windows in 4.13 vSphere clusters, these values should always be
+	// set to this if the cluster platform is vSphere.
```
aravindhp, line 91: Shouldn't we be looking at the storage CR like in https://github.com/openshift/machine-config-operator/pull/3655 instead always defaulting to using the cloud-provider option?

test/e2e/providers/vsphere/vsphere.go
```diff
@@ -133,9 +139,50 @@ func (p *Provider) GetType() config.PlatformType {
...
+	sc, err := p.ensureStorageClass(client)
+	if err != nil {
+		return nil, fmt.Errorf("unable to ensure a usable StorageClass is created: %w", err)
+	}
+	pvcSpec := core.PersistentVolumeClaim{
+		ObjectMeta: meta.ObjectMeta{
+			GenerateName: "e2e-",
+		},
+		Spec: core.PersistentVolumeClaimSpec{
+			AccessModes: []core.PersistentVolumeAccessMode{core.ReadWriteOnce},
```
aravindhp, line 158: Shouldn't we be testing `ReadWriteMany` as we have had requests for that?

test/e2e/providers/vsphere/vsphere.go
```diff
@@ -133,9 +139,50 @@ func (p *Provider) GetType() config.PlatformType {
...
+			Resources: core.ResourceRequirements{
+				Requests: core.ResourceList{core.ResourceStorage: resource.MustParse("2Gi")},
```
aravindhp, line 160: Why is this needed?

test/e2e/providers/vsphere/vsphere.go
```diff
@@ -10,14 +10,20 @@ import (
...
 	client "k8s.io/client-go/kubernetes"
 
 	"github.com/openshift/windows-machine-config-operator/test/e2e/clusterinfo"
 	"github.com/openshift/windows-machine-config-operator/test/e2e/providers/machineset"
 )
 
-const defaultCredentialsSecretName = "vsphere-cloud-credentials"
+const (
+	defaultCredentialsSecretName = "vsphere-cloud-credentials"
+	storageClassName             = "e2e"
```
aravindhp, line 25: Please give a more appropriate name like `ntfs`.

test/e2e/providers/vsphere/vsphere.go
```diff
@@ -133,9 +139,50 @@ func (p *Provider) GetType() config.PlatformType {
...
+func (p *Provider) CreatePVC(client client.Interface, namespace string) (*core.PersistentVolumeClaim, error) {
+	// Use a StorageClass to allow for dynamic volume provisioning
+	// https://docs.openshift.com/container-platform/4.12/storage/dynamic-provisioning.html#about_dynamic-provisioning
+	sc, err := p.ensureStorageClass(client)
+	if err != nil {
+		return nil, fmt.Errorf("unable to ensure a usable StorageClass is created: %w", err)
+	}
+	pvcSpec := core.PersistentVolumeClaim{
+		ObjectMeta: meta.ObjectMeta{
+			GenerateName: "e2e-",
```
aravindhp, line 155: Please use the constant here.

aravindhp (changes requested):

pkg/ignition/ignition.go
jsafrane: IMO even when user opts-in for CSI migration on Linux nodes, Windows nodes should still use in-tree. Or are you going to require CSI driver to be installed on the Windows nodes too?

test/e2e/providers/vsphere/vsphere.go
sebsoto: ReadWriteMany requires the vSphere env to support it. VMC does not support it.
Potentially we could try to determine if the workload is running in an environment that does support it, and use it in that case. I'm not sure how to do that, and would prefer to not have to add that extra work into this PR which is a release blocker.

test/e2e/providers/vsphere/vsphere.go
sebsoto: This is requesting a 2GB slice of storage, its an arbitrary value.

pkg/ignition/ignition.go
sebsoto: @aravindhp if that was done we would have no way to support Windows storage for fresh 4.13 cluster installs in 8.0.0.

test/e2e/providers/vsphere/vsphere.go
sebsoto: i'm adding this and the other changes into a separate commit so they can be cherry-picked into release-4.12 to keep the code in sync.

test/e2e/providers/vsphere/vsphere.go
aravindhp: I am fine not doing that in this PR but we need to track getting this to work in VMC. Maybe a SPLAT ticket? We don't want to wait for a customer issue to be opened to enable this in VMC.

test/e2e/providers/vsphere/vsphere.go
aravindhp: I see this is from a cherry-pick. Sorry for the noise. You can fix this in the separate PR you are going to open to add the comment.

aravindhp (approved):

saifshaikh48 (commented):
/lgtm

PR #1550, opened 2023-04-11
WIP: [release-4.12][e2e][vSphere] Use IPv6-disabled VM template

saifshaikh48:
This PR is just to test the new golden image until we can get access to the IBM vCenter environment. 
The goal is to ensure Windows worker nodes are never assigned IPv6 addresses.


mtnbikenc:
/close
In favor of #1558

PR #1555, opened 2023-04-12
WINC-561: Set kubelet priority flag

alinaryan:
Uses the kubelet priority flag introduced in https://github.com/kubernetes/kubernetes/pull/96051 to set the kubelet priority class to ABOVE_NORMAL_PRIORITY_CLASS to ensure that running Pods do not starve the kubelet of CPU cycles.

### DIFF
--- a/pkg/services/services.go
+++ b/pkg/services/services.go
@@ -203,6 +203,7 @@ func generateKubeletArgs(argsFromIgnition map[string]string, debug bool) ([]stri
 	containerdEndpointValue := "npipe://./pipe/containerd-containerd"
 	certDirectory := "c:\\var\\lib\\kubelet\\pki\\"
 	windowsTaints := "os=Windows:NoSchedule"
+	windowsPriorityClass := "ABOVE_NORMAL_PRIORITY_CLASS"
 	// TODO: Removal of deprecated flags to be done in https://issues.redhat.com/browse/WINC-924
 	kubeletArgs := []string{
 		"--config=" + windows.KubeletConfigPath,
@@ -217,6 +218,10 @@ func generateKubeletArgs(argsFromIgnition map[string]string, debug bool) ([]stri
 		"--container-runtime=remote",
 		"--container-runtime-endpoint=" + containerdEndpointValue,
 		"--resolv-conf=",
+		// Allows the kubelet process to get more CPU time slices when compared to other processes running on the
+		// Windows host.
+		// See: https://kubernetes.io/docs/concepts/configuration/windows-resource-management/#resource-management-cpu
+		"--windows-priorityclass=" + windowsPriorityClass,
 	}
 
 	kubeletArgs = append(kubeletArgs, klogVerbosityArg(debug))
--- a/test/e2e/create_test.go
+++ b/test/e2e/create_test.go
@@ -59,6 +59,7 @@ func creationTestSuite(t *testing.T) {
 	t.Run("Node Logs", testNodeLogs)
 	t.Run("Metrics validation", testMetrics)
 	t.Run("UserData validation", testUserData)
+	t.Run("Kubelet priority class validation", testKubeletPriorityClass)
 }
 
 // testWindowsNodeCreation tests the Windows node creation in the cluster
--- a/test/e2e/validation_test.go
+++ b/test/e2e/validation_test.go
@@ -46,6 +46,37 @@ type winService struct {
 	description string
 }
 
+// testKubeletPriorityClass tests if kubelet priority class is set to "AboveNormal"
+func testKubeletPriorityClass(t *testing.T) {
+	tc, err := NewTestContext()
+	require.NoError(t, err)
+	requiredPriorityClass := "AboveNormal"
+
+	require.Greater(t, len(gc.allNodes()), 0, "test requires at least one Windows node to run")
+	for _, node := range gc.allNodes() {
+		t.Run(node.Name, func(t *testing.T) {
+			out, err := tc.getKubeletPriorityClass(&node)
+			require.NoError(t, err, "error getting kubelet priority class")
+			assert.Containsf(t, out, requiredPriorityClass, "node %s missing required kubelet priority class",
+				node.GetName())
+		})
+	}
+}
+
+// getKubeletPriorityClass returns the priority class of the kubelet service
+func (tc *testContext) getKubeletPriorityClass(node *core.Node) (string, error) {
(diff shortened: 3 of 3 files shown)


alinaryan:
Confirmed process priority was set appropriately on a Windows node using this WMCO image 

Screenshot from 2023-04-13 14-47-33

aravindhp:
> Confirmed process priority was set appropriately on a Windows node using this WMCO image
> 
> Screenshot from 2023-04-13 14-47-33

Shouldn't this be an e2e test?

pkg/services/services.go
```diff
@@ -203,6 +203,7 @@ func generateKubeletArgs(argsFromIgnition map[string]string, debug bool) ([]stri
 	containerdEndpointValue := "npipe://./pipe/containerd-containerd"
 	certDirectory := "c:\\var\\lib\\kubelet\\pki\\"
 	windowsTaints := "os=Windows:NoSchedule"
+	windowsPriorityClass := "ABOVE_NORMAL_PRIORITY_CLASS"
```
aravindhp, line 206: Please link to the doc about Windows priority classes here.

aravindhp (changes requested):
Thanks for working on this, @alinaryan. Like I mentioned in the comment, this requires an e2e test.

test/e2e/validation_test.go
```diff
@@ -46,6 +46,39 @@ type winService struct {
 	description string
 }
 
+// testKubeletPriority tests if kubelet priority class is set to "AboveNormal"
+func testKubeletPriority(t *testing.T) {
+	tc, err := NewTestContext()
+	require.NoError(t, err)
+	requiredPriorityClass := "AboveNormal"
+
+	t.Run("nodes", func(t *testing.T) {
```
saifshaikh48, line 55: consider moving this into the for loop so a separate test is executed for each node (and you can name each test with the node name)

test/e2e/validation_test.go
```diff
@@ -46,6 +46,39 @@ type winService struct {
...
+		if len(gc.allNodes()) == 0 {
+			t.Skip("0 expected nodes")
+		}
+		for _, node := range gc.allNodes() {
+			out, err := tc.getKubeletPriorityClass(&node)
+			require.NoError(t, err, "error getting kubelet priority class")
+			require.Containsf(t, out, requiredPriorityClass, "node %s missing required kubelet priority class",
```
saifshaikh48, line 62: This should be an assert. require will prevent other tests from running after, if something were to fail

test/e2e/validation_test.go
```diff
@@ -46,6 +46,39 @@ type winService struct {
...
+}
+
+// getKubeletPriorityClass returns the PriorityClass of the kubelet service
+func (tc *testContext) getKubeletPriorityClass(node *core.Node) (string, error) {
+	command := "Get-Process kubelet | Select-Object PriorityClass"
+	addr, err := controllers.GetAddress(node.Status.Addresses)
+	if err != nil {
+		return "", fmt.Errorf("error getting node address: %w", err)
+	}
+	out, err := tc.runPowerShellSSHJob("kubelet-query", command, addr)
```
sebsoto, line 75: Please dont name this the same as other jobs in the tests. That makes figuring out the correct pod log to check hard.

lets be specific here, something like "kubelet-priority-class-query"

sebsoto (changes requested):

test/e2e/validation_test.go
```diff
@@ -46,6 +46,39 @@ type winService struct {
...
+func testKubeletPriority(t *testing.T) {
+	tc, err := NewTestContext()
+	require.NoError(t, err)
+	requiredPriorityClass := "AboveNormal"
+
+	if len(gc.allNodes()) == 0 {
+		t.Skip("0 expected nodes")
+	}
+	for _, node := range gc.allNodes() {
+		t.Run("nodes", func(t *testing.T) {
```
saifshaikh48, line 59: > you can name each test with the node name

e.g. `node.Name` instead of `"nodes"`

test/e2e/validation_test.go
```diff
@@ -46,6 +46,39 @@ type winService struct {
...
 
+// testKubeletPriority tests if kubelet priority class is set to "AboveNormal"
+func testKubeletPriority(t *testing.T) {
+	tc, err := NewTestContext()
+	require.NoError(t, err)
+	requiredPriorityClass := "AboveNormal"
+
+	if len(gc.allNodes()) == 0 {
+		t.Skip("0 expected nodes")
+	}
```
saifshaikh48, line 57: Should we skip in this case? Elsewhere in the test suite we `require` > 0 nodes
e.g. 
- require.Greater(t, len(gc.allNodes()), 0, "test requires at least one Windows node to run")
- require.NotEmpty(t, gc.allNodes())

mansikulkarni96:
LGTM

test/e2e/create_test.go
```diff
@@ -59,6 +59,7 @@ func creationTestSuite(t *testing.T) {
 	t.Run("Node Logs", testNodeLogs)
 	t.Run("Metrics validation", testMetrics)
 	t.Run("UserData validation", testUserData)
+	t.Run("Kubelet validation", testKubeletPriority)
```
aravindhp, line 62: Please rename to `Kubelet priority class validation`.

test/e2e/validation_test.go
```diff
@@ -46,6 +46,37 @@ type winService struct {
 	description string
 }
 
+// testKubeletPriority tests if kubelet priority class is set to "AboveNormal"
+func testKubeletPriority(t *testing.T) {
```
aravindhp, line 50: Please rename to `testKubeletPriorityClass`

aravindhp (commented):
Nice work, @alinaryan. Just a couple of name change requests.

aravindhp (commented):
/lgtm

sebsoto (approved):

alinaryan:
/test azure-e2e-operator

cpu usage err

`error: Query '${t_install} cluster:usage:cpu:install:seconds:quantile    label_replace(quantile_over_time(.95,sum(irate(container_cpu_usage_seconds_total{id="/"}[90s:30s]))[${d_all}:${d_test}]),"quantile","0.95","","")' was not valid: d_test: unbound variable`

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/1555/pull-ci-openshift-windows-machine-config-operator-master-azure-e2e-operator/1651291550447046656

PR #1566, opened 2023-04-17
WINC-860: [hack] Automate pre-release version updates

wgahnagl:
There are many places in a number of files where the versions of the WMCO and OCP need to be updated for release. These should be automated so it no longer has to be done manually.

### DIFF
--- a/hack/common.sh
+++ b/hack/common.sh
@@ -33,6 +33,93 @@ get_OCP_version() {
   echo $OCP_VER_MAJOR.$OCP_VER_MINOR
 }
 
+check_git_branch() {
+  # Return if the working branch is dirty
+  if ! git diff --quiet; then
+    echo "branch dirty, exiting to not overwrite work"
+    return 1
+  fi
+}
+
+# This function creates a git branch for automating commits
+# Takes n arguments
+# 1st arg: the branch name prefix, such as submodule-update
+# args [2, n]: the versions to branch from
+# usage: create_git_branch_list submodule-update release-4.10 release-4.12 release-4.13
+create_git_branch_list() {
+  if [ "$#" -lt 2 ]; then
+    echo incorrect parameter count for create_git_branch_list $#
+    return 1
+  fi
+
+  branch_name_prefix=$1
+  new_branch_list=()
+
+  # skip the first argument because that's the branch ID
+  shift
+  for branch in "$@"; do
+    new_branch="$branch-$branch_name_prefix-$(date +%m-%d)"
+    new_branch_list+=($new_branch)
+  done
+  echo "${new_branch_list[@]}"
+}
+
+# This function switches the current git branch, and deletes branches with duplicate names.
+# takes 2 arguments
+# the base branch, and the branch that is going to be created
+# usage: switch_git_branch master new_branch_name
+switch_git_branch() {
+  if [ "$#" -lt 2 ]; then
+    echo incorrect parameter count for switch_git_branch $#
+    return 1
+  fi
+  local base_branch=$1
+  local new_branch=$2
+
+  if git branch -D $new_branch; then
+    echo "Deleted branch $new_branch"
+  fi
+  git checkout $base_branch -b $new_branch
+} 
+
+# This function adds files and generates a commit message 
+# takes n arguments
+# 1st arg: the commit message to add to the commits 
+# args [2, n]: the directories to add to the commit 
+# usage: generate_commit "good commit message" hack bundle 
+generate_commit() {
+  if [ "$#" -lt 2 ]; then
+    echo incorrect parameter count for generate_commit $#
+    return 1
+  fi
+
+  local commit_message="$1"
+  shift
+  for directory in "$@"; do
+    git add $directory
+  done
+
+  # Commit changes if there are any
+  git commit -m "$commit_message"
+}
+
+# This function displays a message to remind the user to push their changes. 
+# takes 1 arg
+# 1st arg: the list of branches to display 
+# usage: display_push_message ${new_branch_list[@]}
+display_push_message() {
+  new_branch_list=("$@")
+  echo "****"
+  echo "New branches created:"
+  echo "    [${new_branch_list[@]}]"
+  echo ""
+  echo "For each branch, you may push to your fork and create a PR against openshift/windows-machine-config-operator"
+  echo "example:"
+  echo "# assumes you have your fork set as the remote 'origin'"
+  echo "git push origin ${new_branch_list[0]}"
+
+}
+
 get_operator_sdk() {
   # Download the operator-sdk binary only if it is not already available
   # We do not validate the version of operator-sdk if it is available already
--- a/hack/pre-release.sh
+++ b/hack/pre-release.sh
@@ -1,0 +1,200 @@
+#!/bin/bash
+set -euo pipefail
+# this script updates the WMCO and OCP versions in the Makefile, CSV, bundle and Dockerfiles.
+
+# Include the common.sh 
+WMCO_ROOT=$(dirname "${BASH_SOURCE}")/..
+source $WMCO_ROOT/hack/common.sh
+
+CVP=false
+BASEIMAGE=false
+
+usage() {
+  echo "usage:"
+  echo "input the name of the branch you want to increment, and the x y or z value to bump"
+  echo "example: ./pre-release.sh master x"
+  echo "this will bump the x stream of master by 1"
+  echo "optional flags: -c"
+  echo "  bumps the versions held back by CVP"
+  echo "optional flags: -b"
+  echo "  bumps the baseimage versions"
+}
+
+# This function updates the WMCO version in all relevant files 
+# takes 2 arguments 
+# 1st arg: stream to update. Can be x, y or z to represent the x y or z stream. 
+# 2nd arg: WMCO version to update.
+update_WMCO_version() { 
+  if [ "$#" -ne 2 ]; then
+    echo incorrect parameter count for update_WMCO_version $#
+    return 1
+  fi
+
+  xyz_stream=$1
+  version=$2 
+  cluster_service_version_yaml=config/manifests/bases/windows-machine-config-operator.clusterserviceversion.yaml
+
+  # Setting the stream var to 1 2 or 3 which is used for incrementing the correct stream
+  # this tells awk which column to act on. Column 1, column 2, or column 3.
+  # which correspond to the x, y and z stream respectively. 
+  if [[ $xyz_stream == "x" ]]; then
+    stream_var=1
+  fi
+  if [[ $xyz_stream == "y" ]]; then
+    stream_var=2
+  fi
+  if [[ $xyz_stream == "z" ]]; then
+    stream_var=3
+  fi
+ 
+  # Increment the version
+  updated_version=$(echo $version |
+    awk -F. -v OFS=. -v stream_var="$stream_var" '{$stream_var += 1 ; print}')
+
+  last_version=$(grep "olm.skipRange" $cluster_service_version_yaml |
+    grep -oP ">=\K([0-9]+\.[0-9]+\.[0-9]+)\s*<" | sed "s/\s*<//")
+
+  echo "WMCO version $version detected from the Makefile"
+
+  echo "Updating the Makefile to version $updated_version"
+  # Example: matches *WMCO_VERSION ?= 9.0.0*
+  sed -i "s/WMCO_VERSION ?= $version/WMCO_VERSION ?= $updated_version/g" Makefile
+
+  echo "Updating the skipRange version in $cluster_service_version_yaml to $updated_version"
+  # Example: matches olm.skipRange: '>=8.0.0 *<9.0.0*'
+  sed -i "s/<$version/<$updated_version/g" $cluster_service_version_yaml
+
+  echo "Updating previous skipRange version in $cluster_service_version_yaml to $version"
+  # Example: matches olm.skipRange: '*>=8.0.0* <9.0.0'
+  sed -i "s/>=$last_version/>=$version/g" $cluster_service_version_yaml
+ 
+  echo "Updating bundle/windows-machine-config-operator.package.yaml to $updated_version"
+  # Example: matches   - currentCSV: windows-machine-*config-operator.v9.0.0*
+  sed -i "s/config-operator.v$version/config-operator.v$updated_version/g" bundle/windows-machine-config-operator.package.yaml
+   
+}
+
+# This function updates the OCP version in all relevant files 
+# takes 2 arguments 
+# 1st arg: stream to update. Can be x, y or z to represent the x y or z stream. 
+# 2nd arg: WMCO version to update.
+update_OCP_version() {
+  if [ "$#" -ne 2 ]; then
+    echo incorrect parameter count for update_OCP_version $#
+    return 1
+  fi
+
+  stream=$1
+  wmco_version=$2 
+  cluster_service_version_yaml=config/manifests/bases/windows-machine-config-operator.clusterserviceversion.yaml
+  
+  # stream_var tells awk which column to act on. Column 2 is the OCP y stream 
+  if [[ $stream == "x" ]]; then
+    stream_var=2
+  else 
+    echo ""
+    echo "no need to update OCP version"
+    echo ""
+    return 0
+  fi
+  
+  version=$(get_OCP_version $wmco_version)
+  echo "using OCP version $version"
+
+  # Increment the version by using the stream_var variable to select which column to increment 
+  updated_version=$(echo $version |
+    awk -F. -v OFS=. -v stream_var="$stream_var" '{$stream_var += 1 ; print}')
+
+  # Find the last version in the dockerfiles
+  last_version=$(sed -n 's/.*\([0-9]\.[0-9][0-9]\).*/\1/p' build/Dockerfile)
+
+  echo "Updating $cluster_service_version_yaml version to $updated_version"
+  # Example: matches on any instance of the OCP version, such as 4.14 
+  sed -i "s/$version/$updated_version/g" $cluster_service_version_yaml
+
+  if [[ $CVP == "true" ]]; then  
+    echo "Updating bundle.Dockerfile version to $updated_version"
+    # Example: matches LABEL com.redhat.openshift.versions="*=v4.13*"
+    sed -i "s/=v$version/=v$updated_version/g" bundle.Dockerfile
+
+    echo "Updating bundle.Dockerfile last version to $version"
+    # Example: matches LABEL com.redhat.openshift.versions="=v4.13"
+    sed -i "s/=v$last_version/=v$version/g" bundle.Dockerfile
+
+    echo "Updating bundle/metadata/annotations.yaml version to $updated_version"
+    # Example: matches com.redhat.openshift.versions: "*=v4.13*"
+    sed -i "s/=v$version/=v$updated_version/g" bundle/metadata/annotations.yaml 
+  else
+    echo ""
+    echo "skipping bundle.Dockerfile and bundle/metadata/annotations.yaml due to CVP being a version behind"
+    echo "if these do need to be bumped, rerun with the -c flag from your original branch"
+    echo ""
+  fi 
+
+  if [[ $BASEIMAGE == "true" ]]; then   
+    baseimages=("Dockerfile" "Dockerfile.base" "Dockerfile.ci" "Dockerfile.wmco")
+    for item in "${baseimages[@]}"; do 
+      echo "Updating build/$item to $updated_version"
+      sed -i "s/openshift-$version/openshift-$updated_version/g" build/$item
+    done
+  else
+    echo ""
+    echo "skipping updating base images"
+    echo "if these do need to be bumped, rerun with the -b flag from your original branch"
+    echo ""
+  fi 
+}
+
+# This sets the CVP and baseimage flags
+while getopts "bc" opt; do 
+  case "$opt" in 
+    c) CVP=true;;
+    b) BASEIMAGE=true;;
+  esac 
+done 
+shift $((OPTIND-1))
+
+# If there are less than two arguments, exit
+if [ ${#@} -lt 2 ]; then
+  echo "Invalid number of args"
+  echo ""
+  usage
+  exit 
+fi
+
+# the base branch comes first, and the stream to update comes second 
+base_branch=$1
+stream=$2
+
+if [[ $stream != "x" && $stream != "y" && $stream != "z" ]]; then
+  echo ""
+  echo "Stream argument can only be x, y or z"
+  echo "input stream argument is $stream"
+  echo ""
+  usage
+  exit 
+fi
+
+# check if the git branch is dirty, and exit if it is
+check_git_branch
+
+initial_branch=$(git branch --show-current)
+new_branch=$(create_git_branch_list $stream-stream-bump $base_branch)
+switch_git_branch $base_branch $new_branch
+
+# Locate and store the WMCO version from the makefile 
+# finds "WMCO_VERSION ?=" and selects text following it
+WMCO_VERSION=$(sed -n -e 's/^.*WMCO_VERSION ?= //p' Makefile)
+update_WMCO_version $stream $WMCO_VERSION
+update_OCP_version $stream $WMCO_VERSION 
+make bundle
+
+commit_message="[release] Update $base_branch $stream stream to version
+This commit was generated by hack/pre-release.sh"
+generate_commit "$commit_message" build bundle Makefile config
+
+# Return to the original branch 
+git checkout $initial_branch
+
+# Display the newly created branches with push instructions 
+display_push_message "$new_branch"
--- a/hack/update_submodules.sh
+++ b/hack/update_submodules.sh
@@ -123,30 +123,25 @@ if [ ! -z "$remote_branch" ]; then
   fi
 fi
 
-# Return if the working branch is dirty
-if ! git diff --quiet; then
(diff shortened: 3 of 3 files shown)


hack/pre-release.sh
```diff
@@ -0,0 +1,72 @@
+# this script updates the WMCO and OCP versions in the Makefile, base CSV, bundle and Dockerfiles.
```
aravindhp, line 1: What shell is this targeted at?

Please ensure that the script will exit on the first error.

hack/pre-release.sh
```diff
@@ -0,0 +1,72 @@
...
+
+# usage: 
+# ./pre-release (x, y or z) 
+# to update a z stream, use ./pre-release.sh z
+# for other streams, use the corresponding x y or z argument.
```
aravindhp, line 6: It is unclear how to use this. Do you literally mean `x`, `y` or `z`? Is there an assumption that you will be in a `release-4.x` branch when executing this?

aravindhp (changes requested):
Thanks for working on this, @wgahnagl. Please address my comments.

hack/pre-release.sh
```diff
@@ -0,0 +1,72 @@
...
+  if [[ $xyz_stream == "z" ]]; then
+    stream_var=3
+  fi
+  
+  version=$(sed -n -e 's/^.*WMCO_VERSION ?= //p' Makefile)
+  updated_version=$(echo $version | 
+    awk -F. -v OFS=. -v stream_var="$stream_var" '{$stream_var += 1 ; print}')
+  last_version=$(grep "olm.skipRange" $cluster_service_version_yaml | 
+    grep -oP ">=\K([0-9]+\.[0-9]+\.[0-9]+)\s*<" | sed "s/\s*<//" | 
+    awk -F. -v OFS=. -v stream_var="$stream_var" '{$stream_var += 1 ; print}')
```
aravindhp, line 27: Please add a comment explaining what these statements are doing.

hack/pre-release.sh
```diff
@@ -0,0 +1,72 @@
...
+  cluster_service_version_yaml=config/manifests/bases/windows-machine-config-operator.clusterserviceversion.yaml
+  
+  if [[ $xy_stream = "x" ]]; then 
+    stream_var=2
+  else
+    exit 
+  fi 
+
+  version=$(grep -o -E 'OCP \b([0-9]\.[0-9]{2})\b' $cluster_service_version_yaml | 
+    sed -E 's/OCP ([0-9]\.[0-9]{2})/\1/')
```
aravindhp, line 52: Is there a jq equivalent that you can use?

aravindhp (changes requested):

hack/pre-release.sh
```diff
@@ -0,0 +1,110 @@
...
+
+  echo "Updating Dockerfile.ci version to $version"
+  sed -i "s/$last_version/$version/g" build/Dockerfile.ci
+
+  echo "Updating Dockerfile.wmco version to $version"
+  sed -i "s/$last_version/$version/g" build/Dockerfile.wmco
+}
+
+if [[ -z $1 ]]; then
+  while ! [[ $STREAM =~ ^(x|y|z)$ ]]
```
aravindhp, line 100: Are you providing an input if the user does not specify `x`, `y` or `z`? Do we really need this sort of nagging rather than just erroring out?

aravindhp (changes requested):
@wgahnagl please add a comment with examples in places where you are using regexes. I will defer to @saifshaikh48 @alinaryan on the release steps.

hack/pre-release.sh
aravindhp: This has not been answered.

hack/pre-release.sh
```diff
@@ -0,0 +1,110 @@
...
+  
+  if [[ $xyz_stream == "x" ]]; then
+    stream_var=1
+  fi 
+  if [[ $xyz_stream == "y" ]]; then
+    stream_var=2
+  fi 
+  if [[ $xyz_stream == "z" ]]; then
+    stream_var=3
+  fi
```
aravindhp, line 22: Please add a comment about why you are setting this to 1, 2 or 3.

hack/pre-release.sh
alinaryan:
```suggestion
  if [ -z $1 ]; then
  echo "Stream version not set"
  exit 1
  fi
```

alinaryan (commented):
Thanks for working on this - See my suggestions

hack/pre-release.sh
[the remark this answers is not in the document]
saifshaikh48: The skipRange still needs to be handled

hack/pre-release.sh
```diff
@@ -0,0 +1,109 @@
...
+}
+
+if [[ -z $1 ]]; then
+echo "usage:"
+echo "on a release-4.x branch, run ./pre-release (x, y or z)"
+echo "to update a z stream, use ./pre-release.sh z"
+echo "for other streams, use the corresponding x y or z argument."
+fi 
+
+update_WMCO_version $STREAM
```
saifshaikh48, line 107: I get an error when I run the script as is
```
$ ./hack/pre-release.sh x
./hack/pre-release.sh: line 107: STREAM: unbound variable
```

saifshaikh48 (commented):
Should we make the script smarter instead of incrementing anything +1? For example, if i run this on master, which has a mix of OCP 4.13 and 4.14 sprinkled throughout the codebase, it bumps them to 4.14 and 4.15 respectively. IMO it should consistently to bring the version value to a particular version

hack/pre-release.sh
aravindhp: This continues to not be answered.

hack/pre-release.sh
wgahnagl: I get the SkipRange here
https://github.com/openshift/windows-machine-config-operator/blob/195b873a7ebeb69670adbb4290cb7b0e914ad4fd/hack/pre-release.sh#L32 

and update it here 
https://github.com/openshift/windows-machine-config-operator/blob/195b873a7ebeb69670adbb4290cb7b0e914ad4fd/hack/pre-release.sh#L45 
Is there somewhere else I'm missing, or should I add a comment to make it more clear ? 

I was thinking of keeping it to that one version because ideally the versions wouldn't be mixed up like that, and it would be difficult to write regexes for each individual occurrence of each version 
If master has to have that mix of versions though, I could figure it out!

saifshaikh48: When I run `./hack/pre-release.sh x` (after fixing the unbounded variable), the skip version is not being edited. I suspect this is because you are making the changes directly to the bundle CSV in `update_WMCO_version` - when you run `make bundle` at the end of the script it will overwrite such changes because they are not being done to the base CSV

saifshaikh48: Also shouldn't the lower bound of the skipRange also be updated?

wgahnagl: https://github.com/openshift/windows-machine-config-operator/blob/195b873a7ebeb69670adbb4290cb7b0e914ad4fd/hack/pre-release.sh#L42 
this line updates the upper bound to the new version

and this line https://github.com/openshift/windows-machine-config-operator/blob/195b873a7ebeb69670adbb4290cb7b0e914ad4fd/hack/pre-release.sh#L45 
updates the lower bound to the current version 

I can add a comment on this to make it make more sense

hack/pre-release.sh
wgahnagl: I'm getting this from a comment inside the clusterserviceversion.yaml (the description [here](https://github.com/openshift/windows-machine-config-operator/blob/cc553e29d3bc16af106392e013e77c3ac4c5428d/config/manifests/bases/windows-machine-config-operator.clusterserviceversion.yaml#L37)) 
If I were to use JQ I would end up searching the comment anyway. 
Where I'm getting this information is arbitrary though, and I could get it from somewhere else, but I figured that getting it from the doc comment would be a reliable source.

hack/pre-release.sh
alinaryan: Adding here what we talked about in Slack DMs:

bundle csv = bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
base csv = config/manifests/bases/windows-machine-config-operator.clusterserviceversion.yaml

You can't directly edit the bundle csv. Changes need to be made in the base csv and then they will be applied to the bundle csv with the make bundle command. It's similar to CI release repo - the jobs need to be edited by editing the config and then you run make jobs to get the job file added

hack/pre-release.sh
aravindhp: I was not asking you to use `jq` but asking if there is a similar utility for processing yaml files. Given the amount of search and replacing with regexes you are doing here, I am wondering if there is a more elegant way of making these changes. Saying you need an utility as a pre-req is acceptable given it will make the code easier to read, understand and maintain. The way it stands, it looks pretty fragile and prone to breakage.

PS: This is a general comment and just specific to this line.

Moreover getting the version from a comment is not advisable. What if we decide to change that comment. Lets figure out a better spot to get the version from. It needn't be from the CSV but elsewhere.

aravindhp: Please add a comment saying what exactly you are looking for with your search statements. It is in no way obvious that you were picking the version from `* OCP 4.14 cluster running on Azure, AWS, GCP or vSphere configured with hybrid OVN Kubernetes networking`. Do this for all the lines.

aravindhp (changes requested):

hack/pre-release.sh
wgahnagl: how about from the gitmodules [here](https://github.com/openshift/windows-machine-config-operator/blob/cc553e29d3bc16af106392e013e77c3ac4c5428d/.gitmodules#L4)

hack/update_submodules.sh
```diff
@@ -123,30 +125,25 @@ if [ ! -z "$remote_branch" ]; then
...
 
-# Return if the working branch is dirty
-if ! git diff --quiet; then
-  echo "branch dirty, exiting to not overwrite work"
-  exit 1
-fi
 
-# For each branch that the user gives (master, release-4.7), create a new branch based off it,
-# with commits updating the submodules
+WMCO_ROOT=$(dirname "${BASH_SOURCE}")/..
```
aravindhp, line 129: Nice work, @wgahnagl. Just some early feedback... this sort of refactor should be done in a separate commit.

hack/pre-release.sh
```diff
@@ -0,0 +1,176 @@
...
+  # Example: matches com.redhat.openshift.versions: "*=v4.13*"
+  sed -i "s/=v$version/=v$updated_version/g" bundle/metadata/annotations.yaml
+
+  echo "Updating config-operator.package.yaml to $updated_version"
+  # Example: matches   - currentCSV: windows-machine-*config-operator.v9.0.0*
+  sed -i "s/config-operator.v$version/config-operator.v$updated_version/g" bundle/windows-machine-config-operator.package.yaml
+}
+
+update_OCP_version() {
+
```
alinaryan, line 71: nit: extra newline

alinaryan (commented):
LGTM, just add comments for the functions you added in common.sh and make sure the commit message is capitalized

hack/common.sh
```diff
@@ -33,6 +33,81 @@ get_OCP_version() {
...
+    echo "branch dirty, exiting to not overwrite work"
+    return 1
+  fi
+}
+
+# This function creates a git branch for automating commits
+# Takes n arguments
+# 1st arg: the ID of the branch, that describes the change, such as submodule-update
+# n+1th arg: the versions to create
+# usage: create_git_branch submodule-update release-4.10 release-4.12 release-4.13
```
saifshaikh48, line 48: create_git_branch --> create_git_branch_list

hack/pre-release.sh
```diff
@@ -0,0 +1,176 @@
...
+# this script updates the WMCO and OCP versions in the Makefile, base CSV, bundle and Dockerfiles.
+
+# usage:
+# on a release-4.x branch, run ./pre-release (x, y or z)
+# to update a z stream, use ./pre-release.sh z
+# for other streams, use the corresponding x y or z argument.
+
+# Include the common.sh 
+  WMCO_ROOT=$(dirname "${BASH_SOURCE}")/..
+  source $WMCO_ROOT/hack/common.sh
```
saifshaikh48, line 12: nit: indent level

hack/pre-release.sh
```diff
@@ -0,0 +1,176 @@
...
+
+update_WMCO_version(){
+  xyz_stream=$1
+  cluster_service_version_yaml=config/manifests/bases/windows-machine-config-operator.clusterserviceversion.yaml
+ 
+  # Setting the stream var to 1 2 or 3 which is used for incrementing the correct stream
+  # this tells awk which column to act on. Column 1, column 2, or column 3.
```
saifshaikh48, line 19: This can be made clearer. The columns mean what? I assume major, minor, and patch version

hack/pre-release.sh
```diff
@@ -0,0 +1,176 @@
+#!/bin/bash
+set -euo pipefail
+# this script updates the WMCO and OCP versions in the Makefile, base CSV, bundle and Dockerfiles.
+
+# usage:
+# on a release-4.x branch, run ./pre-release (x, y or z)
+# to update a z stream, use ./pre-release.sh z
+# for other streams, use the corresponding x y or z argument.
```
saifshaikh48, line 8: you can move the usage() func to the top of the file and get remove this comment block so we don't have to keep 2 places updated

hack/update_submodules.sh
```diff
@@ -123,30 +123,25 @@ if [ ! -z "$remote_branch" ]; then
...
+source $WMCO_ROOT/hack/common.sh
+
+# Check if branch is dirty, and exit if it is
+check_git_branch
```
aravindhp, line 131: If the function returns 1 does it automatically exit here? From what I understand, return exits terminates the subshell. Don't you have to do some return value checking here?

hack/common.sh
```diff
@@ -33,6 +33,93 @@ get_OCP_version() {
   echo $OCP_VER_MAJOR.$OCP_VER_MINOR
 }
 
+check_git_branch(){
```
aravindhp, line 36: Please add a space after `()`. Applies to all the function where the space is not present.

Please add a comment for the function.

hack/common.sh
```diff
@@ -33,6 +33,93 @@ get_OCP_version() {
...
+  # Return if the working branch is dirty
+  if ! git diff --quiet; then
+    echo "branch dirty, exiting to not overwrite work"
+    return 1
+  fi
+}
+
+# This function creates a git branch for automating commits
+# Takes n arguments
+# 1st arg: the ID of the branch, that describes the change, such as submodule-update
```
aravindhp, line 46: I don't follow what this means. Do you mean a branch name prefix?

hack/common.sh
```diff
@@ -33,6 +33,93 @@ get_OCP_version() {
...
+# args [2, n] arg: the versions to create
+# usage: create_git_branch_list submodule-update release-4.10 release-4.12 release-4.13
```
aravindhp, line 48: The usage does not match your function description.

aravindhp (changes requested):
Left a review for the refactor.

hack/pre-release.sh
```diff
@@ -0,0 +1,201 @@
...
+  # Example: matches FROM registry.ci.openshift.org/openshift/release:rhel-9-release-golang-1.19-*openshift-4.13* as build
+  sed -i "s/openshift-$version/openshift-$updated_version/g" build/Dockerfile.base
+
+  echo "Updating Dockerfile.ci version to $updated_version"
+  # Example: matches FROM registry.ci.openshift.org/openshift/release:rhel-9-release-golang-1.19-*openshift-4.13* as build
+  sed -i "s/openshift-$version/openshift-$updated_version/g" build/Dockerfile.ci
+
+  echo "Updating Dockerfile.wmco version to $updated_version"
+  # Example: matches FROM registry.ci.openshift.org/openshift/release:rhel-9-release-golang-1.19-*openshift-4.13* as build
+  sed -i "s/openshift-$version/openshift-$updated_version/g" build/Dockerfile.wmco
```
saifshaikh48, line 129: shouldn't there be a separate subcommand that bumps these?

hack/pre-release.sh
```diff
@@ -0,0 +1,201 @@
...
+   
+  if [[ "$CVP" == true ]] && [[ $# -ne 2 ]]; then
+    update_OCP_version $xy_stream $last_version
```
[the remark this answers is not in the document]
aravindhp: Do we really need to do recursion here?

hack/update_submodules.sh
wgahnagl: because we set the -e at the top, the return exits the subshell, and then since a command exited with a non-zero, the -e exits the program.

hack/common.sh
wgahnagl: I'm not sure what you mean by this :thinking:

aravindhp: Just that the function description and usage together is confusing. What do you mean by `the versions to create` when you specify `release-4.10`? Isn't the `release-4.10` branch already created? From what I can tell `release-4.10` is the branch from which you want to create another branch with a set of commits. That is not clear from your description.

hack/pre-release.sh
```diff
@@ -0,0 +1,200 @@
...
+    stream_var=1
+  fi
+  if [[ $xyz_stream == "y" ]]; then
+    stream_var=2
+  fi
+  if [[ $xyz_stream == "z" ]]; then
+    stream_var=3
+  fi
+ 
+    # Increment the version
```
saifshaikh48, line 45: nit: indentation

hack/pre-release.sh
```diff
@@ -0,0 +1,200 @@
...
+    sed -i "s/openshift-$version/openshift-$updated_version/g" build/Dockerfile.wmco
+  else
+    echo ""
+    echo "skipping updating base images"
+    echo "if these do need to be bumped, rerun with the -baseimage flag from your original branch"
+    echo ""
+  fi 
+}
+
+# This sets the CVP flag and removes it from the arg list
```
saifshaikh48, line 133: comment needs updating with baseimage flag

saifshaikh48 (commented):
Almost LGTM, just a few minor comments. I think the arg parsing can be reworked to be more intuitive (you can take inspiration from some of our other hack scripts) but it's not a blocker for me. Nice work overall

saifshaikh48 (commented):
/lgtm

hack/pre-release.sh
```diff
@@ -0,0 +1,207 @@
...
+
+update_OCP_version() { 
+  xy_stream=$1
+  
+  cluster_service_version_yaml=config/manifests/bases/windows-machine-config-operator.clusterserviceversion.yaml
+  
+  echo "OCP version $version detected"
+
+  # Increment the version by using the stream_var variable to select which column to increment 
+  updated_version=$(echo $version |
```
aravindhp, line 80: Same comment about `$stream_var`

hack/pre-release.sh
```diff
@@ -0,0 +1,207 @@
...
+    echo ""
+    echo "skipping bundle.Dockerfile and bundle/metadata/annotations.yaml due to CVP being a version behind"
+    echo "if these do need to be bumped, rerun with the -cvp flag from your original branch"
+    echo ""
+  fi 
+  
+  if [[ $BASEIMAGE == "true" ]]; then  
+    echo "Updating build/Dockerfile version to $updated_version"
+    # Example: matches FROM registry.ci.openshift.org/openshift/release:rhel-9-release-golang-1.19-*openshift-4.13* as build
+    sed -i "s/openshift-$version/openshift-$updated_version/g" build/Dockerfile
```
aravindhp, line 112: Nit: I suggest making this into a helper function and have a for loop here going over the set of files.

hack/pre-release.sh
```diff
@@ -0,0 +1,207 @@
...
+
+#
+args=()
+for ((i=1; i<=$#; i++)); do
+  if [[ "${!i}" == "-c" ]] || [[ "${!i}" == "-b" ]]; then
+    continue 
+  else
+    args+=("${!i}")
+  fi
+done
```
aravindhp, line 150: What is going on here? 😅

hack/pre-release.sh
```diff
@@ -0,0 +1,207 @@
...
+CVP=false
+BASEIMAGE=false 
+# Locate and store the WMCO version from the makefile 
+# finds "WMCO_VERSION ?=" and selects text following it
+WMCO_VERSION=$(sed -n -e 's/^.*WMCO_VERSION ?= //p' Makefile)
+
+usage() {
+  echo "usage:"
+  echo "input the name of the branch you want to increment, and the x y or z value to bump"
+  echo "example: ./pre-release.sh master x release-4.15 y release-4.10 z"
```
aravindhp, line 18: Do we really need this level of complexity? Why would we need to multiple releases at the same time. It is fine if we just handle `./pre-release.sh <branch> <stream>` i.e. do the updates independently. I think this will simplify some of the code below.

aravindhp (changes requested):

hack/pre-release.sh
```diff
@@ -0,0 +1,193 @@
...
+ 
+  echo "Updating bundle/windows-machine-config-operator.package.yaml to $updated_version"
+  # Example: matches   - currentCSV: windows-machine-*config-operator.v9.0.0*
+  sed -i "s/config-operator.v$version/config-operator.v$updated_version/g" bundle/windows-machine-config-operator.package.yaml
+   
+}
+
+update_OCP_version() { 
+  stream=$1
+  WMCO_VERSION=$2
```
aravindhp, line 74: Why are you overwriting the global `WMCO_VERSION` variable here? This should just be a local variable.

hack/pre-release.sh
```diff
@@ -0,0 +1,193 @@
...
+while getopts "bc" opt; do 
+  case "$opt" in 
+    c) CVP=true;;
+    b) BASEIMAGE=true;;
+  esac 
+done 
+shift $((OPTIND-1))
+
+# If there are less than two arguments, or if the amount of arguments is not divisible by two, exit
+if [ ${#@} -lt 2 ] || [ $(((${#@} - 1) % 2)) -eq 0 ]; then
```
aravindhp, line 154: Why does this have to be divisible by 2?

hack/pre-release.sh
```diff
@@ -0,0 +1,193 @@
...
+  # Example: matches olm.skipRange: '*>=8.0.0* <9.0.0'
+  sed -i "s/>=$last_version/>=$version/g" $cluster_service_version_yaml
+ 
+  echo "Updating bundle/windows-machine-config-operator.package.yaml to $updated_version"
+  # Example: matches   - currentCSV: windows-machine-*config-operator.v9.0.0*
+  sed -i "s/config-operator.v$version/config-operator.v$updated_version/g" bundle/windows-machine-config-operator.package.yaml
+   
+}
+
+update_OCP_version() {
```
aravindhp, line 72: Please add a function doc.

hack/pre-release.sh
```diff
@@ -0,0 +1,193 @@
...
+  sed -i "s/>=$last_version/>=$version/g" $cluster_service_version_yaml
+ 
+  echo "Updating bundle/windows-machine-config-operator.package.yaml to $updated_version"
+  # Example: matches   - currentCSV: windows-machine-*config-operator.v9.0.0*
+  sed -i "s/config-operator.v$version/config-operator.v$updated_version/g" bundle/windows-machine-config-operator.package.yaml
+   
+}
+
+update_OCP_version() { 
+  stream=$1
```
aravindhp, line 73: Please check for inputs

hack/pre-release.sh
```diff
@@ -0,0 +1,193 @@
...
+  echo "input the name of the branch you want to increment, and the x y or z value to bump"
+  echo "example: ./pre-release.sh master x"
+  echo "this will bump the x stream of master by 1"
+  echo "optional flags: -c"
+  echo "  bumps the versions held back by CVP"
+  echo "optional flags: -b"
+  echo "  bumps the baseimage versions"
+}
+
+update_WMCO_version() {
```
aravindhp, line 27: Please add a function doc

aravindhp (changes requested):

aravindhp (approved):
Nice work, @wgahnagl

aravindhp:
/override ci/prow/aws-e2e-upgrade ci/prow/azure-e2e-ccm-install 

The changes in the PR does not affect CI or the  operator

aravindhp:
/hold

For lgtm from @saifshaikh48

hack/pre-release.sh
```diff
@@ -0,0 +1,201 @@
...
+
+# Include the common.sh 
+WMCO_ROOT=$(dirname "${BASH_SOURCE}")/..
+source $WMCO_ROOT/hack/common.sh
+
+CVP=false
+BASEIMAGE=false
+
+# Locate and store the WMCO version from the makefile 
+# finds "WMCO_VERSION ?=" and selects text following it
```
saifshaikh48, line 13: You’d want to move this to where you moved the variable setting

saifshaikh48 (commented):
/lgtm

1 comment but I won’t hold up the PR for it

saifshaikh48 (commented):
/lgtm

1 comment but I won’t hold up the PR for it

hack/pre-release.sh
aravindhp: FTR, we should hold up PRs for something that was overlooked and easily fixed in the current PR itself. Opening a follow up a PR to fix moving of comment is unnecessary overhead for everyone.

aravindhp (approved):
/lgtm

PR #1573, opened 2023-04-18
WINC-1033: Use latest WICD binary during upgrades

saifshaikh48:
This PR enhances the node deconfiguration process to allow for better backward compatibility and failure recovery during upgrades.
- use the new version's WICD binary to cleanup during upgrade
- stop the WICD controller service from running before running WICD cleanup when recovering from node config failure

### DIFF
--- a/pkg/nodeconfig/nodeconfig.go
+++ b/pkg/nodeconfig/nodeconfig.go
@@ -235,7 +235,8 @@ func (nc *nodeConfig) Configure() error {
 	// Stop the kubelet so that the node is marked NotReady in case of an error in configuration. We are stopping all
 	// the required services as they are interdependent and is safer to do so given the node is going to be NotReady.
 	if err != nil {
-		if err := nc.Windows.RunWICDCleanup(nodeConfigCache.apiServerEndpoint, nc.wmcoNamespace); err != nil {
+		err := nc.Windows.RunWICDCleanup(nodeConfigCache.apiServerEndpoint, nc.wmcoNamespace, nodeConfigCache.credentials)
+		if err != nil {
 			nc.log.Info("Unable to mark node as NotReady", "error", err)
 		}
 	}
@@ -441,7 +442,8 @@ func (nc *nodeConfig) Deconfigure() error {
 	}
 
 	// Revert all changes we've made to the instance by removing installed services, files, and the version annotation
-	if err := nc.Windows.Deconfigure(nc.wmcoNamespace, nodeConfigCache.apiServerEndpoint); err != nil {
+	err := nc.Windows.Deconfigure(nc.wmcoNamespace, nodeConfigCache.apiServerEndpoint, nodeConfigCache.credentials)
+	if err != nil {
 		return fmt.Errorf("error deconfiguring instance: %w", err)
 	}
 
--- a/pkg/windows/windows.go
+++ b/pkg/windows/windows.go
@@ -233,9 +233,10 @@ type Windows interface {
 	// ConfigureWICD ensures that the Windows Instance Config Daemon is running on the node
 	ConfigureWICD(string, string, *Authentication) error
 	// Deconfigure removes all files and networks created by WMCO and runs the WICD cleanup command.
-	Deconfigure(string, string) error
-	// RunWICDCleanup runs the WICD cleanup command that ensures all services configured by WICD are stopped.
-	RunWICDCleanup(string, string) error
+	Deconfigure(string, string, *Authentication) error
+	// RunWICDCleanup ensures the WICD service is stopped and runs the cleanup command that ensures all WICD-managed
+	// services are also stopped
+	RunWICDCleanup(string, string, *Authentication) error
 }
 
 // windows implements the Windows interface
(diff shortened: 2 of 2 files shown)


pkg/windows/windows.go
```diff
@@ -404,12 +404,18 @@ func (vm *windows) RunWICDCleanup(apiServerURL string, watchNamespace string) er
...
 }
 
-func (vm *windows) Deconfigure(watchNamespace string, apiServerURL string) error {
+func (vm *windows) Deconfigure(watchNamespace string, apiServerURL string, credentials *Authentication) error {
 	vm.log.Info("deconfiguring")
 
 	if err := vm.deconfigureWICD(); err != nil {
 		return err
 	}
+	if err := vm.ensureWICDExists(credentials); err != nil {
```
sebsoto, line 415: Can you add a commit before this removing the `credentials` argument. it isnt used within ensureWICDExists

pkg/nodeconfig/nodeconfig.go
```diff
@@ -190,7 +190,7 @@ func (nc *nodeConfig) Configure() error {
 			nodeConfigCache.credentials); err != nil {
 			return fmt.Errorf("configuring WICD failed: %w", err)
 		}
-		// Set the desired version annotation, communicating to WICD which Windows services configmap to use
+		// Set the desired version annotation
```
sebsoto, line 193: Why is this being changed?

pkg/daemon/cleanup/cleanup.go
```diff
@@ -62,16 +62,23 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
...
+				return fmt.Errorf("node %s missing desired version annotation", node.Name)
+			}
 		}
 		err = directClient.Get(ctx,
-			client.ObjectKey{Namespace: configMapNamespace, Name: servicescm.NamePrefix + desiredVersion}, cm)
+			client.ObjectKey{Namespace: configMapNamespace, Name: servicescm.NamePrefix + version}, cm)
 		if err != nil {
 			return err
 		}
+		klog.Infof("removing services tied to version: %s", version)
```
sebsoto, line 81: nit: this is being logged to stdout and IMO will be noise when trying to debug the stdout/err stream on WMCO failures. At the least this should be debug level

pkg/windows/windows.go
```diff
@@ -234,6 +234,8 @@ type Windows interface {
 	ConfigureWICD(string, string, *Authentication) error
 	// Deconfigure removes all files and networks created by WMCO and runs the WICD cleanup command.
 	Deconfigure(string, string, *Authentication) error
+	// DeconfigureWICD ensures the WICD service running on the Windows instance is removed
+	DeconfigureWICD() error
 	// RunWICDCleanup runs the WICD cleanup command that ensures all services configured by WICD are stopped.
 	RunWICDCleanup(string, string) error
```
sebsoto, line 240: I don't think we need two separate methods for this. Instead you can just ensure the service is stopped within  RunWICDCleanup.

sebsoto (changes requested):

aravindhp (commented):
@saifshaikh48 you are making multiple changes to the upgrade flow in a single PR and we don't have good upgrade coverage in CI. How can we ensure that this change does not break upgrades? Have you tested it out manually without using an upgrade simulation?

pkg/daemon/cleanup/cleanup.go
saifshaikh48: I'm not seeing debug log capability with klog: https://github.com/kubernetes/klog/blob/v1.0.0/klog.go#L111

pkg/nodeconfig/nodeconfig.go
saifshaikh48: I misunderstood its context (config flow vs deconfig flow), will revert

alinaryan:
LGTM just curious about Aravindh's ask about testing

saifshaikh48:
/test aws-e2e-upgrade

Hit a known flake in the deletion tests
```
timed out waiting for daemonset noop-ds to scale, current status: 
{CurrentNumberScheduled:2 NumberMisscheduled:0 DesiredNumberScheduled:2 NumberReady:1 ObservedGeneration:1 UpdatedNumberScheduled:2 NumberAvailable:1 NumberUnavailable:1 CollisionCount:<nil> Conditions:[]}
```

aravindhp:
/retitle WINC-1033: Use latest WICD binary during upgrades

@saifshaikh48 lets try be succinct in our PR header. You can give more details within the PR.

pkg/windows/windows.go
```diff
@@ -431,7 +431,7 @@ func (vm *windows) Bootstrap(desiredVer, apiServerURL, watchNamespace string, cr
 	}
 	// We have to ensure the WICD files exist separately before the rest of the files because we must ensure services
 	// are stopped and their files closed before modifying them.
-	if err := vm.ensureWICDExists(credentials); err != nil {
+	if err := vm.ensureWICDExists(); err != nil {
```
aravindhp: I know @sebsoto asked for this change but be aware if we ever decided to revert this PR, this orthogonal change which has nothing to do with this PR will get reverted too.

pkg/windows/windows.go
```diff
@@ -233,9 +233,10 @@ type Windows interface {
 	// ConfigureWICD ensures that the Windows Instance Config Daemon is running on the node
 	ConfigureWICD(string, string, *Authentication) error
 	// Deconfigure removes all files and networks created by WMCO and runs the WICD cleanup command.
-	Deconfigure(string, string) error
+	Deconfigure(string, string, *Authentication) error
 	// RunWICDCleanup runs the WICD cleanup command that ensures all services configured by WICD are stopped.
-	RunWICDCleanup(string, string) error
+	// Stops the WICD service prior to executing cleanup, if needed.
```
aravindhp, line 238: `RunWICDCleanup ensures the WICD service  is stopped if needed and runs the cleanup command that ensures WICD managed services are stopped`

aravindhp (changes requested):
Please call out why we need to use the new WICD binary in the `[windows] Stop WICD & copy new .exe before cleanup`. Nit: you can remove `.` from `.exe`

pkg/windows/windows.go
saifshaikh48: updated

aravindhp (approved):

PR #1592, opened 2023-04-28
[nodeconfig] Use normal api URL instead of api-int

sebsoto:
Using api-int is no longer required in WMCO as we do not pull the ignition file from the MCO endpoint.


pkg/nodeconfig/init.go
```diff
@@ -66,10 +66,10 @@ func discoverKubeAPIServerEndpoint() (string, error) {
 		return "", fmt.Errorf("unable to get cluster infrastructure resource: %w", err)
 	}
 	// get API server internal url of format https://api-int.abc.devcluster.openshift.com:6443
```
saifshaikh48, line 68: comment needs updating

sebsoto: Thanks @saifshaikh48, fixed

sebsoto:
Closing for now, will investigate security implications in the future.

PR #1598, opened 2023-05-02
WINC-999: WMCO manages new trusted CA ConfigMap

saifshaikh48:
This PR introduces a new resource to be managed by WMCO in proxied
clusters. This resource will contain a trusted CA injection request label so it
will be updated by the cluster-network-operator (CNO) when the global
Proxy resource changes. This is the first step to supporting custom user
certificates for HTTPS proxies.

It also creates the scaffolding for the proxy test suite, which can be
expanded as more proxy support functionality is added and needs to be tested.

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -196,7 +196,8 @@ func main() {
 		setupLog.Error(err, "error removing invalid annotations from Linux nodes")
 	}
 
-	configMapReconciler, err := controllers.NewConfigMapReconciler(mgr, clusterConfig, watchNamespace)
+	proxyEnabled := cluster.IsProxyEnabled()
+	configMapReconciler, err := controllers.NewConfigMapReconciler(mgr, clusterConfig, watchNamespace, proxyEnabled)
 	if err != nil {
 		setupLog.Error(err, "unable to create ConfigMap reconciler")
 		os.Exit(1)
@@ -243,6 +244,15 @@ func main() {
 	if err := configMapReconciler.EnsureWICDRBAC(); err != nil {
 		setupLog.Error(err, "error ensuring WICD RBAC resources exist", "namespace", watchNamespace)
 		os.Exit(1)
+	}
+
+	// If proxy is enabled, disabled, or edited during WMCO runtime, the WMCO pod will be restarted by OLM. This could
+	// happen in the middle of node configuration, at which the controllers will reconcile once the WMCO pod restarts
+	if proxyEnabled {
+		if err := configMapReconciler.EnsureTrustedCAConfigMapExists(); err != nil {
+			setupLog.Error(err, "error ensuring trusted CA ConfigMap exists", "namespace", watchNamespace)
+			os.Exit(1)
+		}
 	}
 
 	setupLog.Info("starting manager")
--- a/controllers/configmap_controller.go
+++ b/controllers/configmap_controller.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net"
 	"strings"
@@ -48,6 +49,7 @@ import (
 	"github.com/openshift/windows-machine-config-operator/pkg/instance"
 	"github.com/openshift/windows-machine-config-operator/pkg/metrics"
 	"github.com/openshift/windows-machine-config-operator/pkg/nodeconfig"
+	"github.com/openshift/windows-machine-config-operator/pkg/patch"
 	"github.com/openshift/windows-machine-config-operator/pkg/secrets"
 	"github.com/openshift/windows-machine-config-operator/pkg/services"
 	"github.com/openshift/windows-machine-config-operator/pkg/servicescm"
@@ -73,16 +75,22 @@ const (
 	ConfigMapController = "configmap"
 	// wicdRBACResourceName is the name of the resources associated with WICD's RBAC permissions
 	wicdRBACResourceName = "windows-instance-config-daemon"
+	// proxyCertsConfigMap is the name of the ConfigMap that holds the trusted CA bundle for a cluster-wide proxy
+	proxyCertsConfigMap = "trusted-ca"
+	// injectionRequestLabel is used to allow CNO to inject the trusted CA bundle when the global Proxy resource changes
+	injectionRequestLabel = "config.openshift.io/inject-trusted-cabundle"
 )
 
 // ConfigMapReconciler reconciles a ConfigMap object
 type ConfigMapReconciler struct {
 	instanceReconciler
 	servicesManifest *servicescm.Data
+	proxyEnabled     bool
 }
 
 // NewConfigMapReconciler returns a pointer to a ConfigMapReconciler
-func NewConfigMapReconciler(mgr manager.Manager, clusterConfig cluster.Config, watchNamespace string) (*ConfigMapReconciler, error) {
+func NewConfigMapReconciler(mgr manager.Manager, clusterConfig cluster.Config, watchNamespace string,
+	proxyEnabled bool) (*ConfigMapReconciler, error) {
 	clientset, err := kubernetes.NewForConfig(mgr.GetConfig())
 	if err != nil {
 		return nil, fmt.Errorf("error creating kubernetes clientset: %w", err)
@@ -132,6 +140,7 @@ func NewConfigMapReconciler(mgr manager.Manager, clusterConfig cluster.Config, w
 			platform:             clusterConfig.Platform(),
 		},
 		servicesManifest: svcData,
+		proxyEnabled:     proxyEnabled,
 	}, nil
 }
 
@@ -164,8 +173,8 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context,
 	// so it is safe to assume that all ConfigMaps being reconciled are one of:
 	// 1. windows-instances, describing hosts that need to be present in the cluster.
 	// 2. windows-services, describing expected configuration of WMCO-managed services on all Windows instances
-	// 3. kube-apiserver-to-kubelet-client-ca, contains the CA for the kubelet to recognize the kube-apiserver
-	// client certificate
+	// 3. kube-apiserver-to-kubelet-client-ca, contains the CA for the kubelet to recognize the kube-apiserver client cert
+	// 4. trusted-ca, where CNO will publish user-provided certs when there is an active cluster-wide proxy
 	configMap := &core.ConfigMap{}
 	if err := r.client.Get(ctx, req.NamespacedName, configMap); err != nil {
 		if !k8sapierrors.IsNotFound(err) {
@@ -178,6 +187,10 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context,
 				return ctrl.Result{}, err
 			}
 		}
+		if req.NamespacedName.Name == proxyCertsConfigMap {
+			// Create the trusted CA ConfigMap as it is not present
+			return ctrl.Result{}, r.createProxyCertsCM(ctx)
+		}
 	}
 
 	r.log.V(1).Info("Reconciling", "ConfigMap", req.NamespacedName)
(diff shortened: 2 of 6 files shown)


saifshaikh48:
/test vsphere-e2e-operator

Validated locally, proxy tests skipped when cluster-wide proxy is not configured + pass when it is configured 
```
    proxy_test.go:33: cluster-wide proxy is not enabled in this environment
    --- SKIP: TestWMCO/cluster-wide_proxy (0.10s)
```
```
    --- PASS: TestWMCO/cluster-wide_proxy (45.94s)
        --- PASS: TestWMCO/cluster-wide_proxy/Trusted_CA_ConfigMap_validation (45.84s)
            --- PASS: TestWMCO/cluster-wide_proxy/Trusted_CA_ConfigMap_validation/Trusted_CA_ConfigMap_contents (0.21s)
            --- PASS: TestWMCO/cluster-wide_proxy/Trusted_CA_ConfigMap_validation/Trusted_CA_ConfigMap_re-creation (15.24s)
            --- PASS: TestWMCO/cluster-wide_proxy/Trusted_CA_ConfigMap_validation/Invalid_trusted_CA_ConfigMap_patching (30.32s)
```

pkg/cluster/config.go
```diff
@@ -32,6 +33,18 @@ const (
 	MachineAPINamespace = "openshift-machine-api"
 )
 
+var (
+	// acceptableProxyVars is a list of the supported proxy variables
+	acceptableProxyVars = []string{"HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY"}
```
aravindhp, line 38: This is just `proxyEnvVars`. `acceptable` is implicit and need not be called out.

pkg/cluster/config.go
```diff
@@ -340,3 +353,16 @@ func IsCloudControllerOwnedByCCM(oclient configclient.Interface) (bool, error) {
 
 	return ownedByCCM, nil
 }
+
+// getProxyVars returns a map of the proxy variables and values from the WMCO container's environment. The presence of
+// any implies a proxy is enabled, since OLM would inject into the operator spec. Returns an empty map otherwise.
+func getProxyVars() map[string]string {
+	proxyVarsMap := make(map[string]string, len(acceptableProxyVars))
+	for _, envVar := range acceptableProxyVars {
+		value, found := os.LookupEnv(envVar)
```
aravindhp, line 362: Is this run on the WMCO pod? Do we need to do anything in the operator manifests to get CNO to inject these variables into the pod?

pkg/cluster/config.go
```diff
@@ -340,3 +353,16 @@ func IsCloudControllerOwnedByCCM(oclient configclient.Interface) (bool, error) {
 
 	return ownedByCCM, nil
 }
+
+// getProxyVars returns a map of the proxy variables and values from the WMCO container's environment. The presence of
+// any implies a proxy is enabled, since OLM would inject into the operator spec. Returns an empty map otherwise.
+func getProxyVars() map[string]string {
+	proxyVarsMap := make(map[string]string, len(acceptableProxyVars))
```
aravindhp, line 360: Please don't call out the data structure used in the variable now. That is an anti-pattern in go. `proxyVars` will suffice.

pkg/cluster/config.go
```diff
@@ -32,6 +33,18 @@ const (
...
+var (
+	// acceptableProxyVars is a list of the supported proxy variables
+	acceptableProxyVars = []string{"HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY"}
+	// ClusterWideProxyVars is a map of the global egress proxy variables and values from WMCO's environment
+	ClusterWideProxyVars map[string]string
+)
+
+// init runs once, initializing global variables
+func init() {
+	ClusterWideProxyVars = getProxyVars()
```
aravindhp, line 45: This is a one time initialization during WMCO bootup. What happens if these values are changed when WMCO is running?

aravindhp (changes requested):

cmd/operator/main.go
```diff
@@ -196,7 +196,8 @@ func main() {
 		setupLog.Error(err, "error removing invalid annotations from Linux nodes")
 	}
 
-	configMapReconciler, err := controllers.NewConfigMapReconciler(mgr, clusterConfig, watchNamespace)
+	proxyEnabled := len(cluster.ClusterWideProxyVars) > 0
```
sebsoto, line 199: Please add a `proxyEnabled()` function in the cluster package to contain this logic.

sebsoto (changes requested):
Thanks will review again when current comments are addressed as they could cause larger code changes

saifshaikh48:
/hold

until https://github.com/openshift/release/pull/39073 merges

pkg/cluster/config.go
saifshaikh48: The WMCO is restarted by OLM when these values are updated. We never expected a user to update these values

cmd/operator/main.go
```diff
@@ -245,6 +246,13 @@ func main() {
 		os.Exit(1)
 	}
 
+	if proxyEnabled {
```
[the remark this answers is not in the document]
saifshaikh48: Yep, if proxy is enabled when WMCO is running, OLM deletes and recreates the WMCO deployment with the proxy vars injected into the pod environement

pkg/cluster/config.go
saifshaikh48: Yes, the lookup is run on the WMCO pod.

OLM injects the vars into the pod env btw, not CNO. CNO injects the CA bundle into the trusted-ca configmap. We don't need anything in the manifests to enable OLM's injection, it does so to all it's managed operators automatically

controllers/configmap_controller.go
```diff
@@ -73,16 +75,22 @@ const (
 	ConfigMapController = "configmap"
 	// wicdRBACResourceName is the name of the resources associated with WICD's RBAC permissions
 	wicdRBACResourceName = "windows-instance-config-daemon"
+	// proxyCertsConfigMap is the name of the ConfigMap that holds the trusted CA bundle for a cluster-wide proxy
+	proxyCertsConfigMap = "trusted-ca"
+	// injectionRequestLabel is used to allow CNO to inject the trusted CA bundle when the global Proxy resource changes
+	injectionRequestLabel = "config.openshift.io/inject-trusted-cabundle"
```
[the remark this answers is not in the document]
saifshaikh48: They exist at in the [CNO repo](https://github.com/openshift/cluster-network-operator/blob/master/pkg/names/names.go#L131), but I'm hitting issues importing

cmd/operator/main.go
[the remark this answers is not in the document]
aravindhp: Please add a comment about this.

What happens if WMCO is in the middle of a configuration when OLM decides to restart the Pod?

pkg/cluster/config.go
aravindhp: Please  add a comment.

pkg/cluster/config.go
aravindhp: Please add a comment.

aravindhp (changes requested):
Please add comment where appropriate including the commit message about OLM involvement.

cmd/operator/main.go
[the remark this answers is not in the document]
aravindhp: OK, please capture this too in your comment.

pkg/cluster/config.go
saifshaikh48: added to the function doc and comment

pkg/cluster/config.go
saifshaikh48: comment added explaining this

cmd/operator/main.go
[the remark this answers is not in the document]
saifshaikh48: comment added summarizing the above

controllers/configmap_controller.go
```diff
@@ -189,6 +202,8 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context,
 		return ctrl.Result{}, r.reconcileNodes(ctx, configMap)
 	case certificates.KubeAPIServerServingCAConfigMapName:
 		return ctrl.Result{}, r.reconcileKubeletClientCA(ctx, configMap)
+	case proxyCertsConfigMap:
+		return ctrl.Result{}, r.reconcileProxyCertsCM(ctx, configMap)
```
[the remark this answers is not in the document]
saifshaikh48: Nope, in the case of a delete event, lines 190-192 will be executed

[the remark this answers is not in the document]
sebsoto: Okay, it seems that logic should be moved into reconcileProxyCertsCM, as that is code specific to reconciling the ProxyCertsCM

cmd/operator/main.go
```diff
@@ -245,6 +246,15 @@ func main() {
 		os.Exit(1)
 	}
 
+	// If proxy is enabled, disabled, or edited during WMCO runtime, the WMCO pod will be restarted by OLM. This could
+	// happen in the middle of node configuration, at which the controllers will reconcile once the WMCO pod restarts
+	if proxyEnabled {
+		if err := configMapReconciler.EnsureTrustedCAConfigMapExists(); err != nil {
```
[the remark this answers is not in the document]
saifshaikh48: This can be created as part of the operator manifests but we lose the ability to kick back changes (if a user deleted the trusted-ca CM or removed its expected injection label). If we are fine with that, I can updated the PR to go the operator manifest route

aravindhp:
/test nutanix-e2e-operator

Testing to get a signal

saifshaikh48:

previous run hit 4hr timeout, took 2 hours to acquire a vsphere lease...

cmd/operator/main.go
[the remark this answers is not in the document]
saifshaikh48: > Can this be created as part of the operator manifests instead?

resolution: we decided against this for the above reason

controllers/configmap_controller.go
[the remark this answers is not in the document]
saifshaikh48: Let me know if you feel strongly about this, I can do a small refactor.

For context, I followed the existing pattern to only enter the switch statement if the CM is set. See comment above the switch `// At this point configMap will be set properly`. We do something similar for the services CM (lines 184-189).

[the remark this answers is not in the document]
sebsoto: Right, but I don't think the existing pattern makes sense. At least in this case.

You're introducing a new function to reconcile the CM, but you're creating the CM setting the state as we want it to be beforehand. So there's not going to be a reason to call reconcileProxyCertsCM, in the delete case it will always be a no-op as the state has already been reconciled beforehand.

Seeing that it has been deleted and re-creating it is reconciling.

pkg/cluster/config.go
```diff
@@ -32,6 +33,20 @@ const (
 	MachineAPINamespace = "openshift-machine-api"
 )
 
+var (
+	// proxyEnvVars is a list of the supported proxy-related environment variables
+	proxyEnvVars = []string{"HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY"}
```
sebsoto, line 38: Does this need to be a global variable?
Seems like its specific to the `getProxyVars` function

sebsoto (approved):
Please address the comment I left before removing the draft label

pkg/cluster/config.go
saifshaikh48: Guess I was thinking ahead here. This var will likely be exported+used in future proxy PRs. I'll move it to getProxyVars()

saifshaikh48:

install failure, hoping this is a flake as there were a bunch of successful installs before this

saifshaikh48:
/hold

holding as the proxy test passing is a requirement for this PR to merge

saifshaikh48:

took 2 hours to acquire vSphere lease so test timed out. Will have to bring this up with infra team if this persists today

```
INFO[2023-05-11T20:44:45Z] Acquiring leases for test vsphere-proxy-e2e-operator: [vsphere-quota-slice] 
INFO[2023-05-11T22:40:20Z] Acquired 1 lease(s) for vsphere-quota-slice: [ci-segment-90] 
```

mansikulkarni96:
/lgtm

saifshaikh48:
/retest-required

saw cluster spinup failures in all non-azure platforms, 1 retest to see if this persists

saifshaikh48:
/retest-required

regular vsphere test timed out, 2hrs to secure a lease

PR #1603, opened 2023-05-04
Azure 4.10 test

aravindhp:
Do not merge


aravindhp:
/close

confirmed that the image with Docker is still available for now. Not sure when Microsoft will delete the images.

PR #1612, opened 2023-05-08
[release-4.10] WINC-1036: Drop Azure support in WMCO 5.x and lower

aravindhp:
Explain why we are dropping Azure support in WMCO 5.x and lower.

(manually cherry picked from commit 5e754f1)


saifshaikh48 (commented):
/lgtm

PR #1614, opened 2023-05-11
WIP: WINC-886: Make WICD Bootstrap command not a method of ServiceController

alinaryan:
Changes the Bootstrap command so that it is not a method of ServiceController since the service controller object only needs to be instantiated once.


mtnbikenc:
Jira issue closed.  Won't do.

PR #1621, opened 2023-05-18
Reuse testContext across the test suite

sebsoto:
Previously many subtests were creating their own testContext. Everytime a testContext is created there is a cost of initializing multiple k8s API clients, as well as the Provider struct used by the testContext. But more importantly doing this is causing a lot of boilerplate code at the start of each subtest, to initialize the testContext and check for an error.

This commit addresses that by creating a testContext at each top level test for each test suite. The subtests have been made methods of testContext, a pattern that was being used by some of the subtests already.

This also makes the reciever of testContext consistent across the tests, using `tc` instead of the mix of `tc` and `testCtx` that was present before.


aravindhp (commented):
I am not sure if this is the correct approach. You are now basically making the test context in to the global context. Please take a look at the doc for `NewTestContext()` and lets revisit.

Does this cause issues if I want to run individual tests?

sebsoto:
> Does this cause issues if I want to run individual tests?

It does not. Because all the e2e tests are a subtest of TestWMCO, in order to run them, TestWMCO needs to be ran up until the part of TestWMCO that invokes the subtest.

> You are now basically making the test context in to the global context

That's a good point.
I've found the mixed use of the globalContext and TestContext frustrating and unintuitive. For example having to call waitForWindowsNodes, a method of TestContext, in order to populate the machineNodes and byohNodes in the globalContext, in order for tests to function properly. Example of this: https://github.com/openshift/windows-machine-config-operator/blob/0ecb2e1014c32c69574f5aed970460213bc3e12e/test/e2e/network_test.go#L31-L36

> Please take a look at the doc for NewTestContext()

I think the comment still holds true: `This data structure should be instantiated by every test suite...` In practice that's still occurring as we only ever invoke the tests one suite at a time. At least in CI and in my personal use. I'm not sure if others ever directly call
```
go test ./test/e2e/... -run=TestWMCO ...
````

I will only ever run each individual test suite at a time due to the very long turn around to get any feedback from the tests
```
go test ./test/e2e/... -run=TestWMCO/network ...
go test ./test/e2e/... -run=TestWMCO/storage ...
````

All that said, I'm not really seeing a good reason to have both testContext and globalContext, I don't see any reason to not use the same TestContext across all the suites, other than globalContext existing as it currently does.  I think moving the responsibilities of globalContext into testContext would help improve the code, which is something I can do in this PR. Those responsibilities being keeping track of the expected amount of Windows nodes, and the path to the private key.

**Review by aravindhp — APPROVED** [review]

File: test/e2e/wmco_test.go [hunk 1]
```diff
@@ -49,15 +49,15 @@ func TestWMCO(t *testing.T) {
 	// currently wait for nodes to fully reconcile after changing the private key back to the valid key. Any tests
 	// added/moved in between these two suites may fail.
 	// This limitation will be removed with https://issues.redhat.com/browse/WINC-655
-	t.Run("reconfigure", reconfigurationTestSuite)
-	t.Run("destroy", deletionTestSuite)
+	t.Run("reconfigure", tc.reconfigurationTestSuite)
+	t.Run("destroy", tc.deletionTestSuite)
```
saifshaikh48, line 53: Any reason for only doing this to the reconfigure and deletion suites? IMO we should be consistent in either creating a new test context for each suite or reusing the top level test context across all suites (creation, network, upgrade, storage etc)

saifshaikh48:
Thanks for working on this!

/lgtm

aravindhp:
/hold

Until https://github.com/openshift/windows-machine-config-operator/pull/1637 merges which will unblock the vsphere-e2e-operator test

PR #1622, opened 2023-05-18
[release-4.12] OCPBUGS-13790: Address potential unitialized variables

sebsoto:


saifshaikh48 (commented):
/lgtm

pkg/nodeconfig/init.go
```diff
@@ -39,7 +39,6 @@ func init() {
 	kubeAPIServerEndpoint, err := discoverKubeAPIServerEndpoint()
 	if err != nil {
 		log.Error(err, "unable to find kube api server endpoint")
-		return
```
aravindhp, line 42: Isn't it safer to return?

pkg/nodeconfig/nodeconfig.go
```diff
@@ -129,6 +129,14 @@ func NewNodeConfig(c client.Client, clientset *kubernetes.Clientset, clusterServ
 		return nil, fmt.Errorf("error instantiating Windows instance from VM: %w", err)
 	}
 
+	// initialize the cache now if it failed to initialize
```
aravindhp, line 132: Don't we need to do the same for `nodeConfigCache.credentials`?

aravindhp (commented):
Thanks for the quick turnaround on this, @sebsoto

aravindhp (commented):
Sorry I missed the 2nd commit

pkg/nodeconfig/nodeconfig.go
aravindhp: Sorry I missed the 2nd commit

pkg/nodeconfig/init.go
sebsoto: thats probably for the best

PR #1624, opened 2023-05-22
[release-4.13] WINC-1020: Update version to 8.0.1

wgahnagl:

### DIFF
--- a/bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
@@ -8,7 +8,7 @@ metadata:
     certified: "false"
     createdAt: REPLACE_DATE
     description: An operator that enables Windows container workloads on OCP
-    olm.skipRange: '>=7.0.0 <8.0.0'
+    olm.skipRange: '>=7.0.0 <8.0.1'
     operatorframework.io/cluster-monitoring: "true"
     operatorframework.io/suggested-namespace: openshift-windows-machine-config-operator
     operators.openshift.io/valid-subscription: '["Red Hat OpenShift support for Windows
@@ -17,7 +17,7 @@ metadata:
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v3
     repository: https://github.com/openshift/windows-machine-config-operator
     support: Red Hat
-  name: windows-machine-config-operator.v8.0.0
+  name: windows-machine-config-operator.v8.0.1
   namespace: placeholder
 spec:
   apiservicedefinitions: {}
@@ -470,4 +470,4 @@ spec:
   minKubeVersion: 1.26.0
   provider:
     name: Red Hat
-  version: 8.0.0
+  version: 8.0.1
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
 # - use the WMCO_VERSION as arg of the bundle target (e.g make bundle WMCO_VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export WMCO_VERSION=0.0.2)
-WMCO_VERSION ?= 8.0.0
+WMCO_VERSION ?= 8.0.1
 
 # *_GIT_VERSION are the k8s versions. Any update to the build line could potentially require an update to the sed
 # command in generate_k8s_version_commit() in hack/update_submodules.sh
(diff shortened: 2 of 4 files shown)


bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
```diff
@@ -6,9 +6,9 @@ metadata:
     capabilities: Seamless Upgrades
     categories: OpenShift Optional
     certified: "false"
-    createdAt: REPLACE_DATE
+    createdAt: "2023-05-22T16:06:59Z"
```
saifshaikh48, line 9: why is this being changed?

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
```diff
@@ -6,9 +6,9 @@ metadata:
...
     description: An operator that enables Windows container workloads on OCP
-    olm.skipRange: '>=7.0.0 <8.0.0'
+    olm.skipRange: '>=8.0.0 <8.0.1'
```
saifshaikh48, line 11: why bump the lower end of the skipRange here? what are the implications of this change?

saifshaikh48 (commented):
Few comments. The PR should target the release branch rather than community btw

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
mtnbikenc: In other words, this should not be changed.  This date is updated by `render_templates` in [dist-git](https://pkgs.devel.redhat.com/cgit/containers/windows-machine-config-operator-bundle/tree/render_templates?h=rhaos-4.13-rhel-9&id=b6b1019ef95ba89e9537535f5dc1cbfafba2f0a4#n60).

wgahnagl: It seems that all of the changes being made by make bundle aren't needed, but is mentioned in the WMCO releasing doc as part of the post release tasks. Should we keep running make bundle and just ignore the changes it makes to the files, or remove it from the script entirely?

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
wgahnagl: you mentioned [here](https://github.com/openshift/windows-machine-config-operator/pull/1566#discussion_r1176816796) that that's the expected functionality of the script, and these changes are from that script.

saifshaikh48: imagine we were to run this script again with current functionality to prepare for version `8.0.2` -- the skipRange would be bumped to '>=9.0.0 <8.0.2' which doesn't look right
I suggest looking at [OLM upgrade process](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/how-to-update-operators.md), and how the skipRange looks in [other WMCO branches](https://github.com/openshift/windows-machine-config-operator/blob/release-4.11/bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml)

wgahnagl: the way the bump works is that it takes the current version and puts it at the lower range, and then sets the updated version to the higher range. On an 8.0.2 bump, the skipRange would be >=8.0.1 <8.0.2
I can add a condition to only update that for the x version though. Should the lower range stay the same on minor versions?

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
```diff
@@ -367,7 +367,9 @@ spec:
           - use
         serviceAccountName: windows-machine-config-operator
       deployments:
-      - name: windows-machine-config-operator
+      - label:
+          name: windows-machine-config-operator
+        name: windows-machine-config-operator
```
alinaryan, line 372: Are these supposed to be here twice?

alinaryan (commented):
looking good other than my comment

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
mansikulkarni96: yes it should stay the same for minor versions, in short we do not need to touch skiprange for a minor version bump unless there is a need to do so.

mtnbikenc:
/retitle [release-4.13] WINC-1020: Update version to 8.0.1

aravindhp (approved):
/lgtm
/hold cancel

aravindhp:
/hold cancel

There is no ordering issue

PR #1627, opened 2023-05-24
WINC-860: [hack] Bugfixes for the post-release script

wgahnagl:
Adds failure conditon for incorrect operator-sdk version and
updates skipVersion's lower bound only on x releases


hack/pre-release.sh
```diff
@@ -63,11 +64,13 @@ update_WMCO_version() {
...
   sed -i "s/<$version/<$updated_version/g" $cluster_service_version_yaml
-
-  echo "Updating previous skipRange version in $cluster_service_version_yaml to $version"
-  # Example: matches olm.skipRange: '*>=8.0.0* <9.0.0'
-  sed -i "s/>=$last_version/>=$version/g" $cluster_service_version_yaml
- 
+  
+  if [[ $xyz_stream == "x" ]]; then 
+    echo "Updating previous skipRange version in $cluster_service_version_yaml to $version"
+    # Example: matches olm.skipRange: '*>=8.0.0* <9.0.0'
```
aravindhp, line 70: Please call out what you are replacing in addition. You might want to do this in other places too in a separate PR.

hack/pre-release.sh
```diff
@@ -185,11 +188,29 @@ switch_git_branch $base_branch $new_branch
...
+# if not, update it to the correct version 
+operator_sdk_current_version=$(operator-sdk version |  grep -oP 'operator-sdk version: "\K[^"]+' )
+if [[ "$operator_sdk_current_version" != "$OPERATOR_SDK_VERSION" ]]; then
+  echo "incorrect version. operator-sdk version must be $OPERATOR_SDK_VERSION"  
+  echo "downloading and installing operator-sdk version $OPERATOR_SDK_VERSION"
+
+  export ARCH=$(case $(uname -m) in x86_64) echo -n amd64 ;; aarch64) echo -n arm64 ;; *) echo -n $(uname -m) ;; esac)
+  export OS=$(uname | awk '{print tolower($0)}')
+
+  export OPERATOR_SDK_DL_URL=https://github.com/operator-framework/operator-sdk/releases/download/$OPERATOR_SDK_VERSION
```
aravindhp, line 205: What's the reason for exporting instead of just using variables?

hack/pre-release.sh
```diff
@@ -185,11 +188,29 @@ switch_git_branch $base_branch $new_branch
...
+  curl -LO ${OPERATOR_SDK_DL_URL}/operator-sdk_${OS}_${ARCH}
+
+  chmod +x operator-sdk_${OS}_${ARCH} && sudo mv operator-sdk_${OS}_${ARCH} /usr/local/bin/operator-sdk
```
aravindhp, line 208: I suggest just exiting if the SDK version does not match. We don't want to be overwriting the SDK version without user permission.

hack/pre-release.sh
```diff
@@ -185,11 +188,29 @@ switch_git_branch $base_branch $new_branch
 # Locate and store the WMCO version from the makefile 
 # finds "WMCO_VERSION ?=" and selects text following it
 WMCO_VERSION=$(sed -n -e 's/^.*WMCO_VERSION ?= //p' Makefile)
+
```
aravindhp, line 191: Unnecessary change?

hack/pre-release.sh
```diff
@@ -63,11 +64,13 @@ update_WMCO_version() {
...
   # Example: matches olm.skipRange: '>=8.0.0 *<9.0.0*'
   sed -i "s/<$version/<$updated_version/g" $cluster_service_version_yaml
-
-  echo "Updating previous skipRange version in $cluster_service_version_yaml to $version"
-  # Example: matches olm.skipRange: '*>=8.0.0* <9.0.0'
-  sed -i "s/>=$last_version/>=$version/g" $cluster_service_version_yaml
- 
+  
+  if [[ $xyz_stream == "x" ]]; then 
+    echo "Updating previous skipRange version in $cluster_service_version_yaml to $version"
```
aravindhp, line 69: This message is confusing. What do you mean by `previous skipRange version`? I realize this issue was there before but this is bringing it more to light. You need to clearly call out what is being updated here vs the previous `Updating the skipRange version in` message i.e. lower range vs upper range.

aravindhp (changes requested):

hack/pre-release.sh
aravindhp: Moreover you are looking at the OS and architecture. So how can you assume `/usr/local/bin` across OSes?

hack/pre-release.sh
```diff
@@ -187,9 +190,20 @@ switch_git_branch $base_branch $new_branch
 WMCO_VERSION=$(sed -n -e 's/^.*WMCO_VERSION ?= //p' Makefile)
 update_WMCO_version $stream $WMCO_VERSION
 update_OCP_version $stream $WMCO_VERSION 
+
+# check to see if the opeator-sdk version is correct 
+# if not, update it to the correct version
```
aravindhp, line 195: Please update the comment, commit and PR messages now that you are not automatically installing the SDK.

hack/pre-release.sh
```diff
@@ -187,9 +190,20 @@ switch_git_branch $base_branch $new_branch
 WMCO_VERSION=$(sed -n -e 's/^.*WMCO_VERSION ?= //p' Makefile)
 update_WMCO_version $stream $WMCO_VERSION
 update_OCP_version $stream $WMCO_VERSION 
+
+# check to see if the opeator-sdk version is correct 
+# if not, update it to the correct version 
+operator_sdk_current_version=$(operator-sdk version |  grep -oP 'operator-sdk version: "\K[^"]+' )
+if [[ "$operator_sdk_current_version" != "$OPERATOR_SDK_VERSION" ]]; then
```
aravindhp, line 197: Does this account for the SDK not being installed also?

aravindhp (changes requested):

hack/pre-release.sh
wgahnagl: if operator-sdk isn't installed at all, the command that would get the version would fail, and the script would exit. Should we handle this in a different way?

aravindhp: I suggest making the error message more generic.

hack/pre-release.sh
```diff
@@ -187,9 +190,19 @@ switch_git_branch $base_branch $new_branch
 WMCO_VERSION=$(sed -n -e 's/^.*WMCO_VERSION ?= //p' Makefile)
 update_WMCO_version $stream $WMCO_VERSION
 update_OCP_version $stream $WMCO_VERSION 
+
+# check to see if the opeator-sdk version is correct 
+operator_sdk_current_version=$(operator-sdk version |  grep -oP 'operator-sdk version: "\K[^"]+' )
+if [[ "$operator_sdk_current_version" != "$OPERATOR_SDK_VERSION" ]]; then
+  echo "incorrect version. operator-sdk version must be $OPERATOR_SDK_VERSION"
```
aravindhp, line 197:
```suggestion
  echo "operator-sdk $OPERATOR_SDK_VERSION needs to be installed" 
```

aravindhp (changes requested):

aravindhp (approved):

hack/pre-release.sh
```diff
@@ -187,9 +190,19 @@ switch_git_branch $base_branch $new_branch
...
+  echo "operator-sdk $OPERATOR_SDK_VERSION must be installed."  
+  echo "exiting."
+  git checkout $initial_branch
+  exit 
+fi
+
 make bundle
 
-commit_message="[release] Update $base_branch $stream stream to version
+commit_message="[release] Update $base_branch $stream stream to $WMCO_VERSION
```
aravindhp, line 205:
```suggestion
commit_message="[$base_branch] Update version to $WMCO_VERSION`
```
Including the stream in the commit message when you mentioning the full version is a bit odd.

aravindhp (changes requested):

hack/pre-release.sh
aravindhp: This has the added advantage of having the branch prefix when you the open the version bump PR.

PR #1639, opened 2023-06-02
[vendor] Update goio to v1.3.3

aravindhp:
Updated goio entry in go.mod and ran "go mod tidy && go mod vendor"


sebsoto (approved):

aravindhp:
/hold

Until https://github.com/openshift/windows-machine-config-operator/pull/1637 merges which will unblock the vsphere-e2e-operator test

PR #1649, opened 2023-06-08
WINC-950: Test WMCO version upgrades

sebsoto:
Adds the upgrade-test-setup and upgrade-test make targets, which are
designed to be ran before a WMCO upgrade occurs and after,
respectively. These targets will do basic testing on WMCO to ensure that
WMCO is able to upgrade nodes.

### DIFF
--- a/hack/run-ci-e2e-test.sh
+++ b/hack/run-ci-e2e-test.sh
@@ -58,8 +58,8 @@ while getopts ":m:c:b:st:w:" opt; do
       ;;
     t ) # test to run. Defaults to all. Other options are basic and upgrade.
       TEST=$OPTARG
-      if [[ "$TEST" != "all" && "$TEST" != "basic" && "$TEST" != "upgrade" ]]; then
-        echo "Invalid -t option $TEST. Valid options are all, basic or upgrade"
+      if [[ "$TEST" != "all" && "$TEST" != "basic" && "$TEST" != "upgrade" && "$TEST" != "upgrade-setup" && "$TEST" != "upgrade-test" ]]; then
+        echo "Invalid -t option $TEST. Valid options are all, basic, upgrade, upgrade-setup, and upgrade-test"
         exit 1
       fi
       ;;
@@ -140,12 +140,15 @@ GO_TEST_ARGS="$BYOH_NODE_COUNT_OPTION $MACHINE_NODE_COUNT_OPTION --private-key-p
 # Test that the operator is running when the private key secret is not present
 printf "\n####### Testing operator deployed without private key secret #######\n" >> "$ARTIFACT_DIR"/wmco.log
 go test ./test/e2e/... -run=TestWMCO/operator_deployed_without_private_key_secret -v -args $GO_TEST_ARGS
-# Run the creation tests of the Windows VMs
-printf "\n####### Testing creation #######\n" >> "$ARTIFACT_DIR"/wmco.log
-go test ./test/e2e/... -run=TestWMCO/create -v -timeout=90m -args $GO_TEST_ARGS
-# Get logs for the creation tests
-printf "\n####### WMCO logs for creation tests #######\n" >> "$ARTIFACT_DIR"/wmco.log
-get_WMCO_logs
+
+if [[ "$TEST" != "upgrade-setup" && "$TEST" != "upgrade-test" ]]; then
+  # Run the creation tests of the Windows VMs
+  printf "\n####### Testing creation #######\n" >> "$ARTIFACT_DIR"/wmco.log
+  go test ./test/e2e/... -run=TestWMCO/create -v -timeout=90m -args $GO_TEST_ARGS
+  # Get logs for the creation tests
+  printf "\n####### WMCO logs for creation tests #######\n" >> "$ARTIFACT_DIR"/wmco.log
+  get_WMCO_logs
+fi
 
 if [[ "$TEST" = "all" || "$TEST" = "basic" ]]; then
   printf "\n####### Testing network #######\n" >> "$ARTIFACT_DIR"/wmco.log
@@ -171,6 +174,16 @@ if [[ "$TEST" = "all" || "$TEST" = "upgrade" ]]; then
   printf "\n####### Testing reconfiguration #######\n" >> "$ARTIFACT_DIR"/wmco.log
   go test ./test/e2e/... -run=TestWMCO/reconfigure -v -timeout=90m -args $GO_TEST_ARGS
 fi
+
+if [[ "$TEST" = "upgrade-setup" ]]; then
+  go test ./test/e2e/... -run=TestWMCO/create/Creation -v -timeout=90m -args $GO_TEST_ARGS
+  go test ./test/e2e/... -run=TestWMCO/create/Nodes_ready_and_schedulable -v -timeout=90m -args $GO_TEST_ARGS
+fi
+
+if [[ "$TEST" = "upgrade-test" ]]; then
+  go test ./test/e2e/... -run=TestUpgrade -v -timeout=20m -args $GO_TEST_ARGS
+fi
+
 
 # Run the deletion tests while testing operator restart functionality. This will clean up VMs created
 # in the previous step
--- a/test/e2e/create_test.go
+++ b/test/e2e/create_test.go
@@ -45,8 +45,8 @@ const (
 func creationTestSuite(t *testing.T) {
 	tc, err := NewTestContext()
 	require.NoError(t, err)
-	// The order of tests here are important. Any node object related tests should be run only after
-	// testWindowsNodeCreation as that initializes the node objects in the global context.
+
+	require.NoError(t, tc.loadExistingNodes(), "error getting the current Windows nodes in the cluster")
 	if !t.Run("Creation", tc.testWindowsNodeCreation) {
 		// No point in running the other tests if creation failed
 		return
@@ -64,8 +64,26 @@ func creationTestSuite(t *testing.T) {
 	t.Run("Kubelet priority class validation", tc.testKubeletPriorityClass)
 }
 
+// loadExistingNodes adds all configured Windows Nodes to the globalContext cache
+func (tc *testContext) loadExistingNodes() error {
+	byohNodes, err := tc.listFullyConfiguredWindowsNodes(true)
+	if err != nil {
+		return err
+	}
+	machineNodes, err := tc.listFullyConfiguredWindowsNodes(false)
+	if err != nil {
+		return err
+	}
+	gc.byohNodes = byohNodes
+	gc.machineNodes = machineNodes
+	return nil
+}
+
 // testWindowsNodeCreation tests the Windows node creation in the cluster
 func (tc *testContext) testWindowsNodeCreation(t *testing.T) {
+	if len(gc.machineNodes) == numberOfMachineNodes && len(gc.byohNodes) == numberOfBYOHNodes {
+		t.Skip("expected nodes already exist in the cluster")
+	}
 	// Create a private key secret with the known private key.
 	require.NoError(t, tc.createPrivateKeySecret(true), "could not create known private key secret")
 
--- a/test/e2e/create_test.go
+++ b/test/e2e/create_test.go
@@ -20,12 +20,9 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
-	cloudproviderapi "k8s.io/cloud-provider/api"
 
 	"github.com/openshift/windows-machine-config-operator/controllers"
-	"github.com/openshift/windows-machine-config-operator/pkg/crypto"
 	"github.com/openshift/windows-machine-config-operator/pkg/metadata"
-	"github.com/openshift/windows-machine-config-operator/pkg/nodeconfig"
 	"github.com/openshift/windows-machine-config-operator/pkg/patch"
 	"github.com/openshift/windows-machine-config-operator/pkg/retry"
 	"github.com/openshift/windows-machine-config-operator/pkg/wiparser"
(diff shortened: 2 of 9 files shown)


test/e2e/create_test.go
```diff
@@ -42,8 +42,8 @@ const (
 func creationTestSuite(t *testing.T) {
 	tc, err := NewTestContext()
 	require.NoError(t, err)
-	// The order of tests here are important. Any node object related tests should be run only after
-	// testWindowsNodeCreation as that initializes the node objects in the global context.
+
+	require.NoError(t, tc.loadExistingNodes(), "error getting the current Windows nodes in the cluster")
 	if !t.Run("Creation", tc.testWindowsNodeCreation) {
```
aravindhp, line 47: Should we separate this into its own test suite and make this a validationTestSuite()?

aravindhp (approved):
This is great work, @sebsoto. I just have one question but I am approving either way to move things forward.

pkg/daemon/controller/controller.go
```diff
@@ -252,19 +270,50 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
...
+// handleReconcileFailure should be ran when a reconciliation fails due to issues on the Windows instance
+func (sc *ServiceController) handleReconcileFailure(node *core.Node) error {
+	sc.failureCount += 1
+	if sc.failureCount < 5 {
+		return nil
+	}
+	if err := drain.RunCordonOrUncordon(sc.drainer, node, true); err != nil {
+		return err
+	}
+	return nil
```
saifshaikh48, line 300: condense to `return drain.RunCordonOrUncordon(sc.drainer, node, true)`

test/e2e/secrets_test.go
aravindhp, line 1: What does this have to do with upgrade testing? IOW, if you want to include this commit in the upgrade PR please update the commit message with how it is helping with the upgrade testing.

test/e2e/create_test.go
aravindhp, line 1: What does this have to do with upgrade testing?

test/e2e/create_test.go
```diff
@@ -13,23 +13,19 @@ import (
...
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
-	cloudproviderapi "k8s.io/cloud-provider/api"
-
-	"github.com/openshift/windows-machine-config-operator/controllers"
```
aravindhp, line 25: Don't we keep the WMCO packages in a separate stanza?

test/e2e/network_test.go
```diff
@@ -29,9 +29,9 @@ import (
 func testNetwork(t *testing.T) {
 	tc, err := NewTestContext()
 	require.NoError(t, err)
-	err = tc.waitForWindowsNodes(gc.numberOfMachineNodes, false, false, false)
+	err = tc.waitForConfiguredWindowsNodes(gc.numberOfMachineNodes, false, false)
```
aravindhp, line 32: This test needs a node that is schedulable. How are you guaranteeing that now that you have moved to just checking if the node has been configured? This becomes especially important if I am just running the networking test without running the create tests but bringing up nodes outside of the test suite.

hack/run-ci-e2e-test.sh
```diff
@@ -140,12 +140,15 @@ GO_TEST_ARGS="$BYOH_NODE_COUNT_OPTION $MACHINE_NODE_COUNT_OPTION --private-key-p
...
+  # Run the creation tests of the Windows VMs
+  printf "\n####### Testing creation #######\n" >> "$ARTIFACT_DIR"/wmco.log
+  go test ./test/e2e/... -run=TestWMCO/create -v -timeout=90m -args $GO_TEST_ARGS
+  # Get logs for the creation tests
+  printf "\n####### WMCO logs for creation tests #######\n" >> "$ARTIFACT_DIR"/wmco.log
+  get_WMCO_logs
+fi
 
 if [[ "$TEST" = "all" || "$TEST" = "basic" ]]; then
   printf "\n####### Testing network #######\n" >> "$ARTIFACT_DIR"/wmco.log
```
aravindhp, line 154: Shouldn't you doing network tests as part of the upgrade suite?

aravindhp (changes requested):

test/e2e/network_test.go
saifshaikh48: +1, might have to run testNodesBecomeReady() here instead

test/e2e/reconfigure_test.go
saifshaikh48, line 1: > WMCO's controllers will return early, and not change the condition state.

Not sure I understand this. Every WMCO controller changes the upgradeable condition state when its Reconcile function returns without error (via a defer call).

test/e2e/network_test.go
sebsoto: > How are you guaranteeing that now

I am not sure why this needs to be guaranteed within this test?
The test itself will fail if a node is not schedulable. This is the networking test, retesting node statuses that we test within the creation/validation tests is overlapping responsibilities.

test/e2e/create_test.go
```diff
@@ -471,33 +471,6 @@ func (tc *testContext) waitForWindowsNodes(nodeCount int32, expectError, checkVe
 		}
 
 		for _, node := range nodes {
-			// check node status
```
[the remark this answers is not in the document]
sebsoto: Ready != schedulable

hack/run-ci-e2e-test.sh
sebsoto: Maybe? There needs to be a followup WMCO PR to dial in the tests, after the release job is created. In a way this PR is just creating the stubs. I will make this clear in the commit message, and perhaps we can workout out the specifics of this test in that follow-up.

test/e2e/network_test.go
sebsoto: To elaborate, schedulable is a part of the node spec that is applied by WMCO or a user. IMO checking for that would be similar to checking for any taint that isn't tolerated.

test/e2e/secrets_test.go
sebsoto: Commit message updated

test/e2e/create_test.go
sebsoto: Commit message updated

test/e2e/reconfigure_test.go
saifshaikh48: I see, in that case there is time where WMCO is not busy but WICD is. Still, wouldn't this only affect the first check? You are removing both
If you saw both checks causing issues it's fine, I'm just trying to understand why

test/e2e/secrets_test.go
```diff
@@ -95,17 +95,17 @@ func (tc *testContext) testUserDataTamper(t *testing.T) {
...
 	for _, node := range gc.machineNodes {
 		oldNodes = append(oldNodes, node.GetName())
 	}
 
-	// waitForWindowsNodes will re-populate gc.machineNodes with the configured nodes found
+	// waitForConfiguredWindowsNodes will re-populate gc.machineNodes with the configured nodes found
 	log.Printf("waiting for existing Machine nodes to be removed and replaced")
 	return wait.Poll(retryInterval, time.Minute*10, func() (done bool, err error) {
-		err = tc.waitForWindowsNodes(gc.numberOfMachineNodes, false, false, false)
+		err = tc.waitForConfiguredWindowsNodes(gc.numberOfMachineNodes, false, false)
```
saifshaikh48, line 108: we no longer ensure the nodes become schedulable after private key change. is this intended?

test/e2e/create_test.go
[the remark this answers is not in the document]
saifshaikh48: I think you've renamed the function since this comment was originally left. The point still stands. We need to check for ready AND schedulable in more than just the initial create suite.

test/e2e/network_test.go
aravindhp: >I am not sure why this needs to be guaranteed within this test?
The test itself will fail if a node is not schedulable. This is the networking test, retesting node statuses that we test within the creation/validation tests is overlapping responsibilities.

You are assuming the creation test suite was run prior. What if I brought up the nodes manually and then manually ran the just networking test suite? Will `loadExistingNodes()` take care of that?

test/e2e/secrets_test.go
aravindhp: +1

test/e2e/reconfigure_test.go
```diff
@@ -46,10 +46,6 @@ func (tc *testContext) reconfigurationTest(t *testing.T) {
 	_, err = tc.client.K8s.CoreV1().Nodes().Patch(context.TODO(), machineNodes[0].Name, types.JSONPatchType,
 		patchData, metav1.PatchOptions{})
 	require.NoError(t, err)
-	// Ensure operator communicates to OLM that upgrade is not safe when processing Machine nodes
-	err = tc.validateUpgradeableCondition(metav1.ConditionFalse)
```
aravindhp, line 50: > If WICD fixes the change by reconciling the node, before WMCO is able to reconcile the Machine or windows instance ConfigMap, WMCO's controllers will return early, and not change the condition state.

I don't follow the commit message. Isn't this a bug then? I guess I am not understanding the timing issue here.

hack/run-ci-e2e-test.sh
aravindhp: Weren't we running networking tests as part of  the current "upgrade" test? So why is this a maybe?

test/e2e/network_test.go
sebsoto: This is still being tested, its just not a separate check. Specifically, it will be tested by the failure of the windows server deployments to be scheduled on the node. The pod events will state that no nodes meet the scheduling criteria.

test/e2e/reconfigure_test.go
sebsoto: Moved to different PR

test/e2e/secrets_test.go
sebsoto: Added the check below

test/e2e/reconfigure_test.go
sebsoto: Moved to separate PR

test/e2e/validation_test.go
```diff
@@ -53,6 +54,53 @@ type winService struct {
 	description string
 }
 
+// testNodesBecomeReady tests that all Windows nodes become ready and schedulable at least once.
```
aravindhp, line 57: I don't understand the `at least once` comment. Why is that there?

test/e2e/validation_test.go
```diff
@@ -53,6 +54,53 @@ type winService struct {
...
+func (tc *testContext) testNodesBecomeReady(t *testing.T) {
+	nodes := gc.allNodes()
+	for index, node := range nodes {
+		t.Run(node.GetName(), func(t *testing.T) {
+			err := wait.PollImmediate(retry.Interval, retry.ResourceChangeTimeout, func() (done bool, err error) {
+				// Get current node state, and update it in the cache
```
aravindhp, line 63: What cache are you updating? Isn't nodes a local variable?

If you are updating the cache it shouldn't be a side effect of testing if nodes are ready.

test/e2e/validation_test.go
```diff
@@ -53,6 +54,53 @@ type winService struct {
...
+				foundNode, err := tc.client.K8s.CoreV1().Nodes().Get(context.TODO(), node.GetName(), meta.GetOptions{})
+				require.NoError(t, err)
+				nodes[index] = *foundNode
+				return tc.nodeReadyAndSchedulable(*foundNode), nil
```
aravindhp, line 67: This implies, the function should be name `testNodesBecomeReadyAndSchedulable()`. Or am I missing something?

test/e2e/validation_test.go
```diff
@@ -53,6 +54,53 @@ type winService struct {
...
+			})
+			assert.NoError(t, err)
+		})
+	}
+}
+
+func (tc *testContext) nodeReadyAndSchedulable(node core.Node) bool {
```
aravindhp, line 74: Please add a doc

test/e2e/validation_test.go
```diff
@@ -742,13 +790,69 @@ func (tc *testContext) getOperatorConditionName() (string, error) {
...
+
+			pubKey, err := tc.pubKeyAnnotationCorrect(&node)
+			require.NoError(t, err)
+			assert.True(t, pubKey)
+		})
+	}
+}
+
+// usernameAnnotationCorrect checks that the username annotation value is decipherable and correct
+func (tc *testContext) usernameAnnotationCorrect(node *core.Node) (bool, error) {
```
aravindhp, line 815: If you want to go with this sort of naming I suggest `isUsernameAnnotationCorrect()` though I prefer `checkUserNameAnnotation()`

test/e2e/validation_test.go
```diff
@@ -742,13 +790,69 @@ func (tc *testContext) getOperatorConditionName() (string, error) {
...
+		return false, err
+	}
+	if username != tc.vmUsername() {
+		return false, nil
+	}
+	return true, nil
+}
+
+// pubKeyAnnotationCorrect that node is annotated with the public key which matches the private key used to configure it
+func (tc *testContext) pubKeyAnnotationCorrect(node *core.Node) (bool, error) {
```
aravindhp, line 836: Same comment applies to the name here.

aravindhp (changes requested):

test/e2e/validation_test.go
sebsoto: Thought process is that schedulable can be changed by WMCO at will. Removing it for clarity

test/e2e/validation_test.go
sebsoto: gc.Nodes is not a local variable. I agree that it shouldn't be a side effect of testing, but thats already the case with waitForConfiguredWindowsNodes. A large portion of the tests depend on there being an updated node cache available to them via gc, which is why waitForConfiguredNodes have been called at various points in the suite.

I will remove the caching, I don't think it will cause any issues. Being this defensive was an artifact of the cancelled work to move cordoning into WICD.

saifshaikh48 (commented):
/lgtm

Thanks for working on this. e2e tests have been nicely refactored too

aravindhp (approved):
Nice work, @sebsoto

PR #1652, opened 2023-06-20
Move node cordoning to WICD

sebsoto:
This PR includes changes to ensure that Node is only marked as fully configured by WICD by applying the version annotation, only once the node is actually fully configured. This includes the node entering the ready state, and being marked as schedulable.

### DIFF
--- a/pkg/daemon/cleanup/cleanup.go
+++ b/pkg/daemon/cleanup/cleanup.go
@@ -23,11 +23,14 @@ import (
 	"fmt"
 
 	core "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/klog/v2"
+	"k8s.io/kubectl/pkg/drain"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/controller"
+	"github.com/openshift/windows-machine-config-operator/pkg/daemon/drainhelper"
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/manager"
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/powershell"
 	"github.com/openshift/windows-machine-config-operator/pkg/metadata"
@@ -51,6 +54,23 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
 	node, err := controller.GetAssociatedNode(directClient, addrs)
 	if err != nil {
 		klog.Infof("no associated node found")
+	}
+	if node != nil {
+		k8sClient, err := kubernetes.NewForConfig(cfg)
+		if err != nil {
+			return fmt.Errorf("unable to initialize client for drainer")
+		}
+		drainer := drainhelper.NewDrainHelper(ctx, k8sClient, klog.NewKlogr())
+		// Make a best effort attempt to cordon and drain the node before removing any services. We shouldn't block
+		// node cleanup if this fails.
+		err = drain.RunCordonOrUncordon(drainer, node, true)
+		if err != nil {
+			klog.Errorf("unable to cordon node: %s", err)
+		}
+		err = drain.RunNodeDrain(drainer, node.GetName())
+		if err != nil {
+			klog.Errorf("unable to drain node: %s", err)
+		}
 	}
 	svcMgr, err := manager.New()
 	if err != nil {
--- a/pkg/daemon/controller/controller.go
+++ b/pkg/daemon/controller/controller.go
@@ -33,11 +33,13 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/util/jsonpath"
 	"k8s.io/klog/v2"
+	"k8s.io/kubectl/pkg/drain"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -48,6 +50,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	"github.com/openshift/windows-machine-config-operator/pkg/daemon/drainhelper"
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/manager"
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/powershell"
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/winsvc"
@@ -65,6 +68,7 @@ type Options struct {
 	Client    client.Client
 	Mgr       manager.Manager
 	cmdRunner powershell.CommandRunner
+	drainer   *drain.Helper
 }
 
 // setDefaults returns an Options based on the received options, with all nil or empty fields filled in with reasonable
@@ -105,6 +109,7 @@ type ServiceController struct {
 	watchNamespace string
 	psCmdRunner    powershell.CommandRunner
 	ctrl           controller.Controller
+	drainer        *drain.Helper
 }
 
 // Bootstrap starts all Windows services marked as necessary for node bootstrapping as defined in the given data
@@ -152,7 +157,13 @@ func RunController(ctx context.Context, watchNamespace, kubeconfig string) error
 	if err != nil {
 		return fmt.Errorf("unable to start manager: %w", err)
 	}
-	sc, err := NewServiceController(ctx, node.Name, watchNamespace, Options{Client: ctrlMgr.GetClient()})
+
+	k8sClient, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		return fmt.Errorf("unable to initialize client for drainer")
+	}
+	sc, err := NewServiceController(ctx, node.Name, watchNamespace, Options{
+		Client: ctrlMgr.GetClient(), drainer: drainhelper.NewDrainHelper(ctx, k8sClient, klog.NewKlogr())})
 	if err != nil {
 		return err
 	}
@@ -173,7 +184,7 @@ func NewServiceController(ctx context.Context, nodeName, watchNamespace string, 
 		return nil, err
 	}
 	return &ServiceController{client: o.Client, Manager: o.Mgr, ctx: ctx, nodeName: nodeName, psCmdRunner: o.cmdRunner,
-		watchNamespace: watchNamespace}, nil
+		watchNamespace: watchNamespace, drainer: o.drainer}, nil
 }
 
 // SetupWithManager sets up the controller with the Manager.
@@ -232,6 +243,12 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
 		// Error reading the object - requeue the request.
 		return ctrl.Result{}, err
 	}
+	// if the node has never been configured, cordon it until the configuration completes
+	if _, present := node.Annotations[metadata.VersionAnnotation]; !present {
+		if err := drain.RunCordonOrUncordon(sc.drainer, &node, true); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
 
 	desiredVersion, present := node.Annotations[metadata.DesiredVersionAnnotation]
 	if !present {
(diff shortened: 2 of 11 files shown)


pkg/daemon/controller/controller.go
```diff
@@ -105,6 +109,8 @@ type ServiceController struct {
 	watchNamespace string
 	psCmdRunner    powershell.CommandRunner
 	ctrl           controller.Controller
+	drainer        *drain.Helper
```
aravindhp, line 112: Does this need to be part of the struct? You could instantiate in the reconcile function and use it there.

pkg/daemon/controller/controller.go
```diff
@@ -232,6 +244,12 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
 		// Error reading the object - requeue the request.
 		return ctrl.Result{}, err
 	}
+	// if the node has never been configured, cordon it until the configuration completes
```
aravindhp, line 247: Wouldn't you always cordon the node even it has been configured before?

We have also discussed if we need to be draining the node in addition. Do we need to do that?

pkg/daemon/controller/controller.go
```diff
@@ -252,19 +270,50 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
...
-	// Version annotation is the indicator that the node was fully configured by this version of the services ConfigMap
-	if err = metadata.ApplyVersionAnnotation(sc.ctx, sc.client, node, desiredVersion); err != nil {
-		return ctrl.Result{}, fmt.Errorf("error updating version annotation on node %s: %w", sc.nodeName, err)
+	if err = sc.markNodeConfigured(&node, desiredVersion); err != nil {
+		return ctrl.Result{}, err
 	}
 	return ctrl.Result{}, nil
 }
 
+// handleReconcileFailure should be ran when a reconciliation fails due to issues on the Windows instance
```
aravindhp, line 291: I don't understand why we need this function? We should cordon the node at the beginning of configuration and only uncordon if it is has been configured fully. I don't understand why cordon only if we fail more than 5 times.

pkg/daemon/controller/controller.go
```diff
@@ -252,19 +270,50 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
...
+}
+
+// markNodeConfigured marks the node as configured and schedulable
+func (sc *ServiceController) markNodeConfigured(node *core.Node, desiredVersion string) error {
+	err := drain.RunCordonOrUncordon(sc.drainer, node, false)
+	if err != nil {
+		return fmt.Errorf("unable to ensure node is uncordoned: %w", err)
+	}
+	// Version annotation is the indicator that the node was fully configured by this version of the services ConfigMap
+	if err = metadata.ApplyVersionAnnotation(sc.ctx, sc.client, *node, desiredVersion); err != nil {
```
aravindhp, line 310: You need to apply the version annotation first and then uncordon the node.

pkg/metrics/metrics.go
```diff
@@ -136,14 +136,13 @@ func (pc *PrometheusNodeConfig) Configure() error {
 		log.Info("install the prometheus-operator to enable Prometheus configuration")
 		return nil
 	}
-	// get list of Windows nodes that are in Ready phase
```
aravindhp, line 139: >WMCO was made to no longer check for a Node's schedulable state when
determining if the Node's address should be added to the
windows-exporter metrics endpoint, as it has no control over the
schedulability, and due to timing issues, could potentially cause a Node
to not be added to the endpoint. This check was already problematic, as
if a user cordoned and uncordoned a node for maintenance reasons, it
could be removed from the metrics list, and not re-added until WMCO
reconciled a node.

You are actually fixing a bug here. Not sure if this should be included as part of this change. 

Would adding a node that is cordoned or not ready to the metrics endpoint cause any issues?

aravindhp (changes requested):

pkg/daemon/controller/controller.go
sebsoto: I disagree, as applying the version annotation is signaling to WMCO that the node has been completely been configured, and setting spec.unschedulable to false is part of that.

pkg/daemon/controller/controller.go
sebsoto: Either the drainer or the client required to instantiate it needs to be part of the struct.

pkg/metrics/metrics.go
sebsoto: I felt this was appropriate here as cordoning is no longer under WMCO's control and thus should be removed from this check. I can split this off into a separate PR.

> Would adding a node that is cordoned or not ready to the metrics endpoint cause any issues?

Cordoning definitely shouldn't. I don't see why it being ready or not should, the prometheus server is ran by windows-exporter and the metrics collector will just time out when trying to reach it. But that is a normal situation to be handled.

pkg/daemon/controller/controller.go
```diff
@@ -441,6 +446,21 @@ func (sc *ServiceController) resolvePowershellVariables(svc servicescm.Service)
...
 }
 
+// waitUntilNodeReady waits until the Node being configured is ready. Returns an error on timeout.
+func (sc *ServiceController) waitUntilNodeReady() error {
+	return wait.PollUntilContextTimeout(sc.ctx, 5*time.Second, time.Minute, true,
+		func(ctx context.Context) (done bool, err error) {
+			var node core.Node
+			err = sc.client.Get(sc.ctx, client.ObjectKey{Name: sc.nodeName}, &node)
+			for _, condition := range node.Status.Conditions {
+				if condition.Type == core.NodeReady && condition.Status == core.ConditionTrue {
```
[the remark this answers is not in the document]
aravindhp: Won't it be in this state if the kubelet is configured but WICD is configuring the other services?

pkg/daemon/controller/controller.go
aravindhp: I am fine keeping the drainer here then

pkg/daemon/controller/controller.go
aravindhp: I see your point. Would there be an issue if the version annotation does not get applied? i.e. the node start getting configured again just because of failure to update version?

pkg/metrics/metrics.go
aravindhp: >I felt this was appropriate here as cordoning is no longer under WMCO's control and thus should be removed from this check. I can split this off into a separate PR.

What does cordoning ownership have to do with this? This is a bug that would have to be fixed irrespective of cordoning ownership.

aravindhp (changes requested):

pkg/daemon/cleanup/cleanup.go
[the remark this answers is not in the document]
mansikulkarni96: I agree, if we can get this in first, this logic can be reused when cordoning/uncordoning node while restarting instance to set proxy variables.

pkg/daemon/controller/controller.go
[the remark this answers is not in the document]
alinaryan: I'm thinking Scheduling Disabled should also be taken into account. That's what's going on with OCPBUGS-13780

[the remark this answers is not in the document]
alinaryan: or if not here then it should be taken into account somewhere

pkg/daemon/cleanup/cleanup.go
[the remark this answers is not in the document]
alinaryan: I also think this could be moved to a separate PR

pkg/daemon/cleanup/cleanup.go
```diff
@@ -62,6 +70,16 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
 			cmData, err = servicescm.Parse(cm.Data)
 			return err
 		}
+		// Make a best effort attempt to cordon and drain the node before removing any services. We shouldn't block
```
alinaryan, line 73: Shouldn't this also be called in the case where we get the node via config map and VersionAnnotation/DesiredVersionAnnotation? 

Maybe add this change into a function and call it here and after line 98

alinaryan: Nevermind - I saw your comment below "as applying the version annotation is signaling to WMCO that the node has been completely been configured, and setting spec.unschedulable to false is part of that."

pkg/daemon/controller/controller.go
sebsoto: There isn't currently a mechanism in the code to identify a faulty service.

The closest thing to that was removed from this PR due to a previous comment, draining and cordoning if WICD fails to successfully reconcile a service multiple times in a row.

pkg/daemon/controller/controller.go
sebsoto: The node getting configured again in that would be the expected behavior, I can't think of any issues related to cordoning because of that. There would be a blip where the node is momentarily uncordoned and then recordened, as the controller goes to reconcile it again, but it should have no consequence. This would only be until the node is successfully configured.

pkg/daemon/cleanup/cleanup.go
```diff
@@ -52,6 +55,23 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
...
+		// node cleanup if this fails.
+		err = drain.RunCordonOrUncordon(drainer, node, true)
+		if err != nil {
+			klog.Errorf("unable to cordon node: %s", err)
+		}
+		err = drain.RunNodeDrain(drainer, node.GetName())
+		if err != nil {
+			klog.Errorf("unable to drain node: %s", err)
+		}
+	}
```
aravindhp, line 74: Please make this a helper method. We should be able to glance at `Deconfigure()` and understand what it does.

pkg/daemon/cleanup/cleanup.go
```diff
@@ -52,6 +55,23 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
...
 		klog.Infof("no associated node found")
 	}
+	if node != nil {
+		k8sClient, err := kubernetes.NewForConfig(cfg)
+		if err != nil {
+			return fmt.Errorf("unable to initialize client for drainer")
+		}
+		drainer := drainhelper.NewDrainHelper(ctx, k8sClient, klog.NewKlogr())
+		// Make a best effort attempt to cordon and drain the node before removing any services. We shouldn't block
+		// node cleanup if this fails.
```
aravindhp, line 65: Do we need to retry?

pkg/daemon/controller/controller.go
aravindhp: To be clear, by faulty service, I mean a service that was tampered with and was corrected by WICD.

pkg/daemon/controller/controller.go
```diff
@@ -258,13 +275,25 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
...
-		return ctrl.Result{}, fmt.Errorf("error updating version annotation on node %s: %w", sc.nodeName, err)
+	if err = sc.markNodeConfigured(&node, desiredVersion); err != nil {
+		return ctrl.Result{}, err
 	}
 	return ctrl.Result{}, nil
 }
 
+// markNodeConfigured marks the node as configured and schedulable
+func (sc *ServiceController) markNodeConfigured(node *core.Node, desiredVersion string) error {
+	err := drain.RunCordonOrUncordon(sc.drainer, node, false)
```
aravindhp, line 286: I wish there were helpers for Cordon and Uncordon. Please add a comment here saying you are uncordoning.

pkg/daemon/controller/controller.go
aravindhp: I feel we go by the k8s pattern a bit too much with retrying all operations just because the final operation failed.

pkg/daemon/cleanup/cleanup.go
```diff
@@ -52,6 +55,23 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
 	if err != nil {
 		klog.Infof("no associated node found")
 	}
+	if node != nil {
+		k8sClient, err := kubernetes.NewForConfig(cfg)
+		if err != nil {
+			return fmt.Errorf("unable to initialize client for drainer")
+		}
+		drainer := drainhelper.NewDrainHelper(ctx, k8sClient, klog.NewKlogr())
```
aravindhp, line 63: This should read `drainhelper.New()`

pkg/daemon/drainhelper/drainhelper.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func (ow OutWriter) Write(p []byte) (n int, err error) {
+	// log info
+	ow.log.Info(string(p))
+	return len(p), nil
+}
+
+// NewDrainHelper returns new drain.Helper
+func NewDrainHelper(ctx context.Context, client kubernetes.Interface, logger klog.Logger) *drain.Helper {
```
aravindhp, line 35: Just `New()` will suffice.

aravindhp (changes requested):

test/e2e/create_test.go
aravindhp, line 1: What does this commit have to do with the cordoning move?

aravindhp (changes requested):
Please think through upgrade scenarios given RBAC is changing across releases.

pkg/daemon/cleanup/cleanup.go
aravindhp, line 1: Please update your commit message

aravindhp (changes requested):
@sebsoto, does the MCD handling the cordoning or draining of nodes or is it MCO? I am now reconsidering if this is the correct approach. Won't it be possible for all nodes to be drained at the same time?

PR #1656, opened 2023-06-22
Ensure node becomes ready when configuring

sebsoto:
WICD should not be labeling a Node as fully configured if it does not go into the 'Ready' state. Applying the version annotation communicates that this is a working, configured node.


aravindhp (approved):

saifshaikh48 (commented):
/lgtm

PR #1663, opened 2023-06-26
Use operator image to extend must-gather collection with WMCO's specific resources

jrvaldes:
As the OpenShift Windows Machine Config operator increases in complexity and number of customers, it is
important to provide an easy-to-use tool to collect information that can be used to troubleshoot issues.

This PR extends the support for OpenShift must-gather in WMCO ([WINC-462](https://issues.redhat.com/browse/WINC-462)) by preparing the operator image as a compatible "must-gather" image so that the collection scripts can be executed with information that is specific to the operator.

In addition, opened a [mid-stream PR](https://gitlab.cee.redhat.com/openshift-winc-midstream/openshift-winc-midstream/-/merge_requests/291) to include the novel collection tool in the WMCO image.

### Motivation

As per the OpenShift [must-gather guidelines](https://github.com/openshift/must-gather/blob/master/README.md?plain=1#L10-L12):

> The data collection scripts should only include collection logic for components that are included as part of the 
> OpenShift CVO payload. Outside components are encouraged to produce a similar "must-gather" image.

WMCO is an optional day-two operator that is not included as part of the OpenShift CVO payload, so introducing the
logic of the collection script in the WMCO repository reduces the complexity of the OpenShift must-gather image and 
allows updates independently of the CVO payload, considering that WMCO follows a different release cycle.

Eventually, the Windows specific logic currently [included](https://github.com/openshift/must-gather/blob/master/collection-scripts/gather_windows_node_logs#L19) in the OpenShift must-gather image could be ported to the proposed must-gather tools within WMCO.

### Pros and Cons

Pros:
 - Easier to maintain and release independently of the OpenShift must-gather image
 - Exactly match a single version of the operator it is inspecting
 - A single, low-arg client command for users to collect information
 - Reduces the complexity of the OpenShift must-gather image
 
Cons:
 - Increase the size of the WMCO image by ~130MB given that it includes required tools: `oc`, `rsync` and `tar`

### Testing

I'm proposing the usage of the novel collection script as part of the E2E test suite with the required adjustment in the images built in CI and for local development using the approach recommended to the users; i.e. invoking the `oc adm must-gather` passing a custom image pointing to the corresponding operator image (`--image="${OPERATOR_IMAGE}"`). See [[e2e] Use must-gather in CI and local development](https://github.com/openshift/windows-machine-config-operator/pull/1663/commits/6f570d915f75fd61a10a0a304066509f9d000a7c). The latter adds more complexity to the E2E but provides a standard report of the operator state during the E2E execution.

A sample output of the must-gather tool running in a CI job can be found [here](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/1663/pull-ci-openshift-windows-machine-config-operator-master-aws-e2e-operator/1673332891666354176/artifacts/aws-e2e-operator/windows-e2e-operator-test/artifacts/).

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -109,8 +109,19 @@ RUN make build-daemon
 #├── windows_exporter.exe
 #└── windows-instance-config-daemon.exe
 
+FROM registry.redhat.io/openshift4/ose-cli:latest AS cli
+
 FROM registry.access.redhat.com/ubi9/ubi-minimal:latest
 LABEL stage=operator
+
+# install tools required for must-gather collection script
+RUN INSTALL_PKGS=" \
+      rsync \
+      tar \
+      " && \
+    microdnf install -y $INSTALL_PKGS && \
+    rpm -V $INSTALL_PKGS && \
+    microdnf clean all
 
 WORKDIR /payload/
 # Copy WICD
@@ -158,6 +169,11 @@ COPY pkg/internal/windows-defender-exclusion.ps1 .
 COPY pkg/internal/hns.psm1 .
 
 WORKDIR /
+
+# Copy must-gather collection scripts and binaries
+COPY must-gather/collection-scripts/* /usr/bin/
+COPY --from=cli /usr/bin/oc /usr/bin
+RUN ln -s /usr/bin/oc /usr/bin/kubectl
 
 ENV OPERATOR=/usr/local/bin/windows-machine-config-operator \
     USER_UID=1001 \
--- a/must-gather/README.md
+++ b/must-gather/README.md
@@ -1,0 +1,111 @@
+# Support for OpenShift Must-Gather
+This tool is shipped in the operator images as a complement for [OpenShift must-gather](https://github.com/openshift/must-gather)
+to expands its capabilities to gather specific information for the OpenShift Windows Machine Config Operator.
+
+## Usage
+To gather only OpenShift Windows Machine Config Operator information use the following command: 
+```shell script
+oc adm must-gather --image="$(oc get packagemanifests openshift-windows-machine-config-operator \
+  -n openshift-marketplace \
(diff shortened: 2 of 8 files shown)


must-gather/README.md
```diff
@@ -0,0 +1,111 @@
+# Support for OpenShift Must-Gather
```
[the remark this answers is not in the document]
jrvaldes: @aravindhp  Thanks for the early review. 

That was my first approach, but after reviewing the must-gather [source code](https://github.com/openshift/must-gather/tree/master) and [enhancement](https://github.com/openshift/enhancements/blob/master/enhancements/oc/must-gather.md), along with other operators following this pattern. e.g. [CLO](https://github.com/openshift/cluster-logging-operator/tree/master/must-gather) moved it to the project's root for consistency. 

Could you elaborate, from your perspective, on the value of having it in the `hack` directory?

build/Dockerfile
```diff
@@ -109,9 +109,20 @@ RUN make build-daemon
...
 LABEL stage=operator
 
+# install tools required for must-gather collection script
+RUN INSTALL_PKGS=" \
+      rsync \
+      tar \
+      " && \
+    microdnf install -y $INSTALL_PKGS && \
+    rpm -V $INSTALL_PKGS && \
+    microdnf clean all
```
sebsoto, line 124: I'm not sure I'm on board with putting this into the WMCO image.
The WMCO image is already bloated, and adding more dependencies should be justified.

Why can't this be a separate image?

aravindhp: I agree with @sebsoto. In fact the enhancement calls for shipping a separate image.

aravindhp (commented):
Thanks for kicking this off, @jrvaldes. In general this looks useful but we might need to hash out some implementation details. I suggest calling a design meeting to talk to this within the context of a sprint.

PR #1664, opened 2023-06-26
[hack] Fix error when a PR links to multiple bugs

mtnbikenc:
In some cases a PR may reference multiple bugs in the title.  This fix returns only the first bug referenced for looking up details.  The error message is also updated to include more detail for troubleshooting.


alinaryan:
lgtm

aravindhp (approved):
Thanks for fixing this, @mtnbikenc. I tested and verified locally.

PR #1666, opened 2023-06-26
WINC-863: [hack] Automate dist-git version updates

wgahnagl:
Adds automation in the pre-release.sh script to automatically generate a patch for updating dist-git versions on pre-release.


hack/pre-release.sh
```diff
@@ -175,40 +305,16 @@ if [[ $stream != "x" && $stream != "y" && $stream != "z" ]]; then
...
-  echo "exiting."
-  git checkout $initial_branch
-  exit 
-fi
-
-make bundle
 
-commit_message="[$base_branch] Update version to $WMCO_VERSION
+github_update $base_branch $stream $WMCO_VERSION
+dist_git_update $stream $WMCO_VERSION
```
saifshaikh48, line 317: do we think it should be useful to be able to run just the dist-git update portion or just the github update portion?

hack/pre-release.sh
```diff
@@ -21,36 +23,48 @@ usage() {
   echo "  bumps the baseimage versions"
 }
 
-# This function updates the WMCO version in all relevant files 
+# This function increments any xyz version given an x y or z 
 # takes 2 arguments 
-# 1st arg: stream to update. Can be x, y or z to represent the x y or z stream. 
-# 2nd arg: WMCO version to update.
-update_WMCO_version() { 
+# 1st arg: version to update.
```
aravindhp, line 28: What does version mean here? Something like 9.0.0?

hack/pre-release.sh
aravindhp: Wouldn't that be a requirement? We need the GitHub update merged first before we make dist-git changes.

hack/pre-release.sh
```diff
@@ -21,36 +23,48 @@ usage() {
...
-  if [[ $xyz_stream == "z" ]]; then
+  if [[ $2 == "z" ]]; then
     stream_var=3
   fi
- 
+
   # Increment the version
-  updated_version=$(echo $version |
-    awk -F. -v OFS=. -v stream_var="$stream_var" '{$stream_var += 1 ; print}')
+  echo $version | awk -F. -v OFS=. -v stream_var="$stream_var" '{$stream_var += 1 ; print}'
```
aravindhp, line 50: I had asked for this during the first review also. Please add a comment about what you are doing here. Splitting on `.` etc.

hack/pre-release.sh
```diff
@@ -21,36 +23,48 @@ usage() {
...
     stream_var=1
   fi
-  if [[ $xyz_stream == "y" ]]; then
+  if [[ $2 == "y" ]]; then
     stream_var=2
   fi
-  if [[ $xyz_stream == "z" ]]; then
+  if [[ $2 == "z" ]]; then
     stream_var=3
   fi
```
saifshaikh48, line 47: consider using a map. could condense this + make it more readable
```
  declare -A stream_var
  stream_var["x"]=1
  stream_var["y"]=2
  stream_var["z"]=3
```

hack/pre-release.sh
```diff
@@ -105,8 +116,7 @@ update_OCP_version() {
   echo "using OCP version $version"
 
   # Increment the version by using the stream_var variable to select which column to increment 
-  updated_version=$(echo $version |
-    awk -F. -v OFS=. -v stream_var="$stream_var" '{$stream_var += 1 ; print}')
+  updated_version=$(increment_version $wmco_version $stream)
```
aravindhp, line 119: I am confused here. Why are you updating the WMCO version in the `update_OCP_version()`.

hack/pre-release.sh
```diff
@@ -148,6 +158,126 @@ update_OCP_version() {
   fi 
 }
 
+# This function handles updating the versions in GitHub
+# takes 3 arguments
+#
+# 1st arg: base branch to operate on 
+# 2nd arg: stream to update. Can be x, y or z to represent the x y or z stream.#
+# 3rd arg: current WMCO version
+github_update(){
```
aravindhp, line 167: Please be consistent. Space after `()`.

hack/pre-release.sh
```diff
@@ -148,6 +158,126 @@ update_OCP_version() {
...
+  wmco_version=$3
+
+  # check if the git branch is dirty, and exit if it is
+  check_git_branch
+
+  new_branch=$(create_git_branch_list $stream-stream-bump $base_branch)
+  switch_git_branch $base_branch $new_branch
+
+  # Locate and store the WMCO version from the makefile
+  # finds "WMCO_VERSION ?=" and selects text following it
```
aravindhp, line 184: How are you doing this with `update_WMCO_version` and `update_OCP_version`?

hack/pre-release.sh
```diff
@@ -148,6 +158,126 @@ update_OCP_version() {
...
+  update_WMCO_version $stream $wmco_version
+  update_OCP_version $stream $wmco_version
+
+  # check to see if the opeator-sdk version is correct
+  operator_sdk_current_version=$(operator-sdk version |  grep -oP 'operator-sdk version: "\K[^"]+' )
+  if [[ "$operator_sdk_current_version" != "$OPERATOR_SDK_VERSION" ]]; then
```
aravindhp, line 190: This should be an initial check.

hack/pre-release.sh
```diff
@@ -148,6 +158,126 @@ update_OCP_version() {
...
+
+  echo "dist-git location set to $DIST_GIT_LOCATION"
+  current_dir=$PWD
+  cd $DIST_GIT_LOCATION
+
+  check_git_branch
+
+  # switch to the dist-git branch that corresponds with the current ocp version
+  # if a rhel-9 branch doesn't exist, switch to rhel-8
+  base_branch=rhaos-$OCP_version-rhel-9
```
aravindhp, line 240: You should be able to authoritatively figure out if we need to use a rhel-9 or rhel-8 branch and not guess.

hack/pre-release.sh
```diff
@@ -148,6 +158,126 @@ update_OCP_version() {
...
+  if [ "$#" -ne 2 ]; then
+    echo incorrect parameter count for dist_git_update $#
+    return 1
+  fi
+  stream=$1
+  wmco_version=$2 
+
+  # clones the dist-git repo into a temporary directory 
+  # this will be cleaned up at the end 
+  mkdir tmp
```
aravindhp, line 226: Why not use `mktemp`?

hack/pre-release.sh
```diff
@@ -148,6 +158,126 @@ update_OCP_version() {
...
+  generate_commit "$commit_message" build bundle Makefile config
+
+  # Display the newly created branches with push instructions
+  push_message=$(display_push_message "$new_branch")
+  echo
+  echo -e "${PINK}${push_message}${RESET}"
+  echo
+}
+
+# This function handles updating the versions in dist_git. Expects you to be connected to the VPN
```
aravindhp, line 212: This function just generates the patch. The user still needs to apply and push the commit.

hack/pre-release.sh
```diff
@@ -148,6 +158,126 @@ update_OCP_version() {
...
+  
+  # Example: matches version: 8.0.1
+  sed -i "s/version: [0-9]\+\(\.[0-9]\+\)\+/version: $updated_version/g" render_templates
+  
+  # Example: matches previous_version="8.0.1"
+  sed -i "s/previous_version=\"[0-9]\+\(\.[0-9]\+\)\+\"/previous_version=\"$wmco_version\"/g" render_templates
+
+  echo "updated $wmco_version to $updated_version"
+
+  commit_message="This commit bumps the version in render_templates to $updated_version and includes the previous version to be replaced as $wmco_version
```
aravindhp, line 263: Please format this correctly.

hack/pre-release.sh
```diff
@@ -148,6 +158,126 @@ update_OCP_version() {
...
+  generate_commit "$commit_message" Dockerfile render_templates
+  
+  # Generates a patch that can be applied and pushed by the users
+  patch_name=dist-git-$wmco_version-to-$updated_version.patch
+  git format-patch -1 --output=../../$patch_name
+
+  cd $current_dir
+  rm -rf tmp
+  echo
+  echo -e "${PINK}Patch $patch_name created. Apply in your dist-git repo using \n
```
aravindhp, line 276: Please display the patch too

aravindhp (changes requested):
Thanks for working on this, @wgahnagl. Please open a test PR using the script and attach the dist-patch in comment.

hack/pre-release.sh
wgahnagl: I was thinking that if they were all run at the same time, you wouldn't  forget a step. Since it's making a patch, it's your job to apply and push at the right time.

hack/pre-release.sh
```diff
@@ -148,6 +154,124 @@ update_OCP_version() {
   fi 
 }
 
+# This function handles updating the versions in GitHub
+# takes 3 arguments
+#
+# 1st arg: base branch to operate on 
+# 2nd arg: stream to update. Can be x, y or z to represent the x y or z stream.#
```
alinaryan, line 161: nit: extra '#'

hack/pre-release.sh
```diff
@@ -148,6 +154,124 @@ update_OCP_version() {
...
+    echo incorrect parameter count for github_update $#
+    return 1
+  fi
+
+  base_branch=$1
+  stream=$2
+  wmco_version=$3
+
+  # check if the git branch is dirty, and exit if it is
+  check_git_branch
```
alinaryan, line 174: Will this exit here too or just exit common.sh? (also asking for the two other places this is called)

hack/pre-release.sh
wgahnagl: should I make a function in common.sh for this? Or just list all of the branches and select the one with the right version?  Since there are some versions that have both rhel-8 and rhel-9 branches, I feel like I would have to pick between them anyway.

hack/pre-release.sh
```diff
@@ -148,6 +154,124 @@ update_OCP_version() {
...
+  generate_commit "$commit_message" build bundle Makefile config
+
+  # Display the newly created branches with push instructions
+  push_message=$(display_push_message "$new_branch")
+  echo
+  echo -e "${PINK}${push_message}${RESET}"
+  echo
+}
+
+# This function handles updating the versions in dist_git and creates a patch. Expects you to be connected to the VPN
```
alinaryan, line 197: Might wanna call out the VPN prerequisite in the script description

alinaryan (commented):
Thanks! See my review comments

hack/pre-release.sh
```diff
@@ -9,6 +9,8 @@ source $WMCO_ROOT/hack/common.sh
 CVP=false
 BASEIMAGE=false
 OPERATOR_SDK_VERSION=v1.15.0
+PINK='\033[95m'
+RESET='\033[0m'
```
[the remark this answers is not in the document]
aravindhp: Nit: You are assuming the `RESET` color instead of figuring out what the current color is. I could be using a different theme that does not use the `RESET` color by default.

hack/pre-release.sh
```diff
@@ -19,6 +23,30 @@ usage() {
...
+  fi
+  version=$1
+  # Setting the stream var to 1 2 or 3 which is used for incrementing the correct stream
+  # this tells awk which column to act on. Column 1, column 2, or column 3.
+  # which correspond to the x, y and z stream respectively.
+  declare -A stream_var
+  stream_var["x"]=1
+  stream_var["y"]=2
+  stream_var["z"]=3
+  stream_var=${stream_var[$2]}
```
aravindhp, line 46: Are you now overwriting the `stream_var` here? Won't using a new variable be more clear? Or is there some shell magic happening here which is not obvious?

aravindhp (changes requested):

hack/pre-release.sh
wgahnagl: I think we talked about removing dist-git branches that are stale as well, which could make this easier. Is there any reason we have both a rhel-8 and a rhel-9 branch for some versions if we only support one version of rhel on each version?

aravindhp: It doesn't matter. Let us pick the branch authoritatively instead of depending on what is there in dist-git.

hack/pre-release.sh
saifshaikh48: A user should be able to run the two workflows separately IMO

hack/pre-release.sh
```diff
@@ -148,6 +159,128 @@ update_OCP_version() {
...
+  new_branch=$(create_git_branch_list $stream-stream-bump $base_branch)
+  switch_git_branch $base_branch $new_branch
+
+  update_WMCO_version $stream $wmco_version
+  update_OCP_version $stream $wmco_version
+
+  git submodule update --init --recursive
+  make bundle
+
+  commit_message="[$base_branch] Update version to $wmco_version
```
saifshaikh48, line 189: `wmco_version` is the old version so the commit message is not right (see the [test PR](https://github.com/openshift/windows-machine-config-operator/pull/1667#discussion_r1244237797))

hack/pre-release.sh
```diff
@@ -148,6 +159,128 @@ update_OCP_version() {
...
+  check_git_branch
+
+  new_branch=$(create_git_branch_list $stream-stream-bump $base_branch)
+  switch_git_branch $base_branch $new_branch
+
+  update_WMCO_version $stream $wmco_version
+  update_OCP_version $stream $wmco_version
+
+  git submodule update --init --recursive
+  make bundle
```
saifshaikh48, line 187: see comment on [test PR](https://github.com/openshift/windows-machine-config-operator/pull/1667#discussion_r1251081794), which is a consequence of make bundle

hack/pre-release.sh
saifshaikh48: I don't think this has been addressed yet/

hack/pre-release.sh
```diff
@@ -175,40 +316,22 @@ if [[ $stream != "x" && $stream != "y" && $stream != "z" ]]; then
...
-operator_sdk_current_version=$(operator-sdk version |  grep -oP 'operator-sdk version: "\K[^"]+' )
-if [[ "$operator_sdk_current_version" != "$OPERATOR_SDK_VERSION" ]]; then
-  echo "operator-sdk $OPERATOR_SDK_VERSION must be installed."  
-  echo "exiting."
-  git checkout $initial_branch
-  exit 
-fi
 
-make bundle
+github_update $base_branch $stream $WMCO_VERSION
```
saifshaikh48, line 330: I would have this run only `if DISTGIT == "false"` (which should be default)

hack/pre-release.sh
```diff
@@ -148,11 +162,139 @@ update_OCP_version() {
...
+  update_OCP_version $stream $wmco_version
+
+  git submodule update --init --recursive
+  make bundle
+  
+  # remove make bundle artifacts 
+  sed -i 's/REPLACE_IMAGE:latest/REPLACE_IMAGE:/' bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
+  sed -i 's/operator-sdk-v1.14.0+git/operator-sdk-v1.15.0+git/' bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
+
+  updated_version=$(increment_version $version $stream)
```
saifshaikh48, line 196: seems odd you have to do this -- this is the third time you call `increment_version` within this function (it is also called above from update_WMCO_version and update_OCP_version). can update_WMCO_version return the updated version?

wgahnagl: I agree, but I was thinking that it wouldn't make sense for just update_wmco_version to return the incremented version 
I could pass the incremented version in as an argument, but I felt like that would be kind of odd too.

saifshaikh48 (commented):
almost lgtm. please post an example distgit patch generated using this script

hack/pre-release.sh
```diff
@@ -148,11 +162,139 @@ update_OCP_version() {
...
+  exit
+fi
+
 # This sets the CVP and baseimage flags
-while getopts "bc" opt; do 
+while getopts "bcd" opt; do 
   case "$opt" in 
     c) CVP=true;;
     b) BASEIMAGE=true;;
+    d) DISTGIT=true;;
```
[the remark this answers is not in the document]
wgahnagl: I'm not totally sure what you mean by this? 
it [defaults to false ](https://github.com/openshift/windows-machine-config-operator/blob/bcb0ef4620ce8bd1a3af898ada2292a1cde741e3/hack/pre-release.sh#L11 )and then sets to true if the flag is enabled.

hack/pre-release.sh
aravindhp: Why does it not make sense for `update_wmco_version` to return the incremented version? Nor do I see passing the version you want to set to `update_wmco_version` as odd. But calling `increment_version` all over the place is definitely inefficient and could lead to bugs. Ideally we should increment the version in one place and then pass it to functions that update files in the repo.

aravindhp (changes requested):

hack/common.sh
```diff
@@ -17,7 +17,6 @@ error-exit() {
     exit 1
 }
 
-# Accepts one argument, the WMCO version.
```
aravindhp, line 20: Is there a reason to remove this comment?

hack/common.sh
```diff
@@ -33,6 +32,16 @@ get_OCP_version() {
   echo $OCP_VER_MAJOR.$OCP_VER_MINOR
 }
 
+# Accepts one argument, the OCP version.
+get_rhel_version(){
```
aravindhp, line 36: This is missing input check

aravindhp (changes requested):

hack/pre-release.sh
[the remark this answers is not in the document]
saifshaikh48: oops misread, carry on

hack/pre-release.sh
```diff
@@ -19,39 +25,53 @@ usage() {
...
+  printf "${START_COLOR}"
 }
 
-# This function updates the WMCO version in all relevant files 
+# This function increments any xyz version given an x y or z
 # takes 2 arguments 
-# 1st arg: stream to update. Can be x, y or z to represent the x y or z stream. 
+# 1st arg: version to update in x.y.z format
+# 2nd arg: stream to update. Can be x, y or z to represent the x y or z stream.
+increment_version() {
```
aravindhp, line 38: Please update doc to say what the function returns.

hack/pre-release.sh
```diff
@@ -148,11 +163,143 @@ update_OCP_version() {
...
+
+  # Example: matches version: 8.0.1
+  sed -i "s/version: [0-9]\+\(\.[0-9]\+\)\+/version: $updated_version/g" render_templates
+
+  # Example: matches previous_version="8.0.1"
+  sed -i "s/previous_version=\"[0-9]\+\(\.[0-9]\+\)\+\"/previous_version=\"$wmco_version\"/g" render_templates
+
+  echo "updated $wmco_version to $updated_version"
+
+  commit_message="[release-$OCP_version]Prepare for $updated_version release
```
aravindhp, line 268:
```suggestion
  commit_message="[release-$OCP_version] Prepare for $updated_version release
```

aravindhp (commented):
Please respond to https://github.com/openshift/windows-machine-config-operator/pull/1667/files/5b4e49f5d65785bc80f249aa3ecab8afd47d3756#r1256500351 to ensure there is no issue with the script.

aravindhp (approved):

aravindhp:
/hold

For @saifshaikh48's review

saifshaikh48:
/lgtm

Nice work

wgahnagl:
/retest-required 

It seems like a flake

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/1666/pull-ci-openshift-windows-machine-config-operator-master-azure-e2e-operator/1679118445662179328

```level=info msg=Waiting up to 30m0s (until 3:08PM UTC) to ensure each cluster operator has finished progressing...
I0712 14:38:18.842025     128 request.go:696] Waited for 1.001190128s due to client-side throttling, not priority and fairness, request: GET:https://api.ci-op-skt95dty-8411c.XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX:6443/apis/config.openshift.io/v1/clusteroperators?fieldSelector=metadata.name%3Dopenshift-samples&limit=500&resourceVersion=0
level=error msg=Cluster operator etcd does not meet stability threshold of Progressing=false for greater than 30 seconds with Reason: "NodeInstaller" and Message: "NodeInstallerProgressing: 2 nodes are at revision 6; 1 nodes are at revision 7"
Installer exit with code 7
```

aravindhp:
Lets hold this until #1649 merges to avoid unnecessary re-tests on that PR. Once that merges, I will override the tests on this PR and get it merged.

PR #1667, opened 2023-06-27
[TEST DO NOT MERGE][release-4.13] Update version to 8.1.0

wgahnagl:

this is a test


Makefile
```diff
@@ -3,7 +3,7 @@
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
```
aravindhp, line 3: The commit message header is incorrect. It should be `[release-4.13] Update version to 8.0.2`

aravindhp (changes requested):
@wgahnagl please fix the commit message in https://github.com/openshift/windows-machine-config-operator/pull/1666

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
```diff
@@ -395,7 +395,7 @@ spec:
                       fieldPath: metadata.name
                 - name: OPERATOR_NAME
                   value: windows-machine-config-operator
-                image: REPLACE_IMAGE
+                image: REPLACE_IMAGE:latest
```
saifshaikh48, line 398: we should avoid committing this

dist-git-8.0.1-to-8.1.1.patch
```diff
@@ -0,0 +1,51 @@
...
+
+diff --git a/Dockerfile b/Dockerfile
+index c5b1bfe..39abdc0 100644
+--- a/Dockerfile
++++ b/Dockerfile
+@@ -24,7 +24,7 @@ LABEL com.redhat.openshift.versions="=v4.13"
+ LABEL com.redhat.delivery.backport=false
+ 
+ # This label gets replaced by render_templates, so it is safer to keep it on its own
+-LABEL version="v8.0.0"
```
aravindhp, line 22: So I see `v8.0.0` here..

dist-git-8.0.1-to-8.1.1.patch
```diff
@@ -0,0 +1,51 @@
...
+ LABEL com.redhat.component="windows-machine-config-operator-bundle-container"
+diff --git a/render_templates b/render_templates
+index 971e48a..b386a01 100755
+--- a/render_templates
++++ b/render_templates
+@@ -45,7 +45,7 @@ update_csv() {
+     sed -i "s|REPLACE_IMAGE|${wmco_image_sha}|g" ${csv}
+     # Following command adds the replaces field in the CSV using the previous released version of the operator.
+     # TODO: Uncomment below for next minor release
+-    sed -i '/version: 8.0.1/i \  replaces: '${wmco}'.v'${previous_version}'' ${csv}
```
aravindhp, line 35: But `8.0.1` here...

dist-git-8.0.1-to-8.1.1.patch
```diff
@@ -0,0 +1,51 @@
...
++    sed -i '/version: 8.1.1/i \  replaces: '${wmco}'.v'${previous_version}'' ${csv}
+     sed -i "s|REPLACE_DATE|`date "+%Y-%m-%dT%H:%M:%SZ"`|g" ${csv}
+     sed -i "/command:/d" ${csv}
+     awk '{print} sub(/- args:/,"- command:")' ${csv} > tmp && mv tmp ${csv}
+@@ -85,7 +85,7 @@ wmco_image_sha=`echo ${build_info} | jq -e .extra.image.index.pull[0]`
+ wmco="windows-machine-config-operator"
+ csv="manifests/${wmco}.clusterserviceversion.yaml"
+ # TODO: Uncomment below for next minor release
+-previous_version="8.0.0"
++previous_version="8.0.1"
```
aravindhp, line 45: And previous version is now `8.0.1` but you started by replacing `v8.0.0`. This does not seem correct.

wgahnagl: so on the 4.13 branch I ran this from, the [WMCO version is 8.0.1](https://github.com/openshift/windows-machine-config-operator/blob/d0be5e812c3e3914d3cf3c464623e688673565fb/Makefile#L6), and this was a y-stream update, so the correct version would be 8.1.1 

I think The 8.0.0 you see at the top was there because of the comment that said it gets replaced by render-templates, but it never got replaced

The matching for the render-templates is done intentionally to match on any version in the right spot, since I've seen that these files can have incorrect versions in them that only get changed when doing the post-release work.

aravindhp: >so on the 4.13 branch I ran this from, the [WMCO version is 8.0.1](https://github.com/openshift/windows-machine-config-operator/blob/d0be5e812c3e3914d3cf3c464623e688673565fb/Makefile#L6), and this was a y-stream update, so the correct version would be 8.1.1

That is incorrect. A Y stream update for `8.0.1` will be `8.1.0`.

>I think The 8.0.0 you see at the top was there because of the comment that said it gets replaced by render-templates, but it never got replaced

This is worrisome. Aren't we releasing 8.0.1? Why is it still at 8.0.0?

>I've seen that these files can have incorrect versions in them that only get changed when doing the post-release work.

Haven't we done the post release work for 8.0.0 which would have bumped it to 8.0.1?

dist-git-8.0.1-to-8.1.0.patch
```diff
@@ -0,0 +1,51 @@
...
+diff --git a/render_templates b/render_templates
+index 0c25692..e526ec6 100755
+--- a/render_templates
++++ b/render_templates
+@@ -45,7 +45,7 @@ update_csv() {
+     sed -i "s|REPLACE_IMAGE|${wmco_image_sha}|g" ${csv}
+     # Following command adds the replaces field in the CSV using the previous released version of the operator.
+     # TODO: Uncomment below for next minor release
+-    # sed -i '/version: 8.0.1/i \  replaces: '${wmco}'.v'${previous_version}'' ${csv}
++    # sed -i '/version: 8.1.0/i \  replaces: '${wmco}'.v'${previous_version}'' ${csv}
```
saifshaikh48, line 36: When releasing minor versions this line (and the previous_version line) below should be uncommented. And only when releasing a `x.0.0` should both be commented out

saifshaikh48: This is a manual step right now as listed in our [release docs](https://docs.google.com/document/d/1F5TXHgvUcca5Pzsto2gx-eA8Hm-Eb47xxStWU35VgeA/edit#heading=h.f1l61286xjes). It would be great if the pre-release script ensured this for us

wgahnagl: :+1: can do!

PR #1668, opened 2023-06-27
[wicd][controller] Fix Reconcile unit test

saifshaikh48:
This commit fixes the WICD controller unit tests by adding a missing assert
for no error in the Reconcile func happy path.
waitUntilNodeReady() was introduced, which waits for a ready/schedulable node.
Before this, the test was passing despite the mock node not having these labels.

### DIFF
--- a/pkg/daemon/controller/controller_test.go
+++ b/pkg/daemon/controller/controller_test.go
@@ -648,6 +648,14 @@ func TestReconcile(t *testing.T) {
 							metadata.DesiredVersionAnnotation: desiredVersion,
 						},
 					},
+					Status: core.NodeStatus{
+						Conditions: []core.NodeCondition{
+							{
+								Type:   core.NodeReady,
+								Status: core.ConditionTrue,
+							},
+						},
+					},
 				},
 				cm,
 			}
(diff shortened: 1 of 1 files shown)


pkg/daemon/controller/controller_test.go
```diff
@@ -648,6 +648,14 @@ func TestReconcile(t *testing.T) {
 							metadata.DesiredVersionAnnotation: desiredVersion,
 						},
 					},
+					Status: core.NodeStatus{
```
sebsoto, line 651: Can you explain why this only causes a 1 min wait and not a failure?

saifshaikh48: Ah looks like we are missing an assert.NoError() below

sebsoto (approved):

aravindhp (approved):

/lgtm

aravindhp:
/override ci/prow/aws-e2e-upgrade /override ci/prow/aws-e2e-upgrade  ci/prow/vsphere-e2e-operator

This is a unit test fix and does not require the e2e tests to pass

aravindhp:
/override ci/prow/aws-e2e-upgrade ci/prow/aws-e2e-upgrade  ci/prow/vsphere-e2e-operator

PR #1676, opened 2023-07-05
Harden userdata tamper test

sebsoto:
The userdata tamper test has been made to retry if previously configured
nodes were found, instead of failing. This prevents timing issues that
are being seen when refactoring the creation suite, to allow for the
addition of "real" upgrade testing.

### DIFF
--- a/test/e2e/secrets_test.go
+++ b/test/e2e/secrets_test.go
@@ -11,6 +11,7 @@ import (
 	"io/ioutil"
 	"log"
 	"testing"
+	"time"
 
 	config "github.com/openshift/api/config/v1"
 	"github.com/stretchr/testify/assert"
@@ -90,7 +91,6 @@ func (tc *testContext) testUserDataTamper(t *testing.T) {
 
 	// Updating the userdata with incorrect contents will cause the Machine nodes to be deleted and recreated, wait
 	// until the Machine is back up.
-	log.Printf("waiting for Machine nodes to be recreated after userdata update")
 	assert.NoError(t, tc.waitForNewMachineNodes(), "error waiting for Machine nodes to be reconfigured")
 	assert.NoError(t, tc.waitForValidUserData(validUserDataSecret), "error waiting for valid userdata")
 }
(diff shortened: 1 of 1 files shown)


aravindhp (approved):

test/e2e/reconfigure_test.go
```diff
@@ -46,10 +46,6 @@ func (tc *testContext) reconfigurationTest(t *testing.T) {
 	_, err = tc.client.K8s.CoreV1().Nodes().Patch(context.TODO(), machineNodes[0].Name, types.JSONPatchType,
 		patchData, metav1.PatchOptions{})
 	require.NoError(t, err)
-	// Ensure operator communicates to OLM that upgrade is not safe when processing Machine nodes
-	err = tc.validateUpgradeableCondition(metav1.ConditionFalse)
```
aravindhp, line 50: In your other PR, you removed both `validateUpgradeableCondition()` calls from the reconfiguration test. Did you only remove one on purpose?

I still don't understand why this is not a bug.

sebsoto: I removed only one because @saifshaikh48 still wanted the second one if it was not causing issues, which it was not.

This isn't a bug because its an issue with our end to end tests, the operator is functioning properly.

aravindhp: @sebsoto like we were discussing on Slack, you have discovered a scenario where WICD is working but WMCO is not and is marking the operator as upgradable. That is a bug. I am fine if you open a bug, comment out the code, add a comment linking the bug and saying this code needs to be uncommented as part of fixing the bug.

test/e2e/reconfigure_test.go
```diff
@@ -46,9 +46,11 @@ func (tc *testContext) reconfigurationTest(t *testing.T) {
 	_, err = tc.client.K8s.CoreV1().Nodes().Patch(context.TODO(), machineNodes[0].Name, types.JSONPatchType,
 		patchData, metav1.PatchOptions{})
 	require.NoError(t, err)
+	// TODO: This is an unreliable check, which will fail if WICD reconciles the node before WMCO is aware of the
```
aravindhp, line 49: This is not an unreliable check but a useful one that found the bug 😄

test/e2e/reconfigure_test.go
aravindhp, line 1: Please reframe your commit message to indicate how the check revealed a bug and not a timing issue. IOW, you have discovered a scenario where WICD is working but WMCO is not and is marking the operator as upgradable.

The PR heading is also misleading. I know this is annoying I would prefer this commit to be it's own PR.

aravindhp (changes requested):

aravindhp (approved):
Thanks for splitting this out, @sebsoto

saifshaikh48 (commented):
/lgtm

PR #1681, opened 2023-07-06
Remove condition check in reconfigure test

sebsoto:
Removes the OperatorCondition check in the reconfigure test. The check
is causing issues, as the reconfigure test is changing the
version annotation, which is managed by WICD. If WICD fixes the change
by reconciling the node, before WMCO is able to reconcile the Machine or
windows-instance ConfigMap, WMCO's controllers will return early, and
not change the condition state.

The test discovered a scenario where WICD is working but WMCO is not
and is marking the operator as upgradable which is a bug and we are
temporarily disabling it. This will be addressed in the bug linked
in the TODO.

The timeline for this issue is as follows:
* Version annotation change happens
* WMCO is busy
* WICD sees that version annotation is incorrect
* WICD corrects version annotation state
* WMCO becomes un-busy
* WMCO sees nothing to reconcile and returns sucessfully

### DIFF
--- a/test/e2e/reconfigure_test.go
+++ b/test/e2e/reconfigure_test.go
@@ -46,9 +46,11 @@ func (tc *testContext) reconfigurationTest(t *testing.T) {
 	_, err = tc.client.K8s.CoreV1().Nodes().Patch(context.TODO(), machineNodes[0].Name, types.JSONPatchType,
 		patchData, metav1.PatchOptions{})
 	require.NoError(t, err)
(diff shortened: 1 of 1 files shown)


aravindhp (approved):
I would like to see the verbiage, `The test discovered a scenario where WICD is working but WMCO is not and is marking the operator as upgradable which is a bug and we are temporarily disabling it.` in the commit message.

saifshaikh48 (commented):
/lgtm

sebsoto:
/test aws-e2e-upgrade

daemonset flake

PR #1683, opened 2023-07-10
[build] Disable cgo in build containers

sebsoto:
cgo is now being enabled by the base image we use. This is causing containerd builds to fail with:
unrecognized command-line option '-mthreads'; did you mean '-pthread'?

In this commit cgo is being explicity disabled to fix this issue and unblock builds.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,6 +1,9 @@
 FROM registry.ci.openshift.org/openshift/release:rhel-9-release-golang-1.20-openshift-4.14 as build
 LABEL stage=build
 
+# Disable cgo as it breaks containerd builds
+ENV CGO_ENABLED=0
+
 # dos2unix is needed to build CNI plugins
 RUN yum install -y dos2unix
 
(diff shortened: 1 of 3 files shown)


build/Dockerfile
```diff
@@ -1,6 +1,9 @@
 FROM registry.ci.openshift.org/openshift/release:rhel-9-release-golang-1.20-openshift-4.14 as build
 LABEL stage=build
 
+# Disable cgo as it breaks containerd builds
+ENV CGO_ENABLED=false
```
vfreex, line 5: Use `CGO_ENABLED=0` instead.

aravindhp: Now that you are doing it here, you can remove `CGO_ENABLED=0` under the CNI plugins.

aravindhp (changes requested):

build/Dockerfile.ci
```diff
@@ -7,7 +7,7 @@ FROM registry.ci.openshift.org/openshift/release:rhel-9-release-golang-1.20-open
 LABEL stage=build
 
 # Disable cgo as it breaks containerd builds
-ENV CGO_ENABLED=false
+ENV CGO_ENABLED=0
```
aravindhp, line 10: @sebsoto feel free to drop this commit. I added just to test the build.

aravindhp:
/close

This is no longer needed as the builder image has been updated with the script to exempt Windows builds.

PR #1689, opened 2023-07-13
[nodeconfig] Set node in NodeConfig constructor

saifshaikh48:
Allow a NodeConfig object to populate its node field on construction. 
This removed the need to call setNode to populate the node object in each of the exported NodeConfig methods. 
It also removed the need to pass in node objects to functions upon the NodeConfig struct, which was a bad pattern.

### DIFF
--- a/pkg/nodeconfig/nodeconfig.go
+++ b/pkg/nodeconfig/nodeconfig.go
@@ -111,9 +111,8 @@ func (ow OutWriter) Write(p []byte) (n int, err error) {
 func NewNodeConfig(c client.Client, clientset *kubernetes.Clientset, clusterServiceCIDR, wmcoNamespace string,
 	instanceInfo *instance.Info, signer ssh.Signer, additionalLabels,
 	additionalAnnotations map[string]string, platformType configv1.PlatformType) (*nodeConfig, error) {
-	var err error
 
-	if err = cluster.ValidateCIDR(clusterServiceCIDR); err != nil {
+	if err := cluster.ValidateCIDR(clusterServiceCIDR); err != nil {
 		return nil, fmt.Errorf("error receiving valid CIDR value for "+
 			"creating new node config: %w", err)
 	}
@@ -129,8 +128,8 @@ func NewNodeConfig(c client.Client, clientset *kubernetes.Clientset, clusterServ
 		return nil, fmt.Errorf("error instantiating Windows instance from VM: %w", err)
 	}
 
-	return &nodeConfig{client: c, k8sclientset: clientset, Windows: win, platformType: platformType,
-		wmcoNamespace: wmcoNamespace, clusterServiceCIDR: clusterServiceCIDR,
+	return &nodeConfig{client: c, k8sclientset: clientset, Windows: win, node: instanceInfo.Node,
+		platformType: platformType, wmcoNamespace: wmcoNamespace, clusterServiceCIDR: clusterServiceCIDR,
 		publicKeyHash: CreatePubKeyHashAnnotation(signer.PublicKey()), log: log, additionalLabels: additionalLabels,
 		additionalAnnotations: additionalAnnotations}, nil
 }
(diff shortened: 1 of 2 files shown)


aravindhp (commented):
Can we remove `setNode()` at this point?

pkg/nodeconfig/nodeconfig.go
```diff
@@ -111,9 +111,8 @@ func (ow OutWriter) Write(p []byte) (n int, err error) {
 func NewNodeConfig(c client.Client, clientset *kubernetes.Clientset, clusterServiceCIDR, wmcoNamespace string,
 	instanceInfo *instance.Info, signer ssh.Signer, additionalLabels,
 	additionalAnnotations map[string]string, platformType configv1.PlatformType) (*nodeConfig, error) {
-	var err error
 
-	if err = cluster.ValidateCIDR(clusterServiceCIDR); err != nil {
+	if err := cluster.ValidateCIDR(clusterServiceCIDR); err != nil {
```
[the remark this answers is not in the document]
saifshaikh48: Yup, adding a `:`

saifshaikh48:
@aravindhp we can't remove setNode() yet as there is still one use case for it -- when configuring an instance for the first time, after we bootstrap/start kubelet, we need to call setNode() to populate the NodeConfig's node

pkg/nodeconfig/nodeconfig.go
```diff
@@ -162,9 +161,11 @@ func (nc *nodeConfig) Configure() error {
 
 	// Perform rest of the configuration with the kubelet running
 	err = func() error {
-		// populate node object in nodeConfig in the case of a new Windows instance
-		if err := nc.setNode(false); err != nil {
-			return fmt.Errorf("error getting node object: %w", err)
+		if nc.node == nil {
+			// populate node object in nodeConfig in the case of a new Windows instance
+			if err := nc.setNode(false); err != nil {
+				return fmt.Errorf("error getting node object: %w", err)
```
aravindhp, line 167: Please make the error message `error setting node object`

aravindhp (changes requested):
>we can't remove setNode() yet as there is still one use case for it -- when configuring an instance for the first time, after we bootstrap/start kubelet, we need to call setNode() to populate the NodeConfig's node

This is hinting at that we need to extract the bootstrap into it's own method outside of `nc.Configure()` and protect `nc.Configure()` to not run if a node object is present. By extension we also extract the post configuration steps like waiting for WICD to finish the configuration. That feels more elegant. Not saying it is easy to do or that it needs to be done as part of this PR.

mansikulkarni96:
LGTM

pkg/nodeconfig/nodeconfig.go
```diff
@@ -244,9 +245,13 @@ func (nc *nodeConfig) Configure() error {
...
 
-// safeReboot safely restarts the instance associated with the given node object, first cordoning and draining the node
-func (nc *nodeConfig) SafeReboot(ctx context.Context, node *core.Node) error {
-	nc.node = node
+// safeReboot safely restarts the underlying instance, first cordoning and draining the associated node.
+// Waits for reboot to take effect before uncordoning the node.
+func (nc *nodeConfig) SafeReboot(ctx context.Context) error {
+	if nc.node == nil {
+		return fmt.Errorf("safe reboot of the instance requires an associated node")
+	}
```
jrvaldes, line 253: Consider adding a unit test with the new signature of the function to cover the edge case where `nc.node == nil`, and an error is thrown.

jrvaldes (commented):
@saifshaikh48 Thanks for working on this. PTAL at my comment, otherwise LGTM.

pkg/nodeconfig/nodeconfig.go
saifshaikh48: IMO the work required to mock the clients and resources needed to unit test this func is too costly compared to the benefit of the unit test

aravindhp (approved):

jrvaldes (approved):
/lgtm

saifshaikh48:
/test vsphere-e2e-operator

machine deletion flake in new vCenter

PR #1691, opened 2023-07-17
[hack] Remove version gen for csi-proxy 

wgahnagl:
Version was being generated for csi-proxy, which is not required.
Removes this requirement.

### DIFF
--- a/hack/update_submodules.sh
+++ b/hack/update_submodules.sh
@@ -92,7 +92,7 @@ function update_submodules_for_branch() {
     # Update the submodule to the latest remote commit
     git submodule update --remote $submodule
     generate_submodule_commit $submodule
-    if [ "$submodule" = "kubelet" ] || [ "$submodule" = "kube-proxy" ] || [ "$submodule" = "containerd" ] || [ "$submodule" = "csi-proxy" ]; then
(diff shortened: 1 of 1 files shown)


hack/update_submodules.sh
```diff
@@ -39,7 +39,8 @@ function generate_version_commit() {
   if [ "$submodule" = "containerd" ]; then
     # Ref: Containerd Version https://github.com/containerd/containerd/blob/main/Makefile#L33
     version=$(git describe --match 'v[0-9]*' --dirty='.m' --always)
-  else
+  fi 
+  if [ "$submodule" = "kubelet" ] || [ "$submodule" = "kube-proxy" ]; then
```
aravindhp, line 43: This is not the correct fix. `generate_version_commit` should not be called for `csi-proxy` in the first place. Please see line 96.

aravindhp (changes requested):

aravindhp (changes requested):
@wgahnagl please update your PR and commit message with the actual fix you are doing. Reflect that in the header too. Example: `[hack] Remove version gen for csi-proxy`

aravindhp (approved):
Nit: Please update commit message: `Remove this requirement`.

aravindhp:

`hack/update_submodules.sh` is not used in CI or in the product

PR #1697, opened 2023-07-19
WINC-945: [hack] Automate gitlab version updates

wgahnagl:
Enables gitlab version updates to be performed from the pre-release script.

### DIFF
--- a/hack/pre-release.sh
+++ b/hack/pre-release.sh
@@ -10,6 +10,7 @@ source $WMCO_ROOT/hack/common.sh
 CVP=false
 BASEIMAGE=false
 DISTGIT=false
+GITLAB=false
 OPERATOR_SDK_VERSION=v1.15.0
 
 # codes for printing and resetting pink text
@@ -27,6 +28,8 @@ usage() {
   echo "  bumps the baseimage versions"
   echo "optional flags: -d"
   echo "  generates a patch to bump the dist-git version"
+  echo "optional flags: -g"
+  echo "  bumps the gitlab versions"
   echo -e "  ${PINK}Requires connection to the RedHat VPN"
   printf "${START_COLOR}"
 }
@@ -65,6 +68,27 @@ increment_version() {
   echo $version | awk -F. -v OFS=. -v stream="$stream" '{$stream += 1 ; print}'
 }
 
+# This function clones a repo to the current dir, and handles errors.
+# takes 1 argument
+# 1st arg: git URL to clone
+clone_repo() {
+  if [ "$#" -ne 1 ]; then
+     echo incorrect parameter count for clone_repo $#
+    return 1
+  fi
+
+  set +o errexit
+  git clone $1
+  if [ $? -ne 0 ]; then
+    echo -e "${PINK}git clone failed. Are you connected to the VPN?"
+    printf "${START_COLOR}"
+    rm -rf $temp
+    exit 1
+  fi
+  set -o errexit
+
+}
+
 # This function updates the WMCO version in all relevant files
 # takes 3 arguments
 # 1st arg: stream to update. Can be x, y or z to represent the x y or z stream.
@@ -81,9 +105,6 @@ update_WMCO_version() {
   updated_version=$3
   cluster_service_version_yaml=config/manifests/bases/windows-machine-config-operator.clusterserviceversion.yaml
  
-  last_version=$(grep "olm.skipRange" $cluster_service_version_yaml |
-    grep -oP ">=\K([0-9]+\.[0-9]+\.[0-9]+)\s*<" | sed "s/\s*<//")
-
   echo "WMCO version $version detected from the Makefile"
 
   echo "Updating the Makefile to version $updated_version"
@@ -95,6 +116,9 @@ update_WMCO_version() {
   sed -i "s/<$version/<$updated_version/g" $cluster_service_version_yaml
   
   if [[ $stream == "x" ]]; then
+    last_version=$(grep "olm.skipRange" $cluster_service_version_yaml |
+    grep -oP ">=\K([0-9]+\.[0-9]+\.[0-9]+)\s*<" | sed "s/\s*<//")
+
     echo "Updating skipRange lower bound in $cluster_service_version_yaml to $version"
     # Example: matches olm.skipRange: '*>=8.0.0* <9.0.0'
     sed -i "s/>=$last_version/>=$version/g" $cluster_service_version_yaml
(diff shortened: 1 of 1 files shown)


hack/pre-release.sh
```diff
@@ -303,6 +318,92 @@ dist_git_update() {
...
+  cd $temp
+  clone_repo "https://gitlab.cee.redhat.com/cpaas-products/openshift-winc.git"
+
+  GITLAB_LOCATION=$temp/openshift-winc
+  cd $GITLAB_LOCATION
+
+  latest_hash=$(git rev-parse HEAD)
+
+  gitlab_initial_branch=$(git branch --show-current)
+  new_branch=$(create_git_branch_list $stream-stream-bump $gitlab_initial_branch)
```
aravindhp, line 349: Is this branch name unique? What happens if this is not deleted after the MR merges?

hack/pre-release.sh
```diff
@@ -303,6 +318,92 @@ dist_git_update() {
...
+
+  git mv versions/release-$wmco_version versions/release-$updated_version
+  rm versions/release-$updated_version/advisory_map.yml
+
+  # Example: matches "Windows Containers 8.0.1"
+  sed -i "s/Windows Containers $wmco_version/Windows Containers $updated_version/" versions/release-$updated_version/release.yml
+
+  commit_message="Update version to $updated_version
+
+  This commit was generated by hack/pre-release.sh"
```
aravindhp, line 360: Does this actually create a well formed commit message? I suspect there will be extra whitespaces added which are not needed.

hack/pre-release.sh
```diff
@@ -303,6 +318,92 @@ dist_git_update() {
...
+
+  # Example: matches version="8.0.1"
+  sed -i "s/version=\"$wmco_version\"/version=\"$updated_version\"/" $dockerfile_location
+
+  # Example: matches commit: d0be5e812c3e3914d3cf3c464623e688673565fb
+  sed -i "s/commit: [0-9a-f]{40}/commit: $latest_hash/"  upstream_sources.yml
+
+  commit_message="[$base_branch] Update version to $updated_version
+
+  This commit was generated by hack/pre-release.sh"
```
aravindhp, line 393: Same question here.

aravindhp (changes requested):
Thanks for working on this, @wgahnagl. Please address my comments.

hack/pre-release.sh
wgahnagl: the [switch_git_branch ](https://github.com/openshift/windows-machine-config-operator/blob/713a399a6d085721458c1d602a287ce4edf8cf4c/hack/common.sh#L85) function in common.sh that I use [here](https://github.com/openshift/windows-machine-config-operator/blob/713a399a6d085721458c1d602a287ce4edf8cf4c/hack/pre-release.sh#L350) deletes the branch if there's a duplicate. This is the same function used in the submodule update script and elsewhere in this file.

hack/pre-release.sh
```diff
@@ -352,10 +454,12 @@ git submodule update --init --recursive  > /dev/null 2>&1
 WMCO_VERSION=$(sed -n -e 's/^.*WMCO_VERSION ?= //p' Makefile)
 updated_version=$(increment_version $WMCO_VERSION $stream)
 
-if [[ $DISTGIT == "false" ]]; then
-  github_update $base_branch $stream $WMCO_VERSION $updated_version
-else
+if [[ $DISTGIT == "true" ]]; then
```
aravindhp, line 457: Nit: I suggest making this a case statement.

aravindhp (approved):

hack/pre-release.sh
```diff
@@ -303,6 +318,92 @@ dist_git_update() {
...
+
+  git mv versions/release-$wmco_version versions/release-$updated_version
+  rm versions/release-$updated_version/advisory_map.yml
+
+  # Example: matches "Windows Containers 8.0.1"
+  sed -i "s/Windows Containers $wmco_version/Windows Containers $updated_version/" versions/release-$updated_version/release.yml
+
+  commit_message="Update version to $updated_version
+
+This commit was generated by hack/pre-release.sh"
```
alinaryan, line 360: Make this in-line with the other lines?

alinaryan (commented):
Thanks for working on this,

hack/pre-release.sh
wgahnagl: if I do that, the commit message is improperly indented 😔

saifshaikh48, line 360: Is it possible to add the command executed to the commit message? something like 
> "Ran `./pre-release.sh master x -d`"

incorporating `$@` which holds all arguments passed into the shell script could be sufficient

wgahnagl: that would also include the arguments passed along to the function as well, which is the old version and the updated one. Would that be OK?

saifshaikh48: no worries, this was more of a "nice to have" anyway

hack/pre-release.sh
```diff
@@ -303,6 +318,92 @@ dist_git_update() {
...
+# takes 2 arguments
+# 1st arg: stream to update. Can be x, y or z to represent the x y or z stream.
+# 2nd arg: current WMCO version
+# 3rd arg: updated WMCO version
+gitlab_update() {
+ if [ "$#" -ne 3 ]; then
+    echo incorrect parameter count for dist_git_update $#
+    return 1
+  fi
+  stream=$1
```
saifshaikh48, line 331: is this arg used? I would expect it to be -- are the the gitlab steps for x stream vs y/z all the same?

wgahnagl: I use it for naming the branch, but I could remove it if we want

saifshaikh48: I see, we should leave it them. What I was trying to imply here was that it'll be used to see if we have to `Change CDN repo number to correct one for release binary` (x stream only, which will be handled later in the story you're creating)

hack/pre-release.sh
```diff
@@ -263,9 +278,6 @@ dist_git_update() {
 
   previous_version=$(grep -Po 'previous_version="\K.*(?=")' render_templates)
 
-  # Example: matches version="v8.0.1"
-  sed -i "s/version=\"v[0-9]\+\(\.[0-9]\+\)\+\"/version=\"v$updated_version\"/g" Dockerfile
-
```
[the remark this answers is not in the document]
saifshaikh48: > and why it's being done.

because render_templates does this for us...

saifshaikh48 (commented):
/lgtm

PR #1705, opened 2023-07-25
WINC-1043: Test upgrading to CSI storage, from in-tree drivers

sebsoto:
Adds various changes required to test upgrading a Windows node from a version of WMCO which supports in-tree storage drivers, to a version that does not.

The changes within `Enable upgrading from in-tree to CSI ` should be reverted when branching for OCP 4.15 occurs, as they will break the tests. Those changes have been labeled with a TODO, and have been purposely been made easily removable.

### DIFF
--- a/hack/run-ci-e2e-test.sh
+++ b/hack/run-ci-e2e-test.sh
@@ -178,6 +178,8 @@ if [[ "$TEST" = "all" || "$TEST" = "upgrade" ]]; then
 fi
 
 if [[ "$TEST" = "upgrade-setup" ]]; then
+  # TODO: Remove this after adding it to the appropriate release jobs
+  export UPGRADE_FROM_IN_TREE=true
   go test ./test/e2e/... -run=TestWMCO/create/Creation -v -timeout=90m -args $GO_TEST_ARGS
   go test ./test/e2e/... -run=TestWMCO/create/Nodes_ready_and_schedulable -v -timeout=90m -args $GO_TEST_ARGS
   # Run the storage test, skipping deletion of the created workload in order to test that it persists across the upgrade
@@ -186,6 +188,8 @@ if [[ "$TEST" = "upgrade-setup" ]]; then
 fi
 
 if [[ "$TEST" = "upgrade-test" ]]; then
+  # TODO: Remove this after adding it to the appropriate release jobs
+  export UPGRADE_FROM_IN_TREE=true
   go test ./test/e2e/... -run=TestUpgrade -v -timeout=20m -args $GO_TEST_ARGS
 fi
 
--- a/hack/run-ci-e2e-test.sh
+++ b/hack/run-ci-e2e-test.sh
@@ -180,6 +180,9 @@ fi
 if [[ "$TEST" = "upgrade-setup" ]]; then
   go test ./test/e2e/... -run=TestWMCO/create/Creation -v -timeout=90m -args $GO_TEST_ARGS
   go test ./test/e2e/... -run=TestWMCO/create/Nodes_ready_and_schedulable -v -timeout=90m -args $GO_TEST_ARGS
+  # Run the storage test, skipping deletion of the created workload in order to test that it persists across the upgrade
+  go test ./test/e2e/... -run=TestWMCO/storage -v -timeout=15m -args $GO_TEST_ARGS --skip-workload-deletion=true
+  go test ./test/e2e/... -run=TestWMCO/create/Node_Logs -v -timeout=10m -args $GO_TEST_ARGS
 fi
 
 if [[ "$TEST" = "upgrade-test" ]]; then
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -27,6 +27,8 @@ var (
 	numberOfBYOHNodes int
 	// privateKeyPath is the path of the private key file used to configure the Windows node
 	privateKeyPath string
+	// skipWorkloadDeletion indicates that worklaods should not be deleted when tests complete
+	skipWorkloadDeletion bool
 	// wmcoNamespace is the namespace WMCO is deployed to
 	wmcoNamespace string
 	// windowsServerVersion is the Windows Server version to test against
@@ -148,6 +150,8 @@ func TestMain(m *testing.M) {
 		"Namespace that WMCO is deployed to")
 	flag.StringVar(&privateKeyPath, "private-key-path", "",
 		"path of the private key file used to configure the Windows node")
+	flag.BoolVar(&skipWorkloadDeletion, "skip-workload-deletion", false,
+		"skips deletion of workloads at the end of tests")
 	flag.Func("windows-server-version", "Windows Server version to test. "+
 		"Supported versions: 2019 or 2022 (default 2022)",
 		func(value string) error {
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -27,6 +27,9 @@ var (
 	numberOfBYOHNodes int
 	// privateKeyPath is the path of the private key file used to configure the Windows node
 	privateKeyPath string
+	// inTreeUpgrade indicates if tests should be ran to test upgrading from an in-tree storage solution to CSI.
+	// This will cause in tree storage volumes and related resources to be deployed where appropriate, instead of CSI.
+	inTreeUpgrade bool
 	// skipWorkloadDeletion indicates that worklaods should not be deleted when tests complete
 	skipWorkloadDeletion bool
 	// wmcoNamespace is the namespace WMCO is deployed to
@@ -166,6 +169,9 @@ func TestMain(m *testing.M) {
 			return nil
 		})
 	flag.Parse()
+	if val := os.Getenv("UPGRADE_FROM_IN_TREE"); val == "true" {
+		inTreeUpgrade = true
+	}
 
 	os.Exit(m.Run())
 }
--- a/test/e2e/providers/vsphere/vsphere.go
+++ b/test/e2e/providers/vsphere/vsphere.go
@@ -151,14 +151,23 @@ func (p *Provider) StorageSupport() bool {
 	return true
 }
 
+// CreateInTreePVC creates a PVC which has the potential to be migrated to a CSI PVC on upgrade
+func (p *Provider) CreateInTreePVC(client client.Interface, namespace string) (*core.PersistentVolumeClaim, error) {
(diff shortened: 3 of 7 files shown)


test/e2e/storage_test.go
```diff
@@ -2,12 +2,14 @@ package e2e
 
 import (
 	"context"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
```
sebsoto, line 5: will fix this

hack/run-ci-e2e-test.sh
```diff
@@ -178,13 +178,17 @@ if [[ "$TEST" = "all" || "$TEST" = "upgrade" ]]; then
 fi
 
 if [[ "$TEST" = "upgrade-setup" ]]; then
+  # TODO: Remove this after adding it to the appropriate release jobs
```
saifshaikh48, line 181: Are all the TODO comments necessary if we will revert the whole commit in thew 4.15 branch? They'd unnecessarily linger in 4.14

saifshaikh48 (commented):
LGTM overall, one question

hack/run-ci-e2e-test.sh
```diff
@@ -86,6 +86,7 @@ fi
 SKIP_NODE_DELETION=${SKIP_NODE_DELETION:-"false"}
 
 # Setup and run the operator if it is not already deployed
+wmco_deployed_by_script=false
```
aravindhp, line 89: Please explain what you are trying to do in the commit message.

test/e2e/upgrade_test.go
```diff
@@ -221,11 +222,20 @@ func TestUpgrade(t *testing.T) {
...
 // testWorkloadsAvailable tests that all workloads deployed on Windows nodes by the test suite are available
 func (tc *testContext) testWorkloadsAvailable(t *testing.T) {
-	deployments, err := tc.client.K8s.AppsV1().Deployments(tc.workloadNamespace).List(context.TODO(),
-		metav1.ListOptions{})
-	require.NoError(t, err)
-	for _, deployment := range deployments.Items {
-		assert.Equalf(t, deployment.Spec.Replicas, deployment.Status.AvailableReplicas,
-			"deployment %s is not available", deployment.GetName())
-	}
+	err := wait.PollImmediate(retry.Interval, retry.ResourceChangeTimeout, func() (bool, error) {
```
aravindhp, line 238: Shouldn't you be using `PollWithContextTimeout`?

test/e2e/upgrade_test.go
```diff
@@ -221,11 +222,20 @@ func TestUpgrade(t *testing.T) {
...
+		deployments, err := tc.client.K8s.AppsV1().Deployments(tc.workloadNamespace).List(context.TODO(),
+			metav1.ListOptions{})
+		if err != nil {
+			log.Printf("error getting deployment list: %s", err)
+		}
+		for _, deployment := range deployments.Items {
+			if deployment.Spec.Replicas == nil ||
+				(*deployment.Spec.Replicas == deployment.Status.AvailableReplicas) {
```
aravindhp, line 246: You are adding an extra nil check here. I don't follow how that applies to not updating the current state of the deployments when polling given we were not polling before.

hack/run-ci-e2e-test.sh
```diff
@@ -180,6 +180,8 @@ fi
 if [[ "$TEST" = "upgrade-setup" ]]; then
   go test ./test/e2e/... -run=TestWMCO/create/Creation -v -timeout=90m -args $GO_TEST_ARGS
   go test ./test/e2e/... -run=TestWMCO/create/Nodes_ready_and_schedulable -v -timeout=90m -args $GO_TEST_ARGS
+  go test ./test/e2e/... -run=TestWMCO/storage -v -timeout=15m -args $GO_TEST_ARGS --skip-workload-deletion=true
```
aravindhp, line 185: Please mention why we want to skip storage workload deletion in the commit and in a comment here.

test/e2e/storage_test.go
```diff
@@ -29,8 +29,10 @@ func testStorage(t *testing.T) {
 	var pvcVolumeSource *core.PersistentVolumeClaimVolumeSource
 	pvc, err := tc.CloudProvider.CreatePVC(tc.client.K8s, tc.workloadNamespace)
 	require.NoError(t, err)
-	defer tc.client.K8s.CoreV1().PersistentVolumeClaims(tc.workloadNamespace).Delete(context.TODO(),
-		pvc.GetName(), meta.DeleteOptions{})
+	if !skipWorkloadDeletion {
+		defer tc.client.K8s.CoreV1().PersistentVolumeClaims(tc.workloadNamespace).Delete(context.TODO(),
+			pvc.GetName(), meta.DeleteOptions{})
```
aravindhp, line 44: I suggest at least logging an error if we are unable to delete the PVCs.

test/e2e/storage_test.go
```diff
@@ -26,14 +28,22 @@ func testStorage(t *testing.T) {
 	// Create the PVC and choose the node the deployment will be scheduled on. This is necessary as ReadWriteOnly
 	// volumes can only be bound to a single node.
 	// https://docs.openshift.com/container-platform/4.12/storage/understanding-persistent-storage.html#pv-access-modes_understanding-persistent-storage
-	var pvcVolumeSource *core.PersistentVolumeClaimVolumeSource
-	pvc, err := tc.CloudProvider.CreatePVC(tc.client.K8s, tc.workloadNamespace)
-	require.NoError(t, err)
+	var pvc *core.PersistentVolumeClaim
+	if inTreeUpgrade {
```
aravindhp, line 32: Can't we test this before waiting for Windows node and fail early when not on vSphere?

test/e2e/upgrade_test.go
```diff
@@ -240,3 +252,32 @@ func (tc *testContext) testWorkloadsAvailable(t *testing.T) {
 	})
 	assert.NoError(t, err)
 }
+
+// waitForCSINodes waits for a CSINode to exist for each node with a driver loaded, indicating CSI functionality has
+// been enabled for the given node.
+func (tc *testContext) waitForCSINodesWithDrivers(nodes []v1.Node) error {
+	return wait.PollImmediate(retry.Interval, 10*time.Minute, func() (bool, error) {
```
aravindhp, line 259: Lets stop introducing a deprecated function.

test/e2e/main_test.go
```diff
@@ -27,6 +27,8 @@ var (
 	numberOfBYOHNodes int
 	// privateKeyPath is the path of the private key file used to configure the Windows node
 	privateKeyPath string
+	// inTreeUpgrade indicates if tests should be ran to test upgrading from an in-tree storage solution to CSI
```
aravindhp, line 30: Please provide more details about what you are testing with in tree upgrades here and in the commit message.

hack/run-ci-e2e-test.sh
aravindhp: If you want to revert this, I suggest introducing it as a separate PR so that we can just revert the PR.

aravindhp (changes requested):
@sebsoto thanks for working on this. You are introducing test fixes in this PR in addition to new testing features. Do these test fixes need to be backported? If yes, please introduce them in a separate PR.

hack/run-ci-e2e-test.sh
sebsoto: This needs to go in at the same time as the previous commit, as this will not work on 4.14 without it. But I have split the rest of the commits off into a separate PR: https://github.com/openshift/windows-machine-config-operator/pull/1733, as they are just general test improvements.

test/e2e/upgrade_test.go
sebsoto: Leaving this as is so it can be backported to 4.13

aravindhp:
/hold

Blocked by https://github.com/openshift/windows-machine-config-operator/pull/1733

saifshaikh48 (commented):
Great work on this, just a few code structure comments

sebsoto:
/test vsphere-e2e-upgrade

Retesting to see whats going on here, failure might be a flake.

test/e2e/providers/vsphere/vsphere.go
```diff
@@ -151,14 +151,23 @@ func (p *Provider) StorageSupport() bool {
 	return true
 }
 
+// CreateInTreePVC creates a PVC which has the potential to be migrated to a CSI PVC on upgrade
+func (p *Provider) CreateInTreePVC(client client.Interface, namespace string) (*core.PersistentVolumeClaim, error) {
```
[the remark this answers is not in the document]
sebsoto: I would usually agree, but in this case I chose not to make this change as this needs to be reverted after this release, and changes to the interface would make that more complicated.

test/e2e/storage_test.go
```diff
@@ -27,17 +33,24 @@ func testStorage(t *testing.T) {
...
+		vsphereProvider, ok := tc.CloudProvider.(*vsphere.Provider)
+		require.True(t, ok, "in tree upgrade must be ran on vSphere")
```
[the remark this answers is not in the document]
sebsoto: The cast is necessary in order to use the method, which is specific to the vSphere provider.

test/e2e/upgrade_test.go
```diff
@@ -216,6 +217,17 @@ func TestUpgrade(t *testing.T) {
 	t.Run("Node annotations", tc.testNodeAnnotations)
 	t.Run("Node Metadata", tc.testNodeMetadata)
 
+	if inTreeUpgrade {
+		// Deploy the CSI drivers and wait for a CSI node to be created
+		// This is a requirement for storage workloads to go back to ready
+		vsphereProvider, ok := tc.CloudProvider.(*vsphere.Provider)
+		require.True(t, ok, "in tree upgrade must be ran on vSphere")
```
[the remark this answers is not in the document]
sebsoto: same as above:

> The cast is necessary in order to use the method, which is specific to the vSphere provider.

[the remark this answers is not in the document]
saifshaikh48: Fair enough, these comments were only applicable assuming we were to add the methods to the CloudProvider interface

saifshaikh48 (commented):
/lgtm

mtnbikenc:
https://github.com/openshift/windows-machine-config-operator/pull/1733 has merged.
/hold cancel

PR #1706, opened 2023-07-26
WINC-998: Clear proxy variables from Windows nodes

mansikulkarni96:
This PR implements the following scenarios:
1. On existing Windows nodes, when an cluster-wide proxy is removed WICD is able to remove the ENV variables on the instance.
2. When a node deconfiguration occurs, WICD cleanup command will remove the ENV vars set on the instance if cluster-wide proxy is enabled. This includes HTTP_PROXY, HTTPS_PROXY, NO_PROXY being removed from the instance along with the keys set in the Windows registry.

### DIFF
--- a/pkg/cluster/config.go
+++ b/pkg/cluster/config.go
@@ -34,8 +34,8 @@ const (
 )
 
 var (
-	// SupportedProxyVars is a list of the supported proxy variables
-	SupportedProxyVars = []string{"HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY"}
+	// WatchedEnvironmentVars is a list of the WMCO watched environment variables
+	WatchedEnvironmentVars = []string{"HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY"}
 	// clusterWideProxyVars is a map of the global egress proxy variables and values from WMCO's environment
 	clusterWideProxyVars map[string]string
 )
@@ -363,7 +363,7 @@ func GetProxyVars() map[string]string {
 	// if clusterWideProxyVars is not already cached, initialize it. We never expect these values to change during
 	// runtime as OLM restarts the operator when the global cluster proxy config changes
 	clusterWideProxyVars = make(map[string]string, 3)
-	for _, envVar := range SupportedProxyVars {
+	for _, envVar := range WatchedEnvironmentVars {
 		value, found := os.LookupEnv(envVar)
 		if found {
 			// on Windows, hostname lists are separated by semicolons rather than the Linux default of commas
--- a/pkg/daemon/cleanup/cleanup.go
+++ b/pkg/daemon/cleanup/cleanup.go
@@ -28,6 +28,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/controller"
+	"github.com/openshift/windows-machine-config-operator/pkg/daemon/envvar"
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/manager"
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/powershell"
 	"github.com/openshift/windows-machine-config-operator/pkg/metadata"
@@ -57,12 +58,23 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
 		klog.Exitf("could not create service manager: %s", err.Error())
 	}
 	defer svcMgr.Disconnect()
-	services, err := getServicesToRemove(ctx, directClient, node, configMapNamespace)
+	mergedCMData, err := getMergedCMData(ctx, directClient, configMapNamespace, node)
 	if err != nil {
 		return err
 	}
-	if err = removeServices(svcMgr, services); err != nil {
+	if err = removeServices(svcMgr, mergedCMData.Services); err != nil {
 		return err
+	}
+	restartRequired, err := ensureEnvVarsAreRemoved(mergedCMData.WatchedEnvironmentVars)
+	if err != nil {
+		return err
+	}
+	// rebooting instance to unset the environment variables at the process level as expected
+	if restartRequired {
+		// Applying the reboot annotation results in an event picked up by WMCO's node controller to reboot the instance
+		if err = metadata.ApplyRebootAnnotation(ctx, directClient, *node); err != nil {
+			return fmt.Errorf("error setting reboot annotation on node %s: %w", node.Name, err)
+		}
 	}
 	cleanupContainers()
 
@@ -72,13 +84,15 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
 	return nil
 }
 
-// getServicesToRemove returns a list of services that should be removed as part of the cleanup process
-// returns the merged Data of the latest ConfigMap, and the ConfigMap specified by the node's version annotation
-func getServicesToRemove(ctx context.Context, cli client.Client, node *core.Node, configMapNamespace string) ([]servicescm.Service, error) {
+// getMergedCMData attempts to get the latest and the version CM data specified by the node's version annotation
+// It returns the merged CM Data containing services and the watched environment variables
+func getMergedCMData(ctx context.Context, cli client.Client,
+	configMapNamespace string, node *core.Node) (*servicescm.Data, error) {
 	// get data from the latest services ConfigMap
 	latestCM, err := servicescm.GetLatest(cli, ctx, configMapNamespace)
 	if err != nil {
-		return nil, fmt.Errorf("cannot get latest services ConfigMap from namespace %s: %w", configMapNamespace, err)
+		return nil, fmt.Errorf("cannot get latest services ConfigMap from namespace %s: %w",
+			configMapNamespace, err)
 	}
 	latestCMData, err := servicescm.Parse(latestCM.Data)
 	if err != nil {
@@ -110,17 +124,18 @@ func getServicesToRemove(ctx context.Context, cli client.Client, node *core.Node
 	if err != nil {
 		klog.Infof("error getting services ConfigMap associated with version annotation, "+
 			"falling back to use latest services ConfigMap: %s", err)
-		return latestCMData.Services, nil
+		return latestCMData, nil
 	}
-
-	// If the instance was configured using latestCM, return the services from that
+	// If the instance was configured using latestCM, return that
 	if versionCM.GetName() == latestCM.GetName() {
-		klog.Infof("removing the services specified in %s", latestCM.GetName())
-		return latestCMData.Services, nil
+		return latestCMData, nil
 	}
-	// merge the two ConfigMaps into one, so all potential services are listed
-	klog.Infof("removing the services specified in %s and %s", versionCM.GetName(), latestCM.GetName())
-	return mergeServices(latestCMData.Services, versionCMData.Services), nil
+	mergedServices := mergeServices(latestCMData.Services, versionCMData.Services)
+	mergedEnvVars := merge(latestCMData.WatchedEnvironmentVars, versionCMData.WatchedEnvironmentVars)
+	return &servicescm.Data{
+		Services:               mergedServices,
+		WatchedEnvironmentVars: mergedEnvVars,
+	}, nil
 }
 
 // mergeServices combines the list of services, prioritizing the data given by s1
@@ -142,6 +157,22 @@ func mergeServices(s1, s2 []servicescm.Service) []servicescm.Service {
 
 }
 
+// merge returns a combined list of the given lists
+func merge(e1, e2 []string) []string {
+	watchedEnvVars := make(map[string]struct{})
+	for _, item := range e1 {
+		watchedEnvVars[item] = struct{}{}
+	}
+	for _, item := range e2 {
+		watchedEnvVars[item] = struct{}{}
+	}
+	var merged []string
+	for item := range watchedEnvVars {
+		merged = append(merged, item)
+	}
+	return merged
+}
+
 // removeServices uses the given manager to remove all the given Windows services from this instance.
 func removeServices(svcMgr manager.Manager, services []servicescm.Service) error {
 	// Build up log message and failures
@@ -161,6 +192,11 @@ func removeServices(svcMgr manager.Manager, services []servicescm.Service) error
 		return fmt.Errorf("%#v", failedRemovals)
 	}
 	return nil
+}
+
+// ensureEnvVarsAreRemoved removes all WICD configured ENV variables from this instance
+func ensureEnvVarsAreRemoved(watchedEnvVars []string) (bool, error) {
+	return envvar.EnsureVarsAreUpToDate(map[string]string{}, watchedEnvVars)
 }
 
 // cleanupContainers makes a best effort to stop all processes with the name containerd-shim-runhcs-v1, stopping
--- a/pkg/daemon/controller/controller.go
+++ b/pkg/daemon/controller/controller.go
@@ -48,7 +48,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
-	"github.com/openshift/windows-machine-config-operator/pkg/cluster"
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/envvar"
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/manager"
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/powershell"
@@ -258,7 +257,7 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
 		return ctrl.Result{}, err
 	}
 
-	awaitingRestart, err := sc.reconcileEnvironmentVariables(cmData.EnvironmentVars, node)
+	awaitingRestart, err := sc.reconcileEnvironmentVariables(cmData.EnvironmentVars, cmData.WatchedEnvironmentVars, node)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
@@ -284,8 +283,9 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
 
 // reconcileEnvironmentVariables makes sure that the proxy variables exist as expected, or are safely rectified.
 // Returns a boolean expressing whether the instance is awaiting a reboot.
-func (sc *ServiceController) reconcileEnvironmentVariables(envVars map[string]string, node core.Node) (bool, error) {
-	restartRequired, err := envvar.EnsureVarsAreUpToDate(envVars)
+func (sc *ServiceController) reconcileEnvironmentVariables(envVars map[string]string, watchedEnvVars []string,
+	node core.Node) (bool, error) {
+	restartRequired, err := envvar.EnsureVarsAreUpToDate(envVars, watchedEnvVars)
 	if err != nil {
 		return false, err
 	}
@@ -301,7 +301,7 @@ func (sc *ServiceController) reconcileEnvironmentVariables(envVars map[string]st
 	err = wait.PollUntilContextTimeout(sc.ctx, 15*time.Second, 5*time.Minute, true,
 		func(ctx context.Context) (done bool, err error) {
 			stillNeedsReboot := false
-			for _, varName := range cluster.WatchedEnvironmentVars {
+			for _, varName := range watchedEnvVars {
 				cmd := fmt.Sprintf("[Environment]::GetEnvironmentVariable('%s', 'Process')", varName)
 				out, err := sc.psCmdRunner.Run(cmd)
 				if err != nil {
--- a/pkg/daemon/controller/controller.go
+++ b/pkg/daemon/controller/controller.go
@@ -301,7 +301,7 @@ func (sc *ServiceController) reconcileEnvironmentVariables(envVars map[string]st
 	err = wait.PollUntilContextTimeout(sc.ctx, 15*time.Second, 5*time.Minute, true,
 		func(ctx context.Context) (done bool, err error) {
 			stillNeedsReboot := false
-			for _, varName := range cluster.SupportedProxyVars {
+			for _, varName := range cluster.WatchedEnvironmentVars {
 				cmd := fmt.Sprintf("[Environment]::GetEnvironmentVariable('%s', 'Process')", varName)
 				out, err := sc.psCmdRunner.Run(cmd)
 				if err != nil {
--- a/pkg/daemon/controller/controller.go
+++ b/pkg/daemon/controller/controller.go
@@ -325,18 +325,17 @@ func (sc *ServiceController) ensureVarsAreUpToDate(envVars map[string]string) (b
 	// systemEnvVarRegistryPath is where system level environment variables are stored in the Windows OS
 	const systemEnvVarRegistryPath = `SYSTEM\CurrentControlSet\Control\Session Manager\Environment`
 	restartRequired := false
-	for key, expectedVal := range envVars {
-		registryKey, err := registry.OpenKey(registry.LOCAL_MACHINE, systemEnvVarRegistryPath, registry.ALL_ACCESS)
-		if err != nil {
-			return false, fmt.Errorf("unable to open Windows system registry key %s: %w", systemEnvVarRegistryPath, err)
+	registryKey, err := registry.OpenKey(registry.LOCAL_MACHINE, systemEnvVarRegistryPath, registry.ALL_ACCESS)
+	if err != nil {
+		return false, fmt.Errorf("unable to open Windows system registry key %s: %w", systemEnvVarRegistryPath, err)
+	}
+	defer func() { // always close the registry key, without swallowing any error returned before the defer call
+		closeErr := registryKey.Close()
+		if closeErr != nil {
+			klog.Errorf("could not close key %v: %v", registryKey, closeErr)
 		}
-		defer func() { // always close the registry key, without swallowing any error returned before the defer call
-			closeErr := registryKey.Close()
-			if closeErr != nil {
-				klog.Errorf("could not close key: %v", closeErr)
-			}
-		}()
-
+	}()
+	for key, expectedVal := range envVars {
 		actualVal, _, err := registryKey.GetStringValue(key)
 		if err != nil && err != registry.ErrNotExist {
 			return false, fmt.Errorf("unable to read environment variable %s: %w", key, err)
--- a/pkg/daemon/controller/controller.go
+++ b/pkg/daemon/controller/controller.go
@@ -26,7 +26,6 @@ import (
 	"strings"
 	"time"
 
-	"golang.org/x/sys/windows/registry"
 	"golang.org/x/sys/windows/svc"
 	"golang.org/x/sys/windows/svc/mgr"
 	core "k8s.io/api/core/v1"
@@ -50,6 +49,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	"github.com/openshift/windows-machine-config-operator/pkg/cluster"
+	"github.com/openshift/windows-machine-config-operator/pkg/daemon/envvar"
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/manager"
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/powershell"
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/winsvc"
@@ -285,7 +285,7 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
 // reconcileEnvironmentVariables makes sure that the proxy variables exist as expected, or are safely rectified.
 // Returns a boolean expressing whether the instance is awaiting a reboot.
 func (sc *ServiceController) reconcileEnvironmentVariables(envVars map[string]string, node core.Node) (bool, error) {
-	restartRequired, err := sc.ensureVarsAreUpToDate(envVars)
+	restartRequired, err := envvar.EnsureVarsAreUpToDate(envVars)
 	if err != nil {
 		return false, err
 	}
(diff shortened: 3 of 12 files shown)


pkg/daemon/cleanup/cleanup.go
```diff
@@ -65,13 +75,47 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
 		return err
 	}
 	cleanupContainers()
-
+	if len(latestCMData.EnvironmentVars) != 0 {
```
saifshaikh48, line 78: Shouldn't we use the CM tied to `VersionAnnotation`? We could use the latest one as backup like we do in  getServicesToRemove()

pkg/daemon/controller/controller.go
```diff
@@ -353,6 +358,23 @@ func (sc *ServiceController) ensureVarsAreUpToDate(envVars map[string]string) (b
...
+// An instance restart is required after they are removed to ensure all processes pick up the updated values.
+func EnsureEnvVarsAreRemoved(registryKey registry.Key, supportedEnvVars map[string]string) (bool, error) {
+	restartRequired := false
+	for _, envVar := range supportedEnvVars {
+		err := registryKey.DeleteValue(envVar)
+		if err != nil {
+			if err != registry.ErrNotExist {
+				return false, err
+			}
+		} else {
```
saifshaikh48, line 371: should we also log that the key was successfully removed

pkg/daemon/powershell/powershell.go
```diff
@@ -16,7 +21,7 @@ type commandRunner struct{}
 // Run runs the command with the PowerShell on PATH
 func (r *commandRunner) Run(cmd string) (string, error) {
 	out, err := exec.Command("powershell", "/c", cmd).CombinedOutput()
-	if err != nil {
+	if err != nil && !strings.Contains(err.Error(), cmdExitNoStatus) {
```
saifshaikh48, line 24: Could you explain why this change is needed?

pkg/daemon/cleanup/cleanup.go
```diff
@@ -72,47 +82,99 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
...
 	if err != nil {
-		return nil, err
+		return err
+	}
+	if len(cmData.EnvironmentVars) == 0 {
+		// no-op
+		return nil
+	}
+	// systemEnvVarRegistryPath is where system level environment variables are stored in the Windows OS
+	const systemEnvVarRegistryPath = `SYSTEM\CurrentControlSet\Control\Session Manager\Environment`
```
saifshaikh48, line 106: same constant exists within a func somewhere in `controller` package. consider making it global over there and referencing that here

pkg/daemon/cleanup/cleanup.go
```diff
@@ -72,47 +82,99 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
...
+		if closeErr != nil {
+			klog.Errorf("could not close key: %v", closeErr)
+		}
+	}()
+	restartRequired, err := controller.EnsureEnvVarsAreRemoved(registryKey, cmData.SupportedEnvVars)
+	if err != nil {
+		return err
+	}
+	if restartRequired {
+		restartCmd := "Restart-Computer -Force"
```
saifshaikh48, line 123: Please add a comment why restart is necessary

pkg/daemon/cleanup/cleanup.go
```diff
@@ -72,47 +82,99 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
...
+	if err != nil {
+		return fmt.Errorf("unable to open Windows system registry key %s: %w", systemEnvVarRegistryPath, err)
+	}
+	defer func() { // always close the registry key, without swallowing any error returned before the defer call
+		closeErr := registryKey.Close()
+		if closeErr != nil {
+			klog.Errorf("could not close key: %v", closeErr)
+		}
+	}()
+	restartRequired, err := controller.EnsureEnvVarsAreRemoved(registryKey, cmData.SupportedEnvVars)
```
saifshaikh48, line 118: do you think we should be merging the `SupportedEnvVars` from the versionCM and the latestCM, then pass that list to `EnsureEnvVarsAreRemoved()` (similar to how services are removed)?

test/e2e/proxy_test.go
```diff
@@ -76,6 +78,28 @@ func testEnvVars(t *testing.T) {
...
+	_, err = tc.client.Config.ConfigV1().Proxies().Patch(
+		context.TODO(),
+		"cluster",
+		types.JSONPatchType,
+		patchData,
+		meta.PatchOptions{},
+	)
+	patchString := string(patchData)
+	require.NoErrorf(t, err, "unable to patch %s", patchString)
+	time.Sleep(5 * time.Minute)
```
saifshaikh48, line 95: please comment why this sleep is needed. Can we call `waitForConfiguredWindowsNodes()` or something similar instead?

pkg/daemon/powershell/powershell.go
mansikulkarni96: Restart-computer always returns error with cmdExitNoStatus

pkg/daemon/cleanup/cleanup.go
mansikulkarni96: I looked into it and the reason why we merged services from both the CM's was there were some services which were installed by WMCO and manged by WICD. Since we are going to have all envVars created and cleaned up by WICD from the beginning, do we need to do that?

pkg/daemon/cleanup/cleanup.go
```diff
@@ -57,13 +58,22 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
 		klog.Exitf("could not create service manager: %s", err.Error())
 	}
 	defer svcMgr.Disconnect()
-	services, err := getServicesToRemove(ctx, directClient, node, configMapNamespace)
+	// get data from the latest services ConfigMap
```
aravindhp, line 61: You are getting the latest service ConfigMap, passing it to `getServicesToRemove` and `ensureEnvVarsAreRemoved` and within each function getting the version ConfigMap and falling back to the latest if that fails. Shouldn't figuring out which ConfigMap to use be done outside of the two functions? This was we also ensure that both methods are using the same ConfigMap.

aravindhp (changes requested):

pkg/daemon/controller/controller.go
```diff
@@ -321,22 +321,26 @@ func (sc *ServiceController) reconcileEnvironmentVariables(envVars map[string]st
...
-			closeErr := registryKey.Close()
-			if closeErr != nil {
-				klog.Errorf("could not close key: %v", closeErr)
-			}
-		}()
+	}()
 
+	if len(envVars) == 0 {
+		return EnsureEnvVarsAreRemoved(registryKey, supportedEnvVars)
+	}
```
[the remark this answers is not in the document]
mansikulkarni96: addressed

pkg/servicescm/servicescm.go
```diff
@@ -27,6 +27,9 @@ const (
 	filesKey = "files"
 	// envVarsKey is an optional key in the services ConfigMap. The value for this key is a string-to-string map.
 	envVarsKey = "environmentVars"
+	// supportedEnvVarsKey is a required key which lists the supported env vars in the services ConfigMap.
```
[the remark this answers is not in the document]
mansikulkarni96: created https://issues.redhat.com/browse/WINC-1102 as a follow-up.

pkg/servicescm/servicescm.go
```diff
@@ -27,6 +27,9 @@ const (
...
+	// The value for this key is a string-to-string map.
```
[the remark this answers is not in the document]
mansikulkarni96: I ran into some json masrhalling issues with this approach earlier but I retried it, confirmed it works and updated.

pkg/servicescm/servicescm.go
```diff
@@ -269,6 +290,17 @@ func (cmData *Data) ValidateExpectedContent(expected *Data) error {
 				cmData.EnvironmentVars[expectedEnvVar])
 		}
 	}
+	if len(cmData.SupportedEnvVars) != len(expected.SupportedEnvVars) {
+		return fmt.Errorf("unexpected number of supported environment variable")
+	}
+
+	for i, expectedEnvVar := range expected.SupportedEnvVars {
+		if cmData.SupportedEnvVars[i] != expected.SupportedEnvVars[i] {
```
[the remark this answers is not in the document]
mansikulkarni96: Right, only way I can think of doing a comparison irrespective of order is doing a deep equal on the slices.

pkg/daemon/cleanup/cleanup.go
mansikulkarni96: updated, the logic was a bit complex for removing the services, I pulled this out.

pkg/daemon/cleanup/cleanup.go
mansikulkarni96: @saifshaikh48 WDYT?

pkg/daemon/cleanup/cleanup.go
```diff
@@ -72,47 +82,99 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
...
+	if err != nil {
+		return err
+	}
+	// rebooting instance to set the environment variables at the process level as expected
+	if restartRequired {
+		restartCmd := "Restart-Computer -Force"
+		cmdRunner := powershell.NewCommandRunner()
+		out, err := cmdRunner.Run(restartCmd)
+		if err != nil {
+			return fmt.Errorf("'%s' failed with out: %s, err: %w", restartCmd, out, err)
```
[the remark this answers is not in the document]
mansikulkarni96: This is the only place it's being used here. Other places in the controller where we reconcile and remove vars, we do the reboot and reinitialize through the Windows VM object.

pkg/servicescm/servicescm.go
```diff
@@ -127,6 +135,32 @@ func GetLatest(c client.Client, ctx context.Context, namespace string) (*core.Co
 	return &servicesCMs[0], nil
 }
 
+// GetVersionCM returns the configMap associated with the node's version annotation
```
aravindhp, line 138: The function is being introduced in this commit but being used in a follow up commit. It should be named `GetVersion()` to avoid stuttering given the call sites will read `servicescm.GetVersion()`.

You don't need to be constrained by the commit prefixes when introducing changes. Introducing them in a logical manner is more important as it makes reviews easier.

pkg/servicescm/servicescm.go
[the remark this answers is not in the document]
aravindhp: This could be prone to false negatives then. Are we okay with that?

pkg/daemon/cleanup/cleanup.go
aravindhp: My comment was more of a question rather than a request for a change. In future please respond if it is making the code more complex before making the changes.

pkg/daemon/cleanup/cleanup.go
```diff
@@ -72,47 +96,55 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
 	return nil
 }
 
-// getServicesToRemove returns a list of services that should be removed as part of the cleanup process
-// returns the merged Data of the latest ConfigMap, and the ConfigMap specified by the node's version annotation
-func getServicesToRemove(ctx context.Context, cli client.Client, node *core.Node, configMapNamespace string) ([]servicescm.Service, error) {
-	// get data from the latest services ConfigMap
-	latestCM, err := servicescm.GetLatest(cli, ctx, configMapNamespace)
+// ensureEnvVarsAreRemoved removes all WICD configured ENV variables from this instance
+func ensureEnvVarsAreRemoved(cm *core.ConfigMap) error {
```
aravindhp, line 100: Ideally we should be following the same pattern with services i.e. get the env variables to remove and  then remove them. I also don't fully understand the split between `ensureEnvVarsAreRemoved()` and `controller.EnsureEnvVarsAreRemoved() `. To be clear I am not asking for any changes here but just finding the logic split of methods a bit hard to follow across packages.

pkg/daemon/cleanup/cleanup.go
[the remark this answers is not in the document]
aravindhp: I find this a bit odd. I will take a look at this deeper outside of this PR.

aravindhp (changes requested):

pkg/servicescm/servicescm.go
mansikulkarni96: agreed, apologies I didn't double-check and this was added in a local rebase conflict.

pkg/daemon/cleanup/cleanup.go
mansikulkarni96: Thanks for the feedback, I ll give it a try if I can make it less complex. I am also working on merging the CM's through a common function, I'll add it to a separate commit for easier review.

aravindhp: I don't think we should add merging of the environment if it's too complex. Plus we don't want to expand the scope of the bug.

pkg/servicescm/servicescm.go
```diff
@@ -95,16 +98,20 @@ type Data struct {
 	Files []FileInfo `json:"files"`
 	// EnvironmentVars will be set as system-level environment variables on each Windows instance
 	EnvironmentVars map[string]string `json:"environmentVars,omitempty"`
+	// SupportedEnvVars contains information about the WMCO supported environment variables
+	SupportedEnvVars []string `json:"supportedEnvVars,omitempty"`
```
sebsoto, line 103: In general I'm not a huge fan of the `supportedEnvVars` naming, specifically `supported`.
From a technical sense `watchedEnvironmentVariables` would be more fitting.
`support` is more of a business term, while what this api is really doing is directing the WICD controller to ensure the state of specific resources.

sebsoto (changes requested):
Leaving these comments for now, still reviewing the third commit

pkg/servicescm/servicescm.go
sebsoto: I don't think this function should be added to this package, this package contains general functions for working with the Windows service CMs, IMO the logic introduced in this function is a bit too specialized, due to its use of `node.Annotations[metadata.VersionAnnotation]`. This should 

I don't see it being used anywhere but within cleanup.go, and would fit better there as a helper function.

pkg/daemon/cleanup/cleanup.go
```diff
@@ -57,13 +57,37 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
...
+		if err != nil {
+			return err
+		}
+		services = latestCMData.Services
+	} else {
+		relevantCM = versionCM
+		services, err = getServicesToRemove(latestCM, versionCM)
+		if err != nil {
+			return err
+		}
```
sebsoto, line 82: Its not clear why this logic has to be done, which indicates it should be broken out into a separate documented function.

sebsoto (changes requested):

pkg/daemon/cleanup/cleanup.go
sebsoto: This function doesn't make it clear that it will restart the computer if theres any change.
Please rename this to reflect what it is doing. `removeAllWatchedEnvVarsAndRestart`

pkg/daemon/cleanup/cleanup.go
```diff
@@ -72,47 +96,55 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
...
-		}
-		versionCMData, err = servicescm.Parse(versionCM.Data)
+	}()
+	restartRequired, err := controller.EnsureEnvVarsAreRemoved(registryKey, cmData.SupportedEnvVars)
+	if err != nil {
+		return err
+	}
+	// rebooting instance to set the environment variables at the process level as expected
+	if restartRequired {
+		restartCmd := "Restart-Computer -Force"
```
sebsoto, line 126: Should WICD be doing this? Doesn't it signal to WMCO that a restart is necessary in other places?

aravindhp: Sigh, this is why I found this odd. WICD cannot be restarting instances otherwise we could have multiple or all nodes restarting at the same time.

pkg/cluster/config.go
```diff
@@ -357,11 +357,6 @@ func IsProxyEnabled() bool {
 // GetProxyVars returns a map of the proxy variables and values from the WMCO container's environment. The presence of
 // any implies a proxy is enabled, as OLM would have injected them into the operator spec. Returns an empty map otherwise.
 func GetProxyVars() map[string]string {
-	if clusterWideProxyVars != nil {
-		return clusterWideProxyVars
-	}
-	// if clusterWideProxyVars is not already cached, initialize it. We never expect these values to change during
-	// runtime as OLM restarts the operator when the global cluster proxy config changes
```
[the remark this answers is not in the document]
mansikulkarni96: I'll move this to another pre-commit.

pkg/servicescm/servicescm.go
aravindhp: @sebsoto we are giving conflicting reviews 😄  I asked for it to be moved here given it has to do with the services ConfigMap. I don't see why it can't be added here as a helper. One alternative would be to make the `VersionAnnotation` an input.

pkg/daemon/cleanup/cleanup.go
mansikulkarni96: AFAIK WICD will not be restating multiple instances through this code path as we only cleanup one instance at a time.

aravindhp: We shouldn't be spreading reboot responsibility in this manner.

pkg/servicescm/servicescm.go
mansikulkarni96: I don't have strong opinion on this. It made sense to me as WICD supports these set of env Vars. I can update it.

pkg/daemon/cleanup/cleanup.go
sebsoto: Or WMCO directly does the restart when it invokes the cleanup command.

mansikulkarni96: The last step of cleanup is deleting node object itself.

sebsoto: In case there is no node to annotate

mansikulkarni96: Are you suggesting WMCO always restart the instance after cleanup is executed and if proxy was enabled?

aravindhp: WMCO is what deletes the node object, so shouldn't it be present? If it is not present, can't we skip rebooting?

pkg/daemon/cleanup/cleanup.go
mansikulkarni96: I am refactoring this function it will now only return if restart is required and not perform it.

pkg/daemon/cleanup/cleanup.go
aravindhp, line 1: This comment is regarding the commit message:
>This commit expands the WICD cleanup command to remove
the ENV vars set on the instance if cluster-wide proxy
is enabled in the cluster.

WICD shouldn't care whether cluster-wide proxy is set or not. It should just be handling the watched environment variables irrespective of the proxy settings.

pkg/daemon/controller/controller.go
```diff
@@ -321,8 +321,7 @@ func (sc *ServiceController) reconcileEnvironmentVariables(envVars map[string]st
 
 // ensureVarsAreUpToDate ensures that the proxy environment variables are set as expected on the instance
 // If there's any changes, an instance restart is required to ensure all processes pick up the updated values.
-func (sc *ServiceController) ensureVarsAreUpToDate(envVars map[string]string) (bool, error) {
-	// systemEnvVarRegistryPath is where system level environment variables are stored in the Windows OS
+func ensureVarsAreUpToDate(envVars map[string]string, watchedEnvVars []string) (bool, error) {
 	const systemEnvVarRegistryPath = `SYSTEM\CurrentControlSet\Control\Session Manager\Environment`
```
aravindhp: You should use the `SystemEnvVarRegistryPath` here also.

aravindhp (changes requested):
Thanks for reworking this, @mansikulkarni96. 

A general note: I noticed that we are opening a registry key in a function, passing it to another function to actually remove it and then closing it in the calling function. Ideally this should all happen in the same function. But seeing how the code is structured it is going to take bit of a rewrite so lets attempt that later.

pkg/daemon/controller/controller.go
mansikulkarni96: this global variable is introduced in  the cleanup commit where it will be used out of this file.

test/e2e/delete_test.go
```diff
@@ -111,6 +115,30 @@ func (tc *testContext) checkNetworksRemoved(address string) (bool, error) {
...
 
+// checkEnvVarsRemoved returns true if the system and service level ENV vars do not exist on the Windows instance
+func (tc *testContext) checkEnvVarsRemoved(address string) (bool, error) {
+	watchedEnvVars := []string{"HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY"}
+	for _, proxyVar := range watchedEnvVars {
+		systemEnvVars, err := tc.getSystemEnvVar(address, proxyVar)
+		if err != nil {
+			return false, fmt.Errorf("error retrieving system level ENV vars: %w", err)
+		}
+		if systemEnvVars[proxyVar] != "" {
```
saifshaikh48, line 126:
```suggestion
		if _, exists := systemEnvVars[proxyVar]; exists {
```
otherwise this would erroneously pass if the var exists with value ""

test/e2e/delete_test.go
```diff
@@ -111,6 +115,30 @@ func (tc *testContext) checkNetworksRemoved(address string) (bool, error) {
...
+			return false, nil
+		}
+		for _, svcName := range windows.RequiredServices {
+			svcEnvVars, err := tc.getProxyEnvVarsFromService(address, svcName)
+			if err != nil {
+				return false, fmt.Errorf("error retrieving service level ENV vars: %w", err)
+			}
+			if svcEnvVars[proxyVar] != "" {
```
saifshaikh48, line 134: same as above

test/e2e/proxy_test.go
```diff
@@ -69,13 +68,48 @@ func testEnvVars(t *testing.T) {
...
 					require.NoErrorf(t, err, "error getting environment variables of service %s", svcName)
-					for _, proxyVar := range supportedProxyVars {
+					for _, proxyVar := range watchedEnvVars {
 						assert.Equalf(t, expectedEnvVars[proxyVar], svcEnvVars[proxyVar], "incorrect value for %s", proxyVar)
 					}
 				})
 			}
 		})
 	}
+	var patches []*patch.JSONPatch
```
saifshaikh48, line 78: flow looks good but could you add some comments to make it easier to understand? or move some logic into its own testEnvVarRemoval test method?

pkg/daemon/cleanup/cleanup.go
```diff
@@ -163,6 +183,25 @@ func removeServices(svcMgr manager.Manager, services []servicescm.Service) error
 	return nil
 }
 
+// ensureEnvVarsAreRemoved removes all WICD configured ENV variables from this instance
+func ensureEnvVarsAreRemoved(latestCMData, versionCMData *servicescm.Data) (bool, error) {
+	registryKey, err := registry.OpenKey(registry.LOCAL_MACHINE, controller.SystemEnvVarRegistryPath, registry.ALL_ACCESS)
```
saifshaikh48, line 188: is it possible to do this instead to reuse existing code
```
func ensureEnvVarsAreRemoved(latestCMData, versionCMData *servicescm.Data) (bool, error) {
    if versionCMData == nil {
        return controller.EnsureVarsAreUpToDate(map[string]string{}, latestCMData.WatchedEnvironmentVars)
    }
    return controller.EnsureVarsAreUpToDate(map[string]string{}, versionCMData.WatchedEnvironmentVars)
}
```

pkg/daemon/cleanup/cleanup.go
```diff
@@ -57,13 +58,24 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
...
 	}
-	if err = removeServices(svcMgr, services); err != nil {
+	if err = removeServices(latestCMData, versionCMData, svcMgr); err != nil {
 		return err
 	}
+	restartRequired, err := ensureEnvVarsAreRemoved(latestCMData, versionCMData)
+	if err != nil {
+		return err
+	}
+	// rebooting instance to set the environment variables at the process level as expected
```
saifshaikh48, line 72: set --> unset

pkg/daemon/cleanup/cleanup.go
```diff
@@ -57,13 +58,24 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
...
+	if restartRequired {
+		// Applying the reboot annotation results in an event picked up by WMCO's node controller to reboot the instance
+		if err = metadata.ApplyRebootAnnotation(ctx, directClient, *node); err != nil {
```
saifshaikh48, line 75: Can this lead to a timing issue where
- we apply the reboot annotation onto the node here as part of cleanup
- WICD cleanup returns
- WMCO deletes the node object, before node controller can process reboot annotation event

mansikulkarni96: I have thought about this scenario but haven't run into this issue during my testing. I ll take a look again at this.

pkg/daemon/controller/controller.go
```diff
@@ -357,6 +370,24 @@ func (sc *ServiceController) ensureVarsAreUpToDate(envVars map[string]string) (b
 	return restartRequired, nil
 }
 
+// EnsureEnvVarsAreRemoved ensures that the given environment variables are removed from the instance's Windows registry
+// An instance restart is required after they are removed to ensure all processes pick up the updated values.
+func EnsureEnvVarsAreRemoved(registryKey registry.Key, envVarsToRemove []string) (bool, error) {
+	restartRequired := false
```
sebsoto: Theres too many functions being named similar names throughout the PR, its getting a bit confusing 😅 
I think it would be best to create a separate package `pkg/daemon/envvar`, and move all the functions and constants that deal specifically with interacting with the Windows registry/API to alter environment variables there. This will help to create some structure, and prevent controller.go from becoming too bloated with general helper functions.

pkg/daemon/cleanup/cleanup.go
```diff
@@ -143,7 +161,8 @@ func mergeServices(s1, s2 []servicescm.Service) []servicescm.Service {
 }
 
 // removeServices uses the given manager to remove all the given Windows services from this instance.
-func removeServices(svcMgr manager.Manager, services []servicescm.Service) error {
+func removeServices(latestCMData, versionCMData *servicescm.Data, svcMgr manager.Manager) error {
+	services := getServicesToRemove(latestCMData, versionCMData)
```
sebsoto, line 165: This is overloading the responsibilities of this function

pkg/daemon/cleanup/cleanup.go
```diff
@@ -72,19 +83,19 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
 	return nil
 }
 
-// getServicesToRemove returns a list of services that should be removed as part of the cleanup process
-// returns the merged Data of the latest ConfigMap, and the ConfigMap specified by the node's version annotation
-func getServicesToRemove(ctx context.Context, cli client.Client, node *core.Node, configMapNamespace string) ([]servicescm.Service, error) {
+// getLatestAndVersionCMData attempts to get the latest and the version CM data that will be used as part of the
+// WICD cleanup process
+func getLatestAndVersionCMData(ctx context.Context, cli client.Client,
+	namespace string, node *core.Node) (*servicescm.Data, *servicescm.Data, error) {
```
sebsoto, line 89: I'm having a hard time understand why this change is being made. Is it because we no longer only need to merge the service list from each CM, but we also need to know the watched environment variables?
If that is the case instead of making this change, I suggest you change the existing function to return a merged servicescm.Data, instead of two separate CM. Because that is what is actually desired by the caller of this function. Doing this will allow you simplify the `removeServices()` and `cleanup.ensureEnvVarsAreRemoved()` functions as well

pkg/daemon/controller/controller.go
mansikulkarni96: @sebsoto I have no issues with creating another package. However, I want to understand the confusing part here. This is the only function introduced in controller package in this PR.  All the other functions related to env vars are pre-existing only they have been now exported or their params have changed through this PR.
The only other ensureEnvVarsAreRemoved exists in cleanup package as that basically goes through the same logic flow calling the functions in this package.

pkg/daemon/cleanup/cleanup.go
mansikulkarni96: @sebsoto this has to be done because we need the latest and version CM to be retrieved for both service deletion and environment var deletion.  I would have done what you are suggesting in first place but we only need the merged CM data for services removal and not for env Var Removal. In the previous discussions on this PR, we have decided not to use merged configmap data for env var removal as it is not a complete solution. Here, if the vesion CM data is not available we fall-back to using just the latest CM data.

pkg/daemon/cleanup/cleanup.go
mansikulkarni96: It was in deconfigure where it was only used to get the services to be removed. I could move it back but I think it is better suited where it is used instead of passing it in.

pkg/daemon/controller/controller.go
sebsoto: The confusing part is across the different commits in this PR, there are 5 separate changes to the function signatures of `controller.ensureEnvVarsAreRemoved`, `ensureVarsAreUpToDate` , and `cleanup.ensureEnvVarsAreRemoved`. Some of that is churn and its hard to keep track of when reviewing.

pkg/daemon/cleanup/cleanup.go
sebsoto: > we need the latest and version CM to be retrieved for both service deletion and environment var deletion

Yes, I see that both of them are passed into both `removeServices` and `ensureEnvVarsAreRemoved`

> but we only need the merged CM data for services removal and not for env Var Removal... we have decided not to use merged configmap data for env var removal as it is not a complete solution

Looking at the code within this PR its not clear to me why that is the case. I don't see how services and envVars would differ in the way they should be handled here. Wouldn't it be safer to use a merged list for the same reasons we need to for services? Partially configured instances which are upgraded will potentially have envVars set by previous versions of WMCO which we need to undo.

pkg/daemon/controller/controller.go
mansikulkarni96: I have moved envVar related functions into separate package. Please lmk if that helps.

pkg/daemon/cleanup/cleanup.go
```diff
@@ -142,8 +147,20 @@ func mergeServices(s1, s2 []servicescm.Service) []servicescm.Service {
...
+func mergeCMData(data1, data2 map[string]string) map[string]string {
+	mergedData := make(map[string]string)
+	for key, value := range data1 {
+		mergedData[key] = value
+	}
+	for key, value := range data2 {
+		mergedData[key] = value
+	}
+	return mergedData
+}
```
sebsoto, line 160: Don't you need to merge the services and envvars independently, and then create a new CM from the merged values? Doing this seems like it will just take everything (files, services, etc) from mergedData, as the keys will be the same.

pkg/daemon/cleanup/cleanup.go
```diff
@@ -142,8 +147,20 @@ func mergeServices(s1, s2 []servicescm.Service) []servicescm.Service {
...
+
 // removeServices uses the given manager to remove all the given Windows services from this instance.
-func removeServices(svcMgr manager.Manager, services []servicescm.Service) error {
+func removeServices(services []servicescm.Service, svcMgr manager.Manager) error {
```
sebsoto, line 163: can revert this change

pkg/daemon/cleanup/cleanup.go
saifshaikh48: Thanks, concern has been resolved. This was addressed by waiting for reboot annotation removal before deleting node object

sebsoto (approved):
Thanks for the hard work. LGTM

pkg/daemon/cleanup/cleanup.go
```diff
@@ -72,13 +84,15 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
 	return nil
 }
 
-// getServicesToRemove returns a list of services that should be removed as part of the cleanup process
-// returns the merged Data of the latest ConfigMap, and the ConfigMap specified by the node's version annotation
-func getServicesToRemove(ctx context.Context, cli client.Client, node *core.Node, configMapNamespace string) ([]servicescm.Service, error) {
+// getmergedCMData attempts to get the latest and the version CM data specified by the node's version annotation
+// It returns the merged CM Data containing services and the watched environment variables
+func getmergedCMData(ctx context.Context, cli client.Client,
```
saifshaikh48, line 89: nit: `getmergedCMData` --> `getMergedCMData`

saifshaikh48 (commented):
/lgtm

mansikulkarni96:
/retest-required
cluster spinup failures

mansikulkarni96:
/retest-required
I see a [green run](https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-openshift-windows-machine-config-operator-master-vsphere-proxy-e2e-operator) on the job board for another PR

PR #1707, opened 2023-07-27
WINC-1090: Support custom CA certificates for cluster-wide proxy

saifshaikh48:
Cluster admins can provide custom certificates that are required to communicate through their man-in-the-middle egress proxy. This PR covers importing certs Windows nodes so they can become fully proxy aware. 
This is done by updating the local trust store of all Windows instances with the certs present in the trusted-ca ConfigMap.

### DIFF
--- a/cmd/daemon/controller.go
+++ b/cmd/daemon/controller.go
@@ -39,6 +39,7 @@ var (
 	}
 	windowsService bool
 	logDir         string
+	caBundle       string
 )
 
 func init() {
@@ -47,6 +48,8 @@ func init() {
 		"if not provided, the command will log to stdout/stderr")
 	controllerCmd.PersistentFlags().BoolVar(&windowsService, "windows-service", false,
 		"Enables running as a Windows service")
+	controllerCmd.PersistentFlags().StringVar(&caBundle, "ca-bundle", "",
+		"the full path to CA bundle file containing certificates trusted by the cluster")
 }
 
 func runControllerCmd(cmd *cobra.Command, args []string) {
@@ -65,7 +68,7 @@ func runControllerCmd(cmd *cobra.Command, args []string) {
 		}
 	}
 	klog.Info("service controller running")
-	if err := controller.RunController(ctx, namespace, kubeconfig); err != nil {
+	if err := controller.RunController(ctx, namespace, kubeconfig, caBundle); err != nil {
 		klog.Error(err)
 		os.Exit(1)
 	}
--- a/pkg/daemon/certs/certs.go
+++ b/pkg/daemon/certs/certs.go
@@ -1,0 +1,198 @@
+//go:build windows
+
+/*
+Copyright 2023.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package certs
+
+import (
+	"bufio"
+	"bytes"
+	"crypto/x509"
+	"encoding/pem"
+	"errors"
+	"fmt"
+	"os"
+	"syscall"
+	"unsafe"
+
+	"golang.org/x/sys/windows"
+	"k8s.io/klog/v2"
+)
+
+// Reconcile ensures the certificates in the given trusted CA bundle file are imported as system certificates.
+// Returns a boolean whether any certs needed to be reconciled
+func Reconcile(caBundle string) (bool, error) {
+	if caBundle == "" {
+		// TODO: ensure that all certs that we've added previously are removed https://issues.redhat.com/browse/WINC-688
+		return false, nil
+	}
+	// Read expected certs from CA trust bundle file
+	expectedCerts, err := readCertsFromFile(caBundle)
+	if err != nil {
+		return false, fmt.Errorf("Failed to read certificate file: %v", err)
+	}
+	return ensureCertsAreImported(expectedCerts)
+}
+
+// readCertsFromFile reads in any PEM-encoded certificates from the given file
+func readCertsFromFile(path string) ([]*x509.Certificate, error) {
+	f, err := os.Open(path)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to open file: %v", err)
+	}
+	defer func() {
+		if err := f.Close(); err != nil {
+			klog.Errorf("Failed to close file %s", f.Name())
+		}
+	}()
+
+	// Process the file one certificate at a time
+	scanner := bufio.NewScanner(f)
+	scanner.Split(splitAtPEMCert())
+	var certs []*x509.Certificate
+	for scanner.Scan() {
+		// Read next token
+		pemData := scanner.Text()
+		block, _ := pem.Decode([]byte(pemData))
+		if block == nil {
+			return nil, fmt.Errorf("No PEM data read from token")
+		}
+		// Decode and parse the certificate
+		cert, err := x509.ParseCertificate(block.Bytes)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to parse certificate: %v\n", err)
+		}
+		certs = append(certs, cert)
+	}
+	return certs, nil
+}
+
+// splitAtPEMCert is a custom closure function to split data into complete PEM-encoded certificates
+func splitAtPEMCert() func(data []byte, atEOF bool) (advance int, token []byte, err error) {
+	const endCertTag = "-----END CERTIFICATE-----"
+	searchBytes := []byte(endCertTag)
+	searchLen := len(searchBytes)
+	return func(data []byte, atEOF bool) (advance int, token []byte, err error) {
+		dataLen := len(data)
+		if atEOF && dataLen == 0 {
+			// Empty file, do nothing
+			return 0, nil, nil
+		}
+
+		// Find next separator and return the token
+		if i := bytes.Index(data, searchBytes); i >= 0 {
+			return i + searchLen, data[0 : i+searchLen], nil
+		}
+
+		if atEOF {
+			return dataLen, data, fmt.Errorf("Hit end of file without finding terminating separator %s", endCertTag)
+		}
+		// Otherwise, continue reading file data
+		return 0, nil, nil
+	}
+}
+
+// ensureCertsAreImported imports the expected certificates into the instance's root system trust store, if not already
+// present. Returns a boolean if any certificates were imported
+func ensureCertsAreImported(expectedCerts []*x509.Certificate) (bool, error) {
+	// Open the root certificate store
+	systemStore, err := windows.CertOpenStore(windows.CERT_STORE_PROV_SYSTEM, 0, 0,
+		windows.CERT_SYSTEM_STORE_LOCAL_MACHINE, uintptr(unsafe.Pointer(syscall.StringToUTF16Ptr("ROOT"))))
+	if err != nil {
+		return false, fmt.Errorf("Failed to open root certificate store: %v", err)
+	}
+	defer func() {
+		if err := windows.CertCloseStore(systemStore, 0); err != nil {
+			klog.Error("Failed to close root system certificate store")
+		}
+	}()
+
+	// Get all existing from system store
+	existingCerts, err := getAllCerts(systemStore)
+	if err != nil {
+		return false, fmt.Errorf("Failed to read certificate file: %v", err)
+	}
+
+	certChange := false
+	for _, cert := range expectedCerts {
+		if containsCert(existingCerts, cert) {
+			// Cert already exists as expected, do nothing
+			continue
+		}
+
+		// Add the certificate to the instance's system-wide trust store
+		if err := addCertToStore(systemStore, cert); err != nil {
+			return false, err
+		}
+		certChange = true
+	}
+	return certChange, nil
+}
+
+// getAllCerts reads all the certs from the given certificate store
+func getAllCerts(store windows.Handle) ([]*x509.Certificate, error) {
+	var existingSystemCerts []*x509.Certificate
+	var certContext *windows.CertContext
+	var err error
+	for {
+		certContext, err = windows.CertEnumCertificatesInStore(store, certContext)
+		if err != nil {
+			if errors.Is(err, windows.Errno(windows.CRYPT_E_NOT_FOUND)) {
+				// Error code implies we have read all certs:
+				// https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-certenumcertificatesinstore
+				break
+			}
+			return nil, fmt.Errorf("Error reading existing system certificate: %v", err)
+		}
+
+		// Convert the certificate bytes to Golang x509.Certificate
+		certBytes := unsafe.Slice(certContext.EncodedCert, certContext.Length)
+		x509Cert, err := x509.ParseCertificate(certBytes)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to parse certificate from bytes: %v", err)
+		}
+		existingSystemCerts = append(existingSystemCerts, x509Cert)
+	}
+	klog.Infof("found %d existing system certificates", len(existingSystemCerts))
+	return existingSystemCerts, nil
+}
+
+// containsCert checks if the target certificate exists in the given slice
+func containsCert(storeContents []*x509.Certificate, target *x509.Certificate) bool {
+	for _, cert := range storeContents {
+		if cert.Equal(target) {
+			return true
+		}
+	}
+	return false
+}
+
+// addCertToStore imports the given certificate into the given certificate store
+func addCertToStore(store windows.Handle, cert *x509.Certificate) error {
+	// Convert x509 certificate to a Windows CertContext
+	certContext, err := windows.CertCreateCertificateContext(
+		windows.X509_ASN_ENCODING|windows.PKCS_7_ASN_ENCODING, &cert.Raw[0], uint32(len(cert.Raw)))
+	if err != nil {
+		return fmt.Errorf("Failed to create content from x509 cert %s: %v\n", cert.Subject.String(), err)
+	}
+	if err = windows.CertAddCertificateContextToStore(store, certContext,
+		windows.CERT_STORE_ADD_REPLACE_EXISTING, nil); err != nil {
+		return fmt.Errorf("Failed to import certificate %s: %v\n", cert.Subject.String(), err)
+	}
+	klog.Infof("Certificate %s imported successfully.", cert.Subject.String())
+	return nil
+}
--- a/pkg/daemon/controller/controller.go
+++ b/pkg/daemon/controller/controller.go
@@ -66,6 +66,7 @@ type Options struct {
 	Client    client.Client
 	Mgr       manager.Manager
 	cmdRunner powershell.CommandRunner
+	caBundle  string
 }
 
 // setDefaults returns an Options based on the received options, with all nil or empty fields filled in with reasonable
@@ -106,6 +107,7 @@ type ServiceController struct {
 	watchNamespace string
 	psCmdRunner    powershell.CommandRunner
 	ctrl           controller.Controller
+	caBundle       string
 }
 
 // Bootstrap starts all Windows services marked as necessary for node bootstrapping as defined in the given data
@@ -124,7 +126,7 @@ func (sc *ServiceController) Bootstrap(desiredVersion string) error {
 }
 
 // RunController is the entry point of WICD's controller functionality
-func RunController(ctx context.Context, watchNamespace, kubeconfig string) error {
+func RunController(ctx context.Context, watchNamespace, kubeconfig, caBundle string) error {
 	cfg, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
 	if err != nil {
 		return err
@@ -153,7 +155,7 @@ func RunController(ctx context.Context, watchNamespace, kubeconfig string) error
 	if err != nil {
 		return fmt.Errorf("unable to start manager: %w", err)
 	}
-	sc, err := NewServiceController(ctx, node.Name, watchNamespace, Options{Client: ctrlMgr.GetClient()})
+	sc, err := NewServiceController(ctx, node.Name, watchNamespace, Options{Client: ctrlMgr.GetClient(), caBundle: caBundle})
 	if err != nil {
 		return err
 	}
@@ -174,7 +176,7 @@ func NewServiceController(ctx context.Context, nodeName, watchNamespace string, 
 		return nil, err
 	}
 	return &ServiceController{client: o.Client, Manager: o.Mgr, ctx: ctx, nodeName: nodeName, psCmdRunner: o.cmdRunner,
-		watchNamespace: watchNamespace}, nil
+		watchNamespace: watchNamespace, caBundle: o.caBundle}, nil
 }
 
 // SetupWithManager sets up the controller with the Manager.
--- a/pkg/daemon/controller/controller.go
+++ b/pkg/daemon/controller/controller.go
@@ -48,6 +48,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	"github.com/openshift/windows-machine-config-operator/pkg/daemon/certs"
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/envvar"
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/manager"
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/powershell"
@@ -259,7 +260,7 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
 		return ctrl.Result{}, err
 	}
 
-	awaitingRestart, err := sc.reconcileEnvironmentVariables(cmData.EnvironmentVars, cmData.WatchedEnvironmentVars, node)
+	awaitingRestart, err := sc.reconcileEnvVarsAndCerts(cmData.EnvironmentVars, cmData.WatchedEnvironmentVars, node)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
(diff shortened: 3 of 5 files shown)


saifshaikh48:
pull-ci-openshift-windows-machine-config-operator-release-4.10-vsphere-e2e-upgrade

saifshaikh48:

cluster spinup failure

saifshaikh48:

check after rebase

pkg/daemon/controller/controller.go
```diff
@@ -282,15 +282,16 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
 	return ctrl.Result{}, nil
 }
 
-// reconcileEnvironmentVariables makes sure that the proxy variables exist as expected, or are safely rectified.
+// reconcileProxySettings makes sure that the proxy variables exist as expected, or are safely rectified.
```
aravindhp: I am not sure about this change. WICD should be generic wrt environment variables. It shouldn't be proxy specific. These variables just so happen to be related to enabling proxy in the environment.

pkg/daemon/certs/certs.go
```diff
@@ -0,0 +1,125 @@
...
+
+// Reconcile ensures the certificates in the given trusted CA bundle file are imported as system certificates.
+// Returns a boolean whether any certs have needed to be imported
+func Reconcile(trustBundle string) (bool, error) {
+	if trustBundle == "" {
+		// TODO: ensure that all certs that we've added previously are removed https://issues.redhat.com/browse/WINC-998
+		return false, nil
+	}
+	// Read expected certs from CA trust bundle file
+	pemData, err := ioutil.ReadFile(trustBundle)
```
aravindhp, line 24: Isn't `ioutil.ReadFile` deprecated? In general we shouldn't we trying to read the whole file in one shot. There could be potential of a DOS. It might be better to use `os.Open()` and then `io.Copy()`. Using a context aware reader will also help.

pkg/daemon/certs/certs.go
```diff
@@ -0,0 +1,125 @@
...
+	"syscall"
+	"unsafe"
+
+	syswindows "golang.org/x/sys/windows"
+	"k8s.io/klog/v2"
+)
+
+// Reconcile ensures the certificates in the given trusted CA bundle file are imported as system certificates.
+// Returns a boolean whether any certs have needed to be imported
+func Reconcile(trustBundle string) (bool, error) {
```
aravindhp, line 18: Can this function be made readable? Can you have a trust bundle object that you act upon?

aravindhp (commented):
@saifshaikh48 just left some early feedback.

pkg/daemon/certs/certs.go
```diff
@@ -0,0 +1,125 @@
...
+	if trustBundle == "" {
+		// TODO: ensure that all certs that we've added previously are removed https://issues.redhat.com/browse/WINC-998
```
mansikulkarni96, line 20: WINC-688*

pkg/daemon/certs/certs.go
```diff
@@ -0,0 +1,125 @@
+package certs
```
mansikulkarni96, line 1: Can you add the copyright text headers here.

pkg/daemon/certs/certs.go
```diff
@@ -0,0 +1,125 @@
...
+			}
+			// Add the certificate to the instances's system-wide trust store
+			if err = syswindows.CertAddCertificateContextToStore(store, certContext,
+				syswindows.CERT_STORE_ADD_REPLACE_EXISTING, nil); err != nil {
+				klog.Errorf("Failed to import certificate %s: %v\n", cert.Subject.String(), err)
+				continue
+			}
+			certChange = true
+			klog.Infof("Certificate %s imported successfully.", cert.Subject.String())
+		}
```
mansikulkarni96, line 69: will we retry the cert import if we don't return errors?

pkg/daemon/controller/controller.go
mansikulkarni96: I see that you are using this function to reconcile certs in the next commit, if so it can be named along the lines of
reconcileEnvVarsAndCerts or you could create another function that calls both of these.

pkg/daemon/certs/certs.go
saifshaikh48: yes, it'll retry any failed ones on the next WICD controller Reconcile() call (every 30 seconds)

pkg/daemon/certs/certs.go
saifshaikh48: Yup, it's deprecated, I'll use os.Open() and then io.Copy() instead.

I also can explore reading in the file one cert at a time, maybe using the `-----END CERTIFICATE-----` as a delimiter to the buffered reader

pkg/daemon/controller/controller.go
saifshaikh48: I'll rename the func and squash this commit with the next so it's clear how it'll be used

pkg/daemon/certs/certs.go
saifshaikh48: need to explore this more

saifshaikh48: Introduced more helper functions to keep this top level func more readable

pkg/daemon/certs/certs.go
```diff
@@ -0,0 +1,191 @@
...
+		return false, fmt.Errorf("Failed to read certificate file: %v", err)
+	}
+
+	// Open the root certificate store
+	systemStore, err := windows.CertOpenStore(windows.CERT_STORE_PROV_SYSTEM, 0, 0,
+		windows.CERT_SYSTEM_STORE_LOCAL_MACHINE, uintptr(unsafe.Pointer(syscall.StringToUTF16Ptr("ROOT"))))
+	if err != nil {
+		return false, fmt.Errorf("Failed to open root certificate store: %v", err)
+	}
+	defer windows.CertCloseStore(systemStore, 0)
```
aravindhp, line 57: Please don't ignore the error. Printing it will suffice.

pkg/daemon/certs/certs.go
```diff
@@ -0,0 +1,191 @@
...
+	return ensureCertsAreUpToDate(systemStore, expectedCerts, existingCerts), nil
+}
+
+// readCertsFromFile reads in any PEM-encoded certificates from the given file
+func readCertsFromFile(path string) ([]*x509.Certificate, error) {
+	f, err := os.Open(path)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to open file: %v", err)
+	}
+	defer f.Close()
```
aravindhp, line 74: Same comment here.

pkg/daemon/certs/certs.go
```diff
@@ -0,0 +1,191 @@
...
+				break
+			}
+			return nil, fmt.Errorf("Error reading existing system certificate: %v", err)
+		}
+
+		// Convert the certificate bytes to Golang x509.Certificate
+		certBytes := unsafe.Slice(certContext.EncodedCert, certContext.Length)
+		x509Cert, err := x509.ParseCertificate(certBytes)
+		if err != nil {
+			klog.Errorf("Failed to parse certificate from bytes %s: %v", certBytes, err)
```
aravindhp, line 144: This implies we are ignoring the error. We need to guarantee that all certs are handled and errors are returned even if one fails.

pkg/daemon/certs/certs.go
```diff
@@ -0,0 +1,191 @@
...
+		}
+		existingSystemCerts = append(existingSystemCerts, x509Cert)
+	}
+	klog.Infof("found %d existing system certificates", len(existingSystemCerts))
+	return existingSystemCerts, nil
+}
+
+// ensureCertsAreUpToDate imports any of the expected certificates that are not already present in the existing list
+// into the given Windows certificate store. Returns a boolean if any certificates were imported
+func ensureCertsAreUpToDate(store windows.Handle, expected, existing []*x509.Certificate) bool {
```
aravindhp, line 155: `ensureCertsAreImported`?

pkg/daemon/controller/controller.go
```diff
@@ -282,15 +283,20 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
...
-func (sc *ServiceController) reconcileEnvironmentVariables(envVars map[string]string, watchedEnvVars []string,
+func (sc *ServiceController) reconcileEnvVarsAndCerts(envVars map[string]string, watchedEnvVars []string,
 	node core.Node) (bool, error) {
-	restartRequired, err := envvar.EnsureVarsAreUpToDate(envVars, watchedEnvVars)
+	certsUpdated, err := certs.Reconcile(sc.trustBundle)
 	if err != nil {
 		return false, err
 	}
-	if restartRequired {
+	envVarsUpdated, err := envvar.EnsureVarsAreUpToDate(envVars, watchedEnvVars)
```
aravindhp, line 294: We should rename the function to `envvar.Reconcile` in a follow up commit.

aravindhp (changes requested):
Thanks for working on this, @saifshaikh48. Please address my comment.s

pkg/daemon/certs/certs.go
```diff
@@ -0,0 +1,191 @@
...
+	if err != nil {
+		return false, fmt.Errorf("Failed to read certificate file: %v", err)
+	}
+
+	return ensureCertsAreUpToDate(systemStore, expectedCerts, existingCerts), nil
+}
+
+// readCertsFromFile reads in any PEM-encoded certificates from the given file
+func readCertsFromFile(path string) ([]*x509.Certificate, error) {
+	f, err := os.Open(path)
```
sebsoto, line 70: I suggest using os.ReadFile() here.
The use of a reader for non streaming data is making this function a lot more complicated than it needs to be.
Readers are very useful for pipes and extremely large files, but in the case of a file that we know won't cause any memory issues, just use ReadFile and string parsing.

```
os.ReadFile()
strings.Split(endCertTag)
// for each:
strings.TrimSpace
```

pkg/daemon/certs/certs.go
sebsoto: +1, I don't see a reason to not return an error here.

pkg/daemon/certs/certs.go
```diff
@@ -0,0 +1,191 @@
...
+// into the given Windows certificate store. Returns a boolean if any certificates were imported
+func ensureCertsAreUpToDate(store windows.Handle, expected, existing []*x509.Certificate) bool {
+	certChange := false
+	for _, cert := range expected {
+		if containsCert(existing, cert) {
+			// Cert already exists as expected, do nothing
+			continue
+		}
+
+		// Convert x509 certificate to a Windows CertContent
```
sebsoto, line 163: CertContent -> CertContext ?

sebsoto (changes requested):

pkg/daemon/certs/certs.go
sebsoto: This will remove the need for the splitAt() function being introduced

pkg/daemon/certs/certs.go
```diff
@@ -0,0 +1,191 @@
...
+		return false, nil
+	}
+	// Read expected certs from CA trust bundle file
+	expectedCerts, err := readCertsFromFile(trustBundle)
+	if err != nil {
+		return false, fmt.Errorf("Failed to read certificate file: %v", err)
+	}
+
+	// Open the root certificate store
+	systemStore, err := windows.CertOpenStore(windows.CERT_STORE_PROV_SYSTEM, 0, 0,
```
jrvaldes, line 52: Does it make sense to cache the `store` handler so that it get initialized once during the lifespan of the ServiceController?

jrvaldes (commented):
Good work @saifshaikh48, left a few comments PTAL.

pkg/windows/windows.go
```diff
@@ -470,7 +471,9 @@ func (vm *windows) ConfigureWICD(watchNamespace, wicdKubeconfigContents string)
 	}
 	wicdServiceArgs := fmt.Sprintf("controller --windows-service --log-dir %s --kubeconfig %s --namespace %s",
 		wicdLogDir, wicdKubeconfigPath, watchNamespace)
-
+	if cluster.IsProxyEnabled() {
+		wicdServiceArgs = fmt.Sprintf("%s --trust-bundle %s", wicdServiceArgs, TrustedCABundlePath)
```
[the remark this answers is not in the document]
saifshaikh48: Yes, CNO decides what certs to inject into the trusted-ca configmap in all proxied environments
So even if only HTTP is used and even if the trust bundle is empty, the file will still exist.

pkg/daemon/certs/certs.go
aravindhp: @saifshaikh48 had used `os.ReadFile()` originally and I had requested the change to not use it. I am not a fan of using `os.ReadFile()` on files whose content we have no control over.

pkg/daemon/certs/certs.go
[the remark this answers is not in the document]
saifshaikh48: Do you see a case where the deferred close doesn't work as intended?

pkg/daemon/certs/certs.go
saifshaikh48: I was initially reading the file directly into memory but @aravindhp brought up a [good point](https://github.com/openshift/windows-machine-config-operator/pull/1707#discussion_r1305338337). Let me know if you feel strongly that the reader is not needed

The cert bundle has the potential to be large, since it is a combination of user provided certs and cluster-wide trusted CA certificates for platform components ([as deduced by CNO](https://access.redhat.com/documentation/en-us/openshift_container_platform/4.12/html/networking/configuring-a-custom-pki))

pkg/daemon/certs/certs.go
```diff
@@ -0,0 +1,191 @@
...
+		if err != nil {
+			klog.Errorf("Failed to create content from x509 cert %s: %v\n", cert.Subject.String(), err)
+			continue
+		}
+		// Add the certificate to the instance's system-wide trust store
+		if err = windows.CertAddCertificateContextToStore(store, certContext,
+			windows.CERT_STORE_ADD_REPLACE_EXISTING, nil); err != nil {
+			klog.Errorf("Failed to import certificate %s: %v\n", cert.Subject.String(), err)
+			continue
+		}
```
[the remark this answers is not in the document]
saifshaikh48: I was thinking that we shouldn't stop reconciling the rest of the certs if one fails, but I've been convinced otherwise

pkg/daemon/certs/certs.go
```diff
@@ -0,0 +1,191 @@
...
+			return 0, nil, nil
+		}
+
+		// Find next separator and return the token
+		if i := bytes.Index(data, searchBytes); i >= 0 {
+			return i + searchLen, data[0 : i+searchLen], nil
+		}
+
+		if atEOF {
+			// Hit end of file before finding another separator, return rest of the data
```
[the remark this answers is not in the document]
saifshaikh48: good, point will return an error

pkg/daemon/certs/certs.go
```diff
@@ -0,0 +1,191 @@
...
+			klog.Errorf("Failed to parse certificate: %v\n", err)
+			continue
+		}
+		certs = append(certs, cert)
+	}
+	return certs, nil
+}
+
+// splitAt is a custom closure function to split data into complete PEM-encoded certificates
+func splitAt(substring string) func(data []byte, atEOF bool) (advance int, token []byte, err error) {
```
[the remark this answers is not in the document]
saifshaikh48: not intended, fixed

pkg/daemon/certs/certs.go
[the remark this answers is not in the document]
mansikulkarni96: I don't, this is something preventive. I recognize this is a followed pattern so I don't want to block the PR on it.
This can be proposed in a follow-up PR and we can get it updated everywhere if needed.

pkg/daemon/certs/certs.go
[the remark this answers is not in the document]
saifshaikh48: need to figure out the best way to unit test this, TODO

test/e2e/proxy_test.go
```diff
@@ -32,6 +34,44 @@ func proxyTestSuite(t *testing.T) {
 
 	t.Run("Trusted CA ConfigMap validation", testTrustedCAConfigMap)
 	t.Run("Environment variables validation", testEnvVars)
+	t.Run("Certificate validation", testCerts)
```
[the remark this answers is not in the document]
saifshaikh48: I looked into this a bit but am unsure what endpoint to use with the probe

pkg/daemon/certs/certs.go
```diff
@@ -0,0 +1,192 @@
...
+			return 0, nil, nil
+		}
+
+		// Find next separator and return the token
+		if i := bytes.Index(data, searchBytes); i >= 0 {
+			return i + searchLen, data[0 : i+searchLen], nil
+		}
+
+		if atEOF {
+			return dataLen, data, fmt.Errorf("Hit end of file without terminating the current token")
```
aravindhp, line 121: Some context here too will help.

pkg/daemon/certs/certs.go
```diff
@@ -0,0 +1,192 @@
...
+				break
+			}
+			return nil, fmt.Errorf("Error reading existing system certificate: %v", err)
+		}
+
+		// Convert the certificate bytes to Golang x509.Certificate
+		certBytes := unsafe.Slice(certContext.EncodedCert, certContext.Length)
+		x509Cert, err := x509.ParseCertificate(certBytes)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to parse certificate from bytes %s: %v", certBytes, err)
```
aravindhp, line 148: Is `certBytes` sensitive?

pkg/daemon/certs/certs.go
```diff
@@ -0,0 +1,192 @@
...
+		if err := windows.CertCloseStore(systemStore, 0); err != nil {
+			klog.Error("Failed to close root system certificate store")
+		}
+	}()
+
+	// Get all existing from system store
+	existingCerts, err := getAllCerts(systemStore)
+	if err != nil {
+		return false, fmt.Errorf("Failed to read certificate file: %v", err)
+	}
```
aravindhp, line 65: This code should ideally be in `ensureCertsAreImported()`. 

PS: Even better have a struct around all of this but I won't insist. If I have some cycles I will try to do a refactor just to show an example.

pkg/daemon/certs/certs.go
```diff
@@ -0,0 +1,192 @@
...
+		if containsCert(existing, cert) {
+			// Cert already exists as expected, do nothing
+			continue
+		}
+
+		// Convert x509 certificate to a Windows CertContext
+		certContext, err := windows.CertCreateCertificateContext(
+			windows.X509_ASN_ENCODING|windows.PKCS_7_ASN_ENCODING, &cert.Raw[0], uint32(len(cert.Raw)))
+		if err != nil {
+			return false, fmt.Errorf("Failed to create content from x509 cert %s: %v\n", cert.Subject.String(), err)
```
aravindhp, line 170: Is the subject always populated?

aravindhp (changes requested):

pkg/daemon/certs/certs.go
sebsoto: Thats fine with me.

test/e2e/proxy_test.go
```diff
@@ -32,6 +34,44 @@ func proxyTestSuite(t *testing.T) {
...
+}
+
+func testCerts(t *testing.T) {
+	tc, err := NewTestContext()
+	require.NoError(t, err)
```
sebsoto, line 42: I missed this in previous proxy PRs, but this pattern of creating a new testContext within each individual test has been removed from the rest of the e2e tests. Instead NewTestContext is called once at the beginning of each test suite, and the tests which require its use are made methods of it.

The reason for this ask is that by using the previous pattern, each test ends up having a repeated boiler plate code within them. 

If you can please add a commit changing this for the rest of the tests. If you rather not do that within this PR I am fine with it being done only for this test. Meaning within this PR changing only this function to be a method of testContext, without changing the others.

test/e2e/proxy_test.go
```diff
@@ -32,6 +34,44 @@ func proxyTestSuite(t *testing.T) {
...
+
+	// Read all certs from CM data
+	trustedCABundle := cm.Data[certificates.CABundleKey]
+	assert.Greater(t, len(trustedCABundle), 0, "no additional user-provided certs in bundle")
+
+	certs := x509.NewCertPool()
+	require.True(t, certs.AppendCertsFromPEM([]byte(trustedCABundle)), "unable to parse certs from trusted CA ConfigMap data")
+	subjects := certs.Subjects()
+	// Ensure each cert has been imported into every Windows instance's system store
+	for _, node := range gc.allNodes() {
```
sebsoto, line 56: Where is `gc.allNodes` being set?
What happens if a developer just wants to run this individual test?

pkg/daemon/certs/certs.go
sebsoto: Its safer to only open the handle when we specifically need it.

pkg/daemon/certs/certs.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+	defer func() {
+		if err := f.Close(); err != nil {
+			klog.Errorf("Failed to close file %s", f.Name())
+		}
+	}()
+
+	// Process the file one certificate at a time
+	scanner := bufio.NewScanner(f)
+	const endCertTag = "-----END CERTIFICATE-----"
```
sebsoto, line 84: nit: this could actually be moved within `splitAt`, and its parameter removed, unless theres a need for it to be a generic function :sweat_smile:

pkg/daemon/certs/certs.go
sebsoto: +1

sebsoto (changes requested):

pkg/daemon/certs/certs.go
sebsoto: I guess it could be an attack vector. :thinking:

pkg/daemon/certs/certs.go
jrvaldes: True

pkg/daemon/certs/certs.go
saifshaikh48: It'll always be non-nil. Worst case Subject.String() will return empty string so no worry of panic

pkg/daemon/certs/certs.go
saifshaikh48: potentially, I'll remove it from the error msg

pkg/daemon/certs/certs.go
```diff
@@ -0,0 +1,192 @@
...
+	scanner.Split(splitAt(endCertTag))
+	var certs []*x509.Certificate
+	for scanner.Scan() {
+		// Read next token
+		pemData := scanner.Text()
+		block, _ := pem.Decode([]byte(pemData))
+		if block == nil {
+			return nil, fmt.Errorf("No PEM data read from token")
```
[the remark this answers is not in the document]
saifshaikh48: It could leak a cert -- IMO best to be cautious and just let the user know the issue is that the file data is not PEM-encoded

test/e2e/proxy_test.go
saifshaikh48: I see, I'll call `tc.loadExistingNodes()` at the start of the proxy test suite so we can use gc.allNodes() safely

We might need to do the same in the storage and network test suites, could you confirm my understanding? If so that can be done in a followup PR

test/e2e/proxy_test.go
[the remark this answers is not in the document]
mansikulkarni96: That depends on the certs we are using, can you remind me where the certs are created for the test suite, is it through the job itself?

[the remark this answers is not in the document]
saifshaikh48: For the test suite, the additional certs are the RHCOS trust bundle (CNO provides this to us via the trusted-ca configmap)

pkg/daemon/certs/certs.go
```diff
@@ -0,0 +1,191 @@
...
+			return nil, fmt.Errorf("Error reading existing system certificate: %v", err)
+		}
+
+		// Convert the certificate bytes to Golang x509.Certificate
+		certBytes := unsafe.Slice(certContext.EncodedCert, certContext.Length)
+		x509Cert, err := x509.ParseCertificate(certBytes)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to parse certificate from bytes: %v", err)
+		}
+		existingSystemCerts = append(existingSystemCerts, x509Cert)
```
aravindhp, line 177: Nit: this would in turn become a helper method: `addCertFromStore()`

aravindhp (commented):
LGTM. I will let @sebsoto approve as he had asked for changes.

test/e2e/proxy_test.go
sebsoto: yeah either tc.loadExistingNodes() or waitForWindowsNodes needs to be called. Things could be a bit more consistent, but both achieve the same thing of populating the nodes within gc.

sebsoto (approved):

pkg/daemon/certs/certs.go
```diff
@@ -0,0 +1,198 @@
...
+	"syscall"
+	"unsafe"
+
+	"golang.org/x/sys/windows"
+	"k8s.io/klog/v2"
+)
+
+// Reconcile ensures the certificates in the given trusted CA bundle file are imported as system certificates.
+// Returns a boolean whether any certs needed to be reconciled
+func Reconcile(caBundle string) (bool, error) {
```
jrvaldes, line 38: s/caBundle/caBundlePAth

to avoid confusion with the ca bundle raw data

cmd/daemon/controller.go
```diff
@@ -39,6 +39,7 @@ var (
 	}
 	windowsService bool
 	logDir         string
+	caBundle       string
```
jrvaldes, line 42: s/caBundle/caBundlePath

pkg/daemon/certs/certs.go
```diff
@@ -0,0 +1,198 @@
...
+
+// readCertsFromFile reads in any PEM-encoded certificates from the given file
+func readCertsFromFile(path string) ([]*x509.Certificate, error) {
+	f, err := os.Open(path)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to open file: %v", err)
+	}
+	defer func() {
+		if err := f.Close(); err != nil {
+			klog.Errorf("Failed to close file %s", f.Name())
```
jrvaldes, line 59: In case of an issue closing the handler  the error is getting swallowed by the log, should this thrown instead?

pkg/daemon/certs/certs.go
```diff
@@ -0,0 +1,198 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("Failed to parse certificate: %v\n", err)
+		}
+		certs = append(certs, cert)
+	}
+	return certs, nil
+}
+
+// splitAtPEMCert is a custom closure function to split data into complete PEM-encoded certificates
+func splitAtPEMCert() func(data []byte, atEOF bool) (advance int, token []byte, err error) {
```
jrvaldes, line 85: It would be nice to cover this with unit tests; especially for the edge cases i.e. `if atEOF && dataLen == 0` , `if atEOF`, etc

pkg/daemon/certs/certs.go
```diff
@@ -0,0 +1,198 @@
...
+		certChange = true
+	}
+	return certChange, nil
+}
+
+// getAllCerts reads all the certs from the given certificate store
+func getAllCerts(store windows.Handle) ([]*x509.Certificate, error) {
+	var existingSystemCerts []*x509.Certificate
+	var certContext *windows.CertContext
+	var err error
```
jrvaldes, line 150: Consider declaring the error and the list as return parameters in the function's signature.

jrvaldes (approved):
LGTM, added minor suggestions.

pkg/daemon/certs/certs.go
saifshaikh48: True, I tried to think through the best way to unit test but couldn't write something clean. Maybe I'll write unit tests for the func that uses this `readCertsFromFile` instead (followup PR)

pkg/daemon/certs/certs.go
saifshaikh48: throwing the err in this defer block would overwrite any other error returned by the func. Logging such errors is a compromise between that and ignoring them completely

saifshaikh48:

Deletion test failure

PR #1710, opened 2023-07-28
WMCO removes reboot annotation instead of WICD

saifshaikh48:
This PR flips the waiting and removal of the reboot annotation, with WICD now waiting on WMCO to remove the annotation. The reason for this is to decouple WMCO's node controller from the WICD controller -- WMCO removes the reboot annotation as soon as the instance is reachable again after rebooting. Now the safeReboot() func returns after it has done just that, safely rebooted, without waiting on some extra WICD signal.

### DIFF
--- a/pkg/daemon/controller/controller.go
+++ b/pkg/daemon/controller/controller.go
@@ -182,17 +182,19 @@ func NewServiceController(ctx context.Context, nodeName, watchNamespace string, 
 func (sc *ServiceController) SetupWithManager(ctx context.Context, mgr ctrl.Manager) error {
 	nodePredicate := predicate.Funcs{
 		// A node's name will never change, so it is fine to use the name for node identification
-		// The node must have a desired-version annotation for it to be reconcilable
+		// The node must have a desired-version annotation and not be waiting for a reboot for it to be reconcilable
 		CreateFunc: func(e event.CreateEvent) bool {
-			return sc.nodeName == e.Object.GetName() && e.Object.GetAnnotations()[metadata.DesiredVersionAnnotation] != ""
+			return sc.nodeName == e.Object.GetName() && !isAwaitingReboot(e.Object) &&
+				e.Object.GetAnnotations()[metadata.DesiredVersionAnnotation] != ""
 		},
 		UpdateFunc: func(e event.UpdateEvent) bool {
-			// Only process update events if the desired version has changed
-			return sc.nodeName == e.ObjectNew.GetName() &&
+			// Only process update events if the desired version has changed and there is no reboot required
+			return sc.nodeName == e.ObjectNew.GetName() && !isAwaitingReboot(e.ObjectNew) &&
 				e.ObjectOld.GetAnnotations()[metadata.DesiredVersionAnnotation] != e.ObjectNew.GetAnnotations()[metadata.DesiredVersionAnnotation]
 		},
 		GenericFunc: func(e event.GenericEvent) bool {
-			return sc.nodeName == e.Object.GetName() && e.Object.GetAnnotations()[metadata.DesiredVersionAnnotation] != ""
+			return sc.nodeName == e.Object.GetName() && !isAwaitingReboot(e.Object) &&
+				e.Object.GetAnnotations()[metadata.DesiredVersionAnnotation] != ""
 		},
 		DeleteFunc: func(e event.DeleteEvent) bool {
 			return false
@@ -200,6 +202,9 @@ func (sc *ServiceController) SetupWithManager(ctx context.Context, mgr ctrl.Mana
 	}
 	cmPredicate := predicate.NewPredicateFuncs(func(object client.Object) bool {
 		return strings.HasPrefix(object.GetName(), servicescm.NamePrefix)
+	})
+	rebootPredicate := predicate.NewPredicateFuncs(func(object client.Object) bool {
+		return !isAwaitingReboot(object)
 	})
 
 	// Keeping this on the longer side, as each reconciliation requires running each service's powershell scripts
@@ -211,7 +216,8 @@ func (sc *ServiceController) SetupWithManager(ctx context.Context, mgr ctrl.Mana
 		For(&core.Node{}, builder.WithPredicates(nodePredicate)).
 		Watches(&core.ConfigMap{}, handler.EnqueueRequestsFromMapFunc(sc.mapToCurrentNode),
 			builder.WithPredicates(cmPredicate)).
-		WatchesRawSource(&source.Channel{Source: eventChan}, handler.EnqueueRequestsFromMapFunc(sc.mapToCurrentNode)).
+		WatchesRawSource(&source.Channel{Source: eventChan}, handler.EnqueueRequestsFromMapFunc(sc.mapToCurrentNode),
+			builder.WithPredicates(rebootPredicate)).
 		Complete(sc)
 }
 
@@ -310,8 +316,7 @@ func (sc *ServiceController) reconcileEnvironmentVariables(envVars map[string]st
 	if err != nil {
 		return true, fmt.Errorf("error waiting for environment vars to get picked up by processes: %w", err)
 	}
-	// Remove the reboot annotation after we know the reboot occurred successfully. No-op if already not present
-	return false, metadata.RemoveRebootAnnotation(sc.ctx, sc.client, node)
+	return false, nil
 }
 
 // ensureVarsAreUpToDate ensures that the proxy environment variables are set as expected on the instance
@@ -651,4 +656,14 @@ func slicesEquivalent(s1, s2 []string) bool {
 	}
 	return reflect.DeepEqual(s1, s2)
 
+}
+
+// isAwaitingReboot returns true if the given object is a node that is awaiting a reboot by WMCO
+func isAwaitingReboot(obj runtime.Object) bool {
+	node, ok := obj.(*core.Node)
+	if !ok {
+		return false
+	}
+	_, exists := node.Annotations[metadata.RebootAnnotation]
+	return exists
 }
--- a/pkg/nodeconfig/nodeconfig.go
+++ b/pkg/nodeconfig/nodeconfig.go
@@ -263,8 +263,8 @@ func (nc *nodeConfig) SafeReboot(ctx context.Context) error {
 	if err := nc.Windows.RebootAndReinitialize(); err != nil {
 		return err
 	}
-	// wait for WICD to remove Reboot annotation
-	if err := metadata.WaitForRebootAnnotationRemoval(ctx, nc.client, nc.node.Name); err != nil {
+	// Remove the reboot annotation after we can re-init an SSH connection so we know the reboot occurred successfully
+	if err := metadata.RemoveRebootAnnotation(ctx, nc.client, *nc.node); err != nil {
 		return err
 	}
 


pkg/nodeconfig/nodeconfig.go
```diff
@@ -258,8 +258,8 @@ func (nc *nodeConfig) SafeReboot(ctx context.Context, node *core.Node) error {
 	if err := nc.Windows.RebootAndReinitialize(); err != nil {
 		return err
 	}
-	// wait for WICD to remove Reboot annotation
-	if err := metadata.WaitForRebootAnnotationRemoval(ctx, nc.client, nc.node.Name); err != nil {
+	// Remove the reboot annotation after we know the reboot occurred successfully
+	if err := metadata.RemoveRebootAnnotation(ctx, nc.client, *nc.node); err != nil {
```
aravindhp, line 262: Do you really need to dereference the pointer here as per https://go.dev/ref/spec#Selectors?

pkg/nodeconfig/nodeconfig.go
```diff
@@ -258,8 +258,8 @@ func (nc *nodeConfig) SafeReboot(ctx context.Context, node *core.Node) error {
...
 		return err
 	}
```
aravindhp, line 265: I don't think this has anything to do with this PR, but are we uncordoning the node too soon before WICD has finished configuring services? The context for the question is for nodes that have already been configured before and has the correct version annotation.

aravindhp (commented):
Thanks for doing this, @saifshaikh48. I believe even with this change only one node will get rebooted at a time as long as [vm.interact.init()](https://github.com/saifshaikh48/windows-machine-config-operator/blob/72f9caa0f8b977f670db3bdec3e076fee73fb6d2/pkg/windows/connectivity.go#L72) always returns an error if it cannot establish a connection.

pkg/nodeconfig/nodeconfig.go
saifshaikh48: The deference is needed unless we change RemoveVersionAnnotation()'s signature to expect pointer

pkg/nodeconfig/nodeconfig.go
saifshaikh48: I'm not sure. If so, then we'd have to introduce some sort of wait but I am not sure how WICD would communicate that it has finished configuring services

aravindhp: Maybe this is fine. If the version annotation matches, we don't kick off a reconfiguration. And if the version annotation has changed, WCMO will wait for WICD to apply the version annotation. Correct?

saifshaikh48: Correct

aravindhp: @sebsoto please review and ack if you agree

sebsoto (approved):
Changes introduced in the PR LGTM

pkg/nodeconfig/nodeconfig.go
sebsoto: The only reasonable solution I have for that issue is WICD applying a `NeedsCordon`/`NeedsDrain` annotation, and WMCO reacts to its presence or lack of presence. This along the lines of what the MCO does.

pkg/daemon/controller/controller.go
```diff
@@ -310,8 +310,8 @@ func (sc *ServiceController) reconcileEnvironmentVariables(envVars map[string]st
 	if err != nil {
 		return true, fmt.Errorf("error waiting for environment vars to get picked up by processes: %w", err)
 	}
-	// Remove the reboot annotation after we know the reboot occurred successfully. No-op if already not present
-	return false, metadata.RemoveRebootAnnotation(sc.ctx, sc.client, node)
+	// Wait for WMCO to remove Reboot annotation
+	return false, metadata.WaitForRebootAnnotationRemoval(sc.ctx, sc.client, node.Name)
```
sebsoto, line 314: shouldn't this be done at the start of reconciliation?

aravindhp (changes requested):

pkg/daemon/controller/controller.go
```diff
@@ -252,15 +257,18 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
...
 	if err != nil {
 		return ctrl.Result{}, err
 	}
-	if awaitingRestart {
-		// if a restart is required, there is nothing more the controller can do until the instance is rebooted
+	if restartRequired {
+		// If there's any changes, an instance restart is required to ensure all processes pick up the updates.
+		// Applying the reboot annotation results in an event picked up by WMCO's node controller to reboot the instance.
+		// If a restart is required, there is nothing more the controller can do until the instance is rebooted
 		klog.Info("waiting for reboot")
```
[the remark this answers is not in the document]
saifshaikh48: Sure I can add it, the "reboot completed" log will be from WMCO's side. For WICD, we will know reboot is done because 1. service will restart and 2. it will start reacting to node events again

pkg/daemon/controller/controller.go
```diff
@@ -652,3 +622,13 @@ func slicesEquivalent(s1, s2 []string) bool {
 	return reflect.DeepEqual(s1, s2)
 
 }
+
+// isAwaitingReboot returns true if the given object is a node that is awaiting a reboot by WMCO
+func isAwaitingReboot(obj runtime.Object) bool {
```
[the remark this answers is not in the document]
saifshaikh48: generally we've been putting these predicate helper funcs in the controller that uses them

pkg/daemon/controller/controller.go
```diff
@@ -276,50 +284,12 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
...
-	// Remove the reboot annotation after we know the reboot occurred successfully. No-op if already not present
-	return false, metadata.RemoveRebootAnnotation(sc.ctx, sc.client, node)
-}
-
-// ensureVarsAreUpToDate ensures that the proxy environment variables are set as expected on the instance
-// If there's any changes, an instance restart is required to ensure all processes pick up the updated values.
-func (sc *ServiceController) ensureVarsAreUpToDate(envVars map[string]string) (bool, error) {
+// reconcileEnvironmentVariables ensures that the environment variables are set as expected on the instance.
+// Returns a boolean if any of the environment variables had to be updated.
+func reconcileEnvironmentVariables(envVars map[string]string) (bool, error) {
```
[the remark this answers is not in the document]
saifshaikh48: I'll move this to be just after the struct methods, but IMO it should stay in this package given its so closely tied to the Reconcile() func

pkg/daemon/controller/controller.go
```diff
@@ -549,6 +480,45 @@ func (sc *ServiceController) waitUntilNodeReady() error {
 		})
 }
 
+// reconcileEnvironmentVariables ensures that the environment variables are set as expected on the instance.
+// Returns true if any of the environment variables were updated.
+func reconcileEnvironmentVariables(envVars map[string]string) (bool, error) {
```
sebsoto, line 485: please rename envVars to `expectedEnvVars` or something along those lines.

pkg/daemon/controller/controller.go
```diff
@@ -549,6 +480,45 @@ func (sc *ServiceController) waitUntilNodeReady() error {
...
+				klog.Errorf("could not close key: %v", closeErr)
+			}
+		}()
+
+		actualVal, _, err := registryKey.GetStringValue(key)
+		if err != nil && err != registry.ErrNotExist {
+			return false, fmt.Errorf("unable to read environment variable %s: %w", key, err)
+		}
+
+		if actualVal != expectedVal {
```
sebsoto, line 506: This misses the corner case where the registry key doesn't exist, but it should exist with a value of `""`.
This can be covered by `err != registry.ErrNotExist || actualVal != expectedVal `

sebsoto (changes requested):

pkg/nodeconfig/nodeconfig.go
```diff
@@ -263,14 +263,15 @@ func (nc *nodeConfig) SafeReboot(ctx context.Context) error {
 	if err := nc.Windows.RebootAndReinitialize(); err != nil {
 		return err
 	}
-	// wait for WICD to remove Reboot annotation
```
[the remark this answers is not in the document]
saifshaikh48: The changes are distinct, this commit adjusts WMCO's behavior and the next addresses WICD. I can squash though since the changes are closely tied together

pkg/daemon/controller/controller.go
```diff
@@ -276,83 +284,6 @@ func (sc *ServiceController) Reconcile(_ context.Context, req ctrl.Request) (res
...
-		}
-		return true, nil
-	}
-	// Wait until the environment variables at the process level are set as expected.
-	// This will only be picked up after WMCO reboots the instance
-	err = wait.PollUntilContextTimeout(sc.ctx, 15*time.Second, 5*time.Minute, true,
-		func(ctx context.Context) (done bool, err error) {
-			stillNeedsReboot := false
-			for _, varName := range cluster.SupportedProxyVars {
-				cmd := fmt.Sprintf("[Environment]::GetEnvironmentVariable('%s', 'Process')", varName)
```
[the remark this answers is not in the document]
saifshaikh48: I always felt like this check was an overreach by WICD. What we were doing with this is ensuring Windows OS is working as expected i.e. that when we reboot, the OS propagates system level env vars to services. We should only be waiting for a reboot

pkg/daemon/controller/controller.go
saifshaikh48: I assume you meant `err == registry.ErrNotExist || actualVal != expectedVal`, I'll add the clause

pkg/daemon/controller/controller.go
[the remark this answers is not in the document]
sebsoto: > What we were doing with this is ensuring Windows OS is working as expected

Not really, WICD needs to check the current state of the VM and respond accordingly. What if WICD crashed right after changing the env variables and failed to apply the annotation? we can't assume a restart happened just because WMCO starts and there is no restart annotation.

pkg/daemon/controller/controller.go
```diff
@@ -310,8 +316,7 @@ func (sc *ServiceController) reconcileEnvironmentVariables(envVars map[string]st
 	if err != nil {
 		return true, fmt.Errorf("error waiting for environment vars to get picked up by processes: %w", err)
 	}
-	// Remove the reboot annotation after we know the reboot occurred successfully. No-op if already not present
-	return false, metadata.RemoveRebootAnnotation(sc.ctx, sc.client, node)
+	return false, nil
```
sebsoto, line 319: sorry if there was a miscommunication. I was saying you could leave this as WaitForRebootAnnotation to be removed

saifshaikh48: I am confused again :laughing: 
Isn't adding the reboot anno to the predicate accomplishing the same thing as waiting for the anno to be removed? We'd only reach this point in the code if the anno is not present already (otherwise we wouldn't even reconcile)

sebsoto: if you just continue wouldn't this result in the node getting marked as configured before it is rebooted (and working properly)?

thats why i thought the wait was necessary

saifshaikh48: That was true before the addition to the predicate.

Now, we are returning `false, nil` here only if the node is safe to be configured. All of the following must be true to hit this line
- there is no reboot anno on the node (reconcile event actually get processed)
- system level env vars are properly set (otherwise function would exit early after applying RebootRequired anno)
- process level env vars are picked up properly (otherwise the Poll would error out)
That's a green light to continue on with the services configuration etc

sebsoto: Thanks for the explanation, makes sense to me

sebsoto (approved):

jrvaldes (approved):
/lgtm

PR #1711, opened 2023-08-01
WINC-633: Copy additional CA bundle to Windows instances

saifshaikh48:
This PR is part 1 of supporting user-provided certs. 

WMCO copies over the CA bundle with all trusted cluster certs to Windows instances. 
This is done, if needed, during node configuration as well as when there are any 
events on the ConfigMap holding the trusted CA.

### DIFF
--- a/controllers/configmap_controller.go
+++ b/controllers/configmap_controller.go
@@ -75,8 +75,6 @@ const (
 	ConfigMapController = "configmap"
 	// wicdRBACResourceName is the name of the resources associated with WICD's RBAC permissions
 	wicdRBACResourceName = "windows-instance-config-daemon"
-	// ProxyCertsConfigMap is the name of the ConfigMap that holds the trusted CA bundle for a cluster-wide proxy
-	ProxyCertsConfigMap = "trusted-ca"
 	// InjectionRequestLabel is used to allow CNO to inject the trusted CA bundle when the global Proxy resource changes
 	InjectionRequestLabel = "config.openshift.io/inject-trusted-cabundle"
 )
@@ -187,7 +185,7 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context,
 				return ctrl.Result{}, err
 			}
 		}
-		if req.NamespacedName.Name == ProxyCertsConfigMap {
+		if req.NamespacedName.Name == certificates.ProxyCertsConfigMap {
 			// Create the trusted CA ConfigMap as it is not present
 			return ctrl.Result{}, r.createProxyCertsCM(ctx)
 		}
@@ -202,8 +200,8 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context,
 		return ctrl.Result{}, r.reconcileNodes(ctx, configMap)
 	case certificates.KubeAPIServerServingCAConfigMapName:
 		return ctrl.Result{}, r.reconcileKubeletClientCA(ctx, configMap)
-	case ProxyCertsConfigMap:
-		return ctrl.Result{}, r.reconcileProxyCertsCM(ctx, configMap)
+	case certificates.ProxyCertsConfigMap:
+		return ctrl.Result{}, r.reconcileProxyCerts(ctx, configMap)
 	default:
 		// Unexpected configmap, log and return no error so we don't requeue
 		r.log.Error(fmt.Errorf("unexpected resource triggered reconcile"), "ConfigMap", req.NamespacedName)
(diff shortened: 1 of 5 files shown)


alinaryan (commented):
Thanks for working on this! Just a couple of comments and one nit: s/cofiguration/configuration/g in the second commit message

aravindhp (changes requested):

cmd/daemon/bootstrap.go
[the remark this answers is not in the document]
saifshaikh48: I wanted to get eyes on it early, I agree though it should be dropped from this PR

pkg/windows/windows.go
```diff
@@ -204,6 +207,11 @@ func GetK8sDir() string {
 	return K8sDir
 }
 
+// GetRemoteDir returns the location of the remote temporary directory created on the Windows VM
+func GetRemoteDir() string {
```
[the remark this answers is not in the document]
saifshaikh48: Vestige from a previous design, will remove

pkg/nodeconfig/nodeconfig.go
```diff
@@ -148,6 +149,17 @@ func (nc *nodeConfig) Configure() error {
 	if err := nc.createBootstrapFiles(); err != nil {
 		return err
 	}
+	if cluster.IsProxyEnabled() {
+		// ensure cert bundle file on instance is populated with up-to-date data
+		trustedCA, err := nc.k8sclientset.CoreV1().ConfigMaps(nc.wmcoNamespace).Get(context.TODO(),
+			certificates.ProxyCertsConfigMap, meta.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("could not get resource %s: %w", certificates.ProxyCertsConfigMap, err)
+		}
```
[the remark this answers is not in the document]
saifshaikh48: EnsureTrustedCABundle is called from elsewhere where we already have the ConfigMap object. I can create a wrapper and move this logic there to make this Configure() func more readable

aravindhp (approved):

pkg/nodeconfig/nodeconfig.go
```diff
@@ -534,6 +540,22 @@ func (nc *nodeConfig) UpdateKubeletClientCA(contents []byte) error {
 	return nil
 }
 
+// ensureTrustedCABundle gets the trsuted CA ConfigMap and ensures the cert bundle on the instance has up-to-date data
+func (nc *nodeConfig) ensureTrustedCABundle() error {
+	trustedCA, err := nc.k8sclientset.CoreV1().ConfigMaps(nc.wmcoNamespace).Get(context.TODO(),
+		certificates.ProxyCertsConfigMap, meta.GetOptions{})
```
sebsoto, line 546: Please use the cached client here. applies to all calls which deal with objects we watch.

controllers/configmap_controller.go
```diff
@@ -512,9 +512,37 @@ func (r *ConfigMapReconciler) reconcileProxyCerts(ctx context.Context, trustedCA
...
+	winInstance, err := r.instanceFromNode(&node)
+	if err != nil {
+		return fmt.Errorf("error creating instance for node %s: %w", node.Name, err)
+	}
+	nc, err := nodeconfig.NewNodeConfig(r.client, r.k8sclientset, r.clusterServiceCIDR, r.watchNamespace,
+		winInstance, r.signer, nil, nil, r.platform)
+	if err != nil {
+		return fmt.Errorf("error creating nodeConfig for instance %s: %w", winInstance.Address, err)
+	}
+	return nc.UpdateTrustedCABundleFile(caData)
```
sebsoto, line 543: The error messages here don't need to each include the node/instance details. that can be wrapped at the call site at the first place in the chain which is dealing with this node object, in this case within`ensureTrustedCABundleInNodes`

This also helps to prevent stuttering, giving a cleaner error message.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -534,6 +540,22 @@ func (nc *nodeConfig) UpdateKubeletClientCA(contents []byte) error {
 	return nil
 }
 
+// ensureTrustedCABundle gets the trsuted CA ConfigMap and ensures the cert bundle on the instance has up-to-date data
+func (nc *nodeConfig) ensureTrustedCABundle() error {
```
sebsoto, line 544: trsuted -> trusted

pkg/nodeconfig/nodeconfig.go
sebsoto: Perhaps a check should be made to ensure that the CM has the key `certificates.CABundleKey`?
Otherwise that could cause issues within `UpdateTrustedCABundleFile`

sebsoto (changes requested):

pkg/nodeconfig/nodeconfig.go
saifshaikh48: > Perhaps a check should be made to ensure that the CM has the key certificates.CABundleKey?
> Otherwise that could cause issues within UpdateTrustedCABundleFile

I didn't add this check because I was thinking that if the key doesn't exist, it could mean a user cleared out their settings/removed all their additional certs. In which case UpdateTrustedCABundleFile(), will reflect this by setting the file contents as empty (and then WICD can react to that)

jrvaldes (approved):
/lgtm

PR #1713, opened 2023-08-02
[hack] Improve vSphere CSI Manifests

alinaryan:
Separates the manifests in the ideal order that they should be
deployed. 
Adds an example deployment spec and custom namespace used
to confirm a Windows pod with a storage mount file can be written to a
desired mount path.

### DIFF
--- a/hack/manifests/csi/vsphere/02-example-namespace.yaml
+++ b/hack/manifests/csi/vsphere/02-example-namespace.yaml
@@ -1,0 +1,9 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: windows-storage-example
+  labels:
+    # Windows storage resources must be deployed in a privileged namespace with a disabled pod security admission label
+    # synchronization
+    security.openshift.io/scc.podSecurityLabelSync: "false"
+    pod-security.kubernetes.io/enforce: privileged
--- a/hack/manifests/csi/vsphere/03-example-sc.yaml
+++ b/hack/manifests/csi/vsphere/03-example-sc.yaml
@@ -1,0 +1,9 @@
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: example-windows-sc
+  namespace: windows-storage-example
+provisioner: csi.vsphere.vmware.com
+parameters:
+  # vSphere Container Storage Plug-in only supports NTFS file system on Windows nodes
+  csi.storage.k8s.io/fstype: "ntfs"
--- a/hack/manifests/csi/vsphere/04-example-pvc.yaml
+++ b/hack/manifests/csi/vsphere/04-example-pvc.yaml
@@ -1,0 +1,13 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: example-windows-pvc
+  namespace: windows-storage-example
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 2Gi
+  storageClassName: example-windows-sc
+  volumeMode: Filesystem
--- a/hack/manifests/csi/vsphere/01-example-driver-daemonset.yaml
+++ b/hack/manifests/csi/vsphere/01-example-driver-daemonset.yaml
@@ -143,28 +143,3 @@ spec:
           operator: Exists
         - effect: NoSchedule
           operator: Exists
----
-apiVersion: v1
-kind: PersistentVolumeClaim
-metadata:
-  name: example-windows-pvc
-  namespace: openshift-cluster-csi-drivers
-spec:
-  accessModes:
-    - ReadWriteOnce
-  resources:
-    requests:
-      storage: 5Gi
-  storageClassName: example-windows-sc
-  volumeMode: Filesystem
----
-kind: StorageClass
-apiVersion: storage.k8s.io/v1
-metadata:
-  name: example-windows-sc
-  namespace: openshift-cluster-csi-drivers
-provisioner: csi.vsphere.vmware.com
-allowVolumeExpansion: true  # Optional: only applicable to vSphere 7.0U1 and above
-parameters:
-  #datastoreurl: "ds:///vmfs/volumes/vsan:52cdfa80721ff516-ea1e993113acfc77/"  # Optional Parameter
-  csi.storage.k8s.io/fstype: "ntfs"
--- a/hack/manifests/csi/vsphere/05-example-deployment.yaml
+++ b/hack/manifests/csi/vsphere/05-example-deployment.yaml
@@ -1,0 +1,43 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: example-windows-pod
+  name: example-windows-pod
+  # must be in the same namespace as the PVC
+  namespace: windows-storage-example
+spec:
+  selector:
+    matchLabels:
+      app: example-windows-pod
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: example-windows-pod
+      name: example-windows-pod
+    spec:
+      os:
+        name: windows
+      tolerations:
+        - key: "os"
+          value: "Windows"
+          Effect: "NoSchedule"
+      containers:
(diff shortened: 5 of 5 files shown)


alinaryan:
updated-vsphere-csi-manifests

alinaryan:
pod-confirmation

hack/manifests/csi/vsphere/05-example-pod.yaml
```diff
@@ -0,0 +1,31 @@
...
+  os:
+    name: windows
+  tolerations:
+  - key: "os"
+    value: "Windows"
+    Effect: "NoSchedule"
+  nodeSelector:
+    kubernetes.io/os: windows
+  containers:
+    - name: test-container
```
jrvaldes, line 17: Consider naming the container according to the running example. e.g. example-windows-container

hack/manifests/csi/vsphere/02-example-namespace.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: windows-webserver
```
jrvaldes, line 4: Not sure about `windows-webserver`  as the correct name for the namespace. Looking at the pod, it does not run a webserver. Consider something like: `example-windows`

hack/manifests/csi/vsphere/02-example-namespace.yaml
```diff
@@ -0,0 +1,7 @@
...
+  labels: 
+    security.openshift.io/scc.podSecurityLabelSync: "false"
+    pod-security.kubernetes.io/enforce: privileged
```
jrvaldes, line 7: Are these labels required?

hack/manifests/csi/vsphere/03-example-sc.yaml
```diff
@@ -0,0 +1,11 @@
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: example-windows-sc
+  namespace: windows-webserver
+provisioner: csi.vsphere.vmware.com
+allowVolumeExpansion: true  # Optional: only applicable to vSphere 7.0U1 and above
```
jrvaldes, line 7: Consider removing this optional field, as it is irrelevant to the example.

hack/manifests/csi/vsphere/03-example-sc.yaml
```diff
@@ -0,0 +1,11 @@
...
+parameters:
+  csi.storage.k8s.io/fstype: "ntfs" # vSphere Container Storage Plug-in only supports NTFS file system on Windows nodes
```
jrvaldes, line 9:
```suggestion
  # vSphere Container Storage Plug-in only supports NTFS file system on Windows nodes
  csi.storage.k8s.io/fstype: "ntfs" 
```

hack/manifests/csi/vsphere/04-example-pvc.yaml
```diff
@@ -0,0 +1,13 @@
...
+kind: PersistentVolumeClaim
+metadata:
+  name: example-windows-pvc
+  namespace: windows-webserver
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 5Gi
```
jrvaldes, line 11: I recommend going with something smaller as it may lower the odds of the datastore running out of storage capacity.

jrvaldes (commented):
@alinaryan  Thanks for working on this. Mostly LGTM. I left a few comments. PTAL

> Separates the manifests in the ideal order that they should be deployed.

WRT to splitting the resource into separate files; what are the advantages versus having one file with the resource definition in the correct order?

hack/manifests/csi/vsphere/05-example-pod.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: v1
+kind: Pod
```
sebsoto, line 2: This should really be a deployment and not a pod

hack/manifests/csi/vsphere/02-example-namespace.yaml
sebsoto: `windows-storage-example` ?

hack/manifests/csi/vsphere/05-example-pod.yaml
sebsoto: https://kubernetes.io/docs/concepts/workloads/pods/#workload-resources-for-managing-pods

sebsoto (changes requested):

hack/manifests/csi/vsphere/02-example-namespace.yaml
alinaryan: Yes, they are required so the pod has the appropriate privileges/sccs

hack/manifests/csi/vsphere/04-example-pvc.yaml
alinaryan: Changed to 2Gi

hack/manifests/csi/vsphere/05-example-pod.yaml
alinaryan: Changed to deployment

alinaryan:
@jrvaldes 
>WRT to splitting the resource into separate files; what are the advantages versus having one file with the resource definition in the correct order?

The advantage is the user can choose which file(s) they want or don't want without having to download the entire manifests and remove what doesn't apply to them. For example, someone performing an upgrade from in-tree to CSI will only need the driver-daemonset manifest since their in-tree resources are expected to migrate automatically. 

It is also useful for the blog since it helps paint a picture of the order the steps should happen.

hack/manifests/csi/vsphere/03-example-sc.yaml
```diff
@@ -0,0 +1,11 @@
...
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: example-windows-sc
+  namespace: windows-storage-example
+provisioner: csi.vsphere.vmware.com
+parameters:
+  # vSphere Container Storage Plug-in only supports NTFS file system on Windows nodes
+  csi.storage.k8s.io/fstype: "ntfs"
+
+
```
sebsoto, line 11: extra newline

hack/manifests/csi/vsphere/05-example-pod.yaml
```diff
@@ -0,0 +1,41 @@
...
+  selector:
+    matchLabels:
+      app: example-windows-pod
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: example-windows-pod
+      name: example-windows-pod
+    spec:
```
jrvaldes, line 19: @alinaryan, The `os` field got skipped in [this](https://github.com/openshift/windows-machine-config-operator/compare/bda14fef971b55839baf9f8553ca5b2837c4cb0e..1cb901ef1ee536994b7f65edc9740a65a2ced079#diff-e11646292771aad6465ca3da31522c869be8733b1099542ef50ca43bb4b8cec3L8-L9) push.  I would suggest that you bring it back.

jrvaldes (changes requested):

hack/manifests/csi/vsphere/02-example-namespace.yaml
jrvaldes: Consider stating that inline with a comment.

saifshaikh48:
LGTM but @jrvaldes is better qualified to give the stamp as he has been trying out deploying these resources

jrvaldes (commented):
LGTM, just one minor [suggestion](https://github.com/openshift/windows-machine-config-operator/pull/1713/files#r1285790067).

hack/manifests/csi/vsphere/05-example-pod.yaml
mtnbikenc: Since this was changed to a deployment the filename and other relevant references to `pod` should be updated.

alinaryan:
Screenshot from 2023-08-07 12-37-19
Screenshot from 2023-08-07 12-41-45

PR #1719, opened 2023-08-08
[aws] Avoid Windows Server 2019 `Containers` image

saifshaikh48:
This PR moves away from using Windows Server 2019 that have Docker pre-installed in CI as these images have been pulled from AWS. We had been using them for test coverage as we required pre-installed Docker runtime in 4.10 and below, before moving to containerd (WMCO-managed).
Moved to `Core` instead of `Full` as we use that in for 2022 images and it is more lightweight/faster startup time.

### DIFF
--- a/hack/machineset.sh
+++ b/hack/machineset.sh
@@ -80,7 +80,7 @@ get_aws_ms() {
 
   local filter="Windows_Server-2022-English-Core-Base-????.??.??"
   if [ "$winver" == "2019" ]; then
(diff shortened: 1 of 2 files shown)


saifshaikh48:
/test aws-e2e-upgrade
^ uses 2019

saifshaikh48:
/test aws-e2e-operator
sanity check

jrvaldes (commented):
@saifshaikh48  thanks for opening this so quickly. 

LGTM, awaiting the test to pass

sebsoto (approved):

PR #1743, opened 2023-08-23
Use ParseCompatibleVersion when parsing ignition

sebsoto:
Uses this function instead of Parse(), so that this can be backported to release-4.13. This will stop wmco 8.y.z from crashing upon cluster upgrade to 4.14, due to it being unable to parse the 3.4 ignition spec present on 4.14.

### DIFF
--- a/pkg/ignition/ignition.go
+++ b/pkg/ignition/ignition.go
@@ -52,14 +52,14 @@ func New(c client.Client) (*Ignition, error) {
 		return nil, err
 	}
 
-	configuration, report, err := ignCfg.Parse(renderedWorker.Spec.Config.Raw)
+	configuration, report, err := ignCfg.ParseCompatibleVersion(renderedWorker.Spec.Config.Raw)
 	if err != nil || report.IsFatal() {
 		return nil, fmt.Errorf("failed to parse MachineConfig ignition: %v\nReport: %v", err, report)
 	}
 	ign := &Ignition{
 		config: configuration,
 	}
-	log.V(1).Info("parsed", "machineconfig", renderedWorker.GetName(), "ignition version",
+	log.V(1).Info("parsed", "machineconfig", renderedWorker.GetName(), "using ignition version",
 		configuration.Ignition.Version)
 
 	ccList := mcfg.ControllerConfigList{}


saifshaikh48 (commented):
/lgtm

pkg/ignition/ignition.go
```diff
@@ -52,7 +52,7 @@ func New(c client.Client) (*Ignition, error) {
 		return nil, err
 	}
 
-	configuration, report, err := ignCfg.Parse(renderedWorker.Spec.Config.Raw)
+	configuration, report, err := ignCfg.ParseCompatibleVersion(renderedWorker.Spec.Config.Raw)
```
jrvaldes, line 55: This will parse ignition versions less or equal to 3.4 which makes the log message in [line#62](https://github.com/openshift/windows-machine-config-operator/pull/1743/files#diff-5f8bc59c6ca4160f53648587fbed538adbd305ad0dfb837bd3124f2ee84fde9bR62) inaccurate as it will always log the resulting version not the source one. Consider adding a log showing the actual ignition version in the `renderedWorker`


```go
	log.V(1).Info("parsed", "machineconfig", renderedWorker.GetName(), "ignition version",
		configuration.Ignition.Version)
```

sebsoto: @jrvaldes I'm not sure how you get the actual version from the renderedWorker like you are suggestion, can you help me out here?

saifshaikh48:
LGTM, will let @jrvaldes give the stamp

pkg/ignition/ignition.go
jrvaldes: @sebsoto adding a log message with the ignition version present in the renderedWorker adds another unmarshaling to the execution flow. I recommend just reword the log message in [line#62](https://github.com/openshift/windows-machine-config-operator/pull/1743/files#diff-5f8bc59c6ca4160f53648587fbed538adbd305ad0dfb837bd3124f2ee84fde9bR62) for example:

```go
	log.V(1).Info("processing", "machineconfig", renderedWorker.GetName(), "ignition version",
		configuration.Ignition.Version)
```

jrvaldes:
backported as part of 
- https://github.com/openshift/windows-machine-config-operator/pull/1744

PR #1728, opened 2023-08-14
Bump worker ignition specs to 3.4

jrvaldes:
This change updates the ignition specs given the previous version (3.2) does not support the parsing of the 3.4 spec

No backports required given "[4.13 MCO will always render configs with 3.2.0](https://redhat-internal.slack.com/archives/CM4ERHBJS/p1692039942242409?thread_ts=1692035137.053939&cid=CM4ERHBJS)"

PR #1744, opened 2023-08-23
OCPBUGS-18091: [release-4.13] Update ignition parsing to prevent crash on cluster upgrade

sebsoto:
Contains various backports to ensure that WMCO 8 is compatible with the ignition specs present in both OCP 4.13 and 4.14.


saifshaikh48 (commented):
/lgtm

Thanks for the quick fix

pkg/ignition/ignition.go
```diff
@@ -6,8 +6,8 @@ import (
 	"sort"
 	"strings"
 
-	ignCfg "github.com/coreos/ignition/v2/config/v3_2"
-	ignCfgTypes "github.com/coreos/ignition/v2/config/v3_2/types"
+	ignCfg "github.com/coreos/ignition/v2/config/v3_4"
+	ignCfgTypes "github.com/coreos/ignition/v2/config/v3_4/types"
```
jrvaldes, line 10: I don't think this is needed. [xref](https://github.com/openshift/windows-machine-config-operator/pull/1728#issue-1850346835)

cc @jkyros   Thoughts?

jkyros: Assuming upgrade order is such that the WMCO gets upgraded _after_ the MCO, there is potential at least for the 4.13 WMCO to receive a `3.4.0` version from the 4.14 MCO during an upgrade, so having the 4.13 MCO understand `3.4.0` seems like a good idea?

PR #1747, opened 2023-08-23
[servicescm] Use generics for Services and Files

saifshaikh48:
This PR refactors the isPresentAndCorrect function using generics to reduce code duplication and make the code more object-oriented.

### DIFF
--- a/pkg/servicescm/servicescm.go
+++ b/pkg/servicescm/servicescm.go
@@ -245,7 +245,7 @@ func (cmData *Data) ValidateExpectedContent(expected *Data) error {
 		return fmt.Errorf("unexpected number of services")
 	}
 	for _, expectedSvc := range expected.Services {
-		if !expectedSvc.isPresentAndCorrect(cmData.Services) {
+		if !contains(cmData.Services, expectedSvc) {
 			return fmt.Errorf("required service %s is not present with expected configuration", expectedSvc.Name)
 		}
 	}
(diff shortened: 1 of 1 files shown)


pkg/servicescm/servicescm.go
```diff
@@ -272,21 +277,10 @@ func (cmData *Data) ValidateExpectedContent(expected *Data) error {
...
-	return false
-}
-
-// isPresentAndCorrect checks if the required file exists as expected within the given files slice
-// TODO: When we move to go1.18, consolodate this helper with the above Service.isPresentAndCorrect using generics
-func (f FileInfo) isPresentAndCorrect(files []FileInfo) bool {
-	for _, file := range files {
-		if reflect.DeepEqual(f, file) {
+// isPresentAndCorrect checks if the required item exists as expected within the given slice
+func isPresentAndCorrect[L listField](target L, slice []L) bool {
```
sebsoto, line 281: The way this is being used it doesn't seem like `listField` needs to be defined.
You can just use the `any` type constraint

additionally now that this is a  generic function, it can have a more generic name, as it is just checking if the slice contains the value.
```
func contains[S ~[]E, E any](slice S, target E) bool {
```

can you help me understand the difference in the type declarion syntax you are doing, and the one I see in the go slices package: https://cs.opensource.google/go/x/exp/+/d852ddb8:slices/slices.go;l=116 ?
Is there any functional difference?

saifshaikh48: > can you help me understand the difference in the type declarion syntax you are doing, and the one I see in the go slices package

The 2 differences I see:
-  the slices pkg's syntax explicitly ensures that the `target` is an object that implements the [comparable](https://pkg.go.dev/builtin#comparable) interface
- the slices pkg's syntax allows for slices of structs that are `E` types under the hood through the `~` operator (if we had an object of type `customTypeE` where `struct customTypeE E`, you could safely pass the object into the Contains func)

PR #1750, opened 2023-08-28
[WIP] WINC-1098: Block upgrades if it will break storage workloads

sebsoto:
Blocks the upgrade of Nodes if it will disrupt workloads which have an attached storage volume. Workloads will be disrupted if WMCO is upgrading a node to a kubelet version which is using CSI, from a version which uses in tree volumes. WMCO will unblock the upgrade when it sees that there is an installed CSI driver on the node, through the CSINode cluster object.

In the case of a false positive, users can apply the label windowsmachineconfig.openshift.io/force-upgrade if they wish to override the upgrade block, and have WMCO upgrade the Node. This allows for WMCO to be a bit more cautious when determining if the upgrade will break anything.

### DIFF
--- a/controllers/controllers.go
+++ b/controllers/controllers.go
@@ -4,10 +4,13 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"strconv"
+	"strings"
 
 	"github.com/go-logr/logr"
 	config "github.com/openshift/api/config/v1"
 	"golang.org/x/crypto/ssh"
+	"golang.org/x/mod/semver"
 	core "k8s.io/api/core/v1"
 	kubeTypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
@@ -25,6 +28,16 @@ import (
 	"github.com/openshift/windows-machine-config-operator/pkg/nodeconfig"
 	"github.com/openshift/windows-machine-config-operator/pkg/secrets"
 	"github.com/openshift/windows-machine-config-operator/version"
+)
+
+var (
+	// CSIAnnotation indicates that this Node was configured by a version of WMCO with CSI storage. When upgrading this
+	// Node CSI storage migration does not have to be accounted for.
+	CSIAnnotation = "windowsmachineconfig.openshift.io/configured-with-csi"
+	// UpgradeBlockedAnnotation indicates that the Node's upgrade has been blocked by WMCO
+	UpgradeBlockedAnnotation = "windowsmachineconfig.openshift.io/upgrade-blocked"
+	// AllowUpgradeLabel should be applied on a Node by a user, if they wish to unblock an upgrade
+	AllowUpgradeLabel = "windowsmachineconfig.openshift.io/allow-upgrade"
 )
 
 // instanceReconciler contains everything needed to perform actions on a Windows instance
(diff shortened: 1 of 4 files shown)


controllers/controllers.go
```diff
@@ -73,6 +88,18 @@ func (r *instanceReconciler) ensureInstanceIsUpToDate(instanceInfo *instance.Inf
 	// Check if the instance was configured by a previous version of WMCO and must be deconfigured before being
 	// configured again.
 	if instanceInfo.UpgradeRequired() {
+		blocked, err := r.upgradeBlocked(instanceInfo.Node, csiEnabled)
```
aravindhp, line 91: Does `csiEnabled` imply `upgradingToCSI`? I find that a bit confusing.

controllers/controllers.go
```diff
@@ -84,6 +111,46 @@ func (r *instanceReconciler) ensureInstanceIsUpToDate(instanceInfo *instance.Inf
 	return nc.Configure()
 }
 
+var OverrideLabel = "windowsmachineconfig.openshift.io/force-upgrade"
+
+// upgradeBlocked returns whether the upgrade should be blocked if an upgrade will break storage functionality
+func (r *instanceReconciler) upgradeBlocked(node *core.Node, upgradingToCSI bool) (bool, error) {
```
aravindhp, line 117: Are we sure that the `node` is up to date and is not stale?

aravindhp (commented):
>In the case of a false positive, users can apply the label
windowsmachineconfig.openshift.io/force-upgrade if they wish to
override the upgrade block, and have WMCO upgrade the Node. This allows
for WMCO to be a bit more cautious when determining if the upgrade will
break anything.

Shouldn't a customer rather open a bug about this instead of trying to override it?

controllers/controllers.go
```diff
@@ -63,7 +71,14 @@ func (r *instanceReconciler) ensureInstanceIsUpToDate(instanceInfo *instance.Inf
...
 	}
-
+	// Check if the running version of WMCO is one where in tree storage is deprecated in favor of CSI.
+	csiEnabled, err := r.usesCSI()
+	if err != nil {
+		return err
+	}
+	if csiEnabled {
+		annotationsToApply[csiAnnotation] = "true"
+	}
```
saifshaikh48, line 81: Since `usesCSI()` is only needs the `instanceReconciler` platform, could we parameterize the platform and make this a regular helper function? This way we call it a single time from `ensureInstancesAreUpToDate()` instead of once per node (and pass in the csiAnnotation to this func as part of annotationsToApply)

controllers/controllers.go
```diff
@@ -84,6 +111,46 @@ func (r *instanceReconciler) ensureInstanceIsUpToDate(instanceInfo *instance.Inf
 	return nc.Configure()
 }
 
+var OverrideLabel = "windowsmachineconfig.openshift.io/force-upgrade"
```
saifshaikh48, line 114: nit: move to top of file with `csiAnnotation`

jkyros, line 114: are there other reasons why we might force an upgrade? since this is generic (i.e. not storage-specific), I'm wondering if what we're doing here is "adding a general upgrade force label" that _right now_ only happens to apply to the CSI storage block?

controllers/controllers.go
```diff
@@ -84,6 +111,46 @@ func (r *instanceReconciler) ensureInstanceIsUpToDate(instanceInfo *instance.Inf
...
+
+// upgradeBlocked returns whether the upgrade should be blocked if an upgrade will break storage functionality
+func (r *instanceReconciler) upgradeBlocked(node *core.Node, upgradingToCSI bool) (bool, error) {
+	if !upgradingToCSI {
+		return false, nil
+	}
+	if _, present := node.GetLabels()[OverrideLabel]; present {
```
jkyros, line 121: the "other" MCO has a "forcefile" that folks use to get into all sorts of trouble (and then bring us their broken clusters). 

It might not be terrible to make this "verbosely apparent" that the customer has overridden this ( so when they come to us with a bug or a gather we know that they miiiight have forced it and broken it :smile:  )

sebsoto:
Closing in favor of https://github.com/openshift/windows-machine-config-operator/pull/1794

This should be merged directly to 4.14 and backported to 4.13. There is no CSI migration in 4.15

PR #1765, opened 2023-08-31
OCPBUGS-18294: WMCO is not reacting to new environment variables

mansikulkarni96:
This PR fixes a bug where WMCO is not reacting to changes in the env vars as expected by deleting and re-creating the
invalid services CM object. This was caused by a flawed comparison between the expected and existing CM object.
Fixes: #OCPBUGS-18294

### DIFF
--- a/pkg/servicescm/servicescm.go
+++ b/pkg/servicescm/servicescm.go
@@ -282,12 +282,9 @@ func (cmData *Data) ValidateExpectedContent(expected *Data) error {
 	if len(cmData.EnvironmentVars) != len(expected.EnvironmentVars) {
 		return fmt.Errorf("unexpected number of environment variable")
 	}
-	for _, expectedEnvVar := range expected.EnvironmentVars {
-		if cmData.EnvironmentVars[expectedEnvVar] != expected.EnvironmentVars[expectedEnvVar] {
-			return fmt.Errorf("required environment variable %s is not present as expected."+
-				"expected: %s, actual: %s", expectedEnvVar, expected.EnvironmentVars[expectedEnvVar],
-				cmData.EnvironmentVars[expectedEnvVar])
-		}
+	if !reflect.DeepEqual(cmData.EnvironmentVars, expected.EnvironmentVars) {
+		return fmt.Errorf("required environment variables are not present as expected "+
+			"expected: %v, actual: %v", cmData.EnvironmentVars, expected.EnvironmentVars)
 	}
 	if len(cmData.WatchedEnvironmentVars) != len(expected.WatchedEnvironmentVars) {
 		return fmt.Errorf("unexpected number of watched environment variable")
--- a/pkg/servicescm/servicescm_test.go
+++ b/pkg/servicescm/servicescm_test.go
@@ -198,6 +198,30 @@ func TestGenerate(t *testing.T) {
 			assert.NoError(t, parsed.ValidateExpectedContent(data))
 		})
 	}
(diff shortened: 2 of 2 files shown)


pkg/servicescm/servicescm.go
```diff
@@ -282,12 +282,9 @@ func (cmData *Data) ValidateExpectedContent(expected *Data) error {
 	if len(cmData.EnvironmentVars) != len(expected.EnvironmentVars) {
 		return fmt.Errorf("unexpected number of environment variable")
 	}
-	for _, expectedEnvVar := range expected.EnvironmentVars {
-		if cmData.EnvironmentVars[expectedEnvVar] != expected.EnvironmentVars[expectedEnvVar] {
-			return fmt.Errorf("required environment variable %s is not present as expected."+
-				"expected: %s, actual: %s", expectedEnvVar, expected.EnvironmentVars[expectedEnvVar],
-				cmData.EnvironmentVars[expectedEnvVar])
-		}
+	if !reflect.DeepEqual(cmData.EnvironmentVars, expected.EnvironmentVars) {
```
jrvaldes, line 285: Consider adding a unit test to cover the scenario where the loop failed to match the expected vars.

jrvaldes (commented):
@mansikulkarni96  thanks for working on this. LGTM.

pkg/servicescm/servicescm.go
mansikulkarni96: thanks @jrvaldes I was working on it, updated now.

jrvaldes (approved):
/lgtm

pkg/servicescm/servicescm_test.go
```diff
@@ -156,46 +156,61 @@ func TestGenerate(t *testing.T) {
...
-
 			// Ensure that the ConfigMap we generate passes our own validation functions
 			parsed, err := Parse(configMap.Data)
 			require.NoError(t, err)
-			assert.NoError(t, parsed.ValidateExpectedContent(data))
+			if test.expectedErr {
+				assert.Error(t, parsed.ValidateExpectedContent(data0))
+				return
+			}
+			assert.NoError(t, parsed.ValidateExpectedContent(data0))
```
sebsoto, line 213: I'm having a hard time following these test steps, and what specifically you are trying to test with these changes. It seems like you're adding a somewhat different test workflow in order to test a specific case, and trying to make it work for all the tests.


It might be better to test case as a subtest separate from the rest of the test cases.
```
 	for _, test := range testCases {
		t.Run(test.name, func(t *testing.T) { ... })
	}
	t.Run("new test being added in this PR) {
	}
 ```
 
Please also make the variable names more descriptive, envVars0/1 and data0/1 doesn't give any information about the difference between the two.

**Review by sebsoto — CHANGES_REQUESTED** [review]
Thanks for working on this. Functionality changes LGTM, please address the unit test comments.

File: pkg/servicescm/servicescm_test.go [hunk 3]
```diff
@@ -191,13 +191,37 @@ func TestGenerate(t *testing.T) {
 			require.NoError(t, err)
 			configMap, err := Generate(Name, "testNamespace", data)
 			require.NoError(t, err)
-
+
```
aravindhp, line 194: I guess this was accidental. Please revert.

aravindhp (approved):

saifshaikh48 (commented):
/lgtm

new unit test looks good

PR #1756, opened 2023-08-29
[release-4.11] submodule update 08-29

alinaryan:
- [[submodule][ovn-kubernetes] Update to a5c0becfc](https://github.com/openshift/windows-machine-config-operator/commit/492dce4f167d8ca2c8e1ee2aec9a053402edf60e) 
- [[submodule][promu] Update to d5383c5](https://github.com/openshift/windows-machine-config-operator/commit/ee70dec64f1b6d3a625878b4723fd32874f82ada) 
- [[submodule][kubelet] Update to 7aa7ea9ff1f](https://github.com/openshift/windows-machine-config-operator/commit/1b9e255cb91cab1e08c4c919b709e6b9ce06805f) 
- [[build] Update kubelet version to v1.24.16+7aa7ea9](https://github.com/openshift/windows-machine-config-operator/commit/9ba340b30bbf382f013a39a47cb37397c947d9df) 
- [[submodule][containerd] Update to 6dbd1da32](https://github.com/openshift/windows-machine-config-operator/commit/bd9d17dbd40d10c62874b2104dd355a696aca0c5) 
- [[build] Update containerd version to v1.6.23-6-g6dbd1da32](https://github.com/openshift/windows-machine-config-operator/commit/f3a74c336144b3bad756a413b3ff855d2f9b2cdb) 
- [[submodule][hcsshim] Update to 9cd2a016](https://github.com/openshift/windows-machine-config-operator/commit/006c8f0b61d2c15b3604a4f5e2e630699e70b8c4)

PR #1766, opened 2023-09-01
Test change

aravindhp:


aravindhp:
/close

Image build succeeded. There definitely seems to be an issue with building the kubelet in the image in https://github.com/openshift/windows-machine-config-operator/pull/1756

PR #1772, opened 2023-09-07
OCPBUGS-15988: Restore routes on AWS Windows 2022 nodes

aravindhp:
Run the EC2Launch v2 service as a hack to restore routes after the HNS networks have been deleted during deconfiguration. This allows the same VM to be configured as a node again.

Fixes OCPBUGS-15988

### DIFF
--- a/pkg/nodeconfig/nodeconfig.go
+++ b/pkg/nodeconfig/nodeconfig.go
@@ -524,6 +524,15 @@ func (nc *nodeConfig) Deconfigure() error {
 	if err := nc.Windows.Deconfigure(nc.wmcoNamespace, wicdKC); err != nil {
 		return fmt.Errorf("error deconfiguring instance: %w", err)
 	}
+
+	// Windows Server 2022 VMs on AWS have a non-persistent route to the metadata endpoint. This is lost when the HNS
+	// networks are deleted. Explicitly restore them to allow the same VM to be configured as a node again.
+	if nc.platformType == configv1.AWSPlatformType {
+		if err := nc.Windows.RestoreAWSRoutes(); err != nil {
+			return err
+		}
+	}
+
 	// Wait for reboot annotation removal. This prevents deleting the node until the node no longer needs reboot.
 	if err := metadata.WaitForRebootAnnotationRemoval(context.TODO(), nc.client, nc.node.Name); err != nil {
 		return err
--- a/pkg/windows/windows.go
+++ b/pkg/windows/windows.go
@@ -767,19 +767,23 @@ func (vm *windows) stopService(svc *service) error {
 	}
 
 	// Wait until the service has stopped
-	err = wait.PollImmediate(retry.Interval, retry.Timeout, func() (bool, error) {
-		serviceRunning, err := vm.isRunning(svc.name)
-		if err != nil {
-			vm.log.V(1).Error(err, "unable to check if Windows service is running", "service", svc.name)
-			return false, nil
-		}
-		return !serviceRunning, nil
-	})
-	if err != nil {
+	if err = vm.waitStopped(svc.name); err != nil {
 		return fmt.Errorf("error waiting for the %s service to stop: %w", svc.name, err)
 	}
 
 	return nil
+}
+
+// waitStopped returns once the service has stopped within the retry.Timeout interval otherwise returns an error
+func (vm *windows) waitStopped(serviceName string) error {
+	return wait.PollImmediate(retry.Interval, retry.Timeout, func() (bool, error) {
+		serviceRunning, err := vm.isRunning(serviceName)
+		if err != nil {
+			vm.log.V(1).Error(err, "unable to check if Windows service is running", "service", serviceName)
+			return false, nil
+		}
+		return !serviceRunning, nil
+	})
 }
 
 // deleteService deletes the specified Windows service
--- a/pkg/windows/windows.go
+++ b/pkg/windows/windows.go
@@ -633,8 +633,7 @@ func (vm *windows) transferFiles() error {
 	return nil
 }
 
-// ensureServiceIsRunning ensures a Windows service is running on the VM, creating and starting it if not already so.
-// The service's description will be set as part of this.
+// ensureServiceIsRunning ensures a Windows service is running on the VM, creating and starting it if not already so
 func (vm *windows) ensureServiceIsRunning(svc *service) error {
 	serviceExists, err := vm.serviceExists(svc.name)
 	if err != nil {
@@ -646,19 +645,14 @@ func (vm *windows) ensureServiceIsRunning(svc *service) error {
 			return fmt.Errorf("error creating %s Windows service: %w", svc.name, err)
 		}
 	}
-	if err := vm.setServiceDescription(svc.name); err != nil {
-		return fmt.Errorf("error setting description of the %s Windows service: %w", svc.name, err)
-	}
-	if err := vm.setRecoveryActions(svc); err != nil {
-		return fmt.Errorf("error setting recovery actions for the %s Windows service: %w", svc.name, err)
-	}
 	if err := vm.startService(svc); err != nil {
 		return fmt.Errorf("error starting %s Windows service: %w", svc.name, err)
 	}
 	return nil
 }
 
-// createService creates the service on the Windows VM
+// createService creates the service on the Windows VM. The service's description and recovery action will be set after
+// the service is created.
 func (vm *windows) createService(svc *service) error {
 	if svc == nil {
 		return fmt.Errorf("service object should not be nil")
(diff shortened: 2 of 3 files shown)


pkg/windows/windows.go
```diff
@@ -507,6 +513,33 @@ func (vm *windows) ConfigureWICD(watchNamespace, wicdKubeconfigContents string)
...
+func (vm *windows) RestoreAWSRoutes() error {
+	serviceExists, err := vm.serviceExists(ec2LaunchV2ServiceName)
+	if err != nil {
+		return fmt.Errorf("error checking if %s service exists: %w", ec2LaunchV2ServiceName, err)
+	}
+	// We don't want ensureServiceIsRunning to create the service if it does not exist, so we return without an error.
+	// Returning an error does not make sense as we could have an AMI configured without th EC2Launch service present
+	// and the route created using some other customer specific method which is unknown to us.
+	if !serviceExists {
+		vm.log.Info("EC2Launch v2 service not present")
```
jrvaldes, line 525: Consider adding the name of the service used in the query i.e. `ec2LaunchV2ServiceName`

pkg/windows/windows.go
```diff
@@ -507,6 +513,33 @@ func (vm *windows) ConfigureWICD(watchNamespace, wicdKubeconfigContents string)
...
+		return nil
+	}
+	ec2Launch, err := newService(ec2LaunchV2Path, ec2LaunchV2ServiceName, "", nil, nil, 0)
+	if err != nil {
+		return err
+	}
+	if err = vm.ensureServiceIsRunning(ec2Launch); err != nil {
```
jrvaldes, line 532: There may be a case of an AMI configured with the EC2Launch service present in a different location, not the default one (`ec2LaunchV2Path`), if so `vm.ensureServiceIsRunning()` will fail. I feel this case should be treated like the one above where the service is not present.

pkg/windows/windows.go
```diff
@@ -507,6 +513,33 @@ func (vm *windows) ConfigureWICD(watchNamespace, wicdKubeconfigContents string)
 	return nil
 }
 
+func (vm *windows) RestoreAWSRoutes() error {
+	serviceExists, err := vm.serviceExists(ec2LaunchV2ServiceName)
+	if err != nil {
+		return fmt.Errorf("error checking if %s service exists: %w", ec2LaunchV2ServiceName, err)
+	}
+	// We don't want ensureServiceIsRunning to create the service if it does not exist, so we return without an error.
+	// Returning an error does not make sense as we could have an AMI configured without th EC2Launch service present
```
jrvaldes, line 522:
```suggestion
	// Returning an error does not make sense as we could have an AMI configured without the EC2Launch service present
```

jrvaldes (approved):
@aravindhp  thanks for working on this.  LGTM, added minor suggestions.

pkg/windows/windows.go
jrvaldes: Makes sense.

test/e2e/delete_test.go
```diff
@@ -83,6 +84,12 @@ func (tc *testContext) testBYOHRemoval(t *testing.T) {
 			envVarsRemoved, err := tc.checkEnvVarsRemoved(addr)
 			require.NoError(t, err, "error determining if ENV vars are removed")
 			assert.True(t, envVarsRemoved, "ENV vars not removed")
+
+			if tc.CloudProvider.GetType() == config.AWSPlatformType {
+				routeRestored, err := tc.checkAWSMetadataEndpointRouteIsRestored(addr)
+				require.NoError(t, err, "error determining if metadata endpoint route is restored")
+				assert.True(t, routeRestored, "metadata endpoint route is not restored")
+			}
```
sebsoto, line 92: Consider creating a new subtest for this, which is skipped when tc.CloudProvider.GetType() != config.AWSPlatformType.

Helps with organization and prevents the require from blocking any extra test checks within the BYOH removal test, which may be added after this.

sebsoto (changes requested):

pkg/nodeconfig/nodeconfig.go
```diff
@@ -524,6 +524,15 @@ func (nc *nodeConfig) Deconfigure() error {
 	if err := nc.Windows.Deconfigure(nc.wmcoNamespace, wicdKC); err != nil {
 		return fmt.Errorf("error deconfiguring instance: %w", err)
 	}
+
+	// Windows Server 2022 VMs on AWS has a non-persistent route to the metadata endpoint. This is lost when the HNS
```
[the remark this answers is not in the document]
aravindhp: Have implies plural and there is only a single route present for the metadata endpoint.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -524,6 +524,15 @@ func (nc *nodeConfig) Deconfigure() error {
...
+	// networks are deleted. Explicitly restore them to allow the same VM to be configured as a node again.
+	if nc.platformType == configv1.AWSPlatformType {
```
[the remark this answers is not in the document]
aravindhp: We cannot talk about potential WIP solutions from an external source.

pkg/windows/windows.go
```diff
@@ -133,6 +133,10 @@ const (
 	wicdKubeconfigPath = K8sDir + "\\wicd-kubeconfig"
 	// TrustedCABundlePath is the location of the trusted CA bundle file
 	TrustedCABundlePath = remoteDir + "\\ca-bundle.crt"
+	// EC2LaunchV2Path is the location of the EC2Launch v2 binary
+	ec2LaunchV2Path = "C:\\Program Files\\Amazon\\EC2Launch\\service\\EC2LaunchService.exe"
+	// EC2LaunchV2ServiceName is the name of the EC2Launch v2 service
+	ec2LaunchV2ServiceName = "\"Amazon EC2Launch\""
```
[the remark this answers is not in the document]
aravindhp: Fixed.

test/e2e/delete_test.go
aravindhp: Done.

aravindhp (commented):
@sebsoto I have addressed your comments. PTAL.

test/e2e/delete_test.go
```diff
@@ -83,6 +84,8 @@ func (tc *testContext) testBYOHRemoval(t *testing.T) {
 			envVarsRemoved, err := tc.checkEnvVarsRemoved(addr)
 			require.NoError(t, err, "error determining if ENV vars are removed")
 			assert.True(t, envVarsRemoved, "ENV vars not removed")
+
+			tc.checkAWSMetadataEndpointRouteIsRestored(t, addr)
```
sebsoto, line 88: you should be able to do:
```
t.Run("aws metadata endpoint", func(t *testing.T) { tc.checkAWSMetadataEndpointRouteIsRestored(t,addr)}
```

pkg/nodeconfig/nodeconfig.go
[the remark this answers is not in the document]
sebsoto: Isn't the subject  `VMs`?
`VMs on AWS have a`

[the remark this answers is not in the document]
aravindhp: Maybe 😄

sebsoto (approved):

jrvaldes:
/
/lgtm

pkg/windows/windows.go
```diff
@@ -507,6 +509,35 @@ func (vm *windows) ConfigureWICD(watchNamespace, wicdKubeconfigContents string)
...
+	ec2LaunchV2ServiceName := "\"Amazon EC2Launch\""
+	serviceExists, err := vm.serviceExists(ec2LaunchV2ServiceName)
+	if err != nil {
+		return fmt.Errorf("error checking if %s service exists: %w", ec2LaunchV2ServiceName, err)
+	}
+	// We don't want ensureServiceIsRunning to create the service if it does not exist, so we return without an error.
+	// Returning an error does not make sense as we could have an AMI configured without the EC2Launch service present
+	// and the route created using some other customer specific method which is unknown to us.
+	if !serviceExists {
+		vm.log.Info("missing", "service", ec2LaunchV2ServiceName)
```
jrvaldes, line 522: Wondering if this should be logged with lower versatility (i.e. WARN) now that states something is missing. Before with `service not present` it was fine. IMO.

aravindhp: There is no `log.Warn` option. Are you asking for it to be `log.V(1).Info`?

PS: please don't give lgtm and then suggest a change.

aravindhp: I also want to call out that we always want to see this message and it should not have an option to be suppressed.

jrvaldes: Understood.

aravindhp:
/retest-required

Both AWS jobs failed early before even the Job pod came up.

PR #1770, opened 2023-09-06
OCPBUGS-17217: Fix "WMCO does not wait for instance to reboot properly"

saifshaikh48:
This PR adds a necessary wait to our instance reboot flow. 
After kicking off a reboot, we now first wait for the instance to become unreachable via SSH. 
Then it waits for the instance to be back online before returning the reboot as successful.

Previously WMCO checked if the instance is reachable directly after issuing the reboot request, 
which lead to false positives since the instance was viewed as successfully rebooted before the reboot even began.

PR #1774, opened 2023-09-08
OCPBUGS-18554: Fix "error removing HNS network when cleaning up BYOH proxy nodes"

saifshaikh48:
This PR re-orders a deconfiguration step removing files and HNS networks
to be after the instance has finished rebooting. This way WMCO is guaranteed
to only interact with the instance when it is reachable via SSH.

Previously, we were hitting timing issues where, after WICD cleanup is ran,
WMCO's configmap and node controllers would race, resulting in WMCO issuing
SSH commands while the node is still rebooting. These would fail and cause 
deconfiguration to reconcile again, leading to timeouts in CI for our deletion suite.

### DIFF
--- a/pkg/nodeconfig/nodeconfig.go
+++ b/pkg/nodeconfig/nodeconfig.go
@@ -507,6 +507,7 @@ func (nc *nodeConfig) Deconfigure() error {
 	if nc.node == nil {
 		return fmt.Errorf("instance does not a have an associated node to deconfigure")
 	}
+	nc.log.Info("deconfiguring")
 	// Cordon and drain the Node before we interact with the instance
 	drainHelper := nc.newDrainHelper()
 	if err := drain.RunCordonOrUncordon(drainHelper, nc.node, true); err != nil {
@@ -517,11 +518,10 @@ func (nc *nodeConfig) Deconfigure() error {
 	}
 
 	// Revert all changes we've made to the instance by removing installed services, files, and the version annotation
-	wicdKC, err := nc.generateWICDKubeconfig()
-	if err != nil {
+	if err := nc.cleanupWithWICD(); err != nil {
 		return err
 	}
-	if err := nc.Windows.Deconfigure(nc.wmcoNamespace, wicdKC); err != nil {
+	if err := nc.Windows.RemoveFilesAndNetworks(); err != nil {
 		return fmt.Errorf("error deconfiguring instance: %w", err)
 	}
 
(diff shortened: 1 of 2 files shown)


saifshaikh48:
/hold 

blocked by https://github.com/openshift/windows-machine-config-operator/pull/1770

saifshaikh48:
/unhold

https://github.com/openshift/windows-machine-config-operator/pull/1770 merged

pkg/windows/windows.go
```diff
@@ -234,8 +234,8 @@ type Windows interface {
 	Bootstrap(string, string, string) error
 	// ConfigureWICD ensures that the Windows Instance Config Daemon is running on the node
 	ConfigureWICD(string, string) error
-	// Deconfigure removes all files and networks created by WMCO and runs the WICD cleanup command.
-	Deconfigure(string, string) error
+	// Deconfigure removes all files and networks created by WMCO
+	Deconfigure() error
```
sebsoto, line 238: This isn't really `Deconfigure` at this point. Its no longer undoing changes done by `ConfigureWICD`
It may be more apt to name it `RemoveFilesAndNetworks`

pkg/nodeconfig/nodeconfig.go
```diff
@@ -521,7 +521,15 @@ func (nc *nodeConfig) Deconfigure() error {
...
+		return fmt.Errorf("unable to cleanup the Windows instance: %w", err)
+	}
+
+	// Wait for reboot annotation removal. This prevents deleting the node until the node no longer needs reboot.
+	if err := metadata.WaitForRebootAnnotationRemoval(context.TODO(), nc.client, nc.node.Name); err != nil {
+		return err
+	}
+	if err := nc.Windows.Deconfigure(); err != nil {
 		return fmt.Errorf("error deconfiguring instance: %w", err)
 	}
```
sebsoto, line 534: Theres an ordering requirement thats not clear here.
If its solely between `RunWICDCleanup` and `WaitForRebootAnnotationRemoval` it would be best to move them to a separate documented function.

sebsoto (changes requested):

pkg/windows/windows.go
aravindhp: I agree with this given that we no longer have `Windows.Configure()`. In some ways you are doing the opposite of `windows.Bootstrap()`.

pkg/nodeconfig/nodeconfig.go
aravindhp: At this point even `nc.generateWICDKubeconfig()` needs to move into the separate documented function.

aravindhp (changes requested):

pkg/nodeconfig/nodeconfig.go
saifshaikh48: Makes sense, thanks for pointing this out

aravindhp (approved):

alinaryan (commented):
Does this need any testing? Otherwise lgtm

saifshaikh48:
@alinaryan the testing I did was 
- making sure the proxy job passes e2e
- running the deletion tests a bunch of times to make sure the BYOH removal flake wasn't seen anymore

PR #1778, opened 2023-09-12
OCPBUGS-5002: Copy Azure setup files only on Azure 

wgahnagl:
AzureCloudNodeManagerPath was being copied to all platforms, causing
errors. This change ensures AzureCloudNodeManagerPath is only copied
on Azure.

### DIFF
--- a/pkg/csr/csr.go
+++ b/pkg/csr/csr.go
@@ -317,7 +317,7 @@ func matchesHostname(nodeName string, windowsInstances []*instance.Info,
 // findHostName returns the actual host name of the instance by running the 'hostname' command
 func findHostName(instanceInfo *instance.Info, instanceSigner ssh.Signer) (string, error) {
 	// We don't need to pass most args here as we just need to be able to run commands on the instance.
-	win, err := windows.New("", instanceInfo, instanceSigner)
+	win, err := windows.New("", instanceInfo, instanceSigner, nil)
 	if err != nil {
 		return "", fmt.Errorf("error instantiating Windows instance: %w", err)
 	}
--- a/pkg/nodeconfig/nodeconfig.go
+++ b/pkg/nodeconfig/nodeconfig.go
@@ -125,7 +125,7 @@ func NewNodeConfig(c client.Client, clientset *kubernetes.Clientset, clusterServ
 	}
 
 	log := ctrl.Log.WithName(fmt.Sprintf("nc %s", instanceInfo.Address))
-	win, err := windows.New(clusterDNS, instanceInfo, signer)
+	win, err := windows.New(clusterDNS, instanceInfo, signer, &platformType)
 	if err != nil {
 		return nil, fmt.Errorf("error instantiating Windows instance from VM: %w", err)
 	}
--- a/pkg/windows/windows.go
+++ b/pkg/windows/windows.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	"github.com/go-logr/logr"
+	config "github.com/openshift/api/config/v1"
 	"golang.org/x/crypto/ssh"
 	"k8s.io/apimachinery/pkg/util/wait"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -137,8 +138,6 @@ const (
 )
 
 var (
-	// filesToTransfer is a map of what files should be copied to the Windows VM and where they should be copied to
-	filesToTransfer map[*payload.FileInfo]string
 	// RequiredServices is a list of Windows services installed by WMCO. WICD owns all services aside from itself.
 	// The order of this slice matters due to service dependencies. If a service depends on another service, the
 	// dependent service should be placed before the service it depends on.
@@ -167,11 +166,22 @@ var (
 	}
 )
 
-// getFilesToTransfer returns the properly populated filesToTransfer map. Note this does not include the WICD binary.
-func getFilesToTransfer() (map[*payload.FileInfo]string, error) {
-	if filesToTransfer != nil {
-		return filesToTransfer, nil
+// createPayload returns the map of files to transfer with generated file info
+func createPayload(platform *config.PlatformType) (map[*payload.FileInfo]string, error) {
+	srcDestPairs := getFilesToTransfer(platform)
+	files := make(map[*payload.FileInfo]string)
+	for src, dest := range srcDestPairs {
+		f, err := payload.NewFileInfo(src)
+		if err != nil {
+			return nil, fmt.Errorf("could not create FileInfo object for file %s: %w", src, err)
+		}
+		files[f] = dest
 	}
+	return files, nil
+}
+
+// getFilesToTransfer returns the properly populated filesToTransfer map. Note this does not include the WICD binary.
+func getFilesToTransfer(platform *config.PlatformType) map[string]string {
 	srcDestPairs := map[string]string{
 		payload.GcpGetValidHostnameScriptPath:  remoteDir,
 		payload.WinDefenderExclusionScriptPath: remoteDir,
@@ -183,7 +193,6 @@ func getFilesToTransfer() (map[*payload.FileInfo]string, error) {
 		payload.WinOverlayCNIPlugin:            cniDir,
 		payload.KubeProxyPath:                  K8sDir,
 		payload.KubeletPath:                    K8sDir,
-		payload.AzureCloudNodeManagerPath:      K8sDir,
 		payload.KubeLogRunnerPath:              K8sDir,
 		payload.CSIProxyPath:                   K8sDir,
 		payload.ContainerdPath:                 ContainerdDir,
@@ -191,16 +200,11 @@ func getFilesToTransfer() (map[*payload.FileInfo]string, error) {
 		payload.ContainerdConfPath:             ContainerdDir,
 		payload.NetworkConfigurationScript:     remoteDir,
 	}
-	files := make(map[*payload.FileInfo]string)
-	for src, dest := range srcDestPairs {
-		f, err := payload.NewFileInfo(src)
-		if err != nil {
-			return nil, fmt.Errorf("could not create FileInfo object for file %s: %w", src, err)
-		}
-		files[f] = dest
+
+	if platform != nil && *platform == config.AzurePlatformType {
+		srcDestPairs[payload.AzureCloudNodeManagerPath] = K8sDir
 	}
-	filesToTransfer = files
-	return filesToTransfer, nil
+	return srcDestPairs
 }
 
 // GetK8sDir returns the location of the kubernetes executable directory
@@ -257,15 +261,22 @@ type windows struct {
 	log      logr.Logger
 	// defaultShellPowerShell indicates if the default SSH shell is PowerShell
 	defaultShellPowerShell bool
+	// filesToTransfer is the map of files needed for the windows VM
+	filesToTransfer map[*payload.FileInfo]string
 }
 
 // New returns a new Windows instance constructed from the given WindowsVM
-func New(clusterDNS string, instanceInfo *instance.Info, signer ssh.Signer) (Windows, error) {
+func New(clusterDNS string, instanceInfo *instance.Info, signer ssh.Signer, platform *config.PlatformType) (Windows, error) {
 	log := ctrl.Log.WithName(fmt.Sprintf("wc %s", instanceInfo.Address))
 	log.V(1).Info("initializing SSH connection")
 	conn, err := newSshConnectivity(instanceInfo.Username, instanceInfo.Address, signer, log)
 	if err != nil {
 		return nil, fmt.Errorf("unable to setup VM %s sshConnectivity: %w", instanceInfo.Address, err)
+	}
+
+	files, err := createPayload(platform)
+	if err != nil {
+		return nil, fmt.Errorf("unable to create payload: %w", err)
 	}
 
 	return &windows{
@@ -274,6 +285,7 @@ func New(clusterDNS string, instanceInfo *instance.Info, signer ssh.Signer) (Win
 			instance:               instanceInfo,
 			log:                    log,
 			defaultShellPowerShell: defaultShellPowershell(conn),
+			filesToTransfer:        files,
 		},
 		nil
 }
@@ -652,11 +664,7 @@ func (vm *windows) removeDirectories() error {
 // transferFiles copies various files required for configuring the Windows node, to the VM.
 func (vm *windows) transferFiles() error {
 	vm.log.Info("transferring files")
-	filesToTransfer, err := getFilesToTransfer()
-	if err != nil {
-		return fmt.Errorf("error getting list of files to transfer: %w", err)
-	}
-	for src, dest := range filesToTransfer {
+	for src, dest := range vm.filesToTransfer {
 		if err := vm.EnsureFile(src, dest); err != nil {
 			return fmt.Errorf("error copying %s to %s: %w", src.Path, dest, err)
 		}
--- a/pkg/windows/windows_test.go
+++ b/pkg/windows/windows_test.go
@@ -3,8 +3,44 @@ package windows
 import (
 	"testing"
 
+	config "github.com/openshift/api/config/v1"
 	"github.com/stretchr/testify/assert"
+
+	"github.com/openshift/windows-machine-config-operator/pkg/nodeconfig/payload"
 )
+
+func TestGetFilesToTransfer(t *testing.T) {
+	testCases := []struct {
+		name     string
+		platform *config.PlatformType
+	}{
+		{
+			name:     "test AWS",
+			platform: func() *config.PlatformType { t := config.AWSPlatformType; return &t }(),
+		},
+		{
+			name:     "test Azure",
+			platform: func() *config.PlatformType { t := config.AzurePlatformType; return &t }(),
+		},
+		{
+			name:     "test Nil",
+			platform: nil,
+		},
+	}
+
+	for _, test := range testCases {
+		t.Run(test.name, func(t *testing.T) {
+			files := getFilesToTransfer(test.platform)
+			if test.platform != nil && *test.platform == config.AzurePlatformType {
+				file := files[payload.AzureCloudNodeManagerPath]
+				assert.Equal(t, K8sDir, file)
+			} else {
+				_, exists := files[payload.AzureCloudNodeManagerPath]
+				assert.False(t, exists)
+			}
+		})
+	}
+}
 
 func TestSplitPath(t *testing.T) {
 	testCases := []struct {


pkg/nodeconfig/nodeconfig.go
```diff
@@ -33,6 +33,7 @@ import (
 	"github.com/openshift/windows-machine-config-operator/pkg/ignition"
```
jrvaldes, line 33: Consider splitting the changes into two commits; 1) the refactoring and 2) Add check for file copying on azure

pkg/nodeconfig/nodeconfig.go
```diff
@@ -61,6 +62,153 @@ const (
...
+		csiProxyLogDir,
+		KubeProxyLogDir,
+		wicdLogDir,
+		HybridOverlayLogDir,
+		ContainerdDir,
+		containerdLogDir,
+		podManifestDirectory,
+		K8sDir,
+	}
+)
```
sebsoto, line 210: Bubbling all this information up into the nodeconfig package is not what I'd recommend here. This is overloading the responsibilities of nodeconfig by encoding all of this Windows instance specific information within the package.

sebsoto: Like I said before:

>For a starting point I suggest first experimenting with how you can manipulate the [srcDestPairs](https://github.com/openshift/windows-machine-config-operator/blob/55c20d6c68e4fbc6fc1c9ca62d86645f8b726544/pkg/windows/windows.go#L174) when creating the Windows object.

Rather than removing all of this logic from the package itself, I suggest changing the windows.New() function, and the windows struct itself to reflect what needs to be transferred

pkg/windows/windows_test.go
```diff
@@ -3,9 +3,34 @@ package windows
 import (
 	"testing"
 
+	configv1 "github.com/openshift/api/config/v1"
```
sebsoto, line 6: This should be changed to `config` instead of `configv1`

pkg/csr/csr.go
```diff
@@ -317,7 +318,7 @@ func matchesHostname(nodeName string, windowsInstances []*instance.Info,
 // findHostName returns the actual host name of the instance by running the 'hostname' command
 func findHostName(instanceInfo *instance.Info, instanceSigner ssh.Signer) (string, error) {
 	// We don't need to pass most args here as we just need to be able to run commands on the instance.
-	win, err := windows.New("", instanceInfo, instanceSigner)
+	win, err := windows.New("", instanceInfo, instanceSigner, v1.NonePlatformType)
```
sebsoto, line 321: This can't be hardcoded to `v1.NonePlatformType`. It will cause an issue whenever findHostName is called on another platform.

If this for some reason won't cause an issue, it needs to be documented because that is not clear.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -125,7 +125,8 @@ func NewNodeConfig(c client.Client, clientset *kubernetes.Clientset, clusterServ
 	}
 
 	log := ctrl.Log.WithName(fmt.Sprintf("nc %s", instanceInfo.Address))
-	win, err := windows.New(clusterDNS, instanceInfo, signer)
+
```
sebsoto, line 128: Extra whitespace

pkg/windows/windows.go
```diff
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/go-logr/logr"
+	configv1 "github.com/openshift/api/config/v1"
```
sebsoto, line 12: configv1 -> config

pkg/windows/windows.go
```diff
@@ -256,23 +221,68 @@ type windows struct {
 	log      logr.Logger
 	// defaultShellPowerShell indicates if the default SSH shell is PowerShell
 	defaultShellPowerShell bool
+	// filesToTransfer is the list of files needed for the windows VM
+	filesToTransfer map[*payload.FileInfo]string
+}
+
+func getFilesToTransfer(platform configv1.PlatformType) (map[*payload.FileInfo]string, error) {
+
```
sebsoto, line 229: Extra whitespace

pkg/windows/windows.go
```diff
@@ -256,23 +221,68 @@ type windows struct {
 	log      logr.Logger
 	// defaultShellPowerShell indicates if the default SSH shell is PowerShell
 	defaultShellPowerShell bool
+	// filesToTransfer is the list of files needed for the windows VM
+	filesToTransfer map[*payload.FileInfo]string
+}
+
+func getFilesToTransfer(platform configv1.PlatformType) (map[*payload.FileInfo]string, error) {
```
sebsoto, line 228: By moving this function within the file you have made the diff a lot bigger. Please be careful when doing this as it makes the review more difficult. It should be moved back to where it was, line 169 or so.

sebsoto (changes requested):
Thanks for the PR. PTAL at my comments

aravindhp (changes requested):

pkg/windows/windows.go
```diff
@@ -256,23 +248,42 @@ type windows struct {
 	log      logr.Logger
 	// defaultShellPowerShell indicates if the default SSH shell is PowerShell
 	defaultShellPowerShell bool
+	// filesToTransfer is the list of files needed for the windows VM
```
aravindhp, line 251: This is not a list but a map. Please call out what the key value definitions are here.

pkg/windows/windows.go
```diff
@@ -256,23 +248,42 @@ type windows struct {
...
-func New(clusterDNS string, instanceInfo *instance.Info, signer ssh.Signer) (Windows, error) {
+func New(clusterDNS string, instanceInfo *instance.Info, signer ssh.Signer, platform *config.PlatformType) (Windows, error) {
 	log := ctrl.Log.WithName(fmt.Sprintf("wc %s", instanceInfo.Address))
 	log.V(1).Info("initializing SSH connection")
 	conn, err := newSshConnectivity(instanceInfo.Username, instanceInfo.Address, signer, log)
 	if err != nil {
 		return nil, fmt.Errorf("unable to setup VM %s sshConnectivity: %w", instanceInfo.Address, err)
 	}
 
+	files := filesToTransfer
```
aravindhp, line 264: With your change we shouldn't need the global `filesToTransfer` map anymore. Why are you retaining it?

aravindhp (changes requested):

pkg/windows/windows.go
```diff
@@ -256,23 +248,42 @@ type windows struct {
...
+			return nil, fmt.Errorf("unable to get list of files to transfer: %w", err)
+		}
+		files := make(map[*payload.FileInfo]string)
+		for src, dest := range srcDestPairs {
+			f, err := payload.NewFileInfo(src)
+			if err != nil {
+				return nil, fmt.Errorf("could not create FileInfo object for file %s: %w", src, err)
+			}
+			files[f] = dest
+		}
```
saifshaikh48, line 277: any reason for moving this out of the getFilesToTransfer() func?

aravindhp: Good catch @saifshaikh48. We should refrain from adding inline code to a function especially a top level one.

pkg/windows/windows.go
wgahnagl: Is there a place where the list of files to transfer gets overridden? I thought that was why it was written that way in the first place, and I don't want to create a regression, but if that's not the case, getting rid of it would be good!

pkg/windows/windows.go
wgahnagl: If I moved the fileInfo generation back into the getFilesToTransfer function, I wouldn't be able to unit test, because the NewFileInfo tries to get a real location and a hash for a file that doesn't exist in unit tests. 
I could create another wrapper function maybe?

pkg/windows/windows.go
aravindhp: >Is there a place where the list of files to transfer gets overridden?

Please look for all instances of `filesToTransfer` and figure that out. If there is a regression you won't be able to bring up a node which should be part of your dev testing.

pkg/windows/windows.go
aravindhp: >If I moved the fileInfo generation back into the getFilesToTransfer function, I wouldn't be able to unit test,

This can't be the reason to add a bunch of inline code to a top level function.

>I wouldn't be able to unit test, because the NewFileInfo tries to get a real location and a hash for a file that doesn't exist in unit tests.

I think it is possible to "mock" file creation using a temp file but I am not 100% with your context.

>I could create another wrapper function maybe?

The end result should be a single function call in the `New()` method.

aravindhp (changes requested):

pkg/windows/windows.go
wgahnagl: I think part of the thing that I'm testing is that it's the right windows path though, which makes pointing to a temp file impossible since it would be checking windows directory paths like C://k, or else I would have to find a way to make the path point to the wrong/temp directory which I feel would defeat the purpose of the test, and involve changing the constants.

aravindhp: Understood. But that can't be the reason for you to inline code in a top level function. I am reiterating that the end result should be a single function call in the `New()` method. This doesn't mean that you curtailed to introduce just a single helper method. You can break that helper down too allowing you to unit test the sub-methods.

pkg/windows/windows.go
```diff
@@ -166,11 +165,24 @@ var (
...
+		if err != nil {
+			return nil, fmt.Errorf("could not create FileInfo object for file %s: %w", src, err)
+		}
+		files[f] = dest
+	}
+	return files, nil
+}
+
+// getFilesToTransfer returns the properly populated filesToTransfer map. Note this does not include the WICD binary.
+func getFilesToTransfer(platform *config.PlatformType) (map[string]string, error) {
```
aravindhp, line 185: The restructuring you are doing is hinting that this method also needs to move to the `payload` package. The way to do it is have a pre-commit that moves it and then make changes in a follow up commit.

pkg/windows/windows.go
```diff
@@ -256,23 +262,27 @@ type windows struct {
...
-func New(clusterDNS string, instanceInfo *instance.Info, signer ssh.Signer) (Windows, error) {
+func New(clusterDNS string, instanceInfo *instance.Info, signer ssh.Signer, platform *config.PlatformType) (Windows, error) {
 	log := ctrl.Log.WithName(fmt.Sprintf("wc %s", instanceInfo.Address))
 	log.V(1).Info("initializing SSH connection")
 	conn, err := newSshConnectivity(instanceInfo.Username, instanceInfo.Address, signer, log)
 	if err != nil {
 		return nil, fmt.Errorf("unable to setup VM %s sshConnectivity: %w", instanceInfo.Address, err)
 	}
 
+	files, err := createFileInfo(platform)
```
aravindhp, line 278: Errors returned cannot be ignored

pkg/windows/windows.go
```diff
@@ -166,11 +165,24 @@ var (
 	}
 )
 
-// getFilesToTransfer returns the properly populated filesToTransfer map. Note this does not include the WICD binary.
-func getFilesToTransfer() (map[*payload.FileInfo]string, error) {
-	if filesToTransfer != nil {
-		return filesToTransfer, nil
+func createFileInfo(platform *config.PlatformType) (map[*payload.FileInfo]string, error) {
```
alinaryan, line 168: Can you add a comment describing the function

aravindhp (changes requested):

pkg/windows/windows_test.go
```diff
@@ -3,9 +3,45 @@ package windows
...
+			platform: nil,
+		},
+	}
+
+	for _, test := range testCases {
+		t.Run(test.name, func(t *testing.T) {
+			files, err := getFilesToTransfer(test.platform)
+			if test.platform != nil && *test.platform == config.AzurePlatformType {
+				file := files[payload.AzureCloudNodeManagerPath]
+				assert.Equal(t, "C:\\k", file)
```
saifshaikh48, line 36: let's use the `K8sDir` constant rather than hardocding `"C:\\k"`

pkg/windows/windows_test.go
```diff
@@ -3,9 +3,45 @@ package windows
...
+			} else {
+				assert.Equal(t, "", files[payload.AzureCloudNodeManagerPath])
```
saifshaikh48, line 38: could lead to a false positive, I suggest making this more explicit
```suggestion
				_, exists := files[payload.AzureCloudNodeManagerPath]
				assert.False(t,exists)
```

pkg/windows/windows.go
wgahnagl: if we move the getFilesToTransfer function to payload, we'd have to move all the constants that are in the windows package 
is that a good idea?

aravindhp: The refactor also needs to take of the cyclic dependency that will get introduced. Given the turnaround time it is taking, lets leave it here for now.

pkg/windows/windows.go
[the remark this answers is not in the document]
aravindhp: You can leave this function here too but please rename and document it.

pkg/windows/windows.go
```diff
@@ -166,11 +165,25 @@ var (
...
+		if err != nil {
+			return nil, fmt.Errorf("could not create FileInfo object for file %s: %w", src, err)
+		}
+		files[f] = dest
+	}
+	return files, nil
+}
+
+// getFilesToTransfer returns the properly populated filesToTransfer map. Note this does not include the WICD binary.
+func getFilesToTransfer(platform *config.PlatformType) (map[string]string, error) {
```
saifshaikh48, line 186: This function no longer needs to return `error`. Please adjust the signature to have 1 return value

pkg/windows/windows.go
```diff
@@ -256,23 +263,28 @@ type windows struct {
...
-func New(clusterDNS string, instanceInfo *instance.Info, signer ssh.Signer) (Windows, error) {
+func New(clusterDNS string, instanceInfo *instance.Info, signer ssh.Signer, platform *config.PlatformType) (Windows, error) {
 	log := ctrl.Log.WithName(fmt.Sprintf("wc %s", instanceInfo.Address))
 	log.V(1).Info("initializing SSH connection")
 	conn, err := newSshConnectivity(instanceInfo.Username, instanceInfo.Address, signer, log)
 	if err != nil {
 		return nil, fmt.Errorf("unable to setup VM %s sshConnectivity: %w", instanceInfo.Address, err)
 	}
 
+	files, err := createPayload(platform)
```
saifshaikh48, line 279: need to check the error value
```suggestion
	files, err := createPayload(platform)
	if err != nil {
		return nil, fmt.Errorf("unable to create payload: %w", err)
	}
```

pkg/windows/windows.go
```diff
@@ -166,11 +165,25 @@ var (
...
+		}
+		files[f] = dest
+	}
+	return files, nil
+}
+
+// getFilesToTransfer returns the properly populated filesToTransfer map. Note this does not include the WICD binary.
+func getFilesToTransfer(platform *config.PlatformType) (map[string]string, error) {
 	srcDestPairs := map[string]string{
 		payload.GcpGetValidHostnameScriptPath:  remoteDir,
```
jrvaldes, line 188: This PR could be easily reframed as "Copy only specific setup files given target platform" to avoid copying `payload.GcpGetValidHostnameScriptPath` to all platforms as it is specific to GCP.

ATM is out of the scope of the ticket, consider including it or a follow-up PR works too.

pkg/windows/windows.go
jrvaldes: > return fmt.Errorf("error getting list of files to transfer: %w", err)

I would rather stick to the [existing](https://github.com/openshift/windows-machine-config-operator/pull/1778/commits/956a384bcd1b1421ef04531545af7258e9b3b130#diff-4d8fe3bb66a872f0e4a277f820d60d2efa722f09aa89516be74b98c06b57f3bdL657) message for consistency.

jrvaldes (commented):
@wgahnagl  thanks for working on this. LGTM, added minor suggestions. PTAL

aravindhp (approved):

wgahnagl:
/retest-required

i/o timeout 
```{  011 13:48:44.511343      33 memcache.go:265] couldn't get current server API group list: Get "https://api.ci-op-g84wnp8p-5c595.nutanix-ci.devcluster.openshift.com:6443/api?timeout=32s": dial tcp 10.0.200.14:6443: i/o timeout```

**Comment by aravindhp** [comment]
> /retest-required
> 
> i/o timeout `{ 011 13:48:44.511343 33 memcache.go:265] couldn't get current server API group list: Get "https://api.ci-op-g84wnp8p-5c595.nutanix-ci.devcluster.openshift.com:6443/api?timeout=32s": dial tcp 10.0.200.14:6443: i/o timeout`

@wgahnagl That error is during the must-gather/deprovision phase which implies that the real error happened [before](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/1778/pull-ci-openshift-windows-machine-config-operator-master-nutanix-e2e-operator/1712090780752941056):
```
 waiting for api to be available
level=error
level=error msg=Error: error waiting for vm (c777addc-33df-4ec5-8a74-edee592b453f) to create: error_detail: INTERNAL_ERROR: Internal Server Error., progress_message: create_vm
level=error
level=error msg=  with nutanix_virtual_machine.vm_master[1],
level=error msg=  on main.tf line 65, in resource "nutanix_virtual_machine" "vm_master":
level=error msg=  65: resource "nutanix_virtual_machine" "vm_master" {
level=error
level=error
level=error msg=Error: error waiting for vm (336a4d33-62d2-4b79-8629-fd6d508d71e2) to create: error_detail: INTERNAL_ERROR: Internal Server Error., progress_message: create_vm
level=error
level=error msg=  with nutanix_virtual_machine.vm_master[0],
level=error msg=  on main.tf line 65, in resource "nutanix_virtual_machine" "vm_master":
level=error msg=  65: resource "nutanix_virtual_machine" "vm_master" {
level=error
level=error
level=error msg=Error: error waiting for vm (7830a795-09be-45aa-a6d7-8d08679740a5) to create: error_detail: INTERNAL_ERROR: Internal Server Error., progress_message: create_vm
level=error
level=error msg=  with nutanix_virtual_machine.vm_master[2],
level=error msg=  on main.tf line 65, in resource "nutanix_virtual_machine" "vm_master":
level=error msg=  65: resource "nutanix_virtual_machine" "vm_master" {
level=error
level=error msg=failed to fetch Cluster: failed to generate asset "Cluster": failure applying terraform for "cluster" stage: failed to create cluster: failed to apply Terraform: exit status 1
level=error
level=error msg=Error: error waiting for vm (c777addc-33df-4ec5-8a74-edee592b453f) to create: error_detail: INTERNAL_ERROR: Internal Server Error., progress_message: create_vm
level=error
level=error msg=  with nutanix_virtual_machine.vm_master[1],
level=error msg=  on main.tf line 65, in resource "nutanix_virtual_machine" "vm_master":
level=error msg=  65: resource "nutanix_virtual_machine" "vm_master" {
level=error
level=error
level=error msg=Error: error waiting for vm (336a4d33-62d2-4b79-8629-fd6d508d71e2) to create: error_detail: INTERNAL_ERROR: Internal Server Error., progress_message: create_vm
level=error
level=error msg=  with nutanix_virtual_machine.vm_master[0],
level=error msg=  on main.tf line 65, in resource "nutanix_virtual_machine" "vm_master":
level=error msg=  65: resource "nutanix_virtual_machine" "vm_master" {
level=error
level=error
level=error msg=Error: error waiting for vm (7830a795-09be-45aa-a6d7-8d08679740a5) to create: error_detail: INTERNAL_ERROR: Internal Server Error., progress_message: create_vm
level=error
level=error msg=  with nutanix_virtual_machine.vm_master[2],
level=error msg=  on main.tf line 65, in resource "nutanix_virtual_machine" "vm_master":
level=error msg=  65: resource "nutanix_virtual_machine" "vm_master" {
level=error
level=error
Installer exit with code 4
```

wgahnagl:
/retest-required 

time="2023-10-11T17:47:36Z" level=debug msg="Warning: Disabled Providers: foundation. Please provide required fields in provider configuration to enable them. Refer docs."
time="2023-10-11T17:47:36Z" level=debug
time="2023-10-11T17:47:36Z" level=debug msg="  with provider[\"openshift/local/nutanix\"],"
time="2023-10-11T17:47:36Z" level=debug msg="  on main.tf line 5, in provider \"nutanix\":"
time="2023-10-11T17:47:36Z" level=debug msg="   5: provider \"nutanix\" {"
time="2023-10-11T17:47:36Z" level=debug
time="2023-10-11T17:47:36Z" level=debug msg="(and one more similar warning elsewhere)"
time="2023-10-11T17:47:36Z" level=error
time="2023-10-11T17:47:36Z" level=error msg="Error: error waiting for vm (3101ef98-ad27-479b-9e93-ddf9d6f6139a) to create: error_detail: INTERNAL_ERROR: Internal Server Error., progress_message: create_vm"
time="2023-10-11T17:47:36Z" level=error 

not sure if the warning here is what's causing the failure to create the VM, retesting one more time

wgahnagl:
/retest-required 

the node that the test was being run on was behaving strangely. They've replaced the node, and told me to retest

PR #1781, opened 2023-09-13
Allow possibility of nil Azure zone in CI

sebsoto:
Azure Machines do not have to specify a zone. In this case the zone is nil, and could cause a panic when creating a new MachineSet. This is being addressed by not dereferencing the copied zone pointer, and instead using it as is when creating a AzureMachineProviderSpec.


saifshaikh48 (commented):
/lgtm

jrvaldes (commented):
/lgtm

aravindhp (approved):
Thanks for catching and fixing this, @sebsoto

PR #1792, opened 2023-09-14
Update default vSphere network segment for DevQE vcenter

jrvaldes:
This PR set `devqe-segment-222` as the default network segment for development clusters in the DevQE vcenter.

### DIFF
--- a/test/e2e/providers/vsphere/vsphere.go
+++ b/test/e2e/providers/vsphere/vsphere.go
@@ -111,7 +111,7 @@ func (p *Provider) getWorkspaceFromExistingMachineSet() (*mapi.Workspace, error)
 // getNetwork returns the network that needs to be used in the MachineSet
 func getNetwork() string {
 	// Default network for dev environment
-	networkSegment := "dev-segment"
+	networkSegment := "devqe-segment-222"
 	if os.Getenv("OPENSHIFT_CI") == "true" {
 		// $LEASED_RESOURCE holds the network the CI cluster is in
 		networkSegment = os.Getenv("LEASED_RESOURCE")


test/e2e/providers/vsphere/vsphere.go
```diff
@@ -110,17 +110,20 @@ func (p *Provider) getWorkspaceFromExistingMachineSet() (*mapi.Workspace, error)
 
 // getNetwork returns the network that needs to be used in the MachineSet
 func getNetwork() string {
-	// Default network for dev environment
-	networkSegment := "devqe-segment-222"
+	// $LEASED_RESOURCE holds the network the CI cluster is in and allows configuration of the network segment
+	// to use for testing with development clusters.
```
aravindhp, line 114: I don't follow how `LEASED_RESOURCE` works in conjunction with CI and development clusters. Is this an environment variable that gets set in a CI env i.e., with e2e tests or are you introducing this for dev testing? Or both?

If both, do we need to complicate our workflow even further? Can't we just agree to use `devqe-segment-222`? If yes, then this commit can be dropped and the only commit needed will be to update the default network.

test/e2e/providers/vsphere/vsphere.go
```diff
@@ -111,7 +111,7 @@ func (p *Provider) getWorkspaceFromExistingMachineSet() (*mapi.Workspace, error)
 // getNetwork returns the network that needs to be used in the MachineSet
 func getNetwork() string {
 	// Default network for dev environment
-	networkSegment := "dev-segment"
+	networkSegment := "devqe-segment-222"
```
aravindhp: What is the point in making this change and immediately changing it in a follow up commit in the same PR? This is just unnecessary code churn.

aravindhp (changes requested):

test/e2e/providers/vsphere/vsphere.go
```diff
@@ -110,17 +110,20 @@ func (p *Provider) getWorkspaceFromExistingMachineSet() (*mapi.Workspace, error)
...
+	// otherwise use default values
 	if os.Getenv("OPENSHIFT_CI") == "true" {
-		// $LEASED_RESOURCE holds the network the CI cluster is in
-		networkSegment = os.Getenv("LEASED_RESOURCE")
-		// Default to "ci-segment" if the environment variable is not set.
-		if networkSegment == "" {
-			networkSegment = "ci-segment"
-		}
+		// Default to "ci-segment" if the environment variable is not set while running in CI
+		return "ci-segment"
```
aravindhp, line 123: Is this a valid segment to use anymore?

test/e2e/providers/vsphere/vsphere.go
jrvaldes: Correct, `devqe-segment-222` could be agreed as the default, and remove this commit. 

That is the main goal of this PR, propose `devqe-segment-222` as the new default network segment. I wanted to explore with the team the need to configure the network segment externally. Will drop the commit.

test/e2e/providers/vsphere/vsphere.go
jrvaldes: True, the follow-up commit was to highlight the additional (no longer needed) change.

test/e2e/providers/vsphere/vsphere.go
jrvaldes: IDK, and I have the same query. Anyhow, it's out of the scope here.

aravindhp (approved):

test/e2e/providers/vsphere/vsphere.go
sebsoto: This should be pulled from the other machines in the cluster,  like the workspace is https://github.com/openshift/windows-machine-config-operator/pull/1792/files#diff-43febcc9af1775a8b8390c4d64b28993e558e2e58476773d1e1d4c78b5ff0b75R85

Otherwise if this changes again things will break

sebsoto (changes requested):

test/e2e/providers/vsphere/vsphere.go
aravindhp: Good point @sebsoto

aravindhp (commented):
/approve cancel

test/e2e/providers/vsphere/vsphere.go
jrvaldes: True. I looked at it while fixing this. I didn't feel it was worth the refactoring, plus losing the option of injecting the network name from env in CI (i.e. $LEASED_RESOURCE). 

Happy to implement the recommendation if that is the agreement. LMK.

aravindhp: >True. I looked at it while fixing this. I didn't feel it was worth the refactoring, plus losing the option of injecting the network name from env in CI (i.e. $LEASED_RESOURCE).

What is the advantage of this option? @sebsoto's approach is the better long term fix given other workers will be using the correct  network.

jrvaldes: The current proposal is simpler and does not require refactoring in the e2e code, possibly adding more complexity (I know the `providerSpec` is there already 😄 ). 

I can't think of any implication on not using the `LEASED_RESOURCE`  env var right now, but maybe be some.

aravindhp: >The current proposal is simpler and does not require refactoring in the e2e code, possibly adding more complexity (I know the providerSpec is there already 😄 ).

Simpler yes but it might cause us to revisit this again. In addition it forces developers to use a single segment. Yet another thing for them to keep in mind. The more I think about it, hard coding the segment feels like the wrong approach. Moreover I don't see the complexity in refactoring ` getWorkspaceFromExistingMachineSet()` into `getRequiredFieldsFromExistingMachineSet()` to also return the network segment.

jrvaldes: I have opened a separate PR with the proposed approach.

- https://github.com/openshift/windows-machine-config-operator/pull/1813

jrvaldes:
/close

in favor of https://github.com/openshift/windows-machine-config-operator/pull/1813

PR #1806, opened 2023-09-15
Test Azure CSI storage

sebsoto:
Adds azure-file CSI storage to the e2e tests

### DIFF
--- a/test/e2e/providers/azure/azure.go
+++ b/test/e2e/providers/azure/azure.go
@@ -4,10 +4,14 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"reflect"
 
 	config "github.com/openshift/api/config/v1"
 	mapi "github.com/openshift/api/machine/v1beta1"
+	apps "k8s.io/api/apps/v1"
 	core "k8s.io/api/core/v1"
+	k8sapierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	client "k8s.io/client-go/kubernetes"
 
@@ -27,6 +31,7 @@ const (
 	// "unknown instance type: Standard_D4s_V3" on dev cluster instances.
 	// Use the instance type the other worker machines use.
 	defaultVMSize = "Standard_D2s_v3"
+	csiNamespace  = "openshift-cluster-csi-drivers"
 )
 
 // Provider is a provider struct for testing Azure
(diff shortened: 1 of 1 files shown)


test/e2e/providers/azure/azure.go
```diff
@@ -119,11 +124,218 @@ func (p *Provider) GetType() config.PlatformType {
...
+	}
+	storageClassName := "azurefile-csi"
+	pvcSpec := core.PersistentVolumeClaim{
+		ObjectMeta: meta.ObjectMeta{
+			GenerateName: "e2e" + "-",
+		},
+		Spec: core.PersistentVolumeClaimSpec{
+			AccessModes: []core.PersistentVolumeAccessMode{core.ReadWriteMany},
+			Resources: core.ResourceRequirements{
+				Requests: core.ResourceList{core.ResourceStorage: resource.MustParse("2Gi")},
```
saifshaikh48, line 142: nit: do we want to request less storage for the e2e suite?

saifshaikh48 (commented):
/lgtm

test/e2e/providers/azure/azure.go
alinaryan: I thought 2Gi was already pretty small. Also not sure how much customers will have.

alinaryan (commented):
Thanks for working on this, are we going to worry about/test Azure-Disk too?

sebsoto:
> are we going to worry about/test Azure-Disk too?

I don't think thats necessary in CI for now, maybe eventually

test/e2e/providers/azure/azure.go
aravindhp: 2Gi is small in today's world 😄

aravindhp (approved):

test/e2e/providers/azure/azure.go
```diff
@@ -119,11 +124,218 @@ func (p *Provider) GetType() config.PlatformType {
...
+						{
+							Key:    "os",
+							Value:  "Windows",
+							Effect: core.TaintEffectNoSchedule,
+						},
+					},
+					Containers: []core.Container{
+						{
+							Name:  "node-driver-registrar",
+							Image: "mcr.microsoft.com/oss/kubernetes-csi/csi-node-driver-registrar:v2.8.0",
```
jrvaldes, line 181: Why not using the [latest](https://github.com/kubernetes-csi/node-driver-registrar/releases/tag/v2.9.0) (v2.9.0) as of now?

test/e2e/providers/azure/azure.go
```diff
@@ -119,11 +124,218 @@ func (p *Provider) GetType() config.PlatformType {
...
+								},
+								{
+									Name:      "registration-dir",
+									MountPath: "/registration",
+								},
+							},
+						},
+						{
+							Name:  "azurefile",
+							Image: "mcr.microsoft.com/k8s/csi/azurefile-csi:latest",
```
jrvaldes, line 218: Should we worry about matching compatible image versions of the `azurefile-csi` and `csi-node-driver-registrar`?

test/e2e/providers/azure/azure.go
```diff
@@ -119,11 +124,218 @@ func (p *Provider) GetType() config.PlatformType {
...
+			},
+		},
+	}
+	// See if the DaemonSet already exists in the state we expect it to be in.
+	existingDS, err := client.AppsV1().DaemonSets(csiNamespace).Get(context.TODO(), dsName, meta.GetOptions{})
+	if err != nil {
+		if !k8sapierrors.IsNotFound(err) {
+			return fmt.Errorf("error getting existing Windows CSI DaemonSet: %w", err)
+		}
+	} else if err == nil {
```
jrvaldes, line 322: No need for `if err == nil`

test/e2e/providers/azure/azure.go
```diff
@@ -119,11 +124,218 @@ func (p *Provider) GetType() config.PlatformType {
...
+		if reflect.DeepEqual(existingDS.Spec, ds.Spec) {
+			// DaemonSet is already as expected, nothing to do here.
+			return nil
+		}
+		// Delete the DaemonSet as it has the wrong spec.
+		err = client.AppsV1().DaemonSets(csiNamespace).Delete(context.TODO(), dsName, meta.DeleteOptions{})
+		if err != nil {
+			return fmt.Errorf("error deleting existing Windows CSI DaemonSet: %w", err)
+
+		}
```
jrvaldes, line 332: Should we wait for the existing DaemonSet to be deleted before trying to create one?

test/e2e/providers/azure/azure.go
```diff
@@ -119,11 +124,218 @@ func (p *Provider) GetType() config.PlatformType {
...
+	}
+	_, err = client.AppsV1().DaemonSets(csiNamespace).Create(context.TODO(), &ds, meta.CreateOptions{})
+	if err != nil {
+		return fmt.Errorf("error creating Windows CSI DaemonSet: %w", err)
+	}
+	return nil
```
jrvaldes, line 338: Do you think we should check the containers are running before returning?

jrvaldes (commented):
@sebsoto  thanks for working on this. I left minor comments. LGTM otherwise.

sebsoto:
/test azure-e2e-operator
Cluster come up failure

test/e2e/providers/azure/azure.go
sebsoto: This is following the documentation provided https://github.com/kubernetes-sigs/azurefile-csi-driver/tree/master/deploy

test/e2e/providers/azure/azure.go
sebsoto: It is needed in order to exclude the case where `k8sapierrors.IsNotFound(err)` evaluates to true

test/e2e/providers/azure/azure.go
sebsoto: This is following the documentation provided https://github.com/kubernetes-sigs/azurefile-csi-driver/tree/master/deploy

test/e2e/providers/azure/azure.go
jrvaldes: That case is excluded by the `else` statement. `err == nil` is always true in the block following the `else` in
```
if err != nil {
	
} else {

}
```

sebsoto: yep you are right, fixed

test/e2e/providers/azure/azure.go
sebsoto: No need, the pods will do the correct thing

test/e2e/providers/azure/azure.go
sebsoto: I don't think that's necessary, if the containers don't go to running the workload won't be able to deploy so the failure will be shown there.

sebsoto:
Overriding due to vsphere test flakes, and the fact that this PR will only affect azure clusters

PR #1810, opened 2023-09-17
Prevent reconfigure test from leaving Machine nodes in a bad state

sebsoto:
Running the reconfiguration test causes Machine backed nodes to be deleted, without waiting for a Node to replace them. This is bleeding over to the deletion test suite, and causing failures, and potential false positives there.

### DIFF
--- a/hack/run-ci-e2e-test.sh
+++ b/hack/run-ci-e2e-test.sh
@@ -169,10 +169,6 @@ if [[ "$TEST" = "all" || "$TEST" = "upgrade" ]]; then
   go test ./test/e2e/... -run=TestWMCO/upgrade -v -timeout=90m -args $GO_TEST_ARGS
 
   # Run the reconfiguration test
-  # The reconfiguration suite must be run directly before the deletion suite. This is because we do not
-  # currently wait for nodes to fully reconcile after changing the private key back to the valid key. Any tests
-  # added/moved in between these two suites may fail.
-  # This limitation will be removed with https://issues.redhat.com/browse/WINC-655
   printf "\n####### Testing reconfiguration #######\n" >> "$ARTIFACT_DIR"/wmco.log
   go test ./test/e2e/... -run=TestWMCO/reconfigure -v -timeout=90m -args $GO_TEST_ARGS
 fi
--- a/test/e2e/reconfigure_test.go
+++ b/test/e2e/reconfigure_test.go
@@ -23,10 +23,6 @@ func reconfigurationTestSuite(t *testing.T) {
 	if tc.CloudProvider.GetType() == config.VSpherePlatformType {
 		t.Run("Re-add removed instance", tc.testReAddInstance)
 	}
-	// testPrivateKeyChange must be the last test run in the reconfiguration suite. This is because we do not currently
-	// wait for nodes to fully come back up after changing the private key back to the valid key. Only the deletion test
-	// suite should run after this. Any other tests may result in flakes.
-	// This limitation will be removed with https://issues.redhat.com/browse/WINC-655
 	t.Run("Change private key", testPrivateKeyChange)
 }
 
--- a/test/e2e/secrets_test.go
+++ b/test/e2e/secrets_test.go
@@ -181,9 +181,9 @@ func testPrivateKeyChange(t *testing.T) {
 	err = tc.validateUpgradeableCondition(meta.ConditionTrue)
 	require.NoError(t, err, "operator Upgradeable condition not in proper state")
 
-	// Re-create the known private key so SSH connection can be re-established
-	// TODO: Remove dependency on this secret by rotating keys as part of https://issues.redhat.com/browse/WINC-655
-	require.NoError(t, tc.createPrivateKeySecret(true), "error confirming known private key secret exists")
+	// revert key changes so the test suite is able to SSH into the VMs
+	require.NoError(t, tc.createPrivateKeySecret(true))
+	require.NoError(t, tc.waitForNewMachineNodes())
 }
 
 // waitForBYOHPrivateKeyUpdate waits until all BYOH Nodes annotations are updated to reflect the expected private key


sebsoto:
https://github.com/openshift/windows-machine-config-operator/pull/1810#issuecomment-1722561037
This failure was due to the BYOH node not being ready. I haven't been able to find the root cause. The backing Machine is present, the Node's kubelet just becomes unreachable.

test/e2e/secrets_test.go
```diff
@@ -182,8 +182,9 @@ func testPrivateKeyChange(t *testing.T) {
 	require.NoError(t, err, "operator Upgradeable condition not in proper state")
 
 	// Re-create the known private key so SSH connection can be re-established
```
aravindhp, line 184: This comment is in the incorrect spot now.

test/e2e/secrets_test.go
```diff
@@ -182,8 +182,9 @@ func testPrivateKeyChange(t *testing.T) {
...
-	// TODO: Remove dependency on this secret by rotating keys as part of https://issues.redhat.com/browse/WINC-655
+	require.NoError(t, tc.loadExistingNodes())
```
aravindhp, line 185: This warrants a comment too. I am not sure what this is fixing.

test/e2e/reconfigure_test.go
```diff
@@ -23,10 +23,6 @@ func reconfigurationTestSuite(t *testing.T) {
 	if tc.CloudProvider.GetType() == config.VSpherePlatformType {
 		t.Run("Re-add removed instance", tc.testReAddInstance)
 	}
-	// testPrivateKeyChange must be the last test run in the reconfiguration suite. This is because we do not currently
```
aravindhp, line 26: Isn't this still the case? I am not sure we can move this test around without things going wrong.

hack/run-ci-e2e-test.sh
```diff
@@ -169,10 +169,6 @@ if [[ "$TEST" = "all" || "$TEST" = "upgrade" ]]; then
   go test ./test/e2e/... -run=TestWMCO/upgrade -v -timeout=90m -args $GO_TEST_ARGS
 
   # Run the reconfiguration test
-  # The reconfiguration suite must be run directly before the deletion suite. This is because we do not
```
aravindhp, line 172: Isn't this still true?

aravindhp (changes requested):
>Running the reconfiguration test causes Machine backed nodes to be deleted, without waiting for a Node to replace them.

I don't follow this. Don't we run all the validation tests as part of the reconfiguration test suite?

hack/run-ci-e2e-test.sh
sebsoto: No it is not true anymore. This change is making it so the test waits nodes to fully reconcile after changing the private key back to the valid key.

test/e2e/secrets_test.go
sebsoto: The sequence is
1) loading existing nodes
2) changing the private-key
3) ensuring that new nodes are created is part of this process.

I can either remove the comment or break this process out into its own function.

aravindhp: I am fine if you move this into a new method

test/e2e/secrets_test.go
```diff
@@ -181,9 +182,36 @@ func testPrivateKeyChange(t *testing.T) {
...
+		return fmt.Errorf("error getting cloud private key: %w", err)
+	}
+	knownKey, err := os.ReadFile(gc.privateKeyPath)
+	if err != nil {
+		return fmt.Errorf("unable to read private key data from file %s: %w", gc.privateKeyPath, err)
+	}
+	if bytes.Equal(currentKey, knownKey) {
+		return nil
+	}
+	if err := tc.loadExistingNodes(); err != nil {
```
saifshaikh48, line 203: Could you comment why this is necessary? I can't figure out why the nodes may be stale here

test/e2e/secrets_test.go
```diff
@@ -181,9 +182,36 @@ func testPrivateKeyChange(t *testing.T) {
...
+}
+
+// revertPrivateKeyChange changes the cloud private key to the known key if necessary, and waits for Machine nodes
+// to make use of it
+func (tc *testContext) revertPrivateKeyChange() error {
+	currentKey, err := tc.getCloudPrivateKey()
+	if err != nil {
+		return fmt.Errorf("error getting cloud private key: %w", err)
+	}
+	knownKey, err := os.ReadFile(gc.privateKeyPath)
```
jrvaldes, line 196: The most expensive operation here is reading the file and is happening twice for the worst-case scenario (when there is a mismatch). I suggest you rename this function as `ensureKnownPrivateKeySecret()` and make the adjustments to avoid the comparison and minimize the I/O operations while just resetting the known key and waiting for the expected status of the nodes.

test/e2e/secrets_test.go
```diff
@@ -181,9 +182,36 @@ func testPrivateKeyChange(t *testing.T) {
...
+	if err != nil {
+		return fmt.Errorf("unable to read private key data from file %s: %w", gc.privateKeyPath, err)
+	}
+	if bytes.Equal(currentKey, knownKey) {
+		return nil
+	}
+	if err := tc.loadExistingNodes(); err != nil {
+		return fmt.Errorf("error loading existing nodes: %w", err)
+
```
jrvaldes, line 205: Why is this needed here? Consider adding an inline comment.

jrvaldes (changes requested):
@sebsoto  thanks for working on this. PTAL at the comments.

test/e2e/secrets_test.go
sebsoto: I'd rather not prematurely optimize here given this is a test.
Its important to make this check so that this function can be called safely in the case that the private key is already set.

 ensureKnownPrivateKeySecret() is not possible as the machine node refresh is the important bit here. thats not happening if the key wasnt changed.

test/e2e/secrets_test.go
```diff
@@ -181,9 +182,36 @@ func testPrivateKeyChange(t *testing.T) {
...
+	if err != nil {
+		return fmt.Errorf("error getting cloud private key: %w", err)
+	}
+	knownKey, err := os.ReadFile(gc.privateKeyPath)
+	if err != nil {
+		return fmt.Errorf("unable to read private key data from file %s: %w", gc.privateKeyPath, err)
+	}
+	if bytes.Equal(currentKey, knownKey) {
+		return nil
+	}
```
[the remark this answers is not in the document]
sebsoto: Its a guard, in case this function is called elsewhere, or if the context around the callsite changes.

test/e2e/secrets_test.go
sebsoto: you're right, it doesn't seem necessary. removed

test/e2e/secrets_test.go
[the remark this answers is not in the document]
aravindhp: Then please add a comment about that.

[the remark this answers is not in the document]
sebsoto: Found the amount of code added by the function was unnecessary given that loadExistingNodes was not needed as pointed out above.

saifshaikh48 (commented):
/lgtm

thanks for this fix

test/e2e/secrets_test.go
```diff
@@ -181,9 +181,9 @@ func testPrivateKeyChange(t *testing.T) {
 	err = tc.validateUpgradeableCondition(meta.ConditionTrue)
 	require.NoError(t, err, "operator Upgradeable condition not in proper state")
 
-	// Re-create the known private key so SSH connection can be re-established
-	// TODO: Remove dependency on this secret by rotating keys as part of https://issues.redhat.com/browse/WINC-655
-	require.NoError(t, tc.createPrivateKeySecret(true), "error confirming known private key secret exists")
+	// revert key changes so the test suite is able to SSH into the VMs
+	require.NoError(t, tc.createPrivateKeySecret(true))
+	require.NoError(t, tc.waitForNewMachineNodes())
```
aravindhp, line 186: Do we need to always do this almost like a defer function call? i.e. will we have issues collecting node logs upon one of the require failures above?

sebsoto: It should only be done directly after the private-key change, calling it at another time would throw an error.

aravindhp: Correct, so do we need to call always even if:
```
	err = tc.validateUpgradeableCondition(meta.ConditionTrue)
	require.NoError(t, err, "operator Upgradeable condition not in proper state")
```
fails?

sebsoto: Nope, that is before the private key change here.

If you're asking about the private key change on line 168, I would say that thats not an issue I'm trying to solve in this PR. This is fixing the CI issue where the reconfigure test returns green, and then CI continues onto the deletion test.

In the case that specific validateUpgradeableCondition call fails, the reconfigure test will return an error and the deletion test will not be called. I am not very concerned about the node taking some time to return to a working state after that.

aravindhp: > the reconfigure test will return an error and the deletion test will not be called. 

And we don't try to collect logs at that point by SSHing into nodes? If so, I am fine with the change.

aravindhp (approved):

alinaryan:
LGTM

aravindhp:
/test aws-e2e-upgrade

Cluster install failed with:
```
level=fatal msg=failed to fetch Terraform Variables: failed to fetch dependency of "Terraform Variables": failed to generate asset "Platform Provisioning Check": baseDomain: Invalid value: "origin-ci-int-aws.dev.rhcloud.com": the zone already has record sets for the domain of the cluster: [api.ci-op-3wqwt9zp-4bc28.origin-ci-int-aws.dev.rhcloud.com. (A), \052.apps.ci-op-3wqwt9zp-4bc28.origin-ci-int-aws.dev.rhcloud.com. (A)]
Installer exit with code 1
```
Unclear how this can happen, as `ci-op-3wqwt9zp-4bc28` should be unique each time.

aravindhp:
Looks like the job was interrupted during install: [1703828385496567808](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/1810/pull-ci-openshift-windows-machine-config-operator-master-aws-e2e-upgrade/1703828385496567808) after creating the record set.

jrvaldes:
/test aws-e2e-upgrade

Cluster install failed.

sebsoto:
/test nutanix-e2e-operator
nutanix job is stuck in pending

PR #1821, opened 2023-09-21
[e2e] Wait for deleted machines in Azure

jrvaldes:
This change introduces a wait after changing the private key secret on Azure to check all machines provisioned with the outdated windows-user-data were deleted.

In addition, moves the test that changes private key secret while machine configuration to a separate test.

### DIFF
--- a/test/e2e/create_test.go
+++ b/test/e2e/create_test.go
@@ -160,28 +160,62 @@ func (tc *testContext) testMachineConfiguration(t *testing.T) {
 	}
 	_, err := tc.createWindowsMachineSet(gc.numberOfMachineNodes, false)
 	require.NoError(t, err, "failed to create Windows MachineSet")
-	// We need to cover the case where a user changes the private key secret before the WMCO has a chance to
-	// configure the Machine. In order to simulate that case we need to wait for the MachineSet to be fully
-	// provisioned and then change the key. The correct amount of nodes being configured is proof that the
-	// mismatched Machine created with the mismatched key was deleted and replaced.
-	// Depending on timing and configuration flakes this will either cause all Machines, or all Machines after
-	// the first configured Machines to hit this scenario. This is a platform agonistic test so we run it only on
-	// Azure.
+
+	t.Run("Machine configuration while private key change", tc.testMachineConfigurationWhilePrivateKeyChange)
+
 	machines, err := tc.waitForWindowsMachines(int(gc.numberOfMachineNodes), "Provisioned", false)
 	require.NoError(t, err, "error waiting for Windows Machines to be provisioned")
-	if tc.CloudProvider.GetType() == config.AzurePlatformType {
-		// Replace the known private key with a randomly generated one.
-		err = tc.createPrivateKeySecret(false)
-		require.NoError(t, err, "error replacing private key secret")
-	}
 	err = tc.waitForConfiguredWindowsNodes(gc.numberOfMachineNodes, false, false)
 	assert.NoError(t, err, "Windows node creation failed")
-	if tc.CloudProvider.GetType() == config.AzurePlatformType {
-		// Update the machines list
-		machines, err = tc.waitForWindowsMachines(int(gc.numberOfMachineNodes), "Running", false)
-		require.NoError(t, err, "error waiting for Windows Machines to be running")
-	}
 	tc.machineLogCollection(machines.Items)
+}
+
+// testMachineConfigurationWhilePrivateKeyChange tests that machines which have not yet been configured by WMCO are
+// deleted after the private key is changed, but before WMCO is able to configure them, resulting in WMCO getting an
+// SSH authentication error. This could be considered a platform-agnostic test (except for vSphere where the private
+// key is baked in the VM template) so we run it only on Azure.
+func (tc *testContext) testMachineConfigurationWhilePrivateKeyChange(t *testing.T) {
+	if tc.CloudProvider.GetType() != config.AzurePlatformType {
+		t.Skip("test disabled, exclusively runs on Azure")
+	}
+	machines, err := tc.waitForWindowsMachines(int(gc.numberOfMachineNodes), "Provisioned", false)
+	require.NoError(t, err, "error waiting for Windows Machines to be provisioned")
+
+	err = tc.createPrivateKeySecret(false)
+	require.NoError(t, err, "error replacing private key secret")
+
+	err = tc.waitForMachinesDeleted(machines.Items)
+	require.NoError(t, err, "error waiting for machines deletion after private key secret change")
+}
+
+// waitForMachinesDeleted waits for the given list of machines to be deleted
+func (tc *testContext) waitForMachinesDeleted(machines []mapi.Machine) (err error) {
+	// This is the maximum amount of time for the deletion of all machines in Azure
+	deletionTimeout := time.Minute * 15
+	for _, m := range machines {
+		log.Printf("waiting (timeout: %s) for machine %s to be deleted", deletionTimeout.String(), m.GetName())
+		err = wait.PollUntilContextTimeout(context.TODO(), retry.ResourceChangeTimeout, deletionTimeout, false,
+			func(ctx context.Context) (done bool, err error) {
+				_, err = tc.client.Machine.Machines(clusterinfo.MachineAPINamespace).Get(context.TODO(), m.GetName(),
+					metav1.GetOptions{})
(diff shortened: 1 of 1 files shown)


test/e2e/create_test.go
```diff
@@ -160,30 +160,77 @@ func (tc *testContext) testMachineConfiguration(t *testing.T) {
...
+			for _, toDelete := range machinesToDelete {
+				for _, current := range currentMachines.Items {
+					if current.GetName() == toDelete.GetName() {
+						log.Printf("waiting for machine %s to be deleted", toDelete.GetName())
+						return false, nil
+					}
+				}
+			}
+			// all machines were deleted
+			return true, nil
```
sebsoto, line 230: This approach is somewhat over-engineering things.
You can check if a Machine has been deleted by
```
func (tc* testContext) machineDeleted(name string) bool {
  currentMachines, err := tc.client.Machine.Machines(clusterinfo.MachineAPINamespace).Get(context.TODO(), name, getOptions)
  return apierrors.IsNotFound(err)
}
``` 
You can loop this for all the machines you are waiting for.

Your approach makes 1 api call instead of N api calls, but also complicates the code. Consider that in the context this function is used, this is:
1)  test code, so it doesn't need to be as performant as code within WMCO
2) When the tests are running, 100% of the time in CI the amount of Machines to wait for will be 1. And most of the time when ran by developers it will be 1, with the occasional 0 or 2.

test/e2e/create_test.go
```diff
@@ -160,30 +160,77 @@ func (tc *testContext) testMachineConfiguration(t *testing.T) {
...
+	require.NoError(t, err, "error waiting for machines deletion after private key secret change")
+}
+
+// waitForMachinesDeleted waits for the given list of machines to be deleted
+func (tc *testContext) waitForMachinesDeleted(machinesToDelete []mapi.Machine) error {
+	if len(machinesToDelete) == 0 {
+		return nil
+	}
+	// deletion of machines takes time, no need to poll too frequently
+	deletionRetryInterval := time.Minute * 1
```
sebsoto, line 203: Does this need to be defined? Are the constants in the WMCO retry package not sufficient?

test/e2e/create_test.go
```diff
@@ -160,30 +160,77 @@ func (tc *testContext) testMachineConfiguration(t *testing.T) {
...
+	deletionTimeLimit := time.Duration(len(machinesToDelete)) * time.Minute * 10
```
sebsoto, line 204: Why does the wait need to be multiplied?
Why not just 10 minutes?

test/e2e/create_test.go
```diff
@@ -160,30 +160,77 @@ func (tc *testContext) testMachineConfiguration(t *testing.T) {
...
 }
 
+// testMachineConfigurationWhilePrivateKeyChange tests that machines are deleted after the private key change.
+// We need to cover the case where a user changes the private key secret before the WMCO has a chance to
+// configure the Machine. In order to simulate that case we need to wait for the MachineSet to be fully
+// provisioned and then change the key. The correct amount of nodes being configured is proof that the
+// mismatched Machine created with the mismatched key was deleted and replaced.
+// Depending on timing and configuration flakes this will either cause all Machines, or all Machines after
+// the first configured Machines to hit this scenario. This could be considered a platform-agnostic test, except for
+// vSphere where the private key is baked in the VM template, so we run it only on Azure.
```
sebsoto, line 180: This doc is a bit overwhelming :smile:
Moving it to the function level, we don't need to go into too much detail, as it might actually make it less clear to the caller what is actually happening.

I suggest trimming it down to something like:
```
// testMachineConfigurationWhilePrivateKeyChange tests that machines which have not yet been configured by WMCO are deleted after the private key is changed, but before WMCO is able to configure them, resulting in WMCO getting an SSH authentication error
```

test/e2e/create_test.go
```diff
@@ -160,30 +160,77 @@ func (tc *testContext) testMachineConfiguration(t *testing.T) {
...
 	require.NoError(t, err, "failed to create Windows MachineSet")
-	// We need to cover the case where a user changes the private key secret before the WMCO has a chance to
-	// configure the Machine. In order to simulate that case we need to wait for the MachineSet to be fully
-	// provisioned and then change the key. The correct amount of nodes being configured is proof that the
-	// mismatched Machine created with the mismatched key was deleted and replaced.
-	// Depending on timing and configuration flakes this will either cause all Machines, or all Machines after
-	// the first configured Machines to hit this scenario. This is a platform agonistic test so we run it only on
-	// Azure.
+
+	t.Run("Machine configuration while private key change", tc.testMachineConfigurationWhilePrivateKeyChange)
```
sebsoto, line 164: Thanks for making this its own test

sebsoto (changes requested):
Thanks for opening this PR @jrvaldes, PTAL at my comments

test/e2e/create_test.go
jrvaldes: I proposed a machine-count aware timeout to be on the safe side while using more than 2 machines in the test suite, which has a low prob so fixed as suggested.

test/e2e/create_test.go
jrvaldes: Agree, addressed.

jrvaldes:
Added https://github.com/openshift/windows-machine-config-operator/pull/1821/commits/f0bc50e5db366d4b2fb5df35319a8b29b76f6e77 for testing only and will drop it after the `azure-e2e-operator` passes as validation of the timing [issue](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/1770/pull-ci-openshift-windows-machine-config-operator-master-azure-e2e-operator/1704345828228861952) is not present.

pkg/windows/windows.go
```diff
@@ -2,6 +2,7 @@ package windows
```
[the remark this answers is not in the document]
jrvaldes: See https://github.com/openshift/windows-machine-config-operator/pull/1821#issuecomment-1732986640

test/e2e/create_test.go
```diff
@@ -160,30 +160,66 @@ func (tc *testContext) testMachineConfiguration(t *testing.T) {
...
+func (tc *testContext) testMachineConfigurationWhilePrivateKeyChange(t *testing.T) {
+	if tc.CloudProvider.GetType() != config.AzurePlatformType {
+		t.Skip("test disabled, exclusively runs on Azure")
+	}
+	machines, err := tc.waitForWindowsMachines(int(gc.numberOfMachineNodes), "Provisioned", false)
+	require.NoError(t, err, "error waiting for Windows Machines to be provisioned")
+
+	// Replace the known private key with a randomly generated one.
+	err = tc.createPrivateKeySecret(false)
+	require.NoError(t, err, "error replacing private key secret")
```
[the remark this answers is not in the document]
jrvaldes: Comment removed.

test/e2e/create_test.go
```diff
@@ -160,30 +160,66 @@ func (tc *testContext) testMachineConfiguration(t *testing.T) {
...
+
+	// Wait for machines to be reconciled by deletion
```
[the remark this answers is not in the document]
jrvaldes: Comment removed.

jrvaldes:
> Added [f0bc50e](https://github.com/openshift/windows-machine-config-operator/commit/f0bc50e5db366d4b2fb5df35319a8b29b76f6e77) for testing only and will drop it after the `azure-e2e-operator` passes as validation of the timing [issue](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/1770/pull-ci-openshift-windows-machine-config-operator-master-azure-e2e-operator/1704345828228861952) is not present.


 `azure-e2e-operator`  job  [passed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/1821/pull-ci-openshift-windows-machine-config-operator-master-azure-e2e-operator/1706285165321392128)  with  [f0bc50e](https://github.com/openshift/windows-machine-config-operator/commit/f0bc50e5db366d4b2fb5df35319a8b29b76f6e77); commit dropped.

sebsoto (approved):

saifshaikh48 (commented):
/lgtm

Nice fix

aravindhp (approved):
/lgtm

PR #1800, opened 2023-09-15
OCPBUGS-19502: Enable proxy removal test

mansikulkarni96:
This PR enables running proxy removal test which was disabled due to its unreliable behavior.
It has now been modified to wait for the reboot annotation to be applied and to be removed before running the tests.

PR #1823, opened 2023-09-22
OCPBUGS-19716: Enable proxy cert test to run in CI

saifshaikh48:
Proxy tests were not actually running due to an issue with the proxy enabled check.
When enabled, a few of the tests were failing so one was patched and the other was
disabled with the fix coming in another PR (https://github.com/openshift/windows-machine-config-operator/pull/1800).

### DIFF
--- a/test/e2e/clusterinfo/openshift.go
+++ b/test/e2e/clusterinfo/openshift.go
@@ -93,3 +93,12 @@ func (o *OpenShift) GetInfrastructure() (*config.Infrastructure, error) {
 	}
 	return infra, nil
 }
+
+// ProxyEnabled queries the Proxy resource to see if a cluster-wide proxy is enabled in this environment
+func (o *OpenShift) ProxyEnabled() (bool, error) {
+	clusterProxy, err := o.Config.ConfigV1().Proxies().Get(context.TODO(), "cluster", meta.GetOptions{})
+	if err != nil {
+		return false, err
+	}
+	return clusterProxy.Status.HTTPProxy != "" || clusterProxy.Status.HTTPSProxy != "", nil
+}
--- a/test/e2e/proxy_test.go
+++ b/test/e2e/proxy_test.go
@@ -121,6 +121,9 @@ func (tc *testContext) testEnvVars(t *testing.T) {
 // testEnvVarRemoval tests that on each node the system-level and the process-level environment variables
 // are unset when the cluster-wide proxy is disabled by patching the proxy variables in the cluster proxy object.
 func (tc *testContext) testEnvVarRemoval(t *testing.T) {
+	// This test is flaking consistently after we fix the way we check for proxy enabled in the e2e tests.
+	// Until we address https://issues.redhat.com/browse/OCPBUGS-19502, this test has been temporarily disabled.
+	t.Skip("Env var removal validation test is disabled, pending a fix")
 	var patches []*patch.JSONPatch
 	patches = append(patches, patch.NewJSONPatch("remove", "/spec/httpProxy", "httpProxy"),
 		patch.NewJSONPatch("remove", "/spec/httpsProxy", "httpsProxy"))
--- a/test/e2e/proxy_test.go
+++ b/test/e2e/proxy_test.go
@@ -19,7 +19,6 @@ import (
 
 	"github.com/openshift/windows-machine-config-operator/controllers"
 	"github.com/openshift/windows-machine-config-operator/pkg/certificates"
-	"github.com/openshift/windows-machine-config-operator/pkg/cluster"
 	"github.com/openshift/windows-machine-config-operator/pkg/patch"
 	"github.com/openshift/windows-machine-config-operator/pkg/retry"
 	"github.com/openshift/windows-machine-config-operator/pkg/windows"
@@ -28,12 +27,15 @@ import (
 // proxyTestSuite contains the validation cases for cluster-wide proxy.
 // All subtests are skipped if a proxy is not enabled in the test environment.
 func proxyTestSuite(t *testing.T) {
-	if !cluster.IsProxyEnabled() {
+	tc, err := NewTestContext()
+	require.NoError(t, err)
+
+	proxyEnabled, err := tc.client.ProxyEnabled()
+	require.NoErrorf(t, err, "error checking if proxy is enabled in test environment")
+	if !proxyEnabled {
 		t.Skip("cluster-wide proxy is not enabled in this environment")
 	}
 
-	tc, err := NewTestContext()
-	require.NoError(t, err)
 	// Enables proxy test suite to be run individually on existing Windows nodes
 	require.NoError(t, tc.loadExistingNodes())
 
--- a/test/e2e/proxy_test.go
+++ b/test/e2e/proxy_test.go
@@ -1,13 +1,22 @@
 package e2e
 
 import (
+	"bytes"
 	"context"
+	"crypto/rand"
+	"crypto/rsa"
 	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/base64"
 	"encoding/json"
+	"encoding/pem"
 	"fmt"
+	"math/big"
+	"net"
 	"strconv"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -22,6 +31,12 @@ import (
 	"github.com/openshift/windows-machine-config-operator/pkg/patch"
 	"github.com/openshift/windows-machine-config-operator/pkg/retry"
 	"github.com/openshift/windows-machine-config-operator/pkg/windows"
+)
+
+const (
+	// userCABundleName is the name of the ConfigMap that holds additional user-provided proxy certs
+	userCABundleName      = "user-ca-bundle"
+	userCABundleNamespace = "openshift-config"
 )
 
 // proxyTestSuite contains the validation cases for cluster-wide proxy.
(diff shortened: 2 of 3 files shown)


saifshaikh48:

cherry-picked from https://github.com/openshift/windows-machine-config-operator/pull/1800 which had green CI. let's see if there is a timing issue in the tests

test/e2e/proxy_test.go
```diff
@@ -48,30 +50,41 @@ func (tc *testContext) testCerts(t *testing.T) {
...
+			for scanner.Scan() {
+				// Multi-line certificate data causes issues in the command. Encode to base64 as a workaround
+				expectedCertBase64 := base64.StdEncoding.EncodeToString([]byte(scanner.Text()))
+				commandToRun := fmt.Sprintf("$base64Data=\\\"%s\\\";"+
+					// Decode base64 into cert's actual string data
+					"$certString=[Text.Encoding]::Utf8.GetString([Convert]::FromBase64String($base64Data));"+
+					// Create a Powershell certificate object with the expected cert.
+					// First requires data to be written to a file and then provide the file path the cert constructor
+					"Set-Content C:\\Temp\\cert.pem $certString;"+
+					"$expectedCert=[System.Security.Cryptography.X509Certificates.X509Certificate2]::new(\\\"C:\\Temp\\cert.pem\\\");"+
```
saifshaikh48, line 77: I also tried with passing the cert byte data directly to the `X509Certificate2` constructor and, interestingly enough, reading from a file was consistently speedier.

jrvaldes (commented):
@saifshaikh48  thanks for working on this, PTAL at my comments.

hack/run-ci-e2e-test.sh
```diff
@@ -160,7 +160,7 @@ if [[ "$TEST" = "all" || "$TEST" = "basic" ]]; then
   printf "\n####### Testing service reconciliation #######\n" >> "$ARTIFACT_DIR"/wmco.log
   go test ./test/e2e/... -run=TestWMCO/service_reconciliation -v -timeout=20m -args $GO_TEST_ARGS
   printf "\n####### Testing cluster-wide proxy #######\n" >> "$ARTIFACT_DIR"/wmco.log
-  go test ./test/e2e/... -run=TestWMCO/cluster-wide_proxy -v -timeout=10m -args $GO_TEST_ARGS
+  go test ./test/e2e/... -run=TestWMCO/cluster-wide_proxy -v -timeout=45m -args $GO_TEST_ARGS
```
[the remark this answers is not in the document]
aravindhp: Do you know what part of the test is taking a long time to run?

test/e2e/proxy_test.go
```diff
@@ -38,8 +40,8 @@ func proxyTestSuite(t *testing.T) {
 	require.NoError(t, tc.loadExistingNodes())
 
 	t.Run("Trusted CA ConfigMap validation", tc.testTrustedCAConfigMap)
-	t.Run("Environment variables validation", tc.testEnvVars)
 	t.Run("Certificate validation", tc.testCerts)
+	t.Run("Environment variables validation", tc.testEnvVars)
```
aravindhp, line 46: Please add a comment here about the ordering requirement

test/e2e/proxy_test.go
```diff
@@ -48,30 +50,41 @@ func (tc *testContext) testCerts(t *testing.T) {
...
 	for _, node := range gc.allNodes() {
 		t.Run(node.GetName(), func(t *testing.T) {
 			addr, err := controllers.GetAddress(node.Status.Addresses)
 			require.NoError(t, err, "unable to get node address")
 
-			for i, subjectBytes := range subjects {
-				command := fmt.Sprintf("(Get-ChildItem -Path Cert:\\LocalMachine\\Root | "+
-					"Where-Object {$_.Subject -eq '%s'}).Count", string(subjectBytes))
-				out, err := tc.runPowerShellSSHJob(fmt.Sprintf("get-cert-%d", i), command, addr)
+			// Read in one cert at a time and test it exists in the Windows instance's system store
```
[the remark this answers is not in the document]
aravindhp: I agree with @jrvaldes's comment. Alternatively you could mount the expected certs into the Pod and do all the comparison in PowerShell.

aravindhp (changes requested):

hack/run-ci-e2e-test.sh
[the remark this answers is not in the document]
saifshaikh48: average runtime is 25-30min, cert test is the one that takes a long time as CNO injects 100+ certs into the proxy certs CM. Checking each one on each Win node takes a long while

test/e2e/proxy_test.go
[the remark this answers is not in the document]
saifshaikh48: > fetching all the relevant certs in Windows instance's system store at once and performing the comparison in the e2e code

I tried my best for days to do this but the way certs are stored in Windows OS is just not compatible with their representation in any golang object. The cert subject gets rearranged (and special characters get mangled) so not easily translated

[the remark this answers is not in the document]
saifshaikh48: > Alternatively you could mount the expected certs into the Pod and do all the comparison in PowerShell

Does this imply having a single k8s job (one call to `runPowerShellSSHJob`) and parsing the expected cert bundle through powershell? I briefly explored that and debugging was a nightmare so I thought unless we start hitting the e2e test timeout, it was not worth it for now. Definitely a potential improvement in the future for which I can make a story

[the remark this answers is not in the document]
aravindhp: But can't you mount the trustedCABundle into the Job Pod and do `splitAtPEMCert()` in PowerShell allowing you to do the comparison on the Windows node itself?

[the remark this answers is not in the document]
saifshaikh48: After discussion in the meeting we decided to keep the test logic the same for now but test only the user-provided certs to keep the test duration reasonable. Follow-up story to ensure all certs are imported, with more efficient logic, has been created here https://issues.redhat.com/browse/WINC-1144

saifshaikh48:

Green locally :rocket:

test/e2e/proxy_test.go
```diff
@@ -19,59 +29,78 @@ import (
...
+					"$expectedCert=[System.Security.Cryptography.X509Certificates.X509Certificate2]::new(\\\"C:\\Temp\\cert.pem\\\");"+
+					// Get the number of existing certs equivalent to the expected cert
+					"(Get-ChildItem -Path Cert:\\LocalMachine\\Root | Where-Object {$expectedCert.Equals($_)}).Count",
+					expectedCertBase64)
+				out, err := tc.runPowerShellSSHJob(fmt.Sprintf("get-cert-%d", i), commandToRun, addr)
 				if err != nil {
 					require.NoError(t, err, "error running SSH job: %w", err)
 				}
-				count, err := strconv.Atoi(strings.TrimSpace(out))
+				count, err := strconv.Atoi(finalLine(out))
```
aravindhp, line 99: Please add a comment saying what you are expecting to see in the final line.

aravindhp (changes requested):
>[proxy_test] Disable env var removal test 

Lets be consistent with commit naming. `[tests] Disable proxy env var removal`

test/e2e/main_test.go
```diff
@@ -88,6 +88,8 @@ type testContext struct {
 	// windowsServerVersion is the Windows Server version used in the e2e test suite.
 	// If unset or empty, Windows Server 2022 is assumed for Machine and container images.
 	windowsServerVersion windows.ServerVersion
+	// proxyEnabled indicates whether a cluster-wide proxy is in use in the test environment
+	proxyEnabled bool
```
sebsoto, line 92: Why should this be added to the testContext struct?
Based on the way this is used I suggest just calling `testContext.client.IsProxyEnabled()` instead.
Caching it in this way doesn't seem useful enough to justify, and could cause problems if the proxy status is changed after testContext is initialized.

test/e2e/clusterinfo/openshift.go
```diff
@@ -93,3 +93,12 @@ func (o *OpenShift) GetInfrastructure() (*config.Infrastructure, error) {
 	}
 	return infra, nil
 }
+
+// IsProxyEnabled queries the Proxy resource to see if a cluster-wide proxy is enabled in this environment
+func (o *OpenShift) IsProxyEnabled() (bool, error) {
```
sebsoto, line 98: nit: consider renaming this to `ProxyEnabled`, just personal preference as it seems more natural to me:
```
if o.ProxyEnabled{}
vs
if o.IsProxyEnabled{}
```

test/e2e/proxy_test.go
```diff
@@ -245,6 +261,87 @@ func (tc *testContext) getEnvVar(addr, name, command string) (map[string]string,
...
+	if err != nil {
+		return fmt.Errorf("error creating user-provided CA ConfigMap: %w", err)
+	}
+	return nil
+}
+
+// patchProxyTrustedCA patches the proxy to use the user-provided CA bundle, which CNO will merge with other proxy certs
+func (tc *testContext) patchProxyTrustedCA() error {
+	var patches []*patch.JSONPatch
+	patches = append(patches, patch.NewJSONPatch("replace", "/spec/trustedCA/name", userCABundleName))
```
sebsoto, line 312:
```
patches := []*patch.JSONPatch{patch.NewJSONPatch("replace", "/spec/trustedCA/name", userCABundleName)}
```

test/e2e/proxy_test.go
```diff
@@ -53,40 +53,55 @@ func proxyTestSuite(t *testing.T) {
...
 		t.Run(node.GetName(), func(t *testing.T) {
 			addr, err := controllers.GetAddress(node.Status.Addresses)
 			require.NoError(t, err, "unable to get node address")
 
-			for i, subjectBytes := range subjects {
-				command := fmt.Sprintf("(Get-ChildItem -Path Cert:\\LocalMachine\\Root | "+
-					"Where-Object {$_.Subject -eq '%s'}).Count", string(subjectBytes))
-				out, err := tc.runPowerShellSSHJob(fmt.Sprintf("get-cert-%d", i), command, addr)
+			// Read in one cert at a time and test it exists in the Windows instance's system store
+			scanner := bufio.NewScanner(strings.NewReader(trustedCABundle))
```
sebsoto, line 79: Have you looked at https://pkg.go.dev/encoding/pem#Decode?
It looks like this functionality might be built in.
```
    for block, rest := pem.Decode(data); block != nil; block, rest = pem.Decode(rest) {
        if  block.Type == "CERTIFICATE"{
```
Bringing this up because this code + splitAtPEMCert() seems to be duplicated from the WICD code. Keeping things simpler here and using stdlib functions might make things easier to maintain.

Not sure if decoding the actual data causes a problem for this. Re-encoding the individual blocks with `pem.Encode` before doing the comparison is always an option :)

sebsoto (changes requested):

test/e2e/proxy_test.go
saifshaikh48: Good point, I initially was using pem.Decode in the source code but had to move away from that approach to avoid reading an arbitrary number of certs into memory. I don't see any issue using it in the test code (where we are controlling the # of certs)

test/e2e/proxy_test.go
```diff
@@ -28,12 +27,15 @@ import (
 // proxyTestSuite contains the validation cases for cluster-wide proxy.
 // All subtests are skipped if a proxy is not enabled in the test environment.
 func proxyTestSuite(t *testing.T) {
-	if !cluster.IsProxyEnabled() {
+	tc, err := NewTestContext()
+	require.NoError(t, err)
+
+	proxyEnabled, err := tc.client.ProxyEnabled()
+	require.NoErrorf(t, err, "error checking if proxy is enabled in test environment")
+	if !proxyEnabled {
```
aravindhp, line 50: Should we reference OCPBUGS-19502 in this PR and close https://github.com/openshift/windows-machine-config-operator/pull/1800?

aravindhp (commented):
LGTM from me. I will let @sebsoto give the approval given he had requested changes.

test/e2e/proxy_test.go
mansikulkarni96: I am repurposing #1800 to enable the removal validation test disabled in this PR. I will reframe the PR and ticket associated.

PR #1834, opened 2023-09-26
WINC-688: Support removing proxy certs from Windows nodes

saifshaikh48:
WICD controller reacts to changes in the trusted CA bundle that involve
a certificate removal (cert rotation, user disabling cluster-wide proxy, etc). 
Certificates that are removed from the trusted CA bundle are in turn
removed from each Windows node's local trust store and processes. 
This is also the case when nodes are deconfigured. With this, no outgoing
traffic from the node will try to use removed certificates.

### DIFF
--- a/pkg/daemon/certs/certs.go
+++ b/pkg/daemon/certs/certs.go
@@ -25,7 +25,9 @@ import (
 	"encoding/pem"
 	"errors"
 	"fmt"
+	"io"
 	"os"
+	"reflect"
 	"syscall"
 	"unsafe"
 
@@ -33,19 +35,94 @@ import (
 	"k8s.io/klog/v2"
 )
 
-// Reconcile ensures the certificates in the given trusted CA bundle file are imported as system certificates.
-// Returns a boolean whether any certs needed to be reconciled
-func Reconcile(caBundle string) (bool, error) {
-	if caBundle == "" {
-		// TODO: ensure that all certs that we've added previously are removed https://issues.redhat.com/browse/WINC-688
+// importedCABundleFile tracks all the certs that the operator has imported into the node's local trust store
+const importedCABundleFile = "C:\\k\\imported-certs.pem"
+
+// Reconcile ensures the node's system certificates match the expected state as given by the trusted CA bundle file.
+// Returns a boolean whether any certs needed to be reconciled.
+// If this function returns true and a non-nil error, the instance must be rebooted anyway.
+func Reconcile(caBundlePath string) (bool, error) {
+	// caBundlePath will be empty when a cluster-wide proxy is being removed or not in use
+	certChange, err := reconcileCerts(caBundlePath)
+	if err != nil {
+		return false, err
+	}
+	// An error updating the imported bundle file will be self-corrected on the next reconcile, since it is the source
+	// of truth for determining whether certs need to be updated
+	return certChange, updateImportedCABundle(caBundlePath, certChange)
+}
+
+// reconcileCerts reconciles any discrepency between expected and existing Windows certificates by importing or
+// deleting certificates from the root system store. Returns a boolean if any certificates were imported or deleted
+func reconcileCerts(caBundlePath string) (bool, error) {
+	expectedCerts, err := getExpectedCerts(caBundlePath)
+	if err != nil {
+		return false, err
+	}
+	existingCerts, err := getExistingCerts()
+	if err != nil {
+		return false, err
+	}
+
+	if reflect.DeepEqual(expectedCerts, existingCerts) {
 		return false, nil
 	}
-	// Read expected certs from CA trust bundle file
-	expectedCerts, err := readCertsFromFile(caBundle)
+
+	// Open the root certificate store
+	systemStore, err := windows.CertOpenStore(windows.CERT_STORE_PROV_SYSTEM, 0, 0,
+		windows.CERT_SYSTEM_STORE_LOCAL_MACHINE, uintptr(unsafe.Pointer(syscall.StringToUTF16Ptr("ROOT"))))
 	if err != nil {
-		return false, fmt.Errorf("Failed to read certificate file: %v", err)
+		return false, fmt.Errorf("Failed to open root certificate store: %w", err)
 	}
-	return ensureCertsAreImported(expectedCerts)
+	defer func() {
+		if err := windows.CertCloseStore(systemStore, 0); err != nil {
+			klog.Error("Failed to close root system certificate store")
+		}
+	}()
+
+	certsImported, err := ensureCertsAreImported(systemStore, expectedCerts, existingCerts)
+	if err != nil {
+		return false, fmt.Errorf("Failed to import certificates: %w", err)
+	}
+	certsDeleted, err := ensureCertsAreDeleted(systemStore, expectedCerts, existingCerts)
+	if err != nil {
+		return false, fmt.Errorf("Failed to delete certificates: %w", err)
+	}
+	return (certsImported || certsDeleted), nil
+}
+
+// getExpectedCerts reads expected state of certs from the CA bundle file. Returns an empty slice if no proxy is in use.
+func getExpectedCerts(path string) ([]*x509.Certificate, error) {
+	var expectedCerts []*x509.Certificate
+	var err error
+	if path == "" {
+		return expectedCerts, nil
+	}
+
+	if expectedCerts, err = readCertsFromFile(path); err != nil {
+		return nil, fmt.Errorf("Failed to read certificate file %s: %w", path, err)
+	}
+	return expectedCerts, nil
+}
+
+// getExistingCerts populates existing certs using the previously imported CA trust bundle
+func getExistingCerts() ([]*x509.Certificate, error) {
+	var existingCerts []*x509.Certificate
+	exists, err := fileExists(importedCABundleFile)
+	if err != nil {
+		// If there is an error here due to file corruption, the file will get deleted and re-created when the node is
+		// reconfigured. This has the potential to leave stale certificates behind on the node.
+		return nil, err
+	}
+	// The file will not exist on first reconcile in which case existingCerts will be empty.
+	if !exists {
+		return existingCerts, nil
+	}
+
+	if existingCerts, err = readCertsFromFile(importedCABundleFile); err != nil {
+		return nil, fmt.Errorf("Failed to read certificate file %s: %w", importedCABundleFile, err)
+	}
+	return existingCerts, nil
 }
 
 // readCertsFromFile reads in any PEM-encoded certificates from the given file
@@ -106,28 +183,76 @@ func splitAtPEMCert() func(data []byte, atEOF bool) (advance int, token []byte, 
 	}
 }
 
-// ensureCertsAreImported imports the expected certificates into the instance's root system trust store, if not already
-// present. Returns a boolean if any certificates were imported
-func ensureCertsAreImported(expectedCerts []*x509.Certificate) (bool, error) {
-	// Open the root certificate store
-	systemStore, err := windows.CertOpenStore(windows.CERT_STORE_PROV_SYSTEM, 0, 0,
-		windows.CERT_SYSTEM_STORE_LOCAL_MACHINE, uintptr(unsafe.Pointer(syscall.StringToUTF16Ptr("ROOT"))))
+// fileExists checks if the file at the given path exists. Returns true if it exists, false if not, and error otherwise.
+func fileExists(path string) (bool, error) {
+	_, err := os.Stat(path)
+	if err == nil {
+		return true, nil
+	} else if errors.Is(err, os.ErrNotExist) {
+		return false, nil
+	} else {
+		return false, newFileIOError(fmt.Errorf("failed to check if file %s exists: %w", path, err))
+	}
+}
+
+// updateImportedCABundle updates the previously imported CA bundle file, if needed.
+func updateImportedCABundle(caBundlePath string, certChange bool) error {
+	if caBundlePath == "" {
+		// When proxy is deconfigured, ensure the file is deleted to avoid leaking certs
+		return ensureFileIsRemoved(importedCABundleFile)
+	}
+	if certChange {
+		klog.Infof("copied file contents from %s to %s", caBundlePath, importedCABundleFile)
+		// When certs have been reconciled, update the previously imported CA trust bundle to hold the new cert bundle
+		return copyFile(caBundlePath, importedCABundleFile)
+	}
+	return nil
+}
+
+// ensureFileIsRemoved deletes the file at the given path. No-op if the file already does not exist
+func ensureFileIsRemoved(path string) error {
+	exists, err := fileExists(path)
 	if err != nil {
-		return false, fmt.Errorf("Failed to open root certificate store: %v", err)
+		return err
+	}
+	if !exists {
+		return nil
+	}
+	if err := os.Remove(path); err != nil {
+		return err
+	}
+	klog.Infof("removed file %s", path)
+	return nil
+}
+
+// copyFile copies the file contents from the source file to the destination. Creates the destination file if needed.
+func copyFile(src, dst string) error {
+	r, err := os.Open(src)
+	if err != nil {
+		return err
 	}
 	defer func() {
-		if err := windows.CertCloseStore(systemStore, 0); err != nil {
-			klog.Error("Failed to close root system certificate store")
+		if err := r.Close(); err != nil {
+			klog.Errorf("Failed to close file %s", r.Name())
 		}
 	}()
-
-	// Get all existing from system store
-	existingCerts, err := getAllCerts(systemStore)
+	w, err := os.Create(dst)
 	if err != nil {
-		return false, fmt.Errorf("Failed to read certificate file: %v", err)
+		return err
 	}
+	defer func() {
+		if err := w.Close(); err != nil {
+			klog.Errorf("Failed to close file %s", w.Name())
+		}
+	}()
+	_, err = io.Copy(w, r)
+	return err
+}
 
-	certChange := false
+// ensureCertsAreImported imports the expected certificates into the instance's root system trust store, if not already
+// present. Returns a boolean if any certificates were imported
+func ensureCertsAreImported(store windows.Handle, expectedCerts, existingCerts []*x509.Certificate) (bool, error) {
+	certImported := false
 	for _, cert := range expectedCerts {
 		if containsCert(existingCerts, cert) {
 			// Cert already exists as expected, do nothing
@@ -135,40 +260,28 @@ func ensureCertsAreImported(expectedCerts []*x509.Certificate) (bool, error) {
 		}
 
 		// Add the certificate to the instance's system-wide trust store
-		if err := addCertToStore(systemStore, cert); err != nil {
+		if err := addCertToStore(store, cert); err != nil {
 			return false, err
 		}
-		certChange = true
+		certImported = true
 	}
-	return certChange, nil
+	return certImported, nil
 }
 
-// getAllCerts reads all the certs from the given certificate store
-func getAllCerts(store windows.Handle) ([]*x509.Certificate, error) {
-	var existingSystemCerts []*x509.Certificate
-	var certContext *windows.CertContext
-	var err error
-	for {
-		certContext, err = windows.CertEnumCertificatesInStore(store, certContext)
-		if err != nil {
-			if errors.Is(err, windows.Errno(windows.CRYPT_E_NOT_FOUND)) {
-				// Error code implies we have read all certs:
-				// https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-certenumcertificatesinstore
-				break
+// ensureCertsAreDeleted deletes any existing certificates that are not in the expected list.
+// Returns a boolean if any certificates were deleted.
+func ensureCertsAreDeleted(store windows.Handle, expectedCerts, existingCerts []*x509.Certificate) (bool, error) {
+	certDeleted := false
+	for _, cert := range existingCerts {
+		// Any imported previously certs that are NOT in the new cert bundle should be removed
+		if !containsCert(expectedCerts, cert) {
+			if err := removeCertFromStore(store, cert); err != nil {
+				return false, err
 			}
-			return nil, fmt.Errorf("Error reading existing system certificate: %v", err)
+			certDeleted = true
 		}
-
-		// Convert the certificate bytes to Golang x509.Certificate
-		certBytes := unsafe.Slice(certContext.EncodedCert, certContext.Length)
-		x509Cert, err := x509.ParseCertificate(certBytes)
-		if err != nil {
-			return nil, fmt.Errorf("Failed to parse certificate from bytes: %v", err)
-		}
-		existingSystemCerts = append(existingSystemCerts, x509Cert)
 	}
-	klog.Infof("found %d existing system certificates", len(existingSystemCerts))
-	return existingSystemCerts, nil
+	return certDeleted, nil
 }
 
 // containsCert checks if the target certificate exists in the given slice
@@ -195,4 +308,69 @@ func addCertToStore(store windows.Handle, cert *x509.Certificate) error {
 	}
 	klog.Infof("Certificate %s imported successfully.", cert.Subject.String())
 	return nil
+}
+
+// removeCertFromStore removes the given certificate from its associated certificate store
+func removeCertFromStore(store windows.Handle, cert *x509.Certificate) error {
+	// Windows CertContext holds the store which imported the cert, which is needed when deleting from the system store
+	certContext, err := retrieveCertContext(store, cert)
+	if err != nil {
+		return fmt.Errorf("failed to create content from x509 cert %s: %w", cert.Subject.String(), err)
+	}
+	if certContext == nil {
+		// Cert already doesn't exist, nothing to do
+		return nil
+	}
+
+	if err = windows.CertDeleteCertificateFromStore(certContext); err != nil {
+		return fmt.Errorf("failed to remove certificate %s: %w", cert.Subject.String(), err)
+	}
+	klog.V(1).Infof("Certificate %s removed successfully.", cert.Subject.String())
+	return nil
+}
+
+// retrieveCertContext finds the given certificate in the given store and returns its full context per Windows OS
+// Returns a nil CertContext if the cert is not found.
+func retrieveCertContext(store windows.Handle, targetCert *x509.Certificate) (*windows.CertContext, error) {
+	var certContext *windows.CertContext
+	var err error
+	for {
+		certContext, err = windows.CertEnumCertificatesInStore(store, certContext)
+		if err != nil {
+			if errors.Is(err, windows.Errno(windows.CRYPT_E_NOT_FOUND)) {
+				// Error code implies we have read all certs:
+				// https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-certenumcertificatesinstore
+				break
+			}
+			return nil, fmt.Errorf("error reading existing system certificate: %w", err)
+		}
+
+		// Convert the certificate bytes to Golang x509.Certificate
+		certBytes := unsafe.Slice(certContext.EncodedCert, certContext.Length)
+		current, err := x509.ParseCertificate(certBytes)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse certificate from bytes: %w", err)
+		}
+
+		if current.Equal(targetCert) {
+			klog.Infof("found %s existing system certificate", certContext.CertInfo.Subject)
+			return certContext, nil
+		}
+	}
+	klog.Infof("target cert not found in root system store: %s", targetCert.Subject)
+	return nil, nil
+}
+
+// FileIOError occurs when there is an failure interacting with the Windows filesystem
+type FileIOError struct {
+	err error
+}
+
+func (e *FileIOError) Error() string {
+	return e.err.Error()
+}
+
+// newFileIOError returns a new FileIOError
+func newFileIOError(err error) *FileIOError {
+	return &FileIOError{err: err}
 }
--- a/pkg/daemon/cleanup/cleanup.go
+++ b/pkg/daemon/cleanup/cleanup.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/openshift/windows-machine-config-operator/pkg/daemon/certs"
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/controller"
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/envvar"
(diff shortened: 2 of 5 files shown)


pkg/daemon/certs/certs.go
```diff
@@ -26,26 +26,63 @@ import (
...
-	expectedCerts, err := readCertsFromFile(caBundle)
+
+	// Populate existing certs using the previously imported CA trust bundle.
+	pwd, err := os.Getwd()
 	if err != nil {
-		return false, fmt.Errorf("Failed to read certificate file: %v", err)
+		return false, err
 	}
-	return ensureCertsAreImported(expectedCerts)
+	previousCABundlePath := fmt.Sprintf("%s\\%s", pwd, previousCABundleFileName)
```
saifshaikh48: I could just make this a well known location to WICD instead, since it creates/deletes the file. LMK folks

pkg/daemon/certs/certs.go
```diff
@@ -26,26 +26,63 @@ import (
...
+	klog.Infof("previousCABundlePath: %s", previousCABundlePath)
```
saifshaikh48, line 59: will remove my debug logs later :)

pkg/daemon/certs/certs.go
```diff
@@ -26,26 +26,63 @@ import (
...
+		}
+		klog.Infof("read %d expected Certs from %s", len(expectedCerts), caBundlePath)
+	}
+
+	// Populate existing certs using the previously imported CA trust bundle.
+	pwd, err := os.Getwd()
+	if err != nil {
+		return false, err
+	}
+	previousCABundlePath := fmt.Sprintf("%s\\%s", pwd, previousCABundleFileName)
```
saifshaikh48, line 58: I could just make this a well known location to WICD instead, since it creates/deletes the file. LMK folks

pkg/daemon/certs/certs.go
```diff
@@ -26,26 +26,63 @@ import (
...
+	klog.Infof("previousCABundlePath: %s", previousCABundlePath)
```
saifshaikh48, line 59: will remove my debug logs later :)

pkg/daemon/certs/certs.go
aravindhp, line 1: I suggest prefixing WICD commits with `daemon`

pkg/daemon/certs/certs.go
aravindhp: Please make this a well known location. In addition please use [filepath-securejoin](https://github.com/cyphar/filepath-securejoin?tab=readme-ov-file#filepath-securejoin). This was suggested to me in one of my upstream PRs.

pkg/daemon/certs/certs.go
```diff
@@ -26,26 +26,63 @@ import (
...
+// previousCABundleFileName tracks all the certs that the operator has imported into the node's local trust store
+const previousCABundleFileName = "imported-certs.pem"
+
+// Reconcile ensures the node's system certificates match the expected state as given by the trusted CA bundle file.
 // Returns a boolean whether any certs needed to be reconciled
-func Reconcile(caBundle string) (bool, error) {
-	if caBundle == "" {
-		// TODO: ensure that all certs that we've added previously are removed https://issues.redhat.com/browse/WINC-688
-		return false, nil
+func Reconcile(caBundlePath string) (bool, error) {
```
aravindhp, line 42: Please break this function in sub methods. We should be able to read this method and figure out the overall logic flow, without having to inspect each line of code. Otherwise we will end up with an overly complex `Reconcile()` method.

pkg/daemon/certs/certs.go
```diff
@@ -26,26 +26,63 @@ import (
...
 	"syscall"
 	"unsafe"
 
 	"golang.org/x/sys/windows"
 	"k8s.io/klog/v2"
 )
 
-// Reconcile ensures the certificates in the given trusted CA bundle file are imported as system certificates.
+// previousCABundleFileName tracks all the certs that the operator has imported into the node's local trust store
+const previousCABundleFileName = "imported-certs.pem"
```
aravindhp, line 38: Shouldn't this be `importedCertsFileName` or `importedCABundleFileName`? Or does the filename need to change?

pkg/daemon/certs/certs.go
```diff
@@ -26,26 +26,63 @@ import (
...
+	// Populate existing certs using the previously imported CA trust bundle.
+	pwd, err := os.Getwd()
+	if err != nil {
+		return false, err
+	}
+	previousCABundlePath := fmt.Sprintf("%s\\%s", pwd, previousCABundleFileName)
+	klog.Infof("previousCABundlePath: %s", previousCABundlePath)
+	exists, err := fileExists(previousCABundlePath)
+	if err != nil {
+		return false, err
```
aravindhp, line 62: If there is an error here due to file corruption, what is the remediation path?

pkg/daemon/certs/certs.go
```diff
@@ -106,69 +143,120 @@ func splitAtPEMCert() func(data []byte, atEOF bool) (advance int, token []byte,
...
+	}
+	if exists {
+		return os.Remove(path)
+	}
+	return nil
+}
+
+// ensureCertsAreUpToDate reconciles any discrepency between expected and existing Windows certificates by importing or
+// deleting certificates from the root system store. Returns a boolean if any certificates were imported or deleted
+func ensureCertsAreUpToDate(expectedCerts, existingCerts []*x509.Certificate) (bool, error) {
```
aravindhp, line 196: I suggest naming this `reconcleCerts()`

pkg/daemon/certs/certs.go
```diff
@@ -26,26 +26,63 @@ import (
...
+
+	certChange, err := ensureCertsAreUpToDate(expectedCerts, existingCerts)
 	if err != nil {
-		return false, fmt.Errorf("Failed to read certificate file: %v", err)
+		return false, err
+	}
+
+	// Update previously imported CA trust bundle after all certs have been reconciled
+	if caBundlePath != "" {
+		err = copyFile(caBundlePath, previousCABundlePath)
```
aravindhp, line 79: Does this need to be done always or only when certs are changed?

pkg/daemon/certs/certs.go
```diff
@@ -26,26 +26,63 @@ import (
...
+		klog.Infof("copied file contents from %s to %s", caBundlePath, previousCABundlePath)
+	} else {
+		// To avoid leaking certs, ensure the imported CA trust bundle file is removed when proxy is removed
+		err = ensureFileIsRemoved(previousCABundlePath)
 	}
-	return ensureCertsAreImported(expectedCerts)
+	return certChange, err
```
aravindhp, line 85: How do we recover from the errors here?

pkg/daemon/certs/certs.go
```diff
@@ -106,69 +143,120 @@ func splitAtPEMCert() func(data []byte, atEOF bool) (advance int, token []byte,
...
+	w, err := os.Create(dst)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if err := w.Close(); err != nil {
+			klog.Errorf("Failed to close file %s", w.Name())
+		}
+	}()
+	_, err = w.ReadFrom(r)
```
aravindhp, line 178: Isn't it better to use `io.Copy()` and allow it make the decision to use `ReadFrom()`?

pkg/daemon/certs/certs.go
```diff
@@ -106,69 +143,120 @@ func splitAtPEMCert() func(data []byte, atEOF bool) (advance int, token []byte,
...
-				break
-			}
-			return nil, fmt.Errorf("Error reading existing system certificate: %v", err)
+// ensureCertsAreDeleted deletes any existing certificates that are not in the expected list.
+// Returns a boolean if any certificates were deleted.
+func ensureCertsAreDeleted(store windows.Handle, expectedCerts, existingCerts []*x509.Certificate) (bool, error) {
+	// Any certs that have been imported previously but are NOT in the new cert bundle should be removed
+	certDeleted := false
+	for _, cert := range existingCerts {
+		if containsCert(expectedCerts, cert) {
```
aravindhp, line 249: Shouldn't this be `if !containsCert()`?

aravindhp (changes requested):

pkg/daemon/certs/certs.go
sebsoto: Wouldn't passing in the cert bundle directory via command arguments be the correct thing to do here?

pkg/daemon/certs/certs.go
saifshaikh48: An error here would imply that the `imported-certs.pem` didn't get updated. Since `imported-certs.pem` is one of the sources of truth for determining `certChange`, next reconcile the error would self-correct properly.

pkg/daemon/certs/certs.go
saifshaikh48: good catch, only when certs change

pkg/daemon/certs/certs.go
saifshaikh48: fair point

pkg/daemon/certs/certs.go
saifshaikh48: Here's what would happen:
- error would result in a configuration error thrown to WMCO
- WMCO reacts to such configuration errors by first running WICD cleanup, before re-configuring
- WICD cleanup will delete the corrupt `imported-certs.pem` file
- When WMCO invokes WICD during node re-configuration, everything should be fine

aravindhp: In that case would we potentially leave some stale certs behind on the node?

pkg/daemon/certs/certs.go
aravindhp: Please call this out in a comment. It is not immediately obvious.

pkg/daemon/certs/certs.go
saifshaikh48: Nope, this `if` is deciding when we should do nothing. We're looping over _existing_ certs, so if an existing cert is in the expected list, don't remove it. Bit confusing I know. I'll clarify the misleading comment that's on the following line

ex:
```
existing = [a,b,c]
expected = [a,c]
```
When looking at existing cert `a`, we see `containsCert(expectedCerts, a) = true`, so do nothing. But for cert `b`, `containsCert(expectedCerts, b) = false`, so remove it.

saifshaikh48: alternatively I could flip the condition and change indentation to make it less confusing
```golang
for _, cert := range existingCerts {
	if !containsCert(expectedCerts, cert) {
		if err := removeCertFromStore(store, cert); err != nil {
			return false, err
		}
		certDeleted = true
	}
}
```

pkg/daemon/certs/certs.go
saifshaikh48: Yes, if there were certs in the file when it got corrupted

pkg/daemon/certs/certs.go
saifshaikh48: I don't see the need as this is a file wholly managed by WICD. WMCO need not know about it

pkg/daemon/certs/certs.go
aravindhp: Please add a comment about this

saifshaikh48: added in the helper below

test/e2e/proxy_test.go
```diff
@@ -77,35 +80,114 @@ func (tc *testContext) testCerts(t *testing.T) {
...
+	cm, err := tc.client.K8s.CoreV1().ConfigMaps(userCABundleNamespace).Get(context.TODO(), userCABundleName, meta.GetOptions{})
+	require.NoErrorf(t, err, "error getting user-provided CA ConfigMap: %w", err)
+
+	// Read all expected certs from CM data
+	trustedCABundle := cm.Data[certificates.CABundleKey]
+	assert.Greater(t, len(trustedCABundle), 0, "no additional user-provided certs in bundle")
+
+	require.NoErrorf(t, tc.removeProxyTrustedCA(), "error removing user CA bundle from cluster-wide proxy: %w", err)
+	require.NoErrorf(t, err, "unable to list nodes")
+	require.NoError(t, tc.waitForAllNodesToReboot(gc.allNodes()))
```
mansikulkarni96, line 103: If we are just passing in gc.allNodes here we can get it where needed

pkg/daemon/certs/certs.go
```diff
@@ -106,69 +150,132 @@ func splitAtPEMCert() func(data []byte, atEOF bool) (advance int, token []byte,
...
-				break
+// ensureCertsAreDeleted deletes any existing certificates that are not in the expected list.
+// Returns a boolean if any certificates were deleted.
+func ensureCertsAreDeleted(store windows.Handle, expectedCerts, existingCerts []*x509.Certificate) (bool, error) {
+	certDeleted := false
+	for _, cert := range existingCerts {
+		// Any imported previously certs that are NOT in the new cert bundle should be removed
+		if !containsCert(expectedCerts, cert) {
+			if err := removeCertFromStore(store, cert); err != nil {
+				return false, err
```
aravindhp, line 273: Shouldn't we return `certDeleted` here? I am thinking about the case where there are multiple certs that need to be removed but we error on deleting only one. Or will we reconcile again on error? Same question holds for `ensureCertsAreImported()`

pkg/daemon/certs/certs.go
```diff
@@ -196,3 +303,47 @@ func addCertToStore(store windows.Handle, cert *x509.Certificate) error {
...
+		if err != nil {
+			return nil, fmt.Errorf("Failed to parse certificate from bytes: %w", err)
+		}
+
+		if current.Equal(targetCert) {
+			klog.Infof("found %s existing system certificate", certContext.CertInfo.Subject)
+			return certContext, nil
+		}
+	}
+	return nil, fmt.Errorf("unable to find target cert %s in root system store", targetCert.Subject)
```
aravindhp, line 348: From the error message I take it that the cert we want to delete is not in the system store. Doesn't that imply it is not present? Shouldn't we just log and move on.

aravindhp (changes requested):

pkg/daemon/certs/certs.go
saifshaikh48: We will reconcile again on error. This is consistent with the approach for env vars -- if we fail to read/set/remove one of them, we return error and try the transaction again

pkg/daemon/certs/certs.go
saifshaikh48: Fair point, I'll log instead of returning an error

pkg/daemon/certs/certs.go
```diff
@@ -106,69 +150,132 @@ func splitAtPEMCert() func(data []byte, atEOF bool) (advance int, token []byte,
...
 	if err != nil {
-		return false, fmt.Errorf("Failed to open root certificate store: %v", err)
+		return false, fmt.Errorf("Failed to open root certificate store: %w", err)
 	}
 	defer func() {
 		if err := windows.CertCloseStore(systemStore, 0); err != nil {
 			klog.Error("Failed to close root system certificate store")
 		}
 	}()
+	klog.Infof("opened root cert store")
```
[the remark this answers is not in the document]
saifshaikh48: nope

pkg/daemon/certs/certs.go
```diff
@@ -106,69 +150,132 @@ func splitAtPEMCert() func(data []byte, atEOF bool) (advance int, token []byte,
...
+	} else if errors.Is(err, os.ErrNotExist) {
+		return false, nil
+	} else {
+		return false, fmt.Errorf("Failed to check if file %s exists: %w", path, err)
+	}
+}
+
+// updateImportedCABundle updates the previously imported CA bundle file, if needed.
+func updateImportedCABundle(caBundlePath string, certChange bool) error {
+	if caBundlePath == "" {
```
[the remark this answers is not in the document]
saifshaikh48: yes, in which case ensureFileIsRemoved() will be a no-op

pkg/daemon/certs/certs.go
```diff
@@ -25,27 +25,71 @@ import (
...
+}
+
+// getExistingCerts populates existing certs using the previously imported CA trust bundle
+func getExistingCerts() ([]*x509.Certificate, error) {
+	var existingCerts []*x509.Certificate
+	exists, err := fileExists(importedCABundleFile)
 	if err != nil {
-		return false, fmt.Errorf("Failed to read certificate file: %v", err)
+		// If there is an error here due to file corruption, the file will get deleted and re-created when the node is
+		// reconfigured. This has the potential to leave stale certificates behind on the node.
```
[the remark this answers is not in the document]
saifshaikh48: I think so, [drafted a commit to do this](https://github.com/openshift/windows-machine-config-operator/pull/1834/commits/bbd899fbeeac3020bafb66cf71debd0b3bb5ab32). Let me know what you think

test/e2e/proxy_test.go
```diff
@@ -77,35 +80,113 @@ func (tc *testContext) testCerts(t *testing.T) {
...
+				return false, err
+			}
+			// Add nodes to the seen list based on the presence of the reboot annotation
+			for _, node := range winNodes.Items {
+				_, present := node.Annotations[metadata.RebootAnnotation]
+				if present != waitingForRemoval {
+					seenWinNodes[node.Name] = struct{}{}
+				}
 			}
+			// Poll until all Windows nodes are seen with the reboot annotation
```
mansikulkarni96, line 147: is this comment valid if we are waiting for reboot annotation removal?

test/e2e/proxy_test.go
```diff
@@ -77,35 +80,113 @@ func (tc *testContext) testCerts(t *testing.T) {
...
+			assert.True(t, proxyCertsRemoved, "proxy certs not removed")
+		})
+	}
+}
 
-				assert.Equalf(t, count, 1, "unexpected cert %d count on node %s: expected 1, found %d", i, node, count)
-				i++
+// waitForAllNodesToReboot waits for all Windows nodes in the cluster to be rebooted
+func (tc *testContext) waitForAllNodesToReboot() error {
+	// Wait for for reboot annotation to be applied to all Windows nodes, and then wait for removal
```
mansikulkarni96, line 118: redundant for*

pkg/daemon/certs/certs.go
```diff
@@ -106,69 +182,102 @@ func splitAtPEMCert() func(data []byte, atEOF bool) (advance int, token []byte,
...
+// ensureFileIsRemoved deletes the file at the given path. No-op if the file already does not exist
+func ensureFileIsRemoved(path string) error {
+	exists, err := fileExists(path)
+	if err != nil {
+		return err
+	}
+	if !exists {
+		return nil
+	}
+	klog.Infof("removing file %s", path)
```
aravindhp, line 220: Nit: I am starting to see a pattern where we add logs saying we are doing X but no log saying it succeeded. This is not great when debugging or for the end user. Best is to either follow up with a message saying that the file has been removed or just log when we are actually removed the file after the fact. I had a tough time with this in the HNS removal area where we say we are removing the networks but without an indicator of success.

pkg/daemon/cleanup/cleanup.go
```diff
@@ -69,8 +70,12 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
...
 		return err
 	}
-	// rebooting instance to unset the environment variables at the process level as expected
-	if envVarsRemoved {
+	certsRemoved, err := certs.Reconcile("")
+	if err != nil {
+		return err
+	}
+	// rebooting instance to unset the environment variables and available certifictes at the process level as expected
+	if envVarsRemoved || certsRemoved {
```
aravindhp, line 78: This logically belongs in the `[daemon] Trust bundle cert removal`. I had to return to the commit to figure out how the `""` input is handled.

test/e2e/proxy_test.go
```diff
@@ -77,35 +80,113 @@ func (tc *testContext) testCerts(t *testing.T) {
...
+	if err := tc.waitForRebootAnno(false); err != nil {
+		return err
+	}
+	return tc.waitForRebootAnno(true)
+}
+
+// waitForRebootAnno waits for all nodes in the cluster to have the reboot annotation either applied or removed
+func (tc *testContext) waitForRebootAnno(waitingForRemoval bool) error {
```
aravindhp, line 126: Nit: `Anno` sound weird 😆 `Annotation` won't make the name that much longer.

test/e2e/proxy_test.go
```diff
@@ -77,35 +80,113 @@ func (tc *testContext) testCerts(t *testing.T) {
...
+	allWinNodes := make(map[string]struct{})
+	for _, node := range gc.allNodes() {
+		allWinNodes[node.Name] = struct{}{}
+	}
```
aravindhp, line 130: Wouldn't it be safer to create this map before removing the certs from the cluster-wide proxy configuration?

aravindhp (changes requested):

test/e2e/proxy_test.go
saifshaikh48: I don't see why as gc.allNodes will remain as is (even if WMCO were to implode and delete all node objects)

I think the only way the gc's nodes get updated is by calling `loadExistingNodes()` or `loadwaitForConfiguredWindowsNodes()`

aravindhp (commented):
/lgtm

I will let @sebsoto add the approval given he had comments

test/e2e/proxy_test.go
aravindhp: OK, then it leads to another question... don't we need to update gc.Nodes between `waitForRebootAnnotation()` calls?

pkg/daemon/certs/certs.go
```diff
@@ -106,69 +182,105 @@ func splitAtPEMCert() func(data []byte, atEOF bool) (advance int, token []byte,
...
-		windows.CERT_SYSTEM_STORE_LOCAL_MACHINE, uintptr(unsafe.Pointer(syscall.StringToUTF16Ptr("ROOT"))))
+// fileExists checks if the file at the given path exists. Returns true if it exists, false if not, and error otherwise.
+func fileExists(path string) (bool, error) {
+	_, err := os.Stat(path)
+	if err == nil {
+		return true, nil
+	} else if errors.Is(err, os.ErrNotExist) {
+		return false, nil
+	} else {
+		return false, newFileIOError(fmt.Errorf("Failed to check if file %s exists: %w", path, err))
```
sebsoto, line 193: Failed -> failed
errors should be lowercase

pkg/daemon/certs/certs.go
```diff
@@ -196,3 +308,68 @@ func addCertToStore(store windows.Handle, cert *x509.Certificate) error {
...
+	var err error
+	for {
+		certContext, err = windows.CertEnumCertificatesInStore(store, certContext)
+		if err != nil {
+			if errors.Is(err, windows.Errno(windows.CRYPT_E_NOT_FOUND)) {
+				// Error code implies we have read all certs:
+				// https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-certenumcertificatesinstore
+				break
+			}
+			return nil, fmt.Errorf("Error reading existing system certificate: %w", err)
```
sebsoto, line 344: Error -> error 
errors should be lowercase

pkg/daemon/certs/certs.go
```diff
@@ -196,3 +308,68 @@ func addCertToStore(store windows.Handle, cert *x509.Certificate) error {
...
+		}
+
+		// Convert the certificate bytes to Golang x509.Certificate
+		certBytes := unsafe.Slice(certContext.EncodedCert, certContext.Length)
+		current, err := x509.ParseCertificate(certBytes)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to parse certificate from bytes: %w", err)
```
sebsoto, line 351: Failed -> failed
errors should be lowercase

pkg/daemon/controller/controller.go
```diff
@@ -295,6 +304,12 @@ func (sc *ServiceController) reconcileEnvVarsAndCerts(envVars map[string]string,
 	node core.Node) (bool, error) {
 	certsUpdated, err := certs.Reconcile(sc.caBundle)
 	if err != nil {
+		var fileIOErr *certs.FileIOError
+		if errors.As(err, &fileIOErr) {
```
sebsoto, line 308: nit: errors.As is being used here when the functionality of errors.Is would suffice. This threw me off as everywhere else you use errors.Is

pkg/daemon/certs/certs.go
```diff
@@ -196,3 +308,68 @@ func addCertToStore(store windows.Handle, cert *x509.Certificate) error {
 	klog.Infof("Certificate %s imported successfully.", cert.Subject.String())
 	return nil
 }
+
+// removeCertFromStore removes the given certificate from its associated certificate store
+func removeCertFromStore(store windows.Handle, cert *x509.Certificate) error {
+	// Windows CertContext holds the store which imported the cert, which is needed when deleting from the system store
+	certContext, err := retrieveCertContext(store, cert)
+	if err != nil {
+		return fmt.Errorf("Failed to create contet from x509 cert %s: %w\n", cert.Subject.String(), err)
```
sebsoto, line 317: Failed -> failed
contet -> content
please remove the trailing newline.

pkg/daemon/certs/certs.go
```diff
@@ -196,3 +308,68 @@ func addCertToStore(store windows.Handle, cert *x509.Certificate) error {
...
+	}
+	if certContext == nil {
+		// Cert already doesn't exist, nothing to do
+		return nil
+	}
+
+	if err = windows.CertDeleteCertificateFromStore(certContext); err != nil {
+		return fmt.Errorf("Failed to remove certificate %s: %z\n", cert.Subject.String(), err)
+	}
+	klog.Infof("Certificate %s removed successfully.", cert.Subject.String())
```
sebsoto, line 327: Should this be debug log level?

pkg/daemon/certs/certs.go
```diff
@@ -25,27 +25,103 @@ import (
...
+	}()
+
+	certsImported, err := ensureCertsAreImported(systemStore, expectedCerts, existingCerts)
+	if err != nil {
+		return false, fmt.Errorf("Failed to import certificates: %w", err)
+	}
+	certsDeleted, err := ensureCertsAreDeleted(systemStore, expectedCerts, existingCerts)
+	if err != nil {
+		return false, fmt.Errorf("Failed to delete certificates: %w", err)
+	}
```
sebsoto, line 89: This block makes me think there should be a `CertManager` struct.
Specifically the passing in of `systemStore` to a bunch of functions.
Not a necessity, but could help organize the code a bit.

pkg/daemon/certs/certs.go
```diff
@@ -25,27 +25,103 @@ import (
...
 	"syscall"
 	"unsafe"
 
 	"golang.org/x/sys/windows"
 	"k8s.io/klog/v2"
 )
 
-// Reconcile ensures the certificates in the given trusted CA bundle file are imported as system certificates.
+// importedCABundleFile tracks all the certs that the operator has imported into the node's local trust store
+const importedCABundleFile = "C:\\Temp\\imported-certs.pem"
```
sebsoto, line 39: I can't find any info on this being an actually temp directory.
Any reason to not use `c:\\k`?
I have a feeling we may need to lock our file permissions down more securely, and keeping all WMCO/WICD files contained within c:\k simplifies that.

pkg/daemon/certs/certs.go
```diff
@@ -25,27 +25,103 @@ import (
...
-		// TODO: ensure that all certs that we've added previously are removed https://issues.redhat.com/browse/WINC-688
+func Reconcile(caBundlePath string) (bool, error) {
+	// caBundlePath will be empty when a cluster-wide proxy is being removed or not in use
+	certChange, err := reconcileCerts(caBundlePath)
+	if err != nil {
+		return false, err
+	}
+	// An error updating the imported bundle file will be self-corrected on the next reconcile, since it is the source
+	// of truth for determining whether certs need to be updated
+	return certChange, updateImportedCABundle(caBundlePath, certChange)
```
sebsoto, line 51: What if this errors out after doing the actual cert changes?
It won't restart because at the callsites of Reconcile:
```
	certsRemoved, err := certs.Reconcile("")
	if err != nil {
		return err
	}
```
a restart only happens when no error occurs.

Then on the next run, there will be no cert changes as they were already made.
A restart wont be triggered.
In this case a restart is never triggered and things won't function properly.
I suspect you need to restart even in the case of error, if the returned bool is true. This should be documented in the function doc as well.

test/e2e/delete_test.go
```diff
@@ -65,7 +66,18 @@ func (tc *testContext) testBYOHRemoval(t *testing.T) {
...
+	var trustedCABundle string
+	proxyEnabled, err := tc.client.ProxyEnabled()
+	require.NoErrorf(t, err, "error checking if proxy is enabled in test environment")
+	if proxyEnabled {
+		// TODO: this only tests the user-provided certs, a subset of the required proxy certificates.
+		// Should be addressed with https://issues.redhat.com/browse/WINC-1144
+		cm, err := tc.client.K8s.CoreV1().ConfigMaps(userCABundleNamespace).Get(context.TODO(), userCABundleName, meta.GetOptions{})
+		require.NoErrorf(t, err, "error getting user-provided CA ConfigMap: %w", err)
+		trustedCABundle = cm.Data[certificates.CABundleKey]
+	}
```
sebsoto, line 78: Please make this its own function which returns trustedCABundle

test/e2e/delete_test.go
```diff
@@ -89,6 +101,10 @@ func (tc *testContext) testBYOHRemoval(t *testing.T) {
 			require.NoError(t, err, "error determining if ENV vars are removed")
 			assert.True(t, envVarsRemoved, "ENV vars not removed")
 
+			proxyCertsRemoved, err := tc.checkProxyCerts(addr, []byte(trustedCABundle), 0)
+			require.NoError(t, err, "error determining if proxy certs are removed")
+			assert.True(t, proxyCertsRemoved, "proxy certs not removed")
```
sebsoto, line 106: nit: this should probably be a subtest

sebsoto (changes requested):
Great work, PTAL at my comments

pkg/daemon/certs/certs.go
sebsoto: to be clear you don't have to do this, just a nice to have

pkg/daemon/certs/certs.go
saifshaikh48: no reason, I can change this to c:\k

pkg/daemon/certs/certs.go
saifshaikh48: You've convinced me this is a valid concern. doc'd and changed call sites

sebsoto (approved):
Thanks @saifshaikh48
I approved, but please address my comments before removing the hold

saifshaikh48:
/test aws-e2e-upgrade
/test wicd-unit-vsphere
/test unit

running a subset while awaiting LGTM

mansikulkarni96:
Thanks for working on this.
/lgtm

PR #1840, opened 2023-09-27
[release-4.13] WINC-1098: Block upgrades when migrating from in-tree to CSI

sebsoto:
Manual backport of https://github.com/openshift/windows-machine-config-operator/pull/1794

### DIFF
--- a/controllers/controllers.go
+++ b/controllers/controllers.go
@@ -17,6 +17,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	"github.com/openshift/windows-machine-config-operator/pkg/certificates"
+	"github.com/openshift/windows-machine-config-operator/pkg/cluster"
 	"github.com/openshift/windows-machine-config-operator/pkg/condition"
 	"github.com/openshift/windows-machine-config-operator/pkg/crypto"
 	"github.com/openshift/windows-machine-config-operator/pkg/instance"
@@ -64,6 +65,16 @@ func (r *instanceReconciler) ensureInstanceIsUpToDate(instanceInfo *instance.Inf
 		return nil
 	}
 
+	csiEnabled, err := cluster.CSIStorageEnabled(r.platform)
+	if err != nil {
+		return err
+	}
+	if csiEnabled {
+		if labelsToApply == nil {
+			labelsToApply = make(map[string]string)
+		}
+		labelsToApply[metadata.CSIConfiguredLabel] = "true"
+	}
 	nc, err := nodeconfig.NewNodeConfig(r.client, r.k8sclientset, r.clusterServiceCIDR, r.watchNamespace,
 		instanceInfo, r.signer, labelsToApply, annotationsToApply, r.platform)
 	if err != nil {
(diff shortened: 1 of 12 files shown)


mansikulkarni96:
@sebsoto does this need to be backported to 4.12?

sebsoto:
@mansikulkarni96 No it does not

controllers/controllers.go
```diff
@@ -110,8 +118,23 @@ func (r *instanceReconciler) ensureInstanceIsUpToDate(instanceInfo *instance.Inf
 	return nc.Configure()
 }
 
+// getNodeSummary queries the cAdvisor metrics endpoint for the given Node
+func (r *instanceReconciler) getNodeSummary(nodeName string) (stats.Summary, error) {
+	var nodeStats stats.Summary
```
aravindhp, line 123: Please move this closer to the scope where it is used

controllers/controllers.go
```diff
@@ -126,10 +149,15 @@ func (r *instanceReconciler) upgradeBlocked(node *core.Node, upgradingToCSI bool
 	if value := node.GetLabels()[metadata.CSIConfiguredLabel]; value == "true" {
 		return false
 	}
-	if len(node.Status.VolumesAttached) == 0 {
-		return false
+	// For each pod running on this node, check if there is a PVC backed volume, if one is found, block the upgrade
+	for _, pod := range nodeSummary.Pods {
```
aravindhp, line 153: If you care only about `nodeSummary.Pods` then there is no need to pass the whole object.

controllers/controllers.go
```diff
@@ -110,8 +118,23 @@ func (r *instanceReconciler) ensureInstanceIsUpToDate(instanceInfo *instance.Inf
 	return nc.Configure()
 }
 
+// getNodeSummary queries the cAdvisor metrics endpoint for the given Node
+func (r *instanceReconciler) getNodeSummary(nodeName string) (stats.Summary, error) {
```
aravindhp, line 122: I suggest just returning `stats.Summary.Pods` given that is all we need and rename the function appropriately. Example: `getPodStatsForNode()`.

aravindhp (changes requested):

saifshaikh48 (commented):
/lgtm

Can re-apply LGTM when other comments are addressed

saifshaikh48 (commented):
/lgtm

aravindhp (approved):

controllers/controllers.go
```diff
@@ -110,8 +118,23 @@ func (r *instanceReconciler) ensureInstanceIsUpToDate(instanceInfo *instance.Inf
 	return nc.Configure()
 }
 
+// getPodStatsForNode queries the cAdvisor metrics endpoint for pods on the given Node
+func (r *instanceReconciler) getPodStatsForNode(nodeName string) ([]stats.PodStats, error) {
+	data, err := r.k8sclientset.DiscoveryV1().RESTClient().Get().
+		AbsPath("api/v1/nodes/" + nodeName + "/proxy/stats/summary").DoRaw(context.TODO())
+	if err != nil {
+		return nil, err
```
alinaryan, line 126: can you add an err msg

alinaryan (commented):
just a nit: otherwise lgtm

PR #1849, opened 2023-09-29
OCPBUGS-19943: Fix hybrid-overlay parameters required for certificate management

kyrtapz:
First commit bumps ovn-kubernetes submodule.
Second commit adjusts hybrid-overlay parameters required to use a per-node certificate.

### DIFF
--- a/pkg/services/services.go
+++ b/pkg/services/services.go
@@ -102,8 +102,8 @@ func azureCloudNodeManagerConfiguration() servicescm.Service {
 
 // hybridOverlayConfiguration returns the Service definition for hybrid-overlay
 func hybridOverlayConfiguration(vxlanPort string, debug bool) servicescm.Service {
-	hybridOverlayServiceCmd := fmt.Sprintf("%s --node NODE_NAME --k8s-kubeconfig %s --windows-service "+
-		"--logfile "+"%s\\hybrid-overlay.log", windows.HybridOverlayPath, windows.KubeconfigPath,
+	hybridOverlayServiceCmd := fmt.Sprintf("%s --node NODE_NAME --bootstrap-kubeconfig=%s --cert-dir=%s --cert-duration=24h "+
+		"--windows-service --logfile "+"%s\\hybrid-overlay.log", windows.HybridOverlayPath, windows.KubeconfigPath, windows.CniConfDir,
 		windows.HybridOverlayLogDir)
 	if len(vxlanPort) > 0 {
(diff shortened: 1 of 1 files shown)


sebsoto (approved):
Thanks for the quick fix, LGTM

sebsoto:
This is the fix that is being brought in: https://github.com/openshift/ovn-kubernetes/pull/1919/commits/ac329f818509ca282a5791dd4a4751c1cf2bba84

sebsoto:
/retitle OCPBUGS-19943: Fix hybrid-overlay parameters required for certificate management

pkg/services/services.go
```diff
@@ -102,8 +102,8 @@ func azureCloudNodeManagerConfiguration() servicescm.Service {
 
 // hybridOverlayConfiguration returns the Service definition for hybrid-overlay
 func hybridOverlayConfiguration(vxlanPort string, debug bool) servicescm.Service {
-	hybridOverlayServiceCmd := fmt.Sprintf("%s --node NODE_NAME --k8s-kubeconfig %s --windows-service "+
-		"--logfile "+"%s\\hybrid-overlay.log", windows.HybridOverlayPath, windows.KubeconfigPath,
+	hybridOverlayServiceCmd := fmt.Sprintf("%s --node NODE_NAME --bootstrap-kubeconfig=%s --cert-dir=%s --cert-duration=24h "+
```
jrvaldes, line 105: @kyrtapz  `--cert-duration=24h` means the cert will renew or expire in 24 h? 

Can the `hybrid overlay` service modify annotations after this duration?

jrvaldes (commented):
@kyrtapz  thanks for working on this. LGTM.

pkg/services/services.go
kyrtapz: This means that the certificate will be valid for 24h, before it expires the hybrid-overlay binary will perform automatic rotation so there should be no issues there.

jrvaldes: Thanks, went back to the [thread](https://redhat-internal.slack.com/archives/CDCP2LA9L/p1695977610275759) and found the answer. 😄

aravindhp: How do the certs end up in the CNI conf directory?

aravindhp: Seems like the hybrid-overlay will created them.

kyrtapz: The certificate files are created by hybrid-overlay based on a CSR request that gets approved.

PR #1877, opened 2023-10-09
Improve predicate filtering for WindowsMachine controllers

jrvaldes:
This change updates the predicate filters in the Windows Machine controller to first check for machines
with the Windows label so that the machine validations target only Windows machines. In addition, updates the delete event filter in the Windows Machine predicate to honor the ignore label.

Before, WMCO reacted to Linux machines joining the cluster, for example:
```
DEBUG	controller.windowsmachine	Machine has no phase associated with it	{"name": "jvaldes-c57lx-worker-0-7tlm5"}
DEBUG	controller.windowsmachine	Machine has no phase associated with it	{"name": "jvaldes-c57lx-worker-0-7tlm5"}
DEBUG	controller.windowsmachine	invalid Machine	{"name": "jvaldes-c57lx-worker-0-7tlm5", "error": "no IP addresses defined"}
DEBUG	controller.windowsmachine	invalid Machine	{"name": "jvaldes-c57lx-worker-0-7tlm5", "error": "no IP addresses defined"}
DEBUG	controller.windowsmachine	invalid Machine	{"name": "jvaldes-c57lx-worker-0-7tlm5", "error": "no internal IP address associated"}

```

### DIFF
--- a/controllers/windowsmachine_controller.go
+++ b/controllers/windowsmachine_controller.go
@@ -110,18 +110,18 @@ func (r *WindowsMachineReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		// We need the create event to account for Machines that are in provisioned state but were created
 		// before WMCO started running
 		CreateFunc: func(e event.CreateEvent) bool {
-			return r.isValidMachine(e.Object) && isWindowsMachine(e.Object.GetLabels())
+			return isWindowsMachine(e.Object.GetLabels()) && r.isValidMachine(e.Object)
 		},
 		UpdateFunc: func(e event.UpdateEvent) bool {
-			return r.isValidMachine(e.ObjectNew) && isWindowsMachine(e.ObjectNew.GetLabels())
+			return isWindowsMachine(e.ObjectNew.GetLabels()) && r.isValidMachine(e.ObjectNew)
 		},
 		GenericFunc: func(e event.GenericEvent) bool {
-			return r.isValidMachine(e.Object) && isWindowsMachine(e.Object.GetLabels())
+			return isWindowsMachine(e.Object.GetLabels()) && r.isValidMachine(e.Object)
 		},
 		// process delete event
 		DeleteFunc: func(e event.DeleteEvent) bool {
-			// for Windows machines only
-			return isWindowsMachine(e.Object.GetLabels())
+			// for Windows machines only and avoid processing delete event for machines with ignore label
+			return isWindowsMachine(e.Object.GetLabels()) && e.Object.GetLabels()[IgnoreLabel] != "true"
 		},
 	}
 


controllers/windowsmachine_controller.go
```diff
@@ -120,8 +120,8 @@ func (r *WindowsMachineReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		},
 		// process delete event
 		DeleteFunc: func(e event.DeleteEvent) bool {
-			// for Windows machines only
-			return isWindowsMachine(e.Object.GetLabels())
+			// for Windows machines only without ignore label
+			return isWindowsMachine(e.Object.GetLabels()) && !ignoreLabel(e.Object.GetLabels())
```
saifshaikh48, line 124: any reason not to call `r.isValidMachine(e.Object)` instead of introducing `ignoreLabel()`?

controllers/windowsmachine_controller.go
```diff
@@ -110,13 +110,13 @@ func (r *WindowsMachineReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		// We need the create event to account for Machines that are in provisioned state but were created
 		// before WMCO started running
 		CreateFunc: func(e event.CreateEvent) bool {
-			return r.isValidMachine(e.Object) && isWindowsMachine(e.Object.GetLabels())
+			return isWindowsMachine(e.Object.GetLabels()) && r.isValidMachine(e.Object)
```
aravindhp, line 113: Please add to your commit that the only visible advantage of this change is the debug messages not showing up.

Do we need to have a single helper for this?

controllers/windowsmachine_controller.go
```diff
@@ -175,9 +175,19 @@ func isWindowsMachine(labels map[string]string) bool {
 	return false
 }
 
+// ignoreLabel checks if the ignore label is present or not
+func ignoreLabel(labels map[string]string) bool {
```
aravindhp, line 179: Please don't interleave struct and helper methods. `isWindowsMachine()` is badly placed but no reason to continue that.

controllers/windowsmachine_controller.go
aravindhp: But the first thing that is checked in `r.isValidMachine(e.Object)` is the label. How are the other validations being hit?

aravindhp (changes requested):

controllers/windowsmachine_controller.go
jrvaldes: True, and it sounds like a hack with a side effect. It was counter-intuitive for me to add a statement to validate a deleted machine. So, I took the label logic apart. 

In the same sense, a machine could be valid with or without the ignore label, WMCO just chooses to ignore it. Hence, I see it as two separate things.

controllers/windowsmachine_controller.go
jrvaldes: > Please add to your commit that the only visible advantage of this change is the debug messages not showing up.

.. and possible optimization of the busy time for the WindowsMachine controller since it spent cycles validating a non-windows machine, time which could be used to configure provisioned Windows Machine waiting in the queue.

jrvaldes: > Do we need to have a single helper for this?

Possibly, yes. @openshift/openshift-team-windows-containers  thoughts?

controllers/windowsmachine_controller.go
jrvaldes: Good catch; thanks for highlighting the big picture. _This is the way!_

controllers/windowsmachine_controller.go
aravindhp: >.. and possible optimization of the busy time for the WindowsMachine controller since it spent cycles validating a non-windows machine, time which could be used to configure provisioned Windows Machine waiting in the queue.

Have you measured the cycles? I doubt this has any actual effect given `isValidMachine()` just looks at memory and not making any API calls. While I am not against this change we should always keep the context in mind when optimizing. This is for future optimizations you are thinking about.

PS: I am not asking you to spend time profiling the method.

controllers/windowsmachine_controller.go
aravindhp: What this really calls for is just updating the doc for `isValidMachine()` and not introduce more code. But I won't bikeshed on this.

controllers/windowsmachine_controller.go
```diff
@@ -175,9 +175,19 @@ func isWindowsMachine(labels map[string]string) bool {
...
+	if value, ok := labels[IgnoreLabel]; ok {
+		if value == "true" {
+			return true
+		}
+	}
+	return false
```
sebsoto, line 185: This whole function can be simplified to
```
return labels[IgnoreLabel] == "true"
```

Because of that I suggest this function either be removed, or changed to work on an object like the rest of the predicate functions, instead of a map
```
func shouldIgnore(obj client.Object){
  return ignoreLabel(obj.GetLabels()[IgnoreLabel] == true
}
```

sebsoto (changes requested):

controllers/windowsmachine_controller.go
jrvaldes: I follow.

controllers/windowsmachine_controller.go
jrvaldes: Makes sense.

controllers/windowsmachine_controller.go
jrvaldes: Addressed.

jrvaldes:
@openshift/openshift-team-windows-containers   PTAL

jrvaldes:
/close

low priority.

saifshaikh48 (commented):
/lgtm

PR #1890, opened 2023-10-12
WINC-673: Fix Windows Webserver tester jobs in upgrade test

jrvaldes:
This change fixes the upgrade test suite so that
it checks the correct Linux job that curls the
service exposed by the Windows Webserver.

Before, the label selector was picking no pods and
the number of failed pod was always zero making this
test worthless. Now, the e2e upgrade test fails if the Windows
workloads were unavailable in an upgrade scenario.

In addition, adds the collection of the pod logs of the Windows
Webserver deployment and service curl job during
the cleanup in the upgrade test. Before, there was
no evidence in the artifact dir about such pods.

### DIFF
--- a/test/e2e/upgrade_test.go
+++ b/test/e2e/upgrade_test.go
@@ -39,7 +39,7 @@ func upgradeTestSuite(t *testing.T) {
 	require.NoError(t, err)
 
 	// test if Windows workloads are running by creating a Job that curls the workloads continuously.
-	cleanupWorkloadAndTester, err := tc.deployWindowsWorkloadAndTester()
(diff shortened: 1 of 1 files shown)


jrvaldes:
/close

There is no need to fix this check now, as the upgrade test is moving to a newer upgrade test suite. See the [conversation](https://redhat-internal.slack.com/archives/CM4ERHBJS/p1697127533094079?thread_ts=1697123613.234109&cid=CM4ERHBJS) in Slack.

The new upgrade test will be validated by checking a to-be-introduced node annotation to prevent more than one node from reconciling at the same time.

PR #1901, opened 2023-10-16
OCPBUGS-8996: Introduce `upgrading` label to block concurrent upgrades

jrvaldes:
This PR introduces the `windowsmachineconfig.openshift.io/upgrading` label for Windows nodes to block concurrent upgrades between the WindowsMachine and ConfigMap controllers.

The maximum allowed number of parallel upgrades proposed is fixed to 1. Hence, no matter how many Windows nodes are available, WMCO will upgrade them one at a time. 

ATM, the locking mechanism in the condition package is insufficient as it has a controllers-wide scope that sets the operator condition on each work item across all the controllers. From the OLM perspective, it makes sense, as OLM can track the sub-components of WMCO as a parent entity, but there is a need for sequential control for upgrades.

With the concurrent reconciliation between WindowsMachine and ConfigMap controllers, the level of granularity is more downward and specific to the upgrade scenario. For example, it is acceptable to configure two instances (machine and BYOH) simultaneously with Upgradeable operator condition set to False. Still, the same locking scenario is not valid for an upgrade. So, I inclined the approach to a separate locking mechanism exclusive for locking the control flags used to coordinate the upgrade across controllers.

### DIFF
--- a/controllers/controllers.go
+++ b/controllers/controllers.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"sync"
 
 	"github.com/go-logr/logr"
 	config "github.com/openshift/api/config/v1"
@@ -25,6 +26,17 @@ import (
 	"github.com/openshift/windows-machine-config-operator/pkg/nodeconfig"
 	"github.com/openshift/windows-machine-config-operator/pkg/secrets"
 	"github.com/openshift/windows-machine-config-operator/version"
+)
+
+const (
+	// MaxParallelUpgrades is the default maximum allowed number of nodes that can be upgraded in parallel.
+	// It is a positive integer and cannot be used to stop upgrades, only to limit the number of concurrent upgrades.
+	MaxParallelUpgrades = 1
+)
+
+var (
+	// controllerLocker is used to synchronize upgrades between controllers
+	controllerLocker sync.Mutex
 )
 
 // instanceReconciler contains everything needed to perform actions on a Windows instance
@@ -76,6 +88,9 @@ func (r *instanceReconciler) ensureInstanceIsUpToDate(instanceInfo *instance.Inf
 		// Instance requiring an upgrade indicates that node object is present with the version annotation
 		r.log.Info("instance requires upgrade", "node", instanceInfo.Node.GetName(), "version",
 			instanceInfo.Node.GetAnnotations()[metadata.VersionAnnotation], "expected version", version.Get())
+		if err := markNodeAsUpgrading(context.TODO(), r.client, instanceInfo.Node); err != nil {
+			return err
+		}
 		if err := nc.Deconfigure(); err != nil {
 			return err
 		}
@@ -280,4 +295,39 @@ func markAsFreeOnSuccess(c client.Client, watchNamespace string, recorder record
 		return condition.MarkAsFree(c, watchNamespace, recorder, controllerName)
 	}
 	return err
+}
+
+// markNodeAsUpgrading marks the given node as upgrading by adding an annotation to it. If the number of nodes
+// performing upgrades in parallel exceeds the maximum allowed, an error is returned
+func markNodeAsUpgrading(ctx context.Context, c client.Client, currentNode *core.Node) error {
+	controllerLocker.Lock()
+	defer controllerLocker.Unlock()
+	upgradingNodes, err := findUpgradingNodes(ctx, c)
+	if err != nil {
+		return err
+	}
+	// check if current node is already marked as upgrading
+	for _, node := range upgradingNodes.Items {
+		if node.Name == currentNode.Name {
+			// current node is upgrading, continue with it
+			return nil
+		}
+	}
+	if len(upgradingNodes.Items) >= MaxParallelUpgrades {
+		return fmt.Errorf("cannot mark node %s as upgrading, maximum number of parallel upgrading nodes reached (%d)",
+			currentNode.Name, MaxParallelUpgrades)
+	}
+	return metadata.ApplyUpgradingLabel(ctx, c, currentNode)
+}
+
+// findUpgradingNodes returns a pointer to the resulting list of Windows nodes that are upgrading  i.e. have the
+// upgrading label set to true
+func findUpgradingNodes(ctx context.Context, c client.Client) (*core.NodeList, error) {
+	// get nodes Windows nodes with upgrading label
+	matchingLabels := client.MatchingLabels{core.LabelOSStable: "windows", metadata.UpgradingLabel: "true"}
+	nodeList := &core.NodeList{}
+	if err := c.List(ctx, nodeList, matchingLabels); err != nil {
+		return nil, fmt.Errorf("error listing Windows nodes with upgrading label: %w", err)
+	}
+	return nodeList, nil
 }
--- a/hack/common.sh
+++ b/hack/common.sh
@@ -332,3 +332,28 @@ getWindowsInstanceCountFromConfigMap() {
    -n "${WMCO_DEPLOY_NAMESPACE}" \
    -o json | jq '.data | length'
 }
+
+# creates the a job and required RBAC to check the number of Windows nodes performing
+# parallel upgrade in the test cluster
+createParallelUpgradeCheckerResources() {
+  winNodesCount=$(oc get nodes -l kubernetes.io/os=windows  -o jsonpath='{.items[*].metadata.name}' | wc -w)
+  if [[ winNodesCount -lt 2 ]]; then
+    echo "Skipping parallel upgrade checker job, requires 2 or more nodes. Found ${winNodesCount} nodes"
+    return
+  fi
+  # get the latest tools image from the image stream
+  export TOOLS_IMAGE=$(oc get imagestreamtag tools:latest -n openshift -o jsonpath='{.tag.from.name}')
+  # set job' container image and create the job
+  JOB=$(sed -e "s|REPLACE_WITH_OPENSHIFT_TOOLS_IMAGE|${TOOLS_IMAGE}|g" hack/e2e/resources/parallel-upgrade-checker-job.yaml)
+  cat <<EOF | oc apply -f -
+${JOB}
+EOF
+}
+
+# creates the a job and required RBAC to check the number of Windows nodes performing
+# parallel upgrade in the test cluster
+deleteParallelUpgradeCheckerResources() {
+  oc delete -f hack/e2e/resources/parallel-upgrade-checker-job.yaml || {
+    echo "error deleting parallel upgrade checker job"
+  }
+}
--- a/hack/e2e/resources/parallel-upgrade-checker-job.yaml
+++ b/hack/e2e/resources/parallel-upgrade-checker-job.yaml
@@ -1,0 +1,52 @@
+kind: Job
+apiVersion: batch/v1
+metadata:
+  name: parallel-upgrades-checker
+  namespace: wmco-test
+  labels:
+    batch.kubernetes.io/job-name: parallel-upgrades-checker
+    job-name: parallel-upgrades-checker
+spec:
+  parallelism: 1
+  completions: 1
+  backoffLimit: 3
+  template:
+    metadata:
+      labels:
+        batch.kubernetes.io/job-name: parallel-upgrades-checker
+        job-name: parallel-upgrades-checker
+    spec:
+      nodeSelector:
+        kubernetes.io/os: linux
+      restartPolicy: Never
+      serviceAccountName: wmco-test
+      os:
+        name: linux
+      containers:
+        - name: parallel-upgrades-checker
+          image: 'REPLACE_WITH_OPENSHIFT_TOOLS_IMAGE'
+          command:
+            - bash
+            - '-c'
+            - |
+              #!/bin/bash
+              set -euo pipefail
+
+              # max number of parallel upgrades allowed, fixed to 1. Refer to controllers.MaxParallelUpgrades
+              export MAX_PARALLEL_UPGRADES=1
+
+              # loop indefinitely until count exceeded
+              while true; do
+              	upgradingCount=$(oc get nodes -l kubernetes.io/os=windows  -o jsonpath='{.items[*].metadata.labels.windowsmachineconfig\.openshift\.io/upgrading}' | wc -w)	
+              	if [[ $upgradingCount -gt $MAX_PARALLEL_UPGRADES ]]; then
+              	  echo "error: max upgrading count exceeded"
+              	  exit 1 
+              	fi
+                echo ""
+              	echo "pass: upgrading count $upgradingCount/$MAX_PARALLEL_UPGRADES"
+              	echo "waiting 5s for next check..."
+              	sleep 5
+              done
+          imagePullPolicy: IfNotPresent
+          securityContext:
+            runAsNonRoot: true
--- a/hack/run-ci-e2e-test.sh
+++ b/hack/run-ci-e2e-test.sh
@@ -179,9 +179,11 @@ if [[ "$TEST" = "upgrade-setup" ]]; then
   # Run the storage test, skipping deletion of the created workload in order to test that it persists across the upgrade
   go test ./test/e2e/... -run=TestWMCO/storage -v -timeout=15m -args $GO_TEST_ARGS --skip-workload-deletion=true
   go test ./test/e2e/... -run=TestWMCO/create/Node_Logs -v -timeout=10m -args $GO_TEST_ARGS
+  createParallelUpgradeCheckerResources
 fi
 
 if [[ "$TEST" = "upgrade-test" ]]; then
+  trap deleteParallelUpgradeCheckerResources EXIT
   go test ./test/e2e/... -run=TestUpgrade -v -timeout=20m -args $GO_TEST_ARGS
 fi
 
--- a/pkg/metadata/metadata.go
+++ b/pkg/metadata/metadata.go
@@ -23,6 +23,8 @@ const (
 	DesiredVersionAnnotation = "windowsmachineconfig.openshift.io/desired-version"
 	// RebootAnnotation indicates the node's underlying instance needs to be restarted
 	RebootAnnotation = "windowsmachineconfig.openshift.io/reboot-required"
+	// UpgradingLabel indicates the node's underlying instance is performing an upgrade
+	UpgradingLabel = "windowsmachineconfig.openshift.io/upgrading"
 )
 
 // generatePatch creates a patch applying the given operation onto each given annotation key and value
@@ -42,6 +44,24 @@ func generatePatch(op string, labels, annotations map[string]string) ([]*patch.J
 		}
 	}
 	return patches, nil
+}
+
+// removeLabel removes the given label from the node object
+func removeLabel(ctx context.Context, c client.Client, node *core.Node, label string) error {
+	_, present := node.GetLabels()[label]
+	if !present {
+		// label not present in node, nothing to remove
+		return nil
+	}
+	patchData, err := GenerateRemovePatch([]string{label}, []string{})
+	if err != nil {
+		return fmt.Errorf("error creating label remove patch: %w", err)
+	}
+	err = c.Patch(ctx, node, client.RawPatch(kubeTypes.JSONPatchType, patchData))
+	if err != nil {
+		return fmt.Errorf("error removing label from node %s: %w", node.GetName(), err)
+	}
+	return nil
 }
 
 // GenerateAddPatch creates a comma-separated list of operations to add all given labels and annotations from an object
@@ -173,4 +193,16 @@ func WaitForRebootAnnotationRemoval(ctx context.Context, c client.Client, nodeNa
 		return fmt.Errorf("timeout waiting for %s to be cleared: %w", RebootAnnotation, err)
 	}
 	return nil
+}
+
+// RemoveUpgradingLabel clears the upgrading label from the node reference, indicating the instance is
+// no longer upgrading
+func RemoveUpgradingLabel(ctx context.Context, c client.Client, node *core.Node) error {
+	return removeLabel(ctx, c, node, UpgradingLabel)
+}
+
+// ApplyUpgradingLabel applies the upgrading label to the given node reference indicating the instance
+// is performing an upgrade
+func ApplyUpgradingLabel(ctx context.Context, c client.Client, node *core.Node) error {
+	return ApplyLabelsAndAnnotations(ctx, c, *node, map[string]string{UpgradingLabel: "true"}, nil)
 }
--- a/test/e2e/upgrade_test.go
+++ b/test/e2e/upgrade_test.go
@@ -206,9 +206,9 @@ func (tc *testContext) deployWindowsWorkloadAndTester() (func(), error) {
 func TestUpgrade(t *testing.T) {
 	tc, err := NewTestContext()
 	require.NoError(t, err)
-	err = tc.waitForConfiguredWindowsNodes(int32(numberOfMachineNodes), false, false)
+	err = tc.waitForConfiguredWindowsNodes(int32(numberOfMachineNodes), true, false)
 	assert.NoError(t, err, "timed out waiting for Windows Machine nodes")
-	err = tc.waitForConfiguredWindowsNodes(int32(numberOfBYOHNodes), false, true)
+	err = tc.waitForConfiguredWindowsNodes(int32(numberOfBYOHNodes), true, true)
 	assert.NoError(t, err, "timed out waiting for BYOH Windows nodes")
 
 	// Basic testing to ensure the Node object is in a good state
--- a/test/e2e/upgrade_test.go
+++ b/test/e2e/upgrade_test.go
@@ -30,6 +30,8 @@ const (
 	windowsWorkloadTesterJob = "windows-workload-tester"
 	// outdatedVersion is the 'previous' version in the simulated upgrade that the operator is being upgraded from
 	outdatedVersion = "old-version"
+	// parallelUpgradesCheckerJobName is a fixed name for the job that checks for the number of parallel upgrades
+	parallelUpgradesCheckerJobName = "parallel-upgrades-checker"
 )
 
 // upgradeTestSuite tests behaviour of the operator when an upgrade takes place.
@@ -219,6 +221,24 @@ func TestUpgrade(t *testing.T) {
 	// test that any workloads deployed on the node have not been broken by the upgrade
 	t.Run("Workloads ready", tc.testWorkloadsAvailable)
 	t.Run("Node Logs", tc.testNodeLogs)
+	t.Run("Parallel Upgrades Checker", tc.testParallelUpgradesChecker)
+
+}
+
+// testParallelUpgradesChecker tests that the number of parallel upgrades does not exceed the max allowed
+// in the lifetime of the job execution. This test is run after the upgrade is complete.
+func (tc *testContext) testParallelUpgradesChecker(t *testing.T) {
+	// get current Windows node state
+	require.NoError(t, tc.loadExistingNodes(), "error getting the current Windows nodes in the cluster")
+	if len(gc.allNodes()) < 2 {
+		t.Skipf("Requires 2 or more nodes to run. Found %d nodes", len(gc.allNodes()))
+	}
+	failedPods, err := tc.client.K8s.CoreV1().Pods(tc.workloadNamespace).List(context.TODO(), metav1.ListOptions{
+		LabelSelector: "job-name=" + parallelUpgradesCheckerJobName, FieldSelector: "status.phase=Failed"})
+	require.NoError(t, err)
+	tc.writePodLogs("job-name=" + parallelUpgradesCheckerJobName)
+	require.Equal(t, 0, len(failedPods.Items), "parallel upgrades check failed",
+		"failed pod count", len(failedPods.Items))
 }
 
 // testWorkloadsAvailable tests that all workloads deployed on Windows nodes by the test suite are available
--- a/test/e2e/validation_test.go
+++ b/test/e2e/validation_test.go
@@ -331,6 +331,30 @@ func (tc *testContext) ensureTestRunnerRole() error {
 	return nil
 }
 
+// ensureTestRunnerClusterRole ensures the proper ClusterRole exists, a requirement for listing Windows node
+// noop if the ClusterRole already exists. Nodes are cluster-scoped resources, and only ClusterRoles
+// can grant permissions to cluster-scoped resources.
+func (tc *testContext) ensureTestRunnerClusterRole(ctx context.Context) error {
+	clusterRole := rbac.ClusterRole{
+		TypeMeta: meta.TypeMeta{},
+		ObjectMeta: meta.ObjectMeta{
+			Name: tc.workloadNamespace,
+		},
+		Rules: []rbac.PolicyRule{
+			{
+				Resources: []string{"nodes"},
+				APIGroups: []string{""},
+				Verbs:     []string{"get", "list"},
+			},
+		},
+	}
+	_, err := tc.client.K8s.RbacV1().ClusterRoles().Create(ctx, &clusterRole, meta.CreateOptions{})
+	if err != nil && !apierrors.IsAlreadyExists(err) {
+		return fmt.Errorf("unable to create role: %w", err)
+	}
+	return nil
+}
+
 // ensureTestRunnerRoleBinding ensures the proper RoleBinding exists, a requirement for SSHing into a Windows node
 // noop if the RoleBinding already exists.
 func (tc *testContext) ensureTestRunnerRoleBinding() error {
@@ -356,6 +380,32 @@ func (tc *testContext) ensureTestRunnerRoleBinding() error {
 	return nil
 }
 
+// ensureTestRunnerClusterRoleBinding ensures the proper ClusterRoleBinding exists, a requirement for listing
+// Windows nodes, noop if the ClusterRoleBinding already exists.
+func (tc *testContext) ensureTestRunnerClusterRoleBinding(ctx context.Context) error {
+	crb := rbac.ClusterRoleBinding{
+		TypeMeta: meta.TypeMeta{},
+		ObjectMeta: meta.ObjectMeta{
+			Name: tc.workloadNamespace,
+		},
+		Subjects: []rbac.Subject{{
+			Kind:      rbac.ServiceAccountKind,
+			Name:      tc.workloadNamespace,
+			Namespace: tc.workloadNamespace,
+		}},
+		RoleRef: rbac.RoleRef{
+			APIGroup: rbac.GroupName,
+			Kind:     "ClusterRole",
+			Name:     tc.workloadNamespace,
+		},
+	}
+	_, err := tc.client.K8s.RbacV1().ClusterRoleBindings().Create(ctx, &crb, meta.CreateOptions{})
+	if err != nil && !apierrors.IsAlreadyExists(err) {
+		return fmt.Errorf("unable to create cluster role: %w", err)
+	}
+	return nil
+}
+
 // sshSetup creates all the Kubernetes resources required to SSH into a Windows node
 func (tc *testContext) sshSetup() error {
 	if err := tc.ensureTestRunnerSA(); err != nil {
@@ -365,6 +415,21 @@ func (tc *testContext) sshSetup() error {
 		return fmt.Errorf("error ensuring Role created: %w", err)
 	}
 	if err := tc.ensureTestRunnerRoleBinding(); err != nil {
+		return fmt.Errorf("error ensuring RoleBinding created: %w", err)
+	}
+	return nil
+}
+
+// ensureTestRunnerRBAC creates the RBAC resources required for the test runner service account
+func (tc *testContext) ensureTestRunnerRBAC() error {
+	ctx := context.TODO()
+	if err := tc.ensureTestRunnerSA(); err != nil {
+		return fmt.Errorf("error ensuring SA created: %w", err)
+	}
+	if err := tc.ensureTestRunnerClusterRole(ctx); err != nil {
+		return fmt.Errorf("error ensuring Role created: %w", err)
+	}
+	if err := tc.ensureTestRunnerClusterRoleBinding(ctx); err != nil {
 		return fmt.Errorf("error ensuring RoleBinding created: %w", err)
 	}
 	return nil
--- a/test/e2e/validation_test.go
+++ b/test/e2e/validation_test.go
@@ -311,7 +311,7 @@ func (tc *testContext) ensureTestRunnerSA() error {
 
 // ensureTestRunnerRole ensures the proper Role exists, a requirement for SSHing into a Windows node
 // noop if the Role already exists.
-func (tc *testContext) ensureTestRunnerRole() error {
+func (tc *testContext) ensureTestRunnerRole(ctx context.Context) error {
 	role := rbac.Role{
 		TypeMeta:   meta.TypeMeta{},
 		ObjectMeta: meta.ObjectMeta{Name: tc.workloadNamespace},
@@ -324,7 +324,7 @@ func (tc *testContext) ensureTestRunnerRole() error {
 			},
 		},
 	}
-	_, err := tc.client.K8s.RbacV1().Roles(tc.workloadNamespace).Create(context.TODO(), &role, meta.CreateOptions{})
+	_, err := tc.client.K8s.RbacV1().Roles(tc.workloadNamespace).Create(ctx, &role, meta.CreateOptions{})
 	if err != nil && !apierrors.IsAlreadyExists(err) {
 		return fmt.Errorf("unable to create role: %w", err)
 	}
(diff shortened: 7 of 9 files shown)


pkg/metadata/metadata.go
```diff
@@ -23,6 +23,9 @@ const (
 	DesiredVersionAnnotation = "windowsmachineconfig.openshift.io/desired-version"
 	// RebootAnnotation indicates the node's underlying instance needs to be restarted
 	RebootAnnotation = "windowsmachineconfig.openshift.io/reboot-required"
+	// UnhealthyAnnotation indicates the node's underlying instance is unhealthy i.e. not ready, scheduling disabled or
+	// meet other unhealthy status conditions like OutOfDisk, MemoryPressure, etc.
+	UnhealthyAnnotation = "windowsmachineconfig.openshift.io/unhealthy"
```
sebsoto, line 28: This information is already captured as part of the NodeStatus which is managed by the kubelet.
We should not be updating this annotation everytime the status changes. That is what is implied by this annotation existing.

This annotation should be strictly for the status of upgrading.

sebsoto: I will continue with the review, mentally replacing unhealthy with upgrading

controllers/controllers.go
```diff
@@ -287,3 +296,39 @@ func markAsFreeOnSuccess(c client.Client, watchNamespace string, recorder record
...
 	return err
 }
+
+// markNodeAsUnhealthy marks the given node as unhealthy by adding an annotation to it. If the number of unhealthy nodes
+// exceeds the maximum allowed, an error is returned.
+func markNodeAsUnhealthy(c client.Client, node *core.Node) error {
+	nodeLocker.Lock()
+	defer nodeLocker.Unlock()
+
+	ctx := context.TODO()
```
sebsoto, line 311: we should start passing context's into functions rather than using context.TODO everywhere, please add it as a parameter for `markNodeAsUnhealthy`

controllers/controllers.go
```diff
@@ -287,3 +296,39 @@ func markAsFreeOnSuccess(c client.Client, watchNamespace string, recorder record
...
+		// check annotation present with value set to true
+		if ok && v == "true" {
+			unhealthyCount++
+			// check if node is already marked as unhealthy
+			if currentNode.Name == node.Name {
+				// node is unhealthy, nothing to do
+				return nil
+			}
+		}
+	}
```
sebsoto, line 329: You can break this into a separate function which looks for and returns the node with the annotation.

You can then return from this function if the returned node is the same as the node being reconciled.

sebsoto (commented):
Thanks jose. I don't agree with the use of `unhealthy`, but besides from that the general direction looks good. Lets work through this, and once you take this out of WIP, I will review the e2e tests.

pkg/metadata/metadata.go
aravindhp: In addition, when we we cordon the node causing scheduling Pods on it to be disabled, it does not imply it is unhealthy but rather WMCO is working on it. We do not want to alarm the end user with the unhealthy verbiage.

pkg/metadata/metadata.go
```diff
@@ -121,6 +124,11 @@ func ApplyRebootAnnotation(ctx context.Context, c client.Client, node core.Node)
 	return ApplyLabelsAndAnnotations(ctx, c, node, nil, map[string]string{RebootAnnotation: ""})
 }
 
+// ApplyUnhealthyAnnotation applies an annotation to the given Node communicating that the instance is unhealthy
+func ApplyUnhealthyAnnotation(ctx context.Context, c client.Client, node core.Node) error {
```
aravindhp, line 128: I assume you will include this change in the commit where these method will actually be used. I know you mentioned this is a WIP PR but calling this out so that there is no extra work for you in the future.

controllers/controllers.go
aravindhp: +1

controllers/controllers.go
```diff
@@ -33,6 +34,11 @@ const (
 	maxUnhealthyCount = 1
 )
 
+var (
+	// nodeLocker is used to synchronize updates across nodes from multiple controllers
+	nodeLocker sync.Mutex
```
aravindhp, line 39: What was the result of looking into marking the controllers as busy and reusing that in this context also?

hack/run-ci-e2e-test.sh
```diff
@@ -181,6 +181,13 @@ if [[ "$TEST" = "upgrade-setup" ]]; then
   # Run the storage test, skipping deletion of the created workload in order to test that it persists across the upgrade
   go test ./test/e2e/... -run=TestWMCO/storage -v -timeout=15m -args $GO_TEST_ARGS --skip-workload-deletion=true
   go test ./test/e2e/... -run=TestWMCO/create/Node_Logs -v -timeout=10m -args $GO_TEST_ARGS
+
+  # make the kubeconfig available in the test namespace as a secret so that pods can communicate
+  # with the test cluster. KUBECONFIG must set by in CI and local tests configuration
```
aravindhp, line 186: Please mention why a Pod needs to communicate with the test cluster

test/e2e/upgrade_test.go
```diff
@@ -282,3 +331,98 @@ func (tc *testContext) waitForCSINodesWithDrivers(nodes []v1.Node) error {
 		return true, nil
 	})
 }
+
+// deployUnhealthyNodesCheckerJob deploys a Job that checks the number of nodes marked as unhealthy and fails if the
+// number of unhealthy nodes exceeds a threshold
```
aravindhp, line 336: Why does this have be done with a Pod?

test/e2e/upgrade_test.go
```diff
@@ -282,3 +331,98 @@ func (tc *testContext) waitForCSINodesWithDrivers(nodes []v1.Node) error {
...
+	os := &v1.PodOS{Name: v1.Linux}
+	selector := map[string]string{"kubernetes.io/os": "linux"}
+	runAsNonRootTrue := true
+	checkerCommand := `
+		#!/bin/bash
+		set -euo pipefail
+		# set working directory
+		cd /tmp
+		# download oc
+		curl -sL https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/latest/openshift-client-linux.tar.gz -o oc.tar.gz && tar xzvf oc.tar.gz &>/dev/null
```
aravindhp, line 347: I am a nack on this. At the minimum we should be using an image that has oc.

aravindhp (changes requested):

pkg/metadata/metadata.go
jrvaldes: Makes sense. I had mixed feelings about using `unhealthy` for the anno too. I ended up chasing the idea behind `maxUnhealthyCount`. 

Happy to refactor the anno as: `windowsmachineconfig.openshift.io/upgrading` or `windowsmachineconfig.openshift.io/upgrade-in-progress`.

LMK

controllers/controllers.go
jrvaldes: True and no context is available from the caller block. So it is basically moving the `context.TODO()` to the function call. Is that what you mean? I don't think I can afford the level of refactoring to also make `ensureInstanceIsUpToDate()` context aware, at least in this PR.

pkg/metadata/metadata.go
jrvaldes: Correct, I prefixed the commit accordingly [fixup: [metadata] Introduce unhealthy annotation](https://github.com/openshift/windows-machine-config-operator/pull/1901/commits/3280d32ae3add909ac08b826d4e6daea2a4d7042)

test/e2e/upgrade_test.go
jrvaldes: Agree; just need to find the right one that works both in CI and with local clusters.

jrvaldes: Any recommendations?

test/e2e/upgrade_test.go
aravindhp: I don't understand your explanation. You should be able to make the API calls directly from the Pod from where we are running the tests without having to launch a Job on the test cluster. That by definition is external. This is similar to [testNodesBecomeReadyAndSchedulable()](https://github.com/openshift/windows-machine-config-operator/blob/d775f625191d16730f8e86d42f98a9ebbe921753/test/e2e/validation_test.go#L59C27-L59C27). If you want to run something in the background, you can use a Go subroutine and channels.

test/e2e/upgrade_test.go
aravindhp: Hopefully this won't be required as I really don't see a need for a Job here.

test/e2e/upgrade_test.go
jrvaldes: > I don't understand your explanation

I can set up a meeting for more details. LMK

WRT running in the background with  Go subroutine and channels take into account there is no subroutine and main thread in this case. The `vsphere-e2e-upgrade` job  a) [setup the upgrade](https://github.com/openshift/windows-machine-config-operator/blob/master/hack/run-ci-e2e-test.sh#L176), b) [triggers the upgrade](https://github.com/openshift/release/blob/master/ci-operator/config/openshift/windows-machine-config-operator/openshift-windows-machine-config-operator-master.yaml#L177) and c) [tests the upgrade](https://github.com/openshift/windows-machine-config-operator/blob/master/hack/run-ci-e2e-test.sh#L186C4-L186C4),  where the steps are scattered between the release repo and wmco e2e test suite.

jrvaldes:
/test vsphere-e2e-upgrade

Install failed.

controllers/controllers.go
jrvaldes: Yes, I considered such scenarios, and I don't anticipate any effect on each other. 

The existing operator condition lock plays over the list of controllers. On the other hand,  the proposed lock synchronizes the API calls to get and set the upgrade annotation.

controllers/controllers.go
```diff
@@ -33,6 +34,11 @@ const (
 	maxUnhealthyCount = 1
 )
 
+var (
+	// nodeLocker is used to synchronize updates across nodes from multiple controllers
```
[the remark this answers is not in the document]
jrvaldes: Good point.

pkg/metadata/metadata.go
sebsoto: ` windowsmachineconfig.openshift.io/upgrading` is preferred for me as well

controllers/controllers.go
sebsoto: Yep all I'm asking is for the addition of context to the parameters of this one function.
Not asking for a refactor.

jrvaldes: ACK

pkg/metadata/metadata.go
jrvaldes: Same here.

test/e2e/upgrade_test.go
aravindhp: Yes, please set up a meeting

jrvaldes: Updated https://github.com/openshift/windows-machine-config-operator/pull/1901/commits/830c233d53aef2fe256581deec0949fb1b569ef0 with details in the commit message.

controllers/controllers.go
```diff
@@ -281,3 +296,55 @@ func markAsFreeOnSuccess(c client.Client, watchNamespace string, recorder record
...
+// performing upgrades in parallel exceeds the maximum allowed, an error is returned.
+func markNodeAsUpgrading(ctx context.Context, c client.Client, node *core.Node) error {
+	controllerLocker.Lock()
+	defer controllerLocker.Unlock()
+
+	upgradingNodes, err := findUpgradingNodes(ctx, c, node)
+	if err != nil {
+		return err
+	}
+	if len(upgradingNodes) == 1 && upgradingNodes[0].Name == node.Name {
```
mansikulkarni96, line 310: we should only try to access property of upgradingNodes if we can ensure no of upgradingNodes are greater than 0.

jrvaldes: @mansikulkarni96 I don't fully understand your comment. Do you anticipate any issues?

Here the code is treating a special case of ` findUpgradingNodes` as documented in the function' [comment](https://github.com/openshift/windows-machine-config-operator/pull/1901/files#diff-46c1e02dfda1ba21d1ccabdf287921220f29dfdd6a75639fbf8a1a95bce01782R327) where a single-element collection is returned if the given node is found.

jrvaldes:
/test vsphere-e2e-upgrade

Could not resolve host github.com 😣

controllers/controllers.go
```diff
@@ -281,3 +296,55 @@ func markAsFreeOnSuccess(c client.Client, watchNamespace string, recorder record
...
+	if err := metadata.ApplyUpgradingAnnotation(ctx, c, *node); err != nil {
+		return err
+	}
+	return nil
+}
+
+// findUpgradingNodes returns a list of nodes that are currently upgrading i.e. have the upgrading annotation set to
+// true. As shortcut-circuit if the given node is already marked as upgrading, it is returned as the only element
+// in the list.
+func findUpgradingNodes(ctx context.Context, c client.Client, node *core.Node) ([]core.Node, error) {
```
sebsoto, line 329: This function is doing too much. You can tell as the doc is having to explain behavior that contradicts the name `findUpgradingNodes`.

Please change this to just return a list of upgrading nodes, the caller can then determine if the node is in the list, and handle that correct.

controllers/controllers.go
```diff
@@ -281,3 +296,55 @@ func markAsFreeOnSuccess(c client.Client, watchNamespace string, recorder record
...
+	// get nodes with upgrading annotation
+	nodeList := &core.NodeList{}
+	if err := c.List(ctx, nodeList, client.MatchingLabels{core.LabelOSStable: "windows"}); err != nil {
+		return nil, fmt.Errorf("error listing Windows nodes: %w", err)
+	}
+	upgradingNodes := make([]core.Node, 0)
+	// check if any nodes are upgrading
+	for _, currentNode := range nodeList.Items {
+		v, ok := currentNode.Annotations[metadata.UpgradingAnnotation]
+		// check annotation present with value set to true
```
sebsoto, line 339: unncessary comment, doesn't give any extra information that is not already obvious with `if ok && v == "true"`

controllers/controllers.go
```diff
@@ -281,3 +296,55 @@ func markAsFreeOnSuccess(c client.Client, watchNamespace string, recorder record
...
+		if ok && v == "true" {
+			// collect upgrading nodes
```
sebsoto, line 341: unnecessary comment `append(upgradingNodes, currentNode)` makes this already clear

controllers/controllers.go
sebsoto: In fact this whole function can be removed if a label was used instead of an annotation. [Labels should be used when the use case is to find collections of objects that satisfy certain conditions](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#attaching-metadata-to-objects). Which is the case here.

That simplifies this to:
```
	if err := c.List(ctx, nodeList, client.MatchingLabels{core.LabelOSStable: "windows", metadata.UpgradingAnnotation: "true"}); err != nil {
		return nil, fmt.Errorf("error listing Windows nodes: %w", err)
	}
```

test/e2e/network_test.go
```diff
@@ -696,10 +696,13 @@ func (tc *testContext) createJob(name, image string, command []string, selector
 	return job, err
 }
 
-// deleteJob deletes the job with the given name
+// deleteJob deletes the job with the given name and the dependent pods
```
sebsoto, line 707: Why is this required to fix this bug? Please state this in the commit message

test/e2e/upgrade_test.go
```diff
@@ -219,6 +226,57 @@ func TestUpgrade(t *testing.T) {
 	// test that any workloads deployed on the node have not been broken by the upgrade
 	t.Run("Workloads ready", tc.testWorkloadsAvailable)
 	t.Run("Node Logs", tc.testNodeLogs)
+	t.Run("Parallel Upgrades Checker", tc.testParallelUpgradesChecker)
+
+}
+
+// TestSetupUpgrade prep the cluster to run the upgrade tests
+func TestSetupUpgrade(t *testing.T) {
```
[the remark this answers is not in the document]
sebsoto: Yeah neither this nor `testSetupParallelUpgradesChecker` is a test. It seems like you are using this naming scheme to be able to call them through `go test`. That seems a bit off. I'm leaning towards it being better to define the job in its own yaml file and deploy it via oc in hack/run-ci-e2e-test.sh .

test/e2e/upgrade_test.go
```diff
@@ -241,3 +299,154 @@ func (tc *testContext) testWorkloadsAvailable(t *testing.T) {
...
+func setupParallelUpgradesCheckerRBAC(c *clusterinfo.OpenShift, ctx context.Context, namespace string) error {
+	if err := createParallelUpgradesCheckerServiceAccount(c, ctx, namespace); err != nil {
+		return err
+	}
+	if err := createParallelUpgradesCheckerClusterRole(c, ctx); err != nil {
+		return err
+	}
+	if err := createParallelUpgradesCheckerClusterRoleBinding(c, ctx, namespace); err != nil {
+		return err
+	}
```
sebsoto, line 345: Why do we need another set of RBAC resources when we already have [a Service account](https://github.com/openshift/windows-machine-config-operator/blob/f452bc77797f5826f7f9c2bb557824dd7cdc91d9/test/e2e/validation_test.go#L302C24-L302C42) for the tests?

test/e2e/network_test.go
```diff
@@ -656,17 +657,16 @@ func (tc *testContext) createWindowsServerJob(name, pwshCommand string, affinity
...
 	command := []string{powerShellExe, "-command", pwshCommand}
-	return tc.createJob(name, windowsServerImage, command, windowsNodeSelector, windowsTolerations, affinity, windowsOS)
+	return tc.createJob(name, true, windowsServerImage, "", command, windowsNodeSelector,
+		windowsTolerations, affinity, windowsOS)
 }
 
 // createJob creates a job on the cluster using the given parameters
-func (tc *testContext) createJob(name, image string, command []string, selector map[string]string,
-	tolerations []v1.Toleration, affinity *v1.Affinity, os *v1.PodOS) (*batchv1.Job, error) {
+func (tc *testContext) createJob(name string, nameGenerated bool, image, serviceAccountName string, command []string,
```
sebsoto, line 665: I'm not convinced this is necessary.
Adding more and more bools to all our test setup functions make them harder to read.

test/e2e/upgrade_test.go
```diff
@@ -219,6 +226,57 @@ func TestUpgrade(t *testing.T) {
...
+	log.Printf("created checker job: %s", checkerJob.GetName())
+}
+
+// testParallelUpgradesChecker tests that the number of parallel upgrades does not exceed the max allowed
+// in the lifetime of the job execution. This test is run after the upgrade is complete.
+func (tc *testContext) testParallelUpgradesChecker(t *testing.T) {
+	// get current Windows node state
+	require.NoError(t, tc.loadExistingNodes(), "error getting the current Windows nodes in the cluster")
+	nodeCount := len(gc.machineNodes) + len(gc.byohNodes)
+	if nodeCount < 2 {
```
sebsoto, line 266: if len(gc.allNodes()) < 2

test/e2e/upgrade_test.go
```diff
@@ -219,6 +226,57 @@ func TestUpgrade(t *testing.T) {
...
+		t.Skipf("Requires 2 or more nodes to run. Found %d nodes", nodeCount)
+	}
+
+	failedPods, err := tc.client.K8s.CoreV1().Pods(tc.workloadNamespace).List(context.TODO(), metav1.ListOptions{
+		LabelSelector: "job-name=" + parallelUpgradesCheckerJobName, FieldSelector: "status.phase=Failed"})
+	require.NoError(t, err)
+	// collect logs
```
sebsoto, line 273: unnecessary comment, this is documenting a function call which is already documented. it is not adding any additional information

test/e2e/upgrade_test.go
```diff
@@ -219,6 +226,57 @@ func TestUpgrade(t *testing.T) {
...
+	tc.writePodLogs("job-name=" + parallelUpgradesCheckerJobName)
+	// check if any pods failed
+	require.Equal(t, 0, len(failedPods.Items), "parallel upgrades check failed",
+		"failed pod count", len(failedPods.Items))
+	// best effort to delete checker job
+	_ = tc.deleteJob(parallelUpgradesCheckerJobName)
```
sebsoto, line 279: this should be deferred at the start of the function. It should also log the error if there is one

test/e2e/upgrade_test.go
```diff
@@ -219,6 +226,57 @@ func TestUpgrade(t *testing.T) {
...
+	if nodeCount < 2 {
+		t.Skipf("Requires 2 or more nodes to run. Found %d nodes", nodeCount)
+	}
+
+	failedPods, err := tc.client.K8s.CoreV1().Pods(tc.workloadNamespace).List(context.TODO(), metav1.ListOptions{
+		LabelSelector: "job-name=" + parallelUpgradesCheckerJobName, FieldSelector: "status.phase=Failed"})
+	require.NoError(t, err)
+	// collect logs
+	tc.writePodLogs("job-name=" + parallelUpgradesCheckerJobName)
+	// check if any pods failed
```
sebsoto, line 275: unnecessary comment, this comment is describing what the require is checking. its not adding any additional context or information.

sebsoto (changes requested):

test/e2e/network_test.go
saifshaikh48: +1 to removing the fixed name introduction if possible

test/e2e/upgrade_test.go
```diff
@@ -241,3 +299,154 @@ func (tc *testContext) testWorkloadsAvailable(t *testing.T) {
...
+
+// createParallelUpgradesCheckerJob deploys a Job that checks the number of nodes marked as upgrading and fails if the
+// number of parallel upgrades exceeds a threshold
+func (tc *testContext) createParallelUpgradesCheckerJob() (*batchv1.Job, error) {
+	checkerCommand := `
+		#!/bin/bash
+		set -euo pipefail
+		
+		# max number of parallel upgrades allowed, fixed to 1. Refer to controllers.MaxParallelUpgrades
+		export MAX_PARALLEL_UPGRADES=1
```
saifshaikh48, line 311: Should we access `controllers.MaxParallelUpgrades` and Sprintf the checkerCommand so we can maintain this in 1 place?

test/e2e/upgrade_test.go
```diff
@@ -241,3 +299,154 @@ func (tc *testContext) testWorkloadsAvailable(t *testing.T) {
...
+			  echo "error: max upgrading count exceeded"
+			  exit 1 
+			fi
+		
+			echo "pass: upgrading count $upgradingCount/$MAX_PARALLEL_UPGRADES"
+			echo "waiting 30s for next check..."
+			sleep 30
+		done
+`
+	// command
```
saifshaikh48, line 326: nit: redundant comment

controllers/controllers.go
jrvaldes: The scenario where  no node is found and an empty list is returned, is the happy path and is covered below.  All Windows nodes are available for upgrade and it is a first-come first-serve approach to mark the node as upgrading, usually (at least in my local tests) the machinecontroller reacts faster and mark the machine instances first, the byoh instances fall behind.

controllers/controllers.go
```diff
@@ -281,3 +296,55 @@ func markAsFreeOnSuccess(c client.Client, watchNamespace string, recorder record
...
+		return err
+	}
+	if len(upgradingNodes) == 1 && upgradingNodes[0].Name == node.Name {
+		// node is already marked as upgrading
+		return nil
+	}
+
+	// check if marking the node as upgrading would exceed the maximum allowed
+	if len(upgradingNodes) >= MaxParallelUpgrades {
+		return fmt.Errorf("cannot mark node %s as upgrading, maximum number of parrallel upgrading nodes reached (%d)",
```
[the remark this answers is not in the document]
jrvaldes: Fixed

controllers/controllers.go
jrvaldes: > This function is doing too much. You can tell as the doc is having to explain behavior that contradicts the name `findUpgradingNodes`.
> 
> Please change this to just return a list of upgrading nodes, the caller can then determine if the node is in the list, and handle that correct.

I went this way to avoid another loop to filter outside of the loop collecting the nodes with the annotation.

jrvaldes: > In fact this whole function can be removed if a label was used instead of an annotation. [Labels should be used when the use case is to find collections of objects that satisfy certain conditions](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#attaching-metadata-to-objects). Which is the case here.
> 
> That simplifies this to:
> 
> ```
> 	if err := c.List(ctx, nodeList, client.MatchingLabels{core.LabelOSStable: "windows", metadata.UpgradingAnnotation: "true"}); err != nil {
> 		return nil, fmt.Errorf("error listing Windows nodes: %w", err)
> 	}
> ```

Agree and it makes sense to me. I am unsure why/how the annotation ended up as the consented solution in past design sessions.

I will transition everything to a label; moving the filtering to the server side is a huge win here— @sebsoto, thanks for emphasizing this.

controllers/controllers.go
```diff
@@ -281,3 +296,55 @@ func markAsFreeOnSuccess(c client.Client, watchNamespace string, recorder record
...
+	nodeList := &core.NodeList{}
+	if err := c.List(ctx, nodeList, client.MatchingLabels{core.LabelOSStable: "windows"}); err != nil {
+		return nil, fmt.Errorf("error listing Windows nodes: %w", err)
+	}
+	upgradingNodes := make([]core.Node, 0)
+	// check if any nodes are upgrading
+	for _, currentNode := range nodeList.Items {
+		v, ok := currentNode.Annotations[metadata.UpgradingAnnotation]
+		// check annotation present with value set to true
+		if ok && v == "true" {
```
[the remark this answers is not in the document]
jrvaldes: Given the value is getting applied, wanted to check the value as well for consistency.

test/e2e/network_test.go
jrvaldes: Before, only the job was deleted as part of the cleanup process, leaving the dependent pods
running the test cluster. With the introduction of the parallel upgrade checker
job, not deleting the pods may result in unreliability while collecting
the number of failed pods with several iterations. 

Thinking this twice is not a hard requirement, and the e2e test can live without it.  Will move it to a separate PR, given it is a good practice to remove all the resources created in the e2e test suite.

test/e2e/network_test.go
jrvaldes: I'm fine reverting it. Went with this to avoid repeating the logic for job definition and creation.

test/e2e/upgrade_test.go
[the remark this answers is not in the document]
jrvaldes: Yes, I'm okay moving the definition of the RBAC and job resources to a YAML and deploying them with `oc`

test/e2e/upgrade_test.go
```diff
@@ -219,6 +226,57 @@ func TestUpgrade(t *testing.T) {
...
+func TestSetupUpgrade(t *testing.T) {
+	tc, err := NewTestContext()
+	require.NoError(t, err)
+
+	t.Run("Setup Parallel Upgrades Checker", tc.testSetupParallelUpgradesChecker)
+}
+
+// testSetupParallelUpgradesChecker tests the max number of parallel upgrades is reached. Requires 2 or more nodes
+// to run
+func (tc *testContext) testSetupParallelUpgradesChecker(t *testing.T) {
```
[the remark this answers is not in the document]
jrvaldes: Good suggestion, but this logic will be removed as is moving into a yaml file and deployed  via oc in hack/run-ci-e2e-test.sh

test/e2e/upgrade_test.go
jrvaldes: This was intentional to maintain an external verification of the controllers.MaxParallelUpgrades, given it has a fixed/constant value. So, this test verifies both the behavior and that the constant has the expected value.

test/e2e/upgrade_test.go
jrvaldes: To narrow the checker job RBAC to the minimum required i.e. listing nodes. 

Do you anticipate any issues? If so, I can expand the rules in the existing service account and use it to avoid deploying a  another set of RBAC resources. 

Keep in mind the job definition will be moved to a  YAML file and deployed directly with `oc`  in the script, so it might make sense to have everything there.

test/e2e/upgrade_test.go
```diff
@@ -241,3 +299,154 @@ func (tc *testContext) testWorkloadsAvailable(t *testing.T) {
...
+	}
+	if err := createParallelUpgradesCheckerClusterRoleBinding(c, ctx, namespace); err != nil {
+		return err
+	}
+	return nil
+}
+
+// createParallelUpgradesCheckerClusterRoleBinding creates the ClusterRoleBinding required for the parallel
+// upgrades checker job
+func createParallelUpgradesCheckerClusterRoleBinding(c *clusterinfo.OpenShift, ctx context.Context,
```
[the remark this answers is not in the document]
jrvaldes: Good point, @openshift/openshift-team-windows-containers  WDYT?

hack/common.sh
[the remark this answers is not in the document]
jrvaldes: > commit msg needs to be updated

Could you add more details? What needs an update?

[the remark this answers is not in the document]
aravindhp: Wouldn't the current AWS upgrade job also simulate this scenario?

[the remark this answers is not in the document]
jrvaldes: @aravindhp  I don't think so. I had mentioned my concerns with `e2e-aws-upgrade`  [before](https://redhat-internal.slack.com/archives/CM4ERHBJS/p1697125234619449?thread_ts=1697125147.515179&cid=CM4ERHBJS).

cc @openshift/openshift-team-windows-containers    WDYT?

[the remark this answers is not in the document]
aravindhp: @jrvaldes the Slack thread you are pointing has no responses. This is something you should have escalated and brought up as a parking lot item. Because what you are now saying is that there is no e2e test available to actually prove that your fix is working.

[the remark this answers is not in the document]
jrvaldes: This PR introduces both, the fix and the e2e test that proves the fix is working. Both are designed for the new `vsphere-e2e-upgrade` job.

jrvaldes:
The `vsphere-e2e-upgrade` job [passed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/1901/pull-ci-openshift-windows-machine-config-operator-master-vsphere-e2e-upgrade/1719401908772278272)  which validates the proposed logic to block parallel upgrades as shown in the[ WMCO logs for the upgrade test step](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/1901/pull-ci-openshift-windows-machine-config-operator-master-vsphere-e2e-upgrade/1719401908772278272/artifacts/vsphere-e2e-upgrade/test/artifacts/wmco.log).

```
2023-10-31T19:21:05Z	INFO	controllers.configmap	instance requires upgrade	{"node": "winhost-1ov30", "version": "9.0.0-2f19add-dirty", "expected version": "10.15.0-de956cb-dirty"}
2023-10-31T19:21:05Z	ERROR	Reconciler error	{"controller": "configmap", "controllerGroup": "", "controllerKind": "ConfigMap", "ConfigMap": {"name":"windows-instances","namespace":"openshift-windows-machine-config-operator"}, "namespace": "openshift-windows-machine-config-operator", "name": "windows-instances", "reconcileID": "5a0ca201-b34a-4c79-bfc7-34ae4eadcaa0", "error": "error configuring host with address 10.93.43.143: cannot mark node winhost-1ov30 as upgrading, maximum number of parallel upgrading nodes reached (1)"}
```

In addition, the[ logs of the proposed checker job](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/1901/pull-ci-openshift-windows-machine-config-operator-master-vsphere-e2e-upgrade/1719401908772278272/artifacts/vsphere-e2e-upgrade/test/artifacts/pods/job-name=parallel-upgrades-checker/job-name=parallel-upgrades-checker.log) depict the sequence of the upgrade workflow where there is only one node at a time with the `upgrading` annotation. 


```
parallel-upgrades-checker-vdcsv:
pass: upgrading count 0/1
waiting 30s for next check...
pass: upgrading count 0/1
waiting 30s for next check...
pass: upgrading count 0/1
waiting 30s for next check...
pass: upgrading count 0/1
waiting 30s for next check...
pass: upgrading count 0/1
waiting 30s for next check...
pass: upgrading count 1/1
waiting 30s for next check...
pass: upgrading count 1/1
waiting 30s for next check...
pass: upgrading count 1/1
waiting 30s for next check...
pass: upgrading count 1/1
waiting 30s for next check...
pass: upgrading count 1/1
waiting 30s for next check...
pass: upgrading count 1/1
waiting 30s for next check...
pass: upgrading count 1/1
waiting 30s for next check...
pass: upgrading count 1/1
waiting 30s for next check...
pass: upgrading count 1/1
waiting 30s for next check...
pass: upgrading count 1/1
waiting 30s for next check...
pass: upgrading count 1/1
waiting 30s for next check...
pass: upgrading count 1/1
waiting 30s for next check...
pass: upgrading count 0/1
waiting 30s for next check...
pass: upgrading count 0/1
waiting 30s for next check...
pass: upgrading count 0/1
waiting 30s for next check...
pass: upgrading count 1/1
waiting 30s for next check...
pass: upgrading count 1/1
waiting 30s for next check...
pass: upgrading count 1/1
waiting 30s for next check...
pass: upgrading count 1/1
waiting 30s for next check...
pass: upgrading count 1/1
waiting 30s for next check...
pass: upgrading count 1/1
waiting 30s for next check...
pass: upgrading count 1/1
waiting 30s for next check...
pass: upgrading count 1/1
waiting 30s for next check...
pass: upgrading count 1/1
waiting 30s for next check...
pass: upgrading count 1/1
waiting 30s for next check...
pass: upgrading count 1/1
waiting 30s for next check...
pass: upgrading count 1/1
waiting 30s for next check...
pass: upgrading count 1/1
waiting 30s for next check...
pass: upgrading count 0/1
```

hack/common.sh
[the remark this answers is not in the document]
jrvaldes: > Can we get a test run on this PR? I would like to see some artifacts.


@mansikulkarni96  the  vsphere-e2e-upgrade job [passed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/1901/pull-ci-openshift-windows-machine-config-operator-master-vsphere-e2e-upgrade/1719401908772278272) artifacts are available. PTAL

test/e2e/validation_test.go
```diff
@@ -370,6 +420,21 @@ func (tc *testContext) sshSetup() error {
 	return nil
 }
 
+// ensureTestRunnerRBAC creates the RBAC resources required for the test runner service account
+func (tc *testContext) ensureTestRunnerRBAC() error {
```
sebsoto, line 424: This function is duplicating https://github.com/jrvaldes/windows-machine-config-operator/blob/4bbebad3cc022d157e62d9b2578be289ef960e9d/test/e2e/validation_test.go#L276

please expand and rename that rather than creating this new function

sebsoto: True, it is not duplicating all the exact code, however it is a duplication in intended function.
The PR is making it so there are two separate functions which are creating groups of resources for the test-runner service account. You're introducing a function `ensureTestRunnerRBAC` which does not actually create all the necessary RBAC resources for test-runner.

Whether or not this is a "refactor" is inconsequential, this PR adds additional RBAC resource to be created for the SA. It makes sense to have a function `ensureTestRunnerRBAC()`, but it should create all the resources, including those currently present in `sshSetup()`, which can be removed in favor of this new function

aravindhp: Just want to drop a general note, that while we should refrain from refactoring WMCO **product** code as part of a bug fix unless really required, it is fine to refactor **test** code which by definition is not part of the product. Unless of course there is a time crunch involved. IOW, context is important when it comes to making such decisions and not just blindly applying "rules".

jrvaldes:
Depends on:
- https://github.com/openshift/release/pull/45161

test/e2e/validation_test.go
jrvaldes: > True, it is not duplicating all the exact code, however it is a duplication in intended function. The PR is making it so there are two separate functions which are creating groups of resources for the test-runner service account. You're introducing a function `ensureTestRunnerRBAC` which does not actually create all the necessary RBAC resources for test-runner.
> 
> Whether or not this is a "refactor" is inconsequential, this PR adds additional RBAC resource to be created for the SA. It makes sense to have a function `ensureTestRunnerRBAC()`, but it should create all the resources, including those currently present in `sshSetup()`, which can be removed in favor of this new function



Correct, I agree.

jrvaldes: Addressed in https://github.com/openshift/windows-machine-config-operator/pull/1901/commits/03ce26656619fd8b39712bd55d8d090e51e83cbc

hack/e2e/resources/parallel-upgrade-checker-job.yaml
```diff
@@ -0,0 +1,52 @@
...
+              while true; do
+              	upgradingCount=$(oc get nodes -l kubernetes.io/os=windows  -o jsonpath='{.items[*].metadata.labels.windowsmachineconfig\.openshift\.io/upgrading}' | wc -w)	
+              	if [[ $upgradingCount -gt $MAX_PARALLEL_UPGRADES ]]; then
+              	  echo "error: max upgrading count exceeded"
+              	  exit 1 
+              	fi
+
+              	echo "pass: upgrading count $upgradingCount/$MAX_PARALLEL_UPGRADES"
+              	echo "waiting 30s for next check..."
+              	sleep 30
```
sebsoto, line 48: 30 seconds is a big window. Its unlikely, but possible to miss an upgrade here. Especially if WMCO becomes more efficient in the future. How about 5 seconds?

sebsoto (approved):
Thanks @jrvaldes, great job with this PR.
I have approved with a comment, please address it before removing the hold.

hack/e2e/resources/parallel-upgrade-checker-job.yaml
saifshaikh48: i lean towards more, maybe 10 or 15 seconds. 5 seconds seems very quick for upgrade polling, especially since there is no risk of missing something i.e. no parallel upgrades

sebsoto: How is there no risk of missing something?
Even if we don't miss a whole upgrade, we need to test there is no overlap.

sebsoto: Its better to be safer than not. There is no penalty I can think of to having a few dozen extra checks.

jrvaldes:
/test azure-e2e-upgrade

ipi-install  failed.

hack/e2e/resources/parallel-upgrade-checker-job.yaml
jrvaldes: Addressed in [a2a6f6b](https://github.com/openshift/windows-machine-config-operator/commit/a2a6f6b28ae61e28f226a62d5501d1e8c26acdea)

jrvaldes:
@openshift/openshift-team-windows-containers  the azure-e2e-upgrade     job [passed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/1901/pull-ci-openshift-windows-machine-config-operator-master-azure-e2e-upgrade/1720209089482461184) see the [checker pod logs](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/1901/pull-ci-openshift-windows-machine-config-operator-master-azure-e2e-upgrade/1720209089482461184/artifacts/azure-e2e-upgrade/test/artifacts/pods/job-name=parallel-upgrades-checker/job-name=parallel-upgrades-checker.log) doing the upgrade check every 5s

saifshaikh48 (commented):
/lgtm

jrvaldes:
Not sure why the stage-2 pipeline is not triggered, i.e. the `/test remaing-required` added by the bot.

aravindhp:
> Not sure why the stage-2 pipeline is not triggered, i.e. the `/test remaing-required` added by the bot.

It is possible that the new workflow that @jmguzik was implementing is now in effect I.e. we have to manually add the `/test remaing-required` comment

jrvaldes:
> > Not sure why the stage-2 pipeline is not triggered, i.e. the `/test remaing-required` added by the bot.
> 
> It is possible that the new workflow that @jmguzik was implementing is now in effect I.e. we have to manually add the `/test remaing-required` comment

So, the default is `Opt out the bot dropping the command`.   I will add the command. 🤞🏼

jrvaldes:
/test remaing-required


to trigger stage-2 jobs. PR LGTM'ed and Approved

aravindhp:
Looks like taking the PR out of draft did not remove the WIP label.

aravindhp:
> Looks like taking the PR out of draft did not remove the WIP label.

I am writing this off due to all the strangeness we are seeing on this PR but if it happens consistently please open a bug with the test platform team

jmguzik:
> It is possible that the new workflow that @jmguzik was implementing is now in effect I.e. we have to manually add the /test remaing-required comment

Not yet ready but occasional misses can happen, we are heavily affected by gh rate limiting.

PR #1917, opened 2023-10-30
OCPBUGS-22680: Ensure WICD service account token

jrvaldes:
This change updates the logic that looks for the WICD secret token to avoid relying on the naming convention and proposes a more robust approach by matching the secret type and the corresponding service account name.

Ensures the WICD service account token is present by creating one if not found so that the WICD kubeconfig can be created from the secret token.

### DIFF
--- a/pkg/nodeconfig/nodeconfig.go
+++ b/pkg/nodeconfig/nodeconfig.go
@@ -35,6 +35,7 @@ import (
 	"github.com/openshift/windows-machine-config-operator/pkg/metadata"
 	"github.com/openshift/windows-machine-config-operator/pkg/nodeutil"
 	"github.com/openshift/windows-machine-config-operator/pkg/retry"
+	"github.com/openshift/windows-machine-config-operator/pkg/secrets"
 	"github.com/openshift/windows-machine-config-operator/pkg/windows"
 	"github.com/openshift/windows-machine-config-operator/version"
 )
@@ -293,20 +294,56 @@ func (nc *nodeConfig) getWICDServiceAccountSecret() (*core.Secret, error) {
 		return nil, err
 	}
 	// Go through all the secrets in the WMCO namespace, and find the token secret which contains the auth credentials
-	// for the WICD ServiceAccount. This secret's name will always have the form:
-	// ${service_account_name}-token-${random_string}.
-	tokenSecretPrefix := "windows-instance-config-daemon-token-"
+	// for the WICD ServiceAccount
 	var filteredSecrets []core.Secret
 	for _, secret := range secrets.Items {
-		if strings.HasPrefix(secret.Name, tokenSecretPrefix) {
+		if secret.Type != core.SecretTypeServiceAccountToken {
+			// skip non-serviceAccount token secrets
+			continue
+		}
+		if secret.Annotations[core.ServiceAccountNameKey] == windows.WicdServiceName {
 			filteredSecrets = append(filteredSecrets, secret)
 		}
 	}
-	if len(filteredSecrets) != 1 {
-		return nil, fmt.Errorf("expected 1 secret with '%s' prefix, found %d", tokenSecretPrefix, len(filteredSecrets))
+	if len(filteredSecrets) == 1 {
+		return &filteredSecrets[0], nil
 	}
-	return &filteredSecrets[0], nil
+	if len(filteredSecrets) > 1 {
+		return nil, fmt.Errorf("expected 1 secret for SA '%s', found %d", windows.WicdServiceName,
+			len(filteredSecrets))
+	}
+	// no secret token found for WICD service account, create one
+	return nc.createWICDServiceAccountTokenSecret()
+}
 
+// createWICDServiceAccountTokenSecret creates a secret with a long-lived API token for the WICD ServiceAccount and
+// waits for the secret data to be populated
+func (nc *nodeConfig) createWICDServiceAccountTokenSecret() (*core.Secret, error) {
+	ctx := context.TODO()
+	err := nc.client.Create(ctx, secrets.GenerateServiceAccountTokenSecret(nc.wmcoNamespace, windows.WicdServiceName))
+	if err != nil {
+		return nil, fmt.Errorf("error creating secret for WICD ServiceAccount: %w", err)
+	}
+	secret := &core.Secret{}
+	// wait for the secret data to be populated
+	err = wait.PollUntilContextTimeout(ctx, time.Second, time.Minute, true,
+		func(ctx context.Context) (done bool, err error) {
+			secret, err = nc.k8sclientset.CoreV1().Secrets(nc.wmcoNamespace).Get(ctx, windows.WicdServiceName,
+				meta.GetOptions{})
+			if err != nil {
+				return false, nil
+			}
+			caCert := secret.Data[core.ServiceAccountRootCAKey]
+			if caCert == nil {
+				return false, nil
+			}
+			token := secret.Data[core.ServiceAccountTokenKey]
+			if token == nil {
+				return false, nil
+			}
+			return true, nil
+		})
+	return secret, err
 }
 
 // createBootstrapFiles creates all prerequisite files on the node required to start kubelet using latest ignition spec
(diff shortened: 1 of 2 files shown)


pkg/nodeconfig/nodeconfig.go
```diff
@@ -289,20 +290,41 @@ func (nc *nodeConfig) getWICDServiceAccountSecret() (*core.Secret, error) {
...
+	if len(filteredSecrets) > 1 {
+		return nil, fmt.Errorf("expected 1 secret for SA '%s', found %d", windows.WicdServiceName, len(filteredSecrets))
+	}
+	// no secret token found for WICD service account, create one
+	if err := nc.createWICDServiceAccountToken(); err != nil {
+		return nil, err
 	}
-	return &filteredSecrets[0], nil
+	// retry the secret token for the WICD ServiceAccount
+	return nc.getWICDServiceAccountSecret()
```
saifshaikh48, line 315: would prefer if createWICDServiceAccountToken() returned the Secret obj so this avoids recursion and turns into:
```
	saToken, err := nc.createWICDServiceAccountToken()
	if err != nil {
		return nil, err
	}
	return saToken
```

saifshaikh48: we should also update the `getWICDServiceAccountSecret()` func doc to say it creates the token if it does not exist already

jrvaldes:
> also, do we know what caused this bug to start popping up?

Added a [comment](https://issues.redhat.com/browse/OCPBUGS-22680?focusedId=23371005&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-23371005) in the ticket with some info.

pkg/nodeconfig/nodeconfig.go
jrvaldes: > we should also update the `getWICDServiceAccountSecret()` func doc to say it creates the token if it does not exist already

Good point, will include it.

jrvaldes:
nutanix-e2e-operator job [failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/1917/pull-ci-openshift-windows-machine-config-operator-master-nutanix-e2e-operator/1719003959382773760#1:build-log.txt%3A231) with 
```
=== RUN   TestWMCO/create/Metrics_validation/Windows_Prometheus_rules_validation
    metrics_test.go:157: 
        	Error Trace:	/go/src/github.com/openshift/windows-machine-config-operator/test/e2e/metrics_test.go:157
        	            	could not get 'prometheus-k8s-token' secret
        	Test:       	TestWMCO/create/Metrics_validation/Windows_Prometheus_rules_validation
        	Messages:   	Error getting Prometheus token
```

This validates the current scenario with the kube-controller-manager configuration, where Secret-based Service Account Tokens are not automatically generated.

jrvaldes:
Depends on:
- https://github.com/openshift/windows-machine-config-operator/pull/1955

Alternatively, I can reframe this PR to accommodate both commits.

sebsoto (changes requested):
Thanks @jrvaldes 
Please make sure to create the follow up ticket to manage the tokens ourselves.

saifshaikh48 (commented):
/lgtm

pkg/nodeconfig/nodeconfig.go
```diff
@@ -293,20 +294,58 @@ func (nc *nodeConfig) getWICDServiceAccountSecret() (*core.Secret, error) {
...
+// waits for the secret data to be populated
+func (nc *nodeConfig) createWICDServiceAccountTokenSecret() (*core.Secret, error) {
+	ctx := context.TODO()
+	err := nc.client.Create(ctx, secrets.GenerateServiceAccountTokenSecret(nc.wmcoNamespace, windows.WicdServiceName))
+	if err != nil {
+		return nil, fmt.Errorf("error creating secret for WICD ServiceAccount: %w", err)
+	}
+	secret := &core.Secret{}
+	// wait for the secret data to be populated
+	err = wait.PollUntilContextTimeout(ctx, retry.Interval, retry.Timeout, false,
```
sebsoto, line 329: Please reduce the timeout, 10 minutes is a very long time to be blocking on this.
How long do you expect for this to take? I'd expect this to be almost instant.
Any more than a 30 second to 1 minute wait should cause the reconciliation to fail and requeue.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -293,20 +294,58 @@ func (nc *nodeConfig) getWICDServiceAccountSecret() (*core.Secret, error) {
...
+			continue
+		}
+		if secret.Annotations[core.ServiceAccountNameKey] == windows.WicdServiceName {
 			filteredSecrets = append(filteredSecrets, secret)
 		}
 	}
-	if len(filteredSecrets) != 1 {
-		return nil, fmt.Errorf("expected 1 secret with '%s' prefix, found %d", tokenSecretPrefix, len(filteredSecrets))
+	if len(filteredSecrets) == 1 {
+		return &filteredSecrets[0], nil
```
sebsoto, line 309: Don't you need to check if the secret contains the ServiceAccountRootCAKey?

pkg/nodeconfig/nodeconfig.go
sebsoto: Also please change this to an immediate poll. Not doing so adds in a constant 15 second delay within the reconciliation process that may not be necessary.

sebsoto (changes requested):

pkg/nodeconfig/nodeconfig.go
sebsoto: Ack, thanks for the link

sebsoto (approved):

pkg/nodeconfig/nodeconfig.go
```diff
@@ -293,20 +294,58 @@ func (nc *nodeConfig) getWICDServiceAccountSecret() (*core.Secret, error) {
...
+		return nil, fmt.Errorf("error creating secret for WICD ServiceAccount: %w", err)
+	}
+	secret := &core.Secret{}
+	// wait for the secret data to be populated
+	err = wait.PollUntilContextTimeout(ctx, time.Second, time.Minute, true,
+		func(ctx context.Context) (done bool, err error) {
+			secret, err = nc.k8sclientset.CoreV1().Secrets(nc.wmcoNamespace).Get(ctx, windows.WicdServiceName,
+				meta.GetOptions{})
+			if err != nil {
+				return false, fmt.Errorf("error getting secret %s: %w", windows.WicdServiceName, err)
```
sebsoto, line 334: Sorry missed this, but returning error will cause the loop to break early and not retry anymore.
applies everywhere in this function

pkg/nodeconfig/nodeconfig.go
```diff
@@ -293,20 +294,58 @@ func (nc *nodeConfig) getWICDServiceAccountSecret() (*core.Secret, error) {
...
+// waits for the secret data to be populated
+func (nc *nodeConfig) createWICDServiceAccountTokenSecret() (*core.Secret, error) {
+	ctx := context.TODO()
+	err := nc.client.Create(ctx, secrets.GenerateServiceAccountTokenSecret(nc.wmcoNamespace, windows.WicdServiceName))
+	if err != nil {
+		return nil, fmt.Errorf("error creating secret for WICD ServiceAccount: %w", err)
+	}
+	secret := &core.Secret{}
+	// wait for the secret data to be populated
+	err = wait.PollUntilContextTimeout(ctx, time.Second, time.Minute, true,
```
sebsoto, line 329: nit: a second is a bit frequent. Only an issue if this function ends up logging something.

saifshaikh48 (commented):
/lgtm

jrvaldes:

ipi-install [failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/1917/pull-ci-openshift-windows-machine-config-operator-master-vsphere-proxy-e2e-operator/1733140090361745408)

PR #1945, opened 2023-11-16
Do not merge: manual pipelines test

jmguzik:


jmguzik:
it seems new config is working and wmco is no longer using auto pipelines

PR #1952, opened 2023-11-16
[hack] Introduce merge-order.sh

saifshaikh48:
Script to get merge order of PRs that merged since a specific branch was blocked for release reasons. Also introduces a util script for printing and validations.

### DIFF
--- a/hack/errata-check.sh
+++ b/hack/errata-check.sh
@@ -25,14 +25,11 @@ set -euo pipefail
 # * Update your local clone and branches
 # * Run this script ./hack/errata-check.sh <branch_name>
 
+WMCO_ROOT=$(dirname "${BASH_SOURCE}")/..
+source $WMCO_ROOT/hack/util.sh
+
 # Set variables
 ERRATA_API_URL="https://errata.devel.redhat.com/api/v1/erratum"
-RED=$(tput setaf 1)
-GREEN=$(tput setaf 2)
-YELLOW=$(tput setaf 3)
-BLUE=$(tput setaf 4)
-BRIGHT=$(tput bold)
-NORMAL=$(tput sgr0)
 
 # Define functions
 function help() {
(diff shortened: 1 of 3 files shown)


hack/merge-order.sh
```diff
@@ -0,0 +1,94 @@
...
+if [[ ${total_count} -ne 1 ]]; then
+  print-error "Expected one issue for branch $TARGET_BRANCH, found $total_count instead."
+  exit 1
+fi
+
+print-label "Found blocking Github issue for branch" "$TARGET_BRANCH"
+echo "$rel_issue" | jq -r '[.[] | {Issue_Number: .number, Title: .title, Author: .author.login, Created_Date: .createdAt, URL: .url}] | (first | keys_unsorted) as $keys | $keys, map([.[ $keys[] ]])[] | @tsv' | column -t -s $'\t' | \
+    awk -F'\t' 'NR==1 { printf ("\033[1;33m%s\t%s\t%s\t%s\t%s\033[0m\n", $1, $2, $3, $4, $5) } NR>1 { print }'
+
+date_created=$(date -d "$(echo "${rel_issue}" | jq -r ".[].createdAt")" +"%Y-%m-%dT%H:%M:%SZ")
```
saifshaikh48, line 62: If we modify our process to no longer have a release issue, we would just need to look at the errata instead to get this date -- something like querying the errata for latest date of state transition away from NEW_FILES

hack/util.sh
```diff
@@ -0,0 +1,55 @@
...
+  local text=$1
+  printf "\n${RED}ERROR: %s${NORMAL}\n" "$text" 1>&2
+}
+
+function print_separator() {
+  echo -e "\n-------------------------\n"
+}
+
+# validate-target-branch returns the branch version if it's valid, errirs out
+function validate-target-branch {
```
jrvaldes, line 41: This can be moved where it is getting used. e.i.  merge-order.sh

hack/util.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
```
jrvaldes, line 1: Not a fan of utilities; they tend to grow with unrelated stuff making the file hard to read/maintain. 

In case we really need the fancy messages, I recommend moving the styling to its own file or to common.sh

hack/errata-check.sh
```diff
@@ -25,14 +25,11 @@ set -euo pipefail
...
+source $WMCO_ROOT/hack/util.sh
+
 # Set variables
 ERRATA_API_URL="https://errata.devel.redhat.com/api/v1/erratum"
-RED=$(tput setaf 1)
-GREEN=$(tput setaf 2)
-YELLOW=$(tput setaf 3)
-BLUE=$(tput setaf 4)
-BRIGHT=$(tput bold)
-NORMAL=$(tput sgr0)
```
jrvaldes, line 35: This refactor may warrant a separate commit.

hack/merge-order.sh
```diff
@@ -0,0 +1,94 @@
...
+    print-error "Invalid arguments, no branch provided"
+    help
+    exit 1
+fi
+
+TARGET_BRANCH=$1
+BRANCH_VERSION=$(validate-target-branch "$TARGET_BRANCH")
+
+# Check for a blocking issue for the given branch
+rel_issue=$(gh issue list --state open --search "Merge blocker for active release | branch:$TARGET_BRANCH" \
```
jrvaldes, line 48: This may change, based on the conclusions of [How we manage our build pipeline](https://docs.google.com/document/d/1qE0Ym-syg2uOsEinH0mSujw0MYfUm_GBLZDqRFcTzic/edit)

jrvaldes (commented):
@saifshaikh48  thanks for working on this, PTAL at my comments.

hack/util.sh
saifshaikh48: It's also used in errata-check.sh

hack/util.sh
saifshaikh48: I created util.sh by pulling out some functions from errata-check.sh to re-use in merge-order.sh. I can move all this to common.sh if that's the consensus

hack/merge-order.sh
saifshaikh48: Yep, have an idea for this in https://github.com/openshift/windows-machine-config-operator/pull/1952#discussion_r1397550427

hack/util.sh
jrvaldes: I see, prob common.sh makes more sense.

PR #1955, opened 2023-11-28
Stop using automatically generated service account token in e2e test suite

jrvaldes:
This change replaces the API call to fetch the service account token from the automatically generated secret with a TokenRequest.

As per official docs automatically generated secrets for should be avoided as they might be removed in a future OpenShift Container Platform release.

See https://docs.openshift.com/container-platform/4.14/authentication/using-service-accounts-in-applications.html#auto-generated-sa-token-secrets_using-service-accounts

### DIFF
--- a/test/e2e/metrics_test.go
+++ b/test/e2e/metrics_test.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	auth "k8s.io/api/authentication/v1"
 	core "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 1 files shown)


test/e2e/metrics_test.go
```diff
@@ -216,35 +217,16 @@ func (tc *testContext) testWindowsPrometheusRules(t *testing.T) {
...
-	var secretName string
-	for _, secret := range monitoringSecrets.Items {
-		if strings.Contains(secret.Name, "prometheus-k8s-token") {
-			secretName = secret.Name
-			break
-		}
+	tokenRequest := &auth.TokenRequest{
+		Spec: auth.TokenRequestSpec{},
 	}
+	token, err := tc.client.K8s.CoreV1().ServiceAccounts(monitoringNamespace).CreateToken(context.TODO(),
```
saifshaikh48, line 223: this creates a new token each time it is executed right? do the tokens expire, or do we have to revoke them somehow?

jrvaldes: > this creates a new token each time it is executed right? 

Yes

jrvaldes: > Do we have to revoke them somehow?

No.

> Do the tokens expire, 
Yes, the default ExpirationSeconds is 3600 seconds (1 hour)

mansikulkarni96:
LGTM

test/e2e/metrics_test.go
sebsoto: Please add the token expiration time to the function doc

sebsoto (changes requested):
Thanks @jrvaldes, just have one request

PR #1962, opened 2023-12-01
[bundle] Prep for 9.0.1 release

sebsoto:
Changed version in Makefile, then ran `make bundle`


mtnbikenc:
Shouldn't this use the [hack/pre-release.sh script](https://github.com/openshift/windows-machine-config-operator/blob/master/hack/pre-release.sh)?

PR #1965, opened 2023-12-07
WINC-1158: Prep for 10.16.0 release

wgahnagl:
This change bumps the operator and the OCP version according to WMCO 10.16.0 release.

Ran:
```
  make bundle
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,11 +1,11 @@
 [submodule "ovn-kubernetes"]
 	path = ovn-kubernetes
 	url = https://github.com/openshift/ovn-kubernetes
-	branch = release-4.15
+	branch = release-4.16
 [submodule "containernetworking-plugins"]
 	path = containernetworking-plugins
 	url = https://github.com/openshift/containernetworking-plugins
-	branch = release-4.15
+	branch = release-4.16
 [submodule "promu"]
 	path = promu
 	url = https://github.com/openshift/prometheus-promu
@@ -15,7 +15,7 @@
 [submodule "kubelet"]
 	path = kubelet
 	url = https://github.com/openshift/kubernetes
-	branch = release-4.15
+	branch = release-4.16
 [submodule "kube-proxy"]
 	path = kube-proxy
 	url = https://github.com/openshift/kubernetes
@@ -23,16 +23,16 @@
 [submodule "containerd"]
 	path = containerd
 	url = https://github.com/openshift/containerd
-	branch = release-4.15
+	branch = release-4.16
 [submodule "hcsshim"]
 	path = hcsshim
 	url = https://github.com/openshift/hcsshim
-	branch = release-4.15
+	branch = release-4.16
 [submodule "cloud-provider-azure"]
 	path = cloud-provider-azure
 	url = https://github.com/openshift/cloud-provider-azure
-	branch = release-4.15
+	branch = release-4.16
 [submodule "csi-proxy"]
 	path = csi-proxy
 	url = https://github.com/openshift/csi-proxy
-	branch = release-4.15
+	branch = release-4.16
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,4 +1,5 @@
-FROM registry.ci.openshift.org/openshift/release:rhel-9-release-golang-1.20-openshift-4.16 as build
+FROM registry.ci.openshift.org/openshift/release:rhel-9-release-golang-1.21-openshift-4.16 as build
+
 LABEL stage=build
 
 # Silence go compliance shim output
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.ci.openshift.org/openshift/release:rhel-9-release-golang-1.20-openshift-4.15 as build
+FROM registry.ci.openshift.org/openshift/release:rhel-9-release-golang-1.20-openshift-4.16 as build
 LABEL stage=build
 
 # Silence go compliance shim output
--- a/build/Dockerfile.ci
+++ b/build/Dockerfile.ci
@@ -3,7 +3,7 @@
 # building the operator from the PR source without using the operator-sdk.
 
 # build stage for building binaries
-FROM registry.ci.openshift.org/openshift/release:rhel-9-release-golang-1.20-openshift-4.15 as build
+FROM registry.ci.openshift.org/openshift/release:rhel-9-release-golang-1.20-openshift-4.16 as build
 LABEL stage=build
 WORKDIR /build/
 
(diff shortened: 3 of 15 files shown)


bundle.Dockerfile
```diff
@@ -1,45 +1,21 @@
...
-# backport this content into the old appregistry format and upload it to the quay.io application
-# registry endpoints.
-LABEL com.redhat.delivery.backport=false
-
-# This label gets replaced by render_templates, so it is safer to keep it on its own
-LABEL version="v0.0.0"
-
-# This label maps to the brew build target
-LABEL com.redhat.component="windows-machine-config-operator-bundle-container"
-
```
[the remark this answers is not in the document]
sebsoto: Please undo this change

bundle.Dockerfile
```diff
@@ -1,45 +1,21 @@
...
 LABEL operators.operatorframework.io.bundle.channels.v1=preview,stable
 LABEL operators.operatorframework.io.bundle.channel.default.v1=stable
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.32.0+git
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.32.0
 LABEL operators.operatorframework.io.metrics.mediatype.v1=metrics+v1
 LABEL operators.operatorframework.io.metrics.project_layout=go.kubebuilder.io/v3
 
+# Labels for testing.
+LABEL operators.operatorframework.io.test.mediatype.v1=scorecard+v1
+LABEL operators.operatorframework.io.test.config.v1=tests/scorecard/
```
sebsoto, line 16: This is fine, as long as it doesn't cause any issues.
We don't make use of scorecard, but if make bundle is generating this I don't see a reason to not have it.

config/manifests/bases/windows-machine-config-operator.clusterserviceversion.yaml
```diff
@@ -169,4 +169,4 @@ spec:
   minKubeVersion: 1.28.0
   provider:
     name: Red Hat
-  version: 0.0.0
+  version: 10.16.0
```
sebsoto, line 172: Should this change be happening?

sebsoto (changes requested):

bundle/metadata/annotations.yaml
```diff
@@ -9,7 +9,7 @@ annotations:
   operators.operatorframework.io.metrics.builder: operator-sdk-v1.32.0+git
   operators.operatorframework.io.metrics.mediatype.v1: metrics+v1
   operators.operatorframework.io.metrics.project_layout: go.kubebuilder.io/v3
-  com.redhat.openshift.versions: "=v4.14"
+  com.redhat.openshift.versions: "=v4.16"
```
mtnbikenc, line 12: Moving this to `v4.16` may be an issue with CVP.  I think that is why master was still at `v4.14`, (which should probably now be updated).  @mansikulkarni96 could you confirm my statement?

mansikulkarni96: Correct we can switch to 4.15 now and bump to 4.16 in CY24Q1

.gitmodules
```diff
@@ -15,24 +15,24 @@
 [submodule "kubelet"]
 	path = kubelet
 	url = https://github.com/openshift/kubernetes
-	branch = release-4.15
+	branch = release-4.16
 [submodule "kube-proxy"]
 	path = kube-proxy
 	url = https://github.com/openshift/kubernetes
-	branch = sdn-4.15-kubernetes-1.28.3
+	branch = sdn-4.16-kubernetes-1.28.3
```
mansikulkarni96, line 22: sdn-4.16-kubernetes-1.28.3 is not available yet, please create a ticket to move to it when it's created.

mansikulkarni96 (changes requested):
Thanks for working on this @wgahnagl

saifshaikh48 (commented):
are any of these changes generated by hack scripts? If so I would 
1. break generated changes into their commit
2. include the command that was used in commit message

wgahnagl:
> are any of these changes generated by hack scripts? If so I would
> 
>     1. break generated changes into their commit
> 
>     2. include the command that was used in commit message

Since we changed the versioning system, we'll need to change the hack script to address it, and I figured it would be easier to just do this by hand for now.

build/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.ci.openshift.org/openshift/release:rhel-9-release-golang-1.20-openshift-4.15 as build
+FROM registry.ci.openshift.org/openshift/release:rhel-9-release-golang-1.20-openshift-4.16 as build
```
mansikulkarni96, line 1: We are missing these image updates in build/Dockerfile.ci and rest of the other files.

mansikulkarni96:
Please ensure the midstream Dockerfile is updated to use the 4.16 base image.

mansikulkarni96:
I noticed ART has moved to golang 1.21, please verify if the image is available for use and make the necessary updates.

sebsoto (approved):

saifshaikh48:
@wgahnagl Please create a story to address any required hack script changes so we don't lose track of them

> Since we changed the versioning system, we'll need to change the hack script to address it, and I figured it would be easier to just do this by hand for now.

That's what I meant, split the stuff you updated manually into a separate commit and then manually edit the commit message of the generated commit to include the command you ran to generate it. Was just a suggestion for clarity, not blocking the PR

build/Dockerfile
```diff
@@ -1,4 +1,3 @@
-FROM registry.ci.openshift.org/openshift/release:rhel-9-release-golang-1.20-openshift-4.16 as build
```
mansikulkarni96: is this change intended?

mansikulkarni96 (commented):
Changes LGTM just one question.
nit: It may be easier for reviews if all the image related updates are in a single commit.

wgahnagl:
> Changes LGTM just one question. nit: It may be easier for reviews if all the image related updates are in a single commit.

I was thinking that it would make sense to break out just the golang bump into its own commit, because if it had to be reverted, it would be harder to do if it were split up in a different way.

build/Dockerfile.ci
```diff
@@ -123,7 +123,7 @@ RUN make build-daemon
 #├── windows_exporter.exe
 #└── windows-instance-config-daemon.exe
 
-FROM registry.ci.openshift.org/openshift/release:rhel-9-release-golang-1.20-openshift-4.15
+FROM registry.ci.openshift.org/openshift/release:rhel-9-release-golang-1.20-openshift-4.16
```
mansikulkarni96, line 126: shouldn't this be 1.21?

jrvaldes (changes requested):
@wgahnagl  Thanks for updating the PR. Mostly LGTM. 

Just missing the version bump in [.ci-operator.yaml](https://github.com/openshift/windows-machine-config-operator/blame/master/.ci-operator.yaml) used in CI

jrvaldes (approved):
/lgtm

PR #1979, opened 2023-12-13
OCPBUGS-21606,OCPBUGS-2887: Fix network config script

sebsoto:
The network configuration script contained bad logic to determine if the cni config had the correct contents. The 'expected' config does not contain carriage returns, while the actual created config file does. Because the script did not account for this, the script would always determine that the cni configuration needed to be recreated with the correct contents. This can cause issues with containerd, which needs to reload the configuration everytime this happens.

### DIFF
--- a/pkg/nodeconfig/payload/payload.go
+++ b/pkg/nodeconfig/payload/payload.go
@@ -140,7 +140,7 @@ $cni_template=$cni_template.Replace("provider_address",$provider_address)
 # Compare CNI config with existing file, and replace if necessary
 $existing_config=""
 if(Test-Path -Path CNI_CONFIG_PATH) {
-    $existing_config= Get-Content -Path "CNI_CONFIG_PATH"
+` + "    $existing_config=((Get-Content -Path \"CNI_CONFIG_PATH\" -Raw) -Replace \"`r\",\"\")" + `
 }
 if($existing_config -ne $cni_template){
     Set-Content -Path "CNI_CONFIG_PATH" -Value $cni_template -NoNewline
--- a/pkg/nodeconfig/payload/payload_test.go
+++ b/pkg/nodeconfig/payload/payload_test.go
@@ -74,7 +74,7 @@ $cni_template=$cni_template.Replace("provider_address",$provider_address)
 # Compare CNI config with existing file, and replace if necessary
 $existing_config=""
 if(Test-Path -Path c:\k\cni.conf) {
-    $existing_config= Get-Content -Path "c:\k\cni.conf"
+` + "    $existing_config=((Get-Content -Path \"c:\\k\\cni.conf\" -Raw) -Replace \"`r\",\"\")" + `
 }
 if($existing_config -ne $cni_template){
     Set-Content -Path "c:\k\cni.conf" -Value $cni_template -NoNewline


pkg/nodeconfig/payload/payload_test.go
```diff
@@ -74,7 +74,7 @@ $cni_template=$cni_template.Replace("provider_address",$provider_address)
 # Compare CNI config with existing file, and replace if necessary
 $existing_config=""
 if(Test-Path -Path c:\k\cni.conf) {
-    $existing_config= Get-Content -Path "c:\k\cni.conf"
+` + "    $existing_config=((Get-Content -Path \"c:\\k\\cni.conf\" -Raw) -Replace \"`r\",\"\")" + `
```
jrvaldes, line 77: Does it make sense to test with a cni_template that contains the carriage returns so that the proposed sustitution can be unit tested?

sebsoto: At the moment there is no way I can think of to unit test this script.
Our testing framework isn't set up to do individual tests like this on Windows nodes, and certainly not for parts of the powershell scripts.

sebsoto: As far as confirming this works, outside of my manual testing, I'd like to see a decrease in the number of `failed to reload cni configuration after receiving fs change event` in e2e containerd logs.
[Previous run before this PR](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/1949/pull-ci-openshift-windows-machine-config-operator-master-aws-e2e-operator/1734746775840886784/artifacts/aws-e2e-operator/windows-e2e-operator-test/artifacts/nodes/ip-10-0-28-68.us-west-1.compute.internal/containerd/containerd.log)

pkg/nodeconfig/payload/payload.go
```diff
@@ -140,7 +140,7 @@ $cni_template=$cni_template.Replace("provider_address",$provider_address)
 # Compare CNI config with existing file, and replace if necessary
 $existing_config=""
 if(Test-Path -Path CNI_CONFIG_PATH) {
-    $existing_config= Get-Content -Path "CNI_CONFIG_PATH"
+` + "    $existing_config=((Get-Content -Path \"CNI_CONFIG_PATH\" -Raw) -Replace \"`r\",\"\")" + `
```
saifshaikh48, line 143: Interesting to repalce **`r** rather than **\r**. Is this some powershell/string formatting magic?

saifshaikh48 (commented):
/lgtm

pkg/nodeconfig/payload/payload.go
sebsoto: yep, backtick is the escape character in windows/powershell.
As backslash is used for filepaths

pkg/nodeconfig/payload/payload_test.go
sebsoto: I've confirmed from the logs that after the initial setup, there are no more containerd errors reloading the cni.conf file.

mansikulkarni96:
/approve
Thanks for fixing this @sebsoto

PR #1998, opened 2024-01-04
OCPBUGS-24264: [cluster] Keep comma as delimiter in noProxy list

saifshaikh48:
Reverts an erroneous sanitization step in WMCO that replaced commas with semicolons in the user's cluster-wide proxy configuration. This was causing Windows to ignore the values set in the noProxy environment variable, incorrectly sending traffic through the proxy.

From the Azure PowerShell docs:
`NO_PROXY: a comma-separated list of hostnames that should be excluded from proxying.` [[1]](https://learn.microsoft.com/en-us/powershell/azure/az-powershell-proxy?view=azps-11.1.0&viewFallbackFrom=azps-11.2.0)
`noProxy list should not contain semicolon delimiters (;). Use (,) to delimit the noProxy domains.` [[2]](https://www.powershellgallery.com/packages/Moc/1.0.52/Content/en-US%5CcommonLocalizationMessages.psd1)


jrvaldes (commented):
@saifshaikh48 thanks for working on this. PTAL at my comments.

saifshaikh48:
I was wondering why I thought semicolons were used in PS. Turns out the manual testing we did was using netsh, one of the only tools (along with [Microsoft Edge CLI](https://learn.microsoft.com/en-us/deployedge/edge-learnmore-cmdline-options-proxy-settings)) that uses a [bypass-list](https://learn.microsoft.com/en-us/windows/win32/winhttp/netsh-exe-commands) 
> To list multiple addresses and host names, separate them with blank spaces or semicolons in the proxy bypass string

pkg/cluster/config.go
```diff
@@ -366,9 +366,7 @@ func GetProxyVars() map[string]string {
 	for _, envVar := range WatchedEnvironmentVars {
 		value, found := os.LookupEnv(envVar)
 		if found {
-			// on Windows, hostname lists are separated by semicolons rather than the Linux default of commas
-			sanitizedVal := strings.ReplaceAll(value, ",", ";")
-			clusterWideProxyVars[envVar] = sanitizedVal
+			clusterWideProxyVars[envVar] = value
```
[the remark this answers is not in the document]
saifshaikh48: Managing entities (CVO and/or OLM) do basic validations around this to ensure the URLs/domains are well-formed. No further checks needed from our side

pkg/cluster/config.go
```diff
@@ -366,9 +366,7 @@ func GetProxyVars() map[string]string {
 	for _, envVar := range WatchedEnvironmentVars {
```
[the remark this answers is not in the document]
saifshaikh48: nice find

pkg/cluster/config.go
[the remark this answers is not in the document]
jrvaldes: I was thinking more about the token separator, i.e. `;` or `,`

test/e2e/proxy_test.go
```diff
@@ -61,10 +64,64 @@ func proxyTestSuite(t *testing.T) {
...
 	t.Run("Certificate validation removal", tc.testCertsRemoval)
 	t.Run("Environment variables removal validation", tc.testEnvVarRemoval)
 }
 
+// testProxyRequest tests that requests from each Windows node properly go through the proxy
+func (tc *testContext) testProxyRequest(t *testing.T) {
+	clusterProxy, err := tc.client.Config.ConfigV1().Proxies().Get(context.TODO(), "cluster", meta.GetOptions{})
+	if err != nil {
+		require.NoError(t, err)
+	}
```
sebsoto, line 80: there should be no if statement

test/e2e/proxy_test.go
```diff
@@ -61,10 +64,64 @@ func proxyTestSuite(t *testing.T) {
...
+	for _, node := range gc.allNodes() {
+		t.Run(node.GetName(), func(t *testing.T) {
+			addr, err := controllers.GetAddress(node.Status.Addresses)
+			require.NoError(t, err, "unable to get node address")
+
+			command := "$response = $(Invoke-WebRequest -UseBasicParsing https://raw.githubusercontent.com/openshift/windows-machine-config-operator/master/README.md); $response.Headers['Via']"
+			out, err := tc.runPowerShellSSHJob("get-webrequest-via-header", command, addr)
+			require.NoErrorf(t, err, "unable to retrieve response 'Via' header")
+
+			viaHeader := strings.TrimSpace(out)
```
sebsoto, line 91: Please make a helper function for this.
such as `func (tc *testContext) getOutboundProxy` or `getOutboundHTTPViaHeaders`
A documented helper will help with code readability.

sebsoto: An example of the returned string would be good here.

test/e2e/proxy_test.go
```diff
@@ -61,10 +64,64 @@ func proxyTestSuite(t *testing.T) {
...
+			fmt.Printf("with proxy Via Header: %s", viaHeader)
+			assert.Truef(t, viaHeader == clusterProxy.Status.HTTPProxy || viaHeader == clusterProxy.Status.HTTPSProxy,
+				"expected response header 'Via' to be either '%s' or '%s' but instead held value: '%s'",
+				clusterProxy.Status.HTTPProxy, clusterProxy.Status.HTTPSProxy, viaHeader)
+		})
```
sebsoto, line 96: Can there be multiple via headers?
Seems that way looking at [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Via).
How is that being handled?

test/e2e/proxy_test.go
```diff
@@ -61,10 +64,64 @@ func proxyTestSuite(t *testing.T) {
...
+	}
+}
+
+// testProxyRequest tests that requests from each Windows node properly go through the proxy
+// TODO: circumvent the proxy if the target address is specified in the NO_PROXY list
+func (tc *testContext) testNoProxyRequest(t *testing.T) {
+	clusterProxy, err := tc.client.Config.ConfigV1().Proxies().Get(context.TODO(), "cluster", meta.GetOptions{})
+	if err != nil {
```
sebsoto, line 104: drop the if

test/e2e/proxy_test.go
```diff
@@ -61,10 +64,64 @@ func proxyTestSuite(t *testing.T) {
...
+		require.NoError(t, err)
+	}
+
+	for _, node := range gc.allNodes() {
+		t.Run(node.GetName(), func(t *testing.T) {
```
sebsoto, line 109: subtest please

test/e2e/proxy_test.go
sebsoto: it also stops the duplicated code thats occuring in the noproxy test

test/e2e/proxy_test.go
```diff
@@ -61,10 +64,64 @@ func proxyTestSuite(t *testing.T) {
...
+			addr, err := controllers.GetAddress(node.Status.Addresses)
+			require.NoError(t, err, "unable to get node address")
+
+			command := "$response = $(Invoke-WebRequest -UseBasicParsing https://raw.githubusercontent.com/openshift/windows-machine-config-operator/master/README.md); $response.Headers['Via']"
```
sebsoto, line 87: remind me, is this only for external connections?
can we launch a service and query that via DNS so we arent relying on google and github?

sebsoto (changes requested):

test/e2e/proxy_test.go
```diff
@@ -61,10 +64,64 @@ func proxyTestSuite(t *testing.T) {
...
 
+// testProxyRequest tests that requests from each Windows node properly go through the proxy
+func (tc *testContext) testProxyRequest(t *testing.T) {
+	clusterProxy, err := tc.client.Config.ConfigV1().Proxies().Get(context.TODO(), "cluster", meta.GetOptions{})
+	if err != nil {
+		require.NoError(t, err)
+	}
+
+	for _, node := range gc.allNodes() {
+		t.Run(node.GetName(), func(t *testing.T) {
```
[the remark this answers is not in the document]
saifshaikh48: This should already be happening: [example](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-operator/1998/pull-ci-openshift-windows-machine-config-operator-master-vsphere-proxy-e2e-operator/1744494642423926784)
```
        --- FAIL: TestWMCO/cluster-wide_proxy/Request_routed_through_proxy (10.13s)
            --- FAIL: TestWMCO/cluster-wide_proxy/Request_routed_through_proxy/e2e-wm-jwqv2 (5.06s)
            --- FAIL: TestWMCO/cluster-wide_proxy/Request_routed_through_proxy/winhost-zu9oi (5.06s)
        --- FAIL: TestWMCO/cluster-wide_proxy/Request_not_proxied (10.13s)
            --- FAIL: TestWMCO/cluster-wide_proxy/Request_not_proxied/e2e-wm-jwqv2 (5.06s)
            --- FAIL: TestWMCO/cluster-wide_proxy/Request_not_proxied/winhost-zu9oi (5.06s)
```

test/e2e/proxy_test.go
saifshaikh48: Good catch. I'll change this to search for a via reader whose value is the cluster-wide proxy
along the lines of 
`$response.Headers["Via"] == clusterProxy.Status.HTTPProxy || $response.Headers["Via"]  == clusterProxy.Status.HTTPSProxy`

test/e2e/proxy_test.go
saifshaikh48: I had thought of that initially but, yes external requests only

test/e2e/proxy_test.go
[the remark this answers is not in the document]
sebsoto: Oops, yes it should

test/e2e/proxy_test.go
```diff
@@ -39,6 +39,9 @@ const (
 	// userCABundleName is the name of the ConfigMap that holds additional user-provided proxy certs
 	userCABundleName      = "user-ca-bundle"
 	userCABundleNamespace = "openshift-config"
+
+	// whitelist is a list of domains that are not to be proxied
+	whitelist = "static.redhat.com,redhat.io,google.com"
```
jrvaldes, line 44: s/whitelist/noProxyList

test/e2e/proxy_test.go
```diff
@@ -39,6 +39,9 @@ const (
 	// userCABundleName is the name of the ConfigMap that holds additional user-provided proxy certs
 	userCABundleName      = "user-ca-bundle"
 	userCABundleNamespace = "openshift-config"
+
+	// whitelist is a list of domains that are not to be proxied
```
jrvaldes, line 43: > noProxy is a comma-separated list of hostnames and/or CIDRs and/or IPs for 
> which the proxy should not be used.

As per ` oc explain proxies.spec.noProxy`.

test/e2e/proxy_test.go
```diff
@@ -351,6 +405,21 @@ func (tc *testContext) getEnvVar(addr, name, command string) (map[string]string,
 	return parseWindowsEnvVars(out), nil
 }
 
+// configureNoProxy configures the cluster-wide proxy's bypass list with specific URLs for testing purposes
+func (tc *testContext) configureNoProxy() error {
```
jrvaldes, line 409: Consider passing the no proxy list as a parameter of the function.

test/e2e/proxy_test.go
```diff
@@ -351,6 +405,21 @@ func (tc *testContext) getEnvVar(addr, name, command string) (map[string]string,
...
+	patchData, err := json.Marshal(patches)
+	if err != nil {
+		return fmt.Errorf("unable to generate patch request body for %v: %w", patches, err)
+	}
+	_, err = tc.client.Config.ConfigV1().Proxies().Patch(context.TODO(), "cluster", types.JSONPatchType, patchData,
+		meta.PatchOptions{})
+	if err != nil {
+		return fmt.Errorf("unable to patch %s: %w", string(patchData), err)
+	}
+	return nil
```
jrvaldes, line 420: Consider returning just `err` instead.

test/e2e/proxy_test.go
```diff
@@ -67,97 +67,129 @@ func proxyTestSuite(t *testing.T) {
...
+
+				httpVarInUse, noProxyVarInUse, proxyInViaHeader := verifyCurlOutput(out, httpVal, noProxyVal)
+
+				assert.Truef(t, httpVarInUse, "expected http_proxy %s in request", clusterProxy.Status.HTTPProxy)
+				assert.Truef(t, noProxyVarInUse, "expected no_proxy %s in request", clusterProxy.Status.NoProxy)
+				assert.True(t, proxyInViaHeader, "'Via' header with proxy expected for proxied request")
+			})
+
+			t.Run("NO_PROXY request", func(t *testing.T) {
+				whitelistedUrl := "google.com"
```
mansikulkarni96, line 102: how about permitted/nonproxied URL

mansikulkarni96 (commented):
Thanks for working on this @saifshaikh48 . This looks much clear and concise.

test/e2e/proxy_test.go
saifshaikh48: no need, I'd rather keep the noproxylist var in this package scope

pkg/daemon/controller/controller_test.go
```diff
@@ -665,8 +665,8 @@ func TestReconcile(t *testing.T) {
 			name:                         "Reboot annotation is applied when env vars are changed",
 			configMapServices:            []servicescm.Service{},
 			expectedServicesNameCmdPairs: map[string]string{},
-			existingEnvVars:              map[string]string{"NO_PROXY": "localhost;127.0.0.1", "HTTP_PROXY": "http://example.com"},
-			configMapEnvVars:             map[string]string{"NO_PROXY": "localhost;127.0.0.1"},
+			existingEnvVars:              map[string]string{"NO_PROXY": "localhost,127.0.0.1", "HTTP_PROXY": "http://example.com"},
+			configMapEnvVars:             map[string]string{"NO_PROXY": "localhost,127.0.0.1"},
```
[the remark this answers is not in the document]
saifshaikh48: semicolon-separated value is still valid from WICD unit test/Windows OS perspective. The WMCO e2e integration tests would fail/signal to handle this

alinaryan:
/lgtm
Thanks for working on this and adding the example output

test/e2e/proxy_test.go
```diff
@@ -61,10 +65,126 @@ func proxyTestSuite(t *testing.T) {
...
+	return out, err
+}
+
+// verifyCurlOutput parses the output of a curl request searching for any proxy variables in use and proxy headers
+func verifyCurlOutput(output, expectedHTTPVal, expectedNoProxyVal string) (bool, bool, bool) {
+	httpVarFound := false
+	noProxyVarFound := false
+	viaHeaderFound := false
+	// regex to ensure the squid proxy appears as a redirect in the curl headers
+	regex := regexp.MustCompile(`Via:.*\.in-addr\.arpa \(squid\/4\.4\)`)
```
saifshaikh48, line 171: may need to adjust the regex a bit to be more general, something like `Via:.*squid`, so it works on all platforms (enables better local testing). `in-addr.arpa` is vSphere specific

test/e2e/proxy_test.go
```diff
@@ -61,10 +65,126 @@ func proxyTestSuite(t *testing.T) {
...
+	command := "cmd.exe /c curl -vsIL " + urlToCurl
+	out, err := tc.runPowerShellSSHJob("curl-headers-only", command, nodeAddress)
+	return out, err
+}
+
+// verifyCurlOutput parses the output of a curl request searching for any proxy variables in use and proxy headers
+func verifyCurlOutput(output, expectedHTTPVal, expectedNoProxyVal string) (bool, bool, bool) {
+	httpVarFound := false
+	noProxyVarFound := false
+	viaHeaderFound := false
```
sebsoto, line 169: nit: can just be 
`var httpVarFound, noProxyVarFound, viaHeaderFound bool`

test/e2e/proxy_test.go
```diff
@@ -61,10 +65,126 @@ func proxyTestSuite(t *testing.T) {
...
+	// regex to ensure the squid proxy appears as a redirect in the curl headers
+	regex := regexp.MustCompile(`Via:.*squid`)
```
sebsoto, line 171: Where is squid coming from?
Need more info around this.
Is this going to work if this ran from outside OCP CI?
I.e on a user's cluster who has set up a proxy on their own.

test/e2e/proxy_test.go
```diff
@@ -61,10 +65,126 @@ func proxyTestSuite(t *testing.T) {
...
+				httpVarInUse, noProxyVarInUse, proxyInViaHeader := verifyCurlOutput(out, httpVal, noProxyVal)
+
+				assert.Truef(t, httpVarInUse, "expected http_proxy %s in request", clusterProxy.Status.HTTPProxy)
+				assert.Truef(t, noProxyVarInUse, "expected no_proxy %s in request", clusterProxy.Status.NoProxy)
+				assert.True(t, proxyInViaHeader, "'Via' header with proxy expected for proxied request")
+			})
+
+			t.Run("NO_PROXY request", func(t *testing.T) {
+				// using an endpoint that appears in the noProxyList const
+				nonProxiedUrl := "google.com"
```
sebsoto, line 104: Please derive this from the noProxyList variable.
As it is, if we want to change the URL it needs to be changed in two separate spots and thats not clear.

test/e2e/proxy_test.go
```diff
@@ -61,10 +65,126 @@ func proxyTestSuite(t *testing.T) {
...
 
+// testProxyRequest tests that external requests from each Windows node properly go through the proxy
+// and circumvent the proxy if the target address is specified in the NO_PROXY list
+func (tc *testContext) testProxyRequests(t *testing.T) {
+	clusterProxy, err := tc.client.Config.ConfigV1().Proxies().Get(context.TODO(), "cluster", meta.GetOptions{})
+	require.NoError(t, err)
+
+	// variables to look for in the curl output
+	httpVal := fmt.Sprintf("http_proxy == '%s'", clusterProxy.Status.HTTPProxy)
+	noProxyVal := fmt.Sprintf("no_proxy == '%s'", clusterProxy.Status.NoProxy)
```
sebsoto, line 83: please pass in the URLs, and do this string creation within verifyCurlOutput. No need for this to live outside of it.

sebsoto (changes requested):

test/e2e/proxy_test.go
```diff
@@ -61,10 +65,126 @@ func proxyTestSuite(t *testing.T) {
...
+<
+...
+< HTTP/1.1 200 Connection established
+<
+...
+*/
+func (tc *testContext) curlHeadersOnly(nodeAddress, urlToCurl string) (string, error) {
+	command := "cmd.exe /c curl -vsIL " + urlToCurl
+	out, err := tc.runPowerShellSSHJob("curl-headers-only", command, nodeAddress)
+	return out, err
```
jrvaldes, line 162:
```suggestion
	return tc.runPowerShellSSHJob("curl-headers-only", command, nodeAddress)
```

test/e2e/proxy_test.go
```diff
@@ -61,10 +65,126 @@ func proxyTestSuite(t *testing.T) {
...
+< Via: 1.1 varnish, 1.1 951.27.49.01.in-addr.arpa (squid/4.4)
+< Connection: keep-alive
+<
+...
+< HTTP/1.1 200 Connection established
+<
+...
+*/
+func (tc *testContext) curlHeadersOnly(nodeAddress, urlToCurl string) (string, error) {
+	command := "cmd.exe /c curl -vsIL " + urlToCurl
```
jrvaldes, line 160: Consider building the command with ` fmt.Sprintf` for safer URL escape

test/e2e/proxy_test.go
jrvaldes: +1

saifshaikh48: This is coming from the type of proxy server that is spun up in CI. Just looking for the `Via` header isn't sufficient since sometimes the Agent (web browser) adds a Via header even in non-proxied requests.

> Is this going to work if this ran from outside OCP CI?
> I.e on a user's cluster who has set up a proxy on their own.

Not guaranteed to work. I will instead look for the presence of the `[Proxy-Authorization](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Proxy-Authorization)` header, should solve the problem

test/e2e/proxy_test.go
```diff
@@ -244,7 +364,8 @@ func (tc *testContext) testEnvVars(t *testing.T) {
 func (tc *testContext) testEnvVarRemoval(t *testing.T) {
 	var patches []*patch.JSONPatch
 	patches = append(patches, patch.NewJSONPatch("remove", "/spec/httpProxy", nil),
-		patch.NewJSONPatch("remove", "/spec/httpsProxy", nil))
+		patch.NewJSONPatch("remove", "/spec/httpsProxy", nil),
+		patch.NewJSONPatch("remove", "/spec/noProxy", nil))
```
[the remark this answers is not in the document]
saifshaikh48: does so already on the line above

test/e2e/proxy_test.go
```diff
@@ -61,10 +65,126 @@ func proxyTestSuite(t *testing.T) {
...
+	httpVal := fmt.Sprintf("http_proxy == '%s'", clusterProxy.Status.HTTPProxy)
+	noProxyVal := fmt.Sprintf("no_proxy == '%s'", clusterProxy.Status.NoProxy)
+
+	for _, node := range gc.allNodes() {
+		t.Run(node.GetName(), func(t *testing.T) {
+			addr, err := controllers.GetAddress(node.Status.Addresses)
+			require.NoError(t, err, "unable to get node address")
+
+			t.Run("HTTP_PROXY request", func(t *testing.T) {
+				httpUrl := "http://raw.githubusercontent.com/openshift/windows-machine-config-operator/master/README.md"
```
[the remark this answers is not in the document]
saifshaikh48: I would hope the WMCO repo (and our team) continues to exist :)

> so does it matter if is a valid URL

I was having issues with powershell not outputting the stdout when using a dummy URL

test/e2e/proxy_test.go
```diff
@@ -61,10 +65,124 @@ func proxyTestSuite(t *testing.T) {
...
+<
+...
+*/
+func (tc *testContext) curlHeadersOnly(nodeAddress, urlToCurl string) (string, error) {
+	command := fmt.Sprintf("cmd.exe /c curl -vsIL %s", urlToCurl)
+	return tc.runPowerShellSSHJob("curl-headers-only", command, nodeAddress)
+}
+
+// verifyCurlOutput parses the output of a curl request searching for any proxy variables in use and proxy headers
+func verifyCurlOutput(output, httpProxyURL, noProxyList string) (bool, bool, bool) {
```
sebsoto, line 163: should this be named noProxyList, or noProxyURL?
if a comma separated list is passed in would that work?

sebsoto (approved):

saifshaikh48:
/retest-required

Not sure why none of the e2e tests are kicking off automatically.

jrvaldes:
/retitle WINC-1152, OCPBUGS-24264: [cluster] Keep comma as delimiter in noProxy list

PR #1999, opened 2024-01-04
[controllers] Introduce err check when merging CA configmaps

alinaryan:
reconcileKubeletClientCA() handles reconciling the kube-apiserver certificate rotation by merging the initial and current CA configmaps for the kube API server signer, however, there is currently no error-check after the call to mergeCAsConfigMaps(). This commit adds an error-check after mergeCAsConfigMaps() is called to make certain no unknown failures are able to slip through.

### DIFF
--- a/controllers/controllers.go
+++ b/controllers/controllers.go
@@ -153,6 +153,10 @@ func (r *instanceReconciler) reconcileKubeletClientCA(ctx context.Context, bundl
 	// matches the signer subject.
 	kubeAPIServerServingCABytes, err := certificates.MergeCAsConfigMaps(initialCAConfigMap, bundleCAConfigMap,
 		"kube-apiserver-to-kubelet-signer")
+	if err != nil {
+		return fmt.Errorf("error merging the initial %s and current %s CA ConfigMaps %w", initialCAConfigMap,
+			bundleCAConfigMap, err)
+	}
 
 	// fetch all Windows nodes (Machine and BYOH instances)
 	winNodes := &core.NodeList{}


controllers/controllers.go
```diff
@@ -153,6 +153,11 @@ func (r *instanceReconciler) reconcileKubeletClientCA(ctx context.Context, bundl
 	// matches the signer subject.
 	kubeAPIServerServingCABytes, err := certificates.MergeCAsConfigMaps(initialCAConfigMap, bundleCAConfigMap,
 		"kube-apiserver-to-kubelet-signer")
+	if err != nil {
+		r.log.Info("could not merge the initial and current CA config maps", "initialCAConfigMap",
+			initialCAConfigMap, "bundleCAConfigMap", bundleCAConfigMap)
```
sebsoto, line 158: This log isn't necessary as you are returning the error.

controllers/controllers.go
```diff
@@ -153,6 +153,11 @@ func (r *instanceReconciler) reconcileKubeletClientCA(ctx context.Context, bundl
...
+		return fmt.Errorf("error merging the initial and current CA ConfigMaps %w", err)
```
sebsoto, line 159: I suggest adding the names of the initial and bundleCAConfigMaps to the error message

sebsoto (changes requested):
Good catch with this.
Is this fixing an open bug? If so please add the bug to the PR title.

controllers/controllers.go
```diff
@@ -153,6 +153,10 @@ func (r *instanceReconciler) reconcileKubeletClientCA(ctx context.Context, bundl
 	// matches the signer subject.
 	kubeAPIServerServingCABytes, err := certificates.MergeCAsConfigMaps(initialCAConfigMap, bundleCAConfigMap,
```
jrvaldes, line 154: Consider adding a unit test for `reconcileKubeletClientCA()` to check the propagation of a validation error. i.e., passing an empty subject.

jrvaldes (commented):
@alinaryan, thanks for working on this.  I added a nice-to-have suggestion. 

Introducing an early return of the error is helpful, but I don't feel it was causing any issues, given there is a [fail-check](
https://github.com/alinaryan/windows-machine-config-operator/blob/master/pkg/nodeconfig/nodeconfig.go#L600-L603) for empty contents.

alinaryan:
>Is this fixing an open bug? If so please add the bug to the PR title.

I wouldn't say it fixes https://issues.redhat.com/browse/OCPBUGS-22237, but it adds error checking on the WMCO side since the only errors currently come from the kubelet logs.

alinaryan:
>Introducing an early return of the error is helpful, but I don't feel it was causing any issues, given there is a [fail-check](https://github.com/alinaryan/windows-machine-config-operator/blob/master/pkg/nodeconfig/nodeconfig.go#L600-L603) for empty contents.

@jrvaldes Empty contents might not be the only thing causing this to fail.
e.g if the customer updates the bundle cert manually and it doesn't contain the correct signer subject with the common name: "kube-apiserver-to-kubelet-signer" then the new bundle cert contents are overridden with the old bundle.

See: https://github.com/openshift/windows-machine-config-operator/blob/master/pkg/certificates/certificates.go#L123

controllers/controllers.go
sebsoto: > Introducing an early return of the error is helpful, but I don't feel it was causing any issues, given there is a [fail-check](https://github.com/alinaryan/windows-machine-config-operator/blob/master/pkg/nodeconfig/nodeconfig.go#L600-L603) for empty contents.

If a function returns an error it should always be checked.

sebsoto:
https://github.com/openshift/windows-machine-config-operator/pull/1999#pullrequestreview-1809601509
> Introducing an early return of the error is helpful, but I don't feel it was causing any issues, given there is a [fail-check](https://github.com/alinaryan/windows-machine-config-operator/blob/master/pkg/nodeconfig/nodeconfig.go#L600-L603) for empty contents.

If a function returns an error it should always be checked.

controllers/controllers.go
alinaryan: Thinking of adding a unit test to check that the correct subject signer is in both bundles if we determine that is the cause of the bug: https://issues.redhat.com/browse/OCPBUGS-22237

I could also add what you're suggesting in that same PR.

sebsoto (approved):

jrvaldes:
@alinaryan  any backports?

PR #2006, opened 2024-01-11
OCPBUGS-26761:  Use DSR load balancing in kube-proxy

sebsoto:
Changes how kube-proxy is ran in order to use more stable load balancing methods

PR #2017, opened 2024-01-17
[vsphere_ci] Add metrics port rule to packer script

saifshaikh48:
Update the firewall rules packer script to explicity allow incoming traffic for the Windows Exporter metrics server running on Windows nodes.

Follow-up to https://github.com/openshift/windows-machine-config-operator/pull/2006


saifshaikh48:
/close

will be done in https://github.com/openshift/windows-machine-config-operator/pull/2006, since they will be backported to the same versions anyway

PR #2025, opened 2024-01-22
[test] Open SMB port when creating share

sebsoto:
Allows incoming traffic for SMB so the e2e tests can function properly.


jrvaldes (approved):
/lgtm

saifshaikh48 (commented):
/lgtm

is this something users will now have to do manually if they want to use SMB-based volumes? If so maybe we should update our docs with a blurb

sebsoto:
> is this something users will now have to do manually if they want to use SMB-based volumes? If so maybe we should update our docs with a blurb

That would be outside of the scope of WMCO. If a user wants to run an application on the VM outside of WMCO, we do not provide guidance on that.

jrvaldes (approved):
/lgtm

PR #2039, opened 2024-01-24
[build] Set go toolchain to local

mansikulkarni96:
This PR sets the value of ENV var GOTOOLCHAIN to local. The default value is auto if not set.
Making this change prevents it from downloading the latest version of toolchain which is incompatible
with Cachito still running on go1.19.
This change is made to maintain parity between the github and midtstream Dockerfiles.


jrvaldes (approved):
/lgtm

PR #2051, opened 2024-01-30
OCPBUGS-28575: Fix issue with PS type mismatch

sebsoto:
This commit fixes an issue which would be seen when the cni config file was created with no content. When using the -replace operator with $null, an empty object array was returned. This caused comparison issues with the expected config as it was comparing an array with a string.

### DIFF
--- a/pkg/nodeconfig/payload/payload.go
+++ b/pkg/nodeconfig/payload/payload.go
@@ -140,7 +140,10 @@ $cni_template=$cni_template.Replace("provider_address",$provider_address)
 # Compare CNI config with existing file, and replace if necessary
 $existing_config=""
 if(Test-Path -Path CNI_CONFIG_PATH) {
-` + "    $existing_config=((Get-Content -Path \"CNI_CONFIG_PATH\" -Raw) -Replace \"`r\",\"\")" + `
+    $config_file_content=(Get-Content -Path CNI_CONFIG_PATH -Raw)
+    if($config_file_content -ne $null) {
+` + "        $existing_config=$config_file_content.Replace(\"`r\",\"\")" + `
+    }
 }
 if($existing_config -ne $cni_template){
     Set-Content -Path "CNI_CONFIG_PATH" -Value $cni_template -NoNewline
--- a/pkg/nodeconfig/payload/payload_test.go
+++ b/pkg/nodeconfig/payload/payload_test.go
@@ -74,7 +74,10 @@ $cni_template=$cni_template.Replace("provider_address",$provider_address)
 # Compare CNI config with existing file, and replace if necessary
 $existing_config=""
 if(Test-Path -Path c:\k\cni.conf) {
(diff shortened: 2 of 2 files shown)


pkg/nodeconfig/payload/payload.go
```diff
@@ -140,7 +140,10 @@ $cni_template=$cni_template.Replace("provider_address",$provider_address)
 # Compare CNI config with existing file, and replace if necessary
 $existing_config=""
 if(Test-Path -Path CNI_CONFIG_PATH) {
-` + "    $existing_config=((Get-Content -Path \"CNI_CONFIG_PATH\" -Raw) -Replace \"`r\",\"\")" + `
+    $config_file_content=(Get-Content -Path CNI_CONFIG_PATH -Raw)
+    if($config_file_content -ne $null) {
```
jrvaldes, line 144: should it also check for non empty?

pkg/nodeconfig/payload/payload.go
```diff
@@ -140,7 +140,10 @@ $cni_template=$cni_template.Replace("provider_address",$provider_address)
 # Compare CNI config with existing file, and replace if necessary
 $existing_config=""
 if(Test-Path -Path CNI_CONFIG_PATH) {
-` + "    $existing_config=((Get-Content -Path \"CNI_CONFIG_PATH\" -Raw) -Replace \"`r\",\"\")" + `
+    $config_file_content=(Get-Content -Path CNI_CONFIG_PATH -Raw)
```
jrvaldes, line 143: Shouldn't an existing file without content throw an error?

jrvaldes (commented):
Left few questions, otherwise LGTM

pkg/nodeconfig/payload/payload.go
sebsoto: That is not the case
```
PS C:\Users\Administrator> rm /k/cni/config/cni.conf; New-Item /k/cni/config/cni.conf -type file


    Directory: C:\k\cni\config


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----        1/30/2024   5:23 PM              0 cni.conf


PS C:\Users\Administrator> cat /k/cni/config/cni.conf
PS C:\Users\Administrator> $foo=(cat /k/cni/config/cni.conf)
PS C:\Users\Administrator> $foo
PS C:\Users\Administrator> $foo.gettype()
You cannot call a method on a null-valued expression. 
At line:1 char:1
+ $foo.gettype()
+ ~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [], RuntimeException
    + FullyQualifiedErrorId : InvokeMethodOnNull

```

saifshaikh48 (commented):
/lgtm

mtnbikenc:
```
One or more of Snyk's services may be temporarily unavailable.
```
/test security

PR #2055, opened 2024-02-05
[release-4.14] Change how upgrade-blocked Node is labeled

sebsoto:
Changes how the upgrade blocked Node is labeled when doing an in-tree to csi storage upgrade. This is required to stop the use of affinities, which will cause issues across WMCO upgrade in the case that Machine Nodes have to be recreated. If the Machine Node is replaced, the pod will be unschedulable, as it will only run on the deleted Node.

### DIFF
--- a/test/e2e/storage_test.go
+++ b/test/e2e/storage_test.go
@@ -15,6 +15,7 @@ import (
 	core "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/openshift/windows-machine-config-operator/controllers"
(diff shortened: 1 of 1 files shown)


test/e2e/storage_test.go
```diff
@@ -96,6 +86,37 @@ func testStorage(t *testing.T) {
 			}
 		}()
 	}
+	if inTreeUpgrade {
+		err = tc.labelPodsNode(winServerDeployment.Spec.Selector.MatchLabels, map[string]string{storageTestLabel: "true"})
+		require.NoError(t, err)
+	}
+}
+
+// labelPodNode labels the Node which has the pod with matchLabels scheduled to it. Throws an error if more than one pod
```
mansikulkarni96, line 95: labelPodsNode*

test/e2e/storage_test.go
```diff
@@ -96,6 +86,37 @@ func testStorage(t *testing.T) {
...
+// matches the labels.
+func (tc *testContext) labelPodsNode(matchLabels map[string]string, labelsToApply map[string]string) error {
+	if matchLabels == nil {
+		return fmt.Errorf("nill matchLabels")
```
mansikulkarni96, line 99: nit: nil matchLabels or no labels present

test/e2e/storage_test.go
```diff
@@ -96,6 +86,37 @@ func testStorage(t *testing.T) {
 			}
 		}()
 	}
+	if inTreeUpgrade {
+		err = tc.labelPodsNode(winServerDeployment.Spec.Selector.MatchLabels, map[string]string{storageTestLabel: "true"})
+		require.NoError(t, err)
```
mansikulkarni96, line 91: the previous error msg :"error labeling node for upgrade test" seems helpful and can be retained.

mansikulkarni96 (commented):
Mostly LGTM

test/e2e/storage_test.go
saifshaikh48: +1

test/e2e/storage_test.go
sebsoto: The check is only to ensure that a nil pointer deference doesn't happen which will cause a panic

saifshaikh48: nit: `nill` --> `nil`

saifshaikh48 (commented):
/lgtm

PR #2070, opened 2024-02-07
Test Windows pod egress

sebsoto:
Adds network test cases from a Windows pod, to a Linux pod, both through and without and ClusterIP service


jrvaldes (commented):
/lgtm

test/e2e/network_test.go
```diff
@@ -322,8 +361,9 @@ func (tc *testContext) createService(name string, serviceType v1.ServiceType, se
 			Type: serviceType,
 			Ports: []v1.ServicePort{
 				{
-					Protocol: v1.ProtocolTCP,
-					Port:     80,
+					Protocol:   v1.ProtocolTCP,
+					Port:       80,
```
mansikulkarni96, line 365: can we pass in the same targetport here?

sebsoto: No, the targetport is the container's exposed port. `Port` is the port which is open on the Service. That should always be 80 for http.

saifshaikh48: nit: code comment around this could be useful

saifshaikh48 (commented):
/lgtm

PR #2093, opened 2024-02-28
[release-4.15] WINC-1211: Update version to 10.15.1

jrvaldes:

### DIFF
--- a/.ci-operator.yaml
+++ b/.ci-operator.yaml
@@ -2,4 +2,4 @@
 build_root_image:
   namespace: openshift
   name: release
-  tag: rhel-9-release-golang-1.20-openshift-4.14
+  tag: rhel-9-release-golang-1.20-openshift-4.15
--- a/bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
@@ -18,7 +18,7 @@ metadata:
     features.operators.openshift.io/token-auth-aws: "false"
     features.operators.openshift.io/token-auth-azure: "false"
     features.operators.openshift.io/token-auth-gcp: "false"
-    olm.skipRange: '>=9.0.0 <10.15.0'
+    olm.skipRange: '>=9.0.0 <10.15.1'
     operatorframework.io/cluster-monitoring: "true"
     operatorframework.io/suggested-namespace: openshift-windows-machine-config-operator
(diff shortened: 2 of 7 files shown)


bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
```diff
@@ -18,7 +18,7 @@ metadata:
     features.operators.openshift.io/token-auth-aws: "false"
     features.operators.openshift.io/token-auth-azure: "false"
     features.operators.openshift.io/token-auth-gcp: "false"
-    olm.skipRange: '>=9.0.0 <10.15.0'
+    olm.skipRange: '>=9.0.0 <10.15.1'
```
sebsoto, line 21: Please add the index change as well https://github.com/openshift/windows-machine-config-operator/blob/d40284976de0c015177706f7174999ec6bbac03f/bundle/metadata/annotations.yaml#L12

jrvaldes: addressed

sebsoto (approved):

.ci-operator.yaml
```diff
@@ -2,4 +2,4 @@
 build_root_image:
   namespace: openshift
   name: release
-  tag: rhel-9-release-golang-1.20-openshift-4.14
+  tag: rhel-9-release-golang-1.20-openshift-4.15
```
saifshaikh48, line 5: Can we do this later in a separate PR? We want to do this in the Dockerfiles too but IMO it's not needed for this release

jrvaldes: Good point, I don't anticipate any issues as it has been used already https://github.com/openshift/windows-machine-config-operator/blob/release-4.15/build/Dockerfile.ci#L6 

Happy to drop it if the team decides otherwise.

jrvaldes:
/hold

for .ci-operator.yaml update  [decision](https://github.com/openshift/windows-machine-config-operator/pull/2093#discussion_r1506229050)

.ci-operator.yaml
saifshaikh48: Wait my local branch was out of date, I didn't realize we've already bumped to using 4.15 in the dockerfiles. In that case I am fine with leaving this change in

jrvaldes:
/hold cancel

https://github.com/openshift/windows-machine-config-operator/pull/2093#discussion_r1506240809

PR #2100, opened 2024-02-28
[hack] Fix get_operator_sdk hardcode path issue

jrvaldes:
This change updates the get_operator_sdk function in the hack script to use a temp dir instead of a hard-coded path that may not exist and improve error handling.

In addition, it generalizes the OPERATOR_SDK_VERSION environment variable to be re-used.


saifshaikh48 (commented):
/lgtm

sebsoto (approved):

PR #2115, opened 2024-03-26
OCPBUGS-20255: Let caller of Run handle error logging

sebsoto:
Stops logging within windows.Run(), and instead lets the caller decide to log. This stops the undesired logging of errors, in the case that Run() is expected to fail.

### DIFF
--- a/pkg/windows/windows.go
+++ b/pkg/windows/windows.go
@@ -128,8 +128,6 @@ const (
 	// cmdExitNoStatus is part of the error message returned when a command takes too long to report status back to
 	// PowerShell.
 	cmdExitNoStatus = "command exited without exit status or exit signal"
-	// removeHNSCommand is the Windows command used to remove HNS network.
-	removeHNSCommand = "Remove-HnsNetwork"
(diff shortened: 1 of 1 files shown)


jrvaldes (approved):
/lgtm

pkg/windows/windows.go
```diff
@@ -402,12 +400,6 @@ func (vm *windows) Run(cmd string, psCmd bool) (string, error) {
 
 	out, err := vm.interact.run(cmd)
 	if err != nil {
-		// Hack to not print the error log for "sc.exe qc" returning 1060 for non existent services
-		// and not print error when the command takes too long to return after removing HNS networks.
-		if !(strings.Contains(cmd, serviceQueryCmd) && strings.Contains(out, serviceNotFound)) &&
-			!(strings.Contains(err.Error(), cmdExitNoStatus) && strings.HasSuffix(cmd, removeHNSCommand+";\"")) {
-			vm.log.Error(err, "error running", "cmd", cmd, "out", out)
-		}
```
saifshaikh48, line 410: Is there going to be a follow-up PR to re-introduce this log at the call site(s)?

PR #2117, opened 2024-04-01
OCPBUGS-31626: Stop checking CloudControllerOwner

sebsoto:
This commit reacts to
https://github.com/openshift/cluster-cloud-controller-manager-operator/commit/65c07a6e44bfa1b0cdfadc264d3341c093ffd914

There is no longer a CloudControllerOwner condition on the CCM clusteroperator, as external cloud controllers are now fully default. WMCO will no longer check for this condition, and the logic depending on it should now always be done.

### DIFF
--- a/controllers/configmap_controller.go
+++ b/controllers/configmap_controller.go
@@ -25,7 +25,6 @@ import (
 	"strings"
 
 	config "github.com/openshift/api/config/v1"
-	openshiftClient "github.com/openshift/client-go/config/clientset/versioned"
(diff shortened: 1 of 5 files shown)


saifshaikh48 (commented):
/lgtm

:rocket:

jrvaldes (approved):
/lgtm

sebsoto:
/approve

Self approving now that multiple people have `/lgtm`'d, in order to unblock merges to master.

PR #2114, opened 2024-03-25
WINC-1220: Set config_path for containerd registry settings

saifshaikh48:
This PR adjusts the containerd base config to pick up CRI plugin config so we can customize runtime settings like image registry configuration. The new directory is added to the set created and deleted by WMCO.

PR #2140, opened 2024-04-15
WINC-1221: WMCO generates hosts.toml file content

saifshaikh48:
Goal:
WMCO to create config files compatible with containerd from cluster's mirror registry settings

Approach: 
- introducing a "registry controller" to WMCO which watches to ImageDigestMirrorSet and ImageTagMirrorSet resources
- The controller generates a set of `hosts.toml` containerd config files when cluster settings are updated

### DIFF
--- a/controllers/registry_controller.go
+++ b/controllers/registry_controller.go
@@ -27,6 +27,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
 	"github.com/openshift/windows-machine-config-operator/pkg/cluster"
+	"github.com/openshift/windows-machine-config-operator/pkg/registries"
 )
 
 //+kubebuilder:rbac:groups="config.openshift.io",resources=imagedigestmirrorsets,verbs=get;list;watch
@@ -66,6 +67,18 @@ func NewRegistryReconciler(mgr manager.Manager, clusterConfig cluster.Config,
 // related to image registry config and aims to move the current state of the cluster closer to the desired state.
 func (r *registryReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, err error) {
 	r.log = r.log.WithValues(RegistryController, req.NamespacedName)
+
+	// List all IDMS/ITMS resources
+	imageDigestMirrorSetList := &config.ImageDigestMirrorSetList{}
+	if err = r.client.List(ctx, imageDigestMirrorSetList); err != nil {
+		return ctrl.Result{}, fmt.Errorf("error getting IDMS list: %w", err)
+	}
+	imageTagMirrorSetList := &config.ImageTagMirrorSetList{}
+	if err = r.client.List(ctx, imageTagMirrorSetList); err != nil {
+		return ctrl.Result{}, fmt.Errorf("error getting ITMS list: %w", err)
+	}
+
+	_ = registries.NewRegistryConfig(imageDigestMirrorSetList.Items, imageTagMirrorSetList.Items)
 
 	return ctrl.Result{}, nil
 }
--- a/controllers/registry_controller.go
+++ b/controllers/registry_controller.go
@@ -1,0 +1,79 @@
+/*
+Copyright 2021.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+	"fmt"
+
+	config "github.com/openshift/api/config/v1"
+	"k8s.io/client-go/kubernetes"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+
+	"github.com/openshift/windows-machine-config-operator/pkg/cluster"
+)
+
+//+kubebuilder:rbac:groups="config.openshift.io",resources=imagedigestmirrorsets,verbs=get;list;watch
+//+kubebuilder:rbac:groups="config.openshift.io",resources=imagetagmirrorsets,verbs=get;list;watch
+
+const (
+	// RegistryController is the name of this controller in logs and other outputs.
+	RegistryController = "registry"
+)
+
+// registryReconciler holds the info required to reconcile image registry settings on Windows nodes
+type registryReconciler struct {
+	instanceReconciler
+}
+
+// NewRegistryReconciler returns a pointer to a new registryReconciler
+func NewRegistryReconciler(mgr manager.Manager, clusterConfig cluster.Config,
+	watchNamespace string) (*registryReconciler, error) {
+	clientset, err := kubernetes.NewForConfig(mgr.GetConfig())
+	if err != nil {
+		return nil, fmt.Errorf("error creating kubernetes clientset: %w", err)
+	}
+
+	return &registryReconciler{
+		instanceReconciler: instanceReconciler{
+			client:             mgr.GetClient(),
+			log:                ctrl.Log.WithName("controllers").WithName(RegistryController),
+			k8sclientset:       clientset,
+			clusterServiceCIDR: clusterConfig.Network().GetServiceCIDR(),
+			watchNamespace:     watchNamespace,
+			recorder:           mgr.GetEventRecorderFor(RegistryController),
+		},
+	}, nil
+}
+
+// Reconcile is part of the main kubernetes reconciliation loop which reads that state of the cluster for objects
+// related to image registry config and aims to move the current state of the cluster closer to the desired state.
+func (r *registryReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, err error) {
+	r.log = r.log.WithValues(RegistryController, req.NamespacedName)
+
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *registryReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&config.ImageDigestMirrorSet{}).
+		Watches(&config.ImageTagMirrorSet{}, &handler.EnqueueRequestForObject{}).
+		Complete(r)
+}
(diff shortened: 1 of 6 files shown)


saifshaikh48:
Likely depends on https://github.com/openshift/windows-machine-config-operator/pull/2114 otherwise there will be a small merge conflict

controllers/registry_controller.go
```diff
@@ -97,6 +125,14 @@ func (r *registryReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
 
 	mergedMirrorSets := collectMirrorSets(imageDigestMirrorSetList.Items, imageTagMirrorSetList.Items)
 
+	// configFiles is a map from file path on the Windows node to the file content
+	var configFiles map[string][]byte
+	for _, m := range mergedMirrorSets {
+		filePath := fmt.Sprintf("%s\\%s\\hosts.toml", windows.ContainerdConfigDir, m.source)
+		configFiles[filePath] = []byte(m.generateConfig())
+	}
```
sebsoto, line 133: As this will probably be handled by [EnsureFileContent](https://github.com/openshift/windows-machine-config-operator/blob/e872eecec22f31dbb8942912cb7afc42d0923c67/pkg/windows/windows.go#L237C5-L237C6), I would remove this section of code, as it will most likely be mostly removed and replaced

controllers/registry_controller.go
```diff
@@ -59,6 +60,19 @@ func NewRegistryReconciler(mgr manager.Manager, clusterConfig cluster.Config,
 	}, nil
 }
 
+// mirror represents a mirrored image repo entry in a registry configuration file
+type mirror struct {
+	host        string
+	resolveTags bool
```
sebsoto, line 67: please document resolvetags. Its not clear what it represents.

controllers/registry_controller.go
```diff
@@ -71,9 +85,118 @@ func (r *registryReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
...
+	// Sort mirrors by host alphabetically within each mirrorSet
+	for i := range result {
+		sort.Slice(result[i].mirrors, func(j, k int) bool {
+			return result[i].mirrors[j].host < result[i].mirrors[k].host
+		})
+	}
+	// Sort mirrorSets by source alphabetically
+	sort.Slice(result, func(i, j int) bool {
+		return result[i].source < result[j].source
+	})
```
sebsoto, line 201: Please break this out into a separate sort function

controllers/registry_controller.go
```diff
@@ -71,9 +85,118 @@ func (r *registryReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
...
+			set := &mirrorSet{
+				source:             entry.Source,
+				mirrorSourcePolicy: entry.MirrorSourcePolicy,
+			}
+			for _, image := range entry.Mirrors {
+				set.mirrors = append(set.mirrors, mirror{host: string(image), resolveTags: true})
+			}
+			allMirrorSets = append(allMirrorSets, *set)
+		}
+	}
```
sebsoto, line 166: Wouldn't it be easier to merge the mirrorsets as you are creating the initial list?

start with a map,
check if each new mirrorset already is in the map
if not -> add to map
if it is -> add mirror to map

controllers/registry_controller.go
```diff
@@ -71,9 +85,118 @@ func (r *registryReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
...
+// mergeMirrorSets consolidates duplicate entries in the given slice (based on the source) since we do not want to
+// generate multiple config files for the same source image repo. Output is sorted to ensure it is deterministic.
+func mergeMirrorSets(baseMirrorSets []mirrorSet) []mirrorSet {
+	// Map to keep track of unique mirrorSets by source
+	uniqueMirrorSets := make(map[string]mirrorSet)
+
+	for _, ms := range baseMirrorSets {
+		if existingMS, ok := uniqueMirrorSets[ms.source]; ok {
+			// If the source already exists, merge its mirrors slices and update the map entry
+			existingMS.mirrors = mergeMirrors(existingMS.mirrors, ms.mirrors)
```
sebsoto, line 179: what if the mirror sets have conflicting a mirrorSourcePolicy?

sebsoto (changes requested):

controllers/registry_controller.go
saifshaikh48: Good point, I looked at what MCO does and it looks like "NeverContactSource" wins out if there is conflict. Erring on the side of caution security wise

controllers/registry_controller.go
saifshaikh48: I actually tried this earlier but it made it difficult to unit test the merging logic

controllers/registry_controller.go
saifshaikh48: Fair enough, removed to not make assumptions about future implementation

controllers/registry_controller.go
```diff
@@ -71,9 +86,129 @@ func (r *registryReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
...
+	}
+
+	mergedMirrorSets := collectMirrorSets(imageDigestMirrorSetList.Items, imageTagMirrorSetList.Items)
+
 	return ctrl.Result{}, nil
 }
 
+// collectMirrorSets extracts the contents of the given mirror sets and merges them into a single collection.
+// Each member of the collection represents the registry configuration for a specific source.
+func collectMirrorSets(idmsItems []config.ImageDigestMirrorSet, idtsItems []config.ImageTagMirrorSet) []mirrorSet {
```
mansikulkarni96, line 134: Can these helper methods be placed into another package say registry and be unit tested in that package?

mansikulkarni96 (changes requested):

controllers/registry_controller.go
sebsoto: I was thinking something similar with something like
```
type RegistryConfig struct {
  mirrors [] mirrorset
}
func NewRegistryConfig([]imageDigestMirrorSetList, ImageTagMirrorSetList) (*RegistryConfig,err){
  //Populates and sorts mirrorset list
  ...
}
// with method:

// outputs a config file representation
func (r * RegistryConfig) ToFile() []byte {}


```

controllers/registry_controller.go
saifshaikh48: Don't think so. Here's the package they use for merging and editing registry config https://pkg.go.dev/github.com/openshift/runtime-utils/pkg/registries#EditRegistriesConfig

controllers/registry_controller.go
saifshaikh48: Sure I will create a new `registries` package and top level struct

I don't think I can implement ToFile yet, will likely have to wait for the next story due to the [same problem as here](https://github.com/openshift/windows-machine-config-operator/pull/2140#discussion_r1567940984)

controllers/registry_controller.go
```diff
@@ -0,0 +1,97 @@
...
+// related to image registry config and aims to move the current state of the cluster closer to the desired state.
+func (r *registryReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, err error) {
+	r.log = r.log.WithValues(RegistryController, req.NamespacedName)
+	// Prevent WMCO upgrades while registry configuration settings are being processed
+	if err := condition.MarkAsBusy(r.client, r.watchNamespace, r.recorder, RegistryController); err != nil {
+		return ctrl.Result{}, err
+	}
+	defer func() {
+		err = markAsFreeOnSuccess(r.client, r.watchNamespace, r.recorder, RegistryController, result.Requeue, err)
+	}()
```
sebsoto, line 73: Do upgrades really have to be prevented by this controller?
The whole containerd directory is being wiped on upgrade and re-created.
We should be cautious when adding this lock, as there is the small chance of deadlock when upgrading the cluster due to changes in the payload.

pkg/registries/registries_test.go
```diff
@@ -0,0 +1,328 @@
...
+			assert.Equal(t, out, test.expectedOutput)
+		})
+	}
+}
+
+func TestMergeMirrorSets(t *testing.T) {
+	testCases := []struct {
+		name           string
+		input          []mirrorSet
+		expectedOutput []mirrorSet // source and mirror orders matter since result is expected to be sorted alphabetically
```
sebsoto, line 87: please stick the comment above the line, so we have a consistent code style across the repo.

pkg/registries/registries_test.go
```diff
@@ -0,0 +1,328 @@
...
+		if len(a[i].mirrors) != len(b[i].mirrors) {
+			return false
+		}
+		for j := range a[i].mirrors {
+			if a[i].mirrors[j].host != b[i].mirrors[j].host || a[i].mirrors[j].resolveTags != b[i].mirrors[j].resolveTags {
+				return false
+			}
+		}
+	}
+	return true
```
sebsoto, line 238: Why not assert.Equal? (Uses reflect.DeepEqual under the hood)

sebsoto (changes requested):
Thanks @saifshaikh48 
Almost LGTM.

pkg/registries/registries_test.go
saifshaikh48: makes sense, I had thought for some reason assert.Equal wouldn't do a strict checking of slice indices

controllers/registry_controller.go
saifshaikh48: I was thinking of a case where our wm or cm controllers were upgrading a node (i.e. deleting directories) and, at the same time, the registry controller is trying to write new config to the containerd dir. Wouldn't that clash?

pkg/registries/registries.go
```diff
@@ -0,0 +1,139 @@
...
+
+	// Iterate over existing mirrors and add them to the map
+	for _, m := range existingMirrors {
+		uniqueMirrors[m.host] = m
+	}
+	// Iterate over new mirrors
+	for _, m := range newMirrors {
+		if existingM, ok := uniqueMirrors[m.host]; ok {
+			// If the mirror already exists, check the resolveTags field. Resolving by tag is preferred over by digest.
+			if !existingM.resolveTags && m.resolveTags {
```
mansikulkarni96, line 125: If both existing and new mirrors have resolveTags set to true, would we prefer one over the other based on other attributes, such as any priority or creation time?

saifshaikh48: Don't see why we'd need to. `mirror` structs only have 2 fields, and if both are the same them then the objects are equivalent -- it wouldn't matter which copy we keep

controllers/registry_controller.go
sebsoto: > Wouldn't that clash?
Potentially, but as far as I can tell the impact would be the registry controller fails and retries once the node is upgraded.

saifshaikh48: Fair enough, I've removed this

jrvaldes:
/lgtm
/hold

for `No pip available exiting` issue in CI

saifshaikh48:
/retest-required
/unhold

pip issue is fixed

saifshaikh48:
/hold

pip issue still seen in nutanix job

saifshaikh48:
/retest-required
/unhold

https://github.com/openshift/release/pull/51327 should fix the pip issues in nutanix and vsphere jobs

PR #2142, opened 2024-04-22
WINC-1271: Remove unused CNI config field

sebsoto:
Removes the subnet field in the IPAM configuration, as it is outside the spec, and not used.

### DIFF
--- a/pkg/nodeconfig/payload/payload.go
+++ b/pkg/nodeconfig/payload/payload.go
@@ -91,8 +91,7 @@ $cni_template=@'
         "dns":true
     },
     "ipam":{
-        "type":"host-local",
-        "subnet":"ovn_host_subnet"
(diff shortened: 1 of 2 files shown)


sebsoto:
Spec: https://github.com/containernetworking/plugins/blob/3bc00017e302752374a522221e7feceafa37b4b1/pkg/hns/netconf_windows.go#L32

Usage: https://github.com/selansen/plugins/blob/8b8825bcd81fc20346ae21d049e497bd9cab2465/plugins/main/windows/win-overlay/win-overlay_windows.go

jrvaldes (commented):
@sebsoto  thanks for working on this. As per source code, changes LGTM. 

Looks like the e2e failed, PTAL and check if the failure is related to this change. I will `/lgtm` with  a greener CI signal.

PR #2156, opened 2024-05-02
WINC-588: Expose metrics over https

mansikulkarni96:
Add kube-rbac-proxy container to WMCO deployment
listening on :8443 and forwarding traffic to :9182
Update ServiceMonitor to tell Prometheus to scrape using https.
Update role permissionkube-rbac-proxy container
Copy webconfig.yaml that stores the location of TLS certs.
React to wmco-tls secret created by service-ca operator that holds the serving cert and key used for metrics authentication.


bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
```diff
@@ -441,6 +453,31 @@ spec:
...
+                  requests:
+                    cpu: 10m
+                    memory: 20Mi
+                volumeMounts:
+                - mountPath: /etc/tls/private
+                  name: wmco-tls
+              volumes:
+              - name: wmco-tls
+                secret:
+                  secretName: wmco-tls
```
jrvaldes, line 481: consider `windows-machine-config-operator-tls` to align with the name of the operator.

pkg/tls/tls.go
```diff
@@ -0,0 +1,6 @@
+package tls
+
+const (
+	// Secret is the name of the userData secret that WMCO creates
+	Secret = "wmco-tls"
```
jrvaldes, line 5:
```suggestion
	SecretName = "wmco-tls"
```

pkg/metadata/metadata.go
```diff
@@ -25,6 +25,8 @@ const (
 	RebootAnnotation = "windowsmachineconfig.openshift.io/reboot-required"
 	// UpgradingLabel indicates the node's underlying instance is performing an upgrade
 	UpgradingLabel = "windowsmachineconfig.openshift.io/upgrading"
+	// TlsAnnotation indicates the node's TLS certs need update
+	TlsAnnotation = "windowsmachineconfig.openshift.io/tls-update"
```
jrvaldes, line 29:
```suggestion
	TlsUpdateRequiredAnnotation = "windowsmachineconfig.openshift.io/tls-update-required"
```

so that is clear the node needs attention. Other suggestions:
-  `tls-mismatch`
-  `tls-failure`

pkg/nodeconfig/nodeconfig.go
```diff
@@ -162,12 +163,27 @@ func (nc *nodeConfig) Configure() error {
...
+	tlsSecret := &core.Secret{}
+	err = nc.client.Get(context.TODO(), types.NamespacedName{Name: tls.Secret,
+		Namespace: nc.wmcoNamespace}, tlsSecret)
+	if err != nil {
+		return fmt.Errorf("unable to get secret %s: %w", nc.wmcoNamespace, err)
+	}
+	var tlsPresent bool
+	if _, ok := tlsSecret.Annotations[metadata.TlsAnnotation]; ok {
+		tlsPresent = true
+	}
```
jrvaldes, line 175:
```suggestion
	_, tlsPresent := tlsSecret.Annotations[metadata.TlsAnnotation]
```

pkg/metadata/metadata.go
```diff
@@ -126,6 +128,34 @@ func ApplyRebootAnnotation(ctx context.Context, c client.Client, node core.Node)
 	return ApplyLabelsAndAnnotations(ctx, c, node, nil, map[string]string{RebootAnnotation: ""})
 }
 
+// ApplyTlsAnnotation applies an annotation to the given secret communicating that the TLS certs need update
+func ApplyTlsAnnotation(ctx context.Context, c client.Client, secret *core.Secret) error {
```
jrvaldes, line 132:
```suggestion
func ApplySecretTlsAnnotation(ctx context.Context, c client.Client, secret *core.Secret) error {
```

to avoid confusion with node' annotations helper functions.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -162,12 +163,27 @@ func (nc *nodeConfig) Configure() error {
...
 	}
 
+	tlsSecret := &core.Secret{}
+	err = nc.client.Get(context.TODO(), types.NamespacedName{Name: tls.Secret,
+		Namespace: nc.wmcoNamespace}, tlsSecret)
+	if err != nil {
+		return fmt.Errorf("unable to get secret %s: %w", nc.wmcoNamespace, err)
+	}
+	var tlsPresent bool
+	if _, ok := tlsSecret.Annotations[metadata.TlsAnnotation]; ok {
```
jrvaldes, line 173: @mansikulkarni96  I'm not clear where/who initially sets the  proposed `windowsmachineconfig.openshift.io/tls-update` annotation to the secret?

jrvaldes (changes requested):
@mansikulkarni96  thanks for working on this, PTAL at my comments.

pkg/nodeconfig/nodeconfig.go
mansikulkarni96: It is set in secret_controller.go-> https://github.com/openshift/windows-machine-config-operator/blob/cc01700785ac78e1e47a68ea59823a48d473b767/controllers/secret_controller.go#L165

bundle/manifests/windows-exporter_monitoring.coreos.com_v1_servicemonitor.yaml
```diff
@@ -10,6 +10,11 @@ spec:
     interval: 30s
     path: /metrics
     port: metrics
+    scheme: https
+    bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+    tlsConfig:
+      caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
+      serverName: windows-exporter.openshift-windows-machine-config-operator.svc
```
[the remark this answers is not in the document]
sebsoto: Seems like it to me.

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
```diff
@@ -441,6 +453,31 @@ spec:
                   requests:
                     cpu: 20m
                     memory: 300Mi
+              - name: kube-rbac-proxy
+                image: quay.io/openshift/origin-kube-rbac-proxy:latest
```
sebsoto, line 458: This is being pulled from the internet at runtime. We cannot do this.
Is a build of this image being shipped as part of the payload?

pkg/windows/windows.go
```diff
@@ -54,7 +54,10 @@ const (
 	CniConfDir = cniDir + "\\config"
 	// ContainerdDir is the directory for storing Containerd binary
 	ContainerdDir = K8sDir + "\\containerd"
-	// ContainerdPath is the location of the containerd exe
+	// TlsDir is the directory for storing WMCO tls certs
+	TlsDir = K8sDir + "\\tls"
```
sebsoto, line 58: TLSDir

pkg/windows/windows.go
```diff
@@ -54,7 +54,10 @@ const (
...
+	// TlsConfPath is the location of TLS config files
+	TlsConfPath    = TlsDir + "\\webconfig.yaml"
```
sebsoto, line 60: TLSConfPath

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
```diff
@@ -444,7 +445,7 @@ spec:
                 - name: OPERATOR_NAME
                   value: windows-machine-config-operator
                 image: REPLACE_IMAGE
-                imagePullPolicy: IfNotPresent
+                imagePullPolicy: Always
```
sebsoto, line 448: Why is this change being done?

controllers/secret_controller.go
```diff
@@ -149,6 +155,19 @@ func (r *SecretReconciler) Reconcile(ctx context.Context,
...
+		tlsSecret := &core.Secret{}
+		err := r.client.Get(context.TODO(), kubeTypes.NamespacedName{Name: tls.SecretName,
+			Namespace: r.watchNamespace}, tlsSecret)
+		if err != nil {
+			return reconcile.Result{}, fmt.Errorf("unable to get secret %s: %w", r.watchNamespace, err)
+		}
+		err = metadata.ApplySecretTlsAnnotation(ctx, r.client, tlsSecret)
+		if err != nil {
+			return reconcile.Result{}, fmt.Errorf("unable to apply TLS annotation %s: %w", r.watchNamespace, err)
+		}
```
sebsoto, line 168: You need to return at the end of this segment.

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
```diff
@@ -441,6 +453,31 @@ spec:
                   requests:
                     cpu: 20m
                     memory: 300Mi
+              - name: kube-rbac-proxy
+                image: quay.io/openshift/origin-kube-rbac-proxy:latest
+                imagePullPolicy: Always
+                args:
```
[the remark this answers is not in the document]
mansikulkarni96: Most operators I referred to don't use them like the CMO and cloud credential operator. However, MCO specifies it and I don't see any harm in doing it.

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
```diff
@@ -441,6 +453,31 @@ spec:
...
+                  - --upstream=http://127.0.0.1:9182/
+                  - --tls-cert-file=/etc/tls/private/tls.crt
+                  - --tls-private-key-file=/etc/tls/private/tls.key
+                  - --logtostderr=true
+                  - --v=10
+                ports:
+                  - containerPort: 8443
+                    protocol: TCP
+                    name: metrics
+                resources:
```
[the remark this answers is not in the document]
mansikulkarni96: I couldn't find any example limits and I didn't find them defined in any of the operators.

bundle/manifests/windows-exporter_monitoring.coreos.com_v1_servicemonitor.yaml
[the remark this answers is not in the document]
mansikulkarni96: its based on the service namespace, windows-exporter service object is created in operator namespace.

sebsoto (changes requested):

pkg/windows/windows.go
```diff
@@ -475,10 +476,17 @@ func (vm *windows) Bootstrap(desiredVer, watchNamespace, wicdKubeconfigContents
 	if err := vm.createDirectories(); err != nil {
 		return fmt.Errorf("error creating directories on Windows VM: %w", err)
 	}
+	if tlsPresent {
+		if err := vm.EnsureFileContent(tlsData["tls.crt"], "tls.crt", TlsDir); err != nil {
+			return fmt.Errorf("unable to transfer TLS certs: %w", err)
+		}
+		if err := vm.EnsureFileContent(tlsData["tls.key"], "tls.key", TlsDir); err != nil {
+			return fmt.Errorf("unable to transfer TLS key: %w", err)
+		}
```
sebsoto, line 485: This should be done within transferFiles(). The files to transfer are setup as part of windows.New()

pkg/nodeconfig/nodeconfig.go
```diff
@@ -162,12 +163,24 @@ func (nc *nodeConfig) Configure() error {
 		return err
 	}
 
+	tlsSecret := &core.Secret{}
+	err = nc.client.Get(context.TODO(), types.NamespacedName{Name: tls.SecretName,
+		Namespace: nc.wmcoNamespace}, tlsSecret)
+	if err != nil {
+		return fmt.Errorf("unable to get secret %s: %w", nc.wmcoNamespace, err)
+	}
+	_, tlsPresent := tlsSecret.Annotations[metadata.TlsUpdateRequiredAnnotation]
```
sebsoto, line 172: Why is this necessary?

pkg/metadata/metadata.go
```diff
@@ -126,6 +128,34 @@ func ApplyRebootAnnotation(ctx context.Context, c client.Client, node core.Node)
 	return ApplyLabelsAndAnnotations(ctx, c, node, nil, map[string]string{RebootAnnotation: ""})
 }
 
+// ApplySecretTlsAnnotation applies an annotation to the given secret communicating that the TLS certs need update
```
sebsoto, line 131: > communicating that the TLS certs need update

Communicating to who?

sebsoto (changes requested):

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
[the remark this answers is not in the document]
jrvaldes: With a high number of requests, the reverse proxy pod may run out of resources, I recommend adding limits.

pkg/windows/windows.go
```diff
@@ -120,7 +123,7 @@ const (
 	// WindowsExporterServiceCommand specifies metrics for the windows_exporter service to collect
 	// and expose metrics at endpoint with default port :9182 and default URL path /metrics
 	WindowsExporterServiceCommand = windowsExporterPath + " --collectors.enabled " +
-		"cpu,cs,logical_disk,net,os,service,system,textfile,container,memory,cpu_info"
+		"cpu,cs,logical_disk,net,os,service,system,textfile,container,memory,cpu_info --web.config.file " + TlsConfPath
```
saifshaikh48, line 126: consider fmt.Sprintf for readablity of this command

controllers/secret_controller.go
saifshaikh48: In addition, I suggest pulling this out into a helper func so this call site could become something like 
`return reconcile.Result{}, r.reconcileTLSSecret()`

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
```diff
@@ -441,6 +454,32 @@ spec:
                   requests:
                     cpu: 20m
                     memory: 300Mi
+              - name: kube-rbac-proxy
+                image: quay.io/openshift/origin-kube-rbac-proxy:latest
+                imagePullPolicy: IfNotPresent
+                args:
+                  - --secure-listen-address=0.0.0.0:8443
+                  - --upstream=http://127.0.0.1:9182/
```
jrvaldes, line 462: Looks like the `windows-service` is now [getting tls_server_config](https://github.com/openshift/windows-machine-config-operator/pull/2156/files#diff-13dd767f866337233005dff5673462886531a69cd167f2e4e03dcfcb20f8652aR1) so that it will use it in the default port,  that is 9182. 

Hence, the upstream address in the proxy may need the same set of credentials to communicate with the local port opened by the  `windows-service`.  As per docs:

```
      --upstream-client-cert-file string            If set, the client will be used to authenticate the proxy to upstream. Requires --upstream-client-key-file to be set, too.
      --upstream-client-key-file string             The key matching the certificate from --upstream-client-cert-file. If set, requires --upstream-client-cert-file to be set, too.
 ```

File: controllers/secret_controller.go [hunk 20]
```diff
@@ -79,6 +80,12 @@ func (r *SecretReconciler) SetupWithManager(mgr ctrl.Manager) error {
 					return true
 				}
 			}
+			if isTlsSecret(e.ObjectNew, r.watchNamespace) {
+				if string(e.ObjectOld.(*core.Secret).Data[secrets.TLSSecret]) !=
+					string(e.ObjectNew.(*core.Secret).Data[secrets.TLSSecret]) {
+					return true
+				}
+			}
```
saifshaikh48, line 88: else if isTlsSecret(e.ObjectNew, r.watchNamespace) {
				return string(e.ObjectOld.(*core.Secret).Data[secrets.TLSSecret]) !=
					string(e.ObjectNew.(*core.Secret).Data[secrets.TLSSecret]) 
			}

controllers/secret_controller.go
```diff
@@ -64,9 +65,9 @@ func (r *SecretReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 	}
 
-	privateKeyPredicate := builder.WithPredicates(predicate.Funcs{
+	secretPredicate := builder.WithPredicates(predicate.Funcs{
 		CreateFunc: func(e event.CreateEvent) bool {
-			return isPrivateKeySecret(e.Object, r.watchNamespace)
+			return isPrivateKeySecret(e.Object, r.watchNamespace) || isTlsSecret(e.Object, r.watchNamespace)
 		},
 		DeleteFunc: func(e event.DeleteEvent) bool {
 			return isPrivateKeySecret(e.Object, r.watchNamespace)
```
saifshaikh48, line 73: With no reconcile for deletion event, the deleted secret may still be in use on Windows nodes. Is that the desired behavior?

build/Dockerfile
```diff
@@ -121,7 +121,9 @@ RUN make build-daemon
 #│   ├── gcp-get-hostname.ps1
 #│   ├── windows-defender-exclusion.ps1
 #│   └── hns.psm1
-#├── windows_exporter.exe
+#├── windows-exporter/
+#│   ├── windows_exporter.exe
+#│   ├── webconfig.yaml
```
saifshaikh48, line 126: nittiest of nits: `└──`

pkg/windows/windows.go
```diff
@@ -735,13 +734,19 @@ func (vm *windows) removeDirectories() error {
...
 		if err := vm.EnsureFile(src, dest); err != nil {
 			return fmt.Errorf("error copying %s to %s: %w", src.Path, dest, err)
 		}
 	}
+	if err := vm.EnsureFileContent(tlsData["tls.crt"], "tls.crt", TLSDir); err != nil {
+		return fmt.Errorf("unable to transfer TLS certs: %w", err)
+	}
+	if err := vm.EnsureFileContent(tlsData["tls.key"], "tls.key", TLSDir); err != nil {
+		return fmt.Errorf("unable to transfer TLS key: %w", err)
+	}
```
saifshaikh48, line 749: consider pulling this out into a helper (and maybe making tls.crt and tls.key local constants). will allow you to remove duplicated calls in secret_controller.reconcileTLSSecret()

config/rbac/role.yaml
```diff
@@ -241,3 +242,10 @@ rules:
   - securitycontextconstraints
   verbs:
   - use
+- apiGroups:
+  - monitoring.coreos.com
+  resources:
+  - servicemonitors
+  verbs:
+  - create
+  - get
```
saifshaikh48, line 252: curious how this was generated? do we need to add `//+kubebuilder:rbac` in metrics.go?

pkg/windows/windows.go
mansikulkarni96: Unsure if we should add it to the interface since it will not be a genric helper.

controllers/secret_controller.go
```diff
@@ -191,6 +207,49 @@ func (r *SecretReconciler) Reconcile(ctx context.Context,
...
+	tlsSecret := &core.Secret{}
+	err := r.client.Get(context.TODO(), kubeTypes.NamespacedName{Name: secrets.TLSSecret,
+		Namespace: r.watchNamespace}, tlsSecret)
+	if err != nil {
+		if k8sapierrors.IsNotFound(err) {
+			// Request object not found, could have been deleted after reconcile request.
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			// Return and don't requeue
+			return nil
+		}
```
saifshaikh48, line 220: Will this case be hit on a deletion event? If so, we need to reconcile here to remove the old cert and key files from each Windows node

pkg/nodeconfig/payload/payload.go
```diff
@@ -63,9 +63,13 @@ const (
...
 	// WindowsExporterName is the name of the Windows metrics exporter executable
 	WindowsExporterName = "windows_exporter.exe"
+	// WindowsExporterDirectory is the directory for storing the windows-exporter binary and the TLS webconfig file
+	WindowsExporterDirectory = "windows-exporter/"
 	// WindowsExporterPath contains the path of the windows_exporter binary. The container image should already have
 	// this binary mounted
-	WindowsExporterPath = payloadDirectory + WindowsExporterName
+	WindowsExporterPath = payloadDirectory + WindowsExporterDirectory + WindowsExporterName
+	// TlsConfPath contains the path of the TLS config file
+	TlsConfPath = payloadDirectory + WindowsExporterDirectory + "webconfig.yaml"
```
sebsoto, line 72: `TLSConfPath`

bundle/manifests/windows-exporter_v1_service.yaml
```diff
@@ -5,11 +5,15 @@ metadata:
...
+  annotations:
+    service.alpha.openshift.io/serving-cert-secret-name: windows-machine-config-operator-tls
 spec:
   ports:
   - name: metrics
     port: 9182
     protocol: TCP
     targetPort: 9182
+  selector:
+    app.kubernetes.io/name: windows-exporter
```
sebsoto, line 17: What is this change doing?

bundle/manifests/windows-exporter_v1_service.yaml
```diff
@@ -5,11 +5,15 @@ metadata:
   labels:
     name: windows-exporter
   name: windows-exporter
+  annotations:
+    service.alpha.openshift.io/serving-cert-secret-name: windows-machine-config-operator-tls
```
sebsoto, line 9: This annotation seems unrelated to the servicemonitor addition. If that is true, moving them to separate commits would be best.

pkg/metrics/metrics.go
```diff
@@ -299,3 +304,66 @@ func (c *Config) createEndpoint(subsets []v1.EndpointSubset) error {
...
+// ensureServiceMonitor creates a serviceMonitor object in the operator namespace if it does not exist.
+func (c *Config) ensureServiceMonitor() error {
+	// get existing serviceMonitor object if it exists
+	_, err := c.ServiceMonitors(c.namespace).Get(context.TODO(), WindowsMetricsResource, metav1.GetOptions{})
+	if err == nil {
+		return nil
+	}
+	if err != nil && !apierrors.IsNotFound(err) {
+		return fmt.Errorf("error retrieving %s serviceMonitor: %w", WindowsMetricsResource, err)
+	}
```
sebsoto, line 317: You are not checking if it has the expected state or not.

pkg/metrics/metrics.go
```diff
@@ -299,3 +304,66 @@ func (c *Config) createEndpoint(subsets []v1.EndpointSubset) error {
 	}
 	return nil
 }
+
+// ensureServiceMonitor creates a serviceMonitor object in the operator namespace if it does not exist.
+func (c *Config) ensureServiceMonitor() error {
```
sebsoto, line 309: Why is this being moved from the static config file?
That should be called out in the commit message

pkg/internal/webconfig.yaml
```diff
@@ -0,0 +1,3 @@
+tls_server_config:
```
sebsoto, line 1: The name of this file makes it unclear of its purpose.
Please name it `windows-exporter-config.yaml` or something similar.

controllers/secret_controller.go
sebsoto: Yes this needs cleanup logic which needs to be handled with a finalizer.
Without one the controller will not be able to reliably react to the deletion event before the secret is actually deleted.

pkg/windows/windows.go
```diff
@@ -735,13 +734,19 @@ func (vm *windows) removeDirectories() error {
 }
 
 // transferFiles copies various files required for configuring the Windows node, to the VM.
-func (vm *windows) transferFiles() error {
+func (vm *windows) transferFiles(tlsData map[string][]byte) error {
```
sebsoto, line 737: Rather than making this a special case this should be handled in the same way the other files are. It should be part of `filesToTransfer`, which is generated when the Windows object is created. Doing it this way is adding additional arguments to any function which needs to call `transferFiles`, and is moving away from the established pattern.

controllers/secret_controller.go
```diff
@@ -149,6 +161,10 @@ func (r *SecretReconciler) Reconcile(ctx context.Context,
 			result.Requeue, reconcileErr)
 	}()
 
+	if request.NamespacedName.Name == secrets.TLSSecret {
+		return reconcile.Result{}, r.reconcileTLSSecret(ctx)
+	}
```
sebsoto, line 166: By adding this to the function, this `Reconcile` function is now no longer has a single purpose, but is reconciling different secrets in different ways.
Please move the rest of this function into a new function `reconcileUserDataSecret` so the separation between these two purposes is clear.

pkg/windows/windows.go
sebsoto: If this doesn't make sense for this file, which may be the case, then it should be not be handled by the `transferFiles` function.

sebsoto: Yes looking at this more it seems like this shouldn't be handled here, rather in the createBootstrapFiles function https://github.com/openshift/windows-machine-config-operator/blob/0531500ee3ef8e88fabed3e05087338f733164e6/pkg/nodeconfig/nodeconfig.go#L344

sebsoto (changes requested):

controllers/secret_controller.go
sebsoto: Or does it need cleanup logic? I suppose all that matters is if a new secret is created the certs on the node are replaced.

bundle/manifests/windows-exporter_v1_service.yaml
mansikulkarni96: moved to the commit where this secret is used.

mansikulkarni96:
aws-e2e-operator

controllers/configmap_controller.go
```diff
@@ -160,9 +160,9 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context,
 
 	// Fetch the ConfigMap. The predicate will have filtered out any ConfigMaps that we should not reconcile
 	// so it is safe to assume that all ConfigMaps being reconciled are one of:
-	// 1. windows-instances, describing hosts that need to be present in the cluster.
 	// 2. windows-services, describing expected configuration of WMCO-managed services on all Windows instances
 	// 3. kube-apiserver-to-kubelet-client-ca, contains the CA for the kubelet to recognize the kube-apiserver client cert
+	// 1. windows-instances, describing hosts that need to be present in the cluster.
```
saifshaikh48, line 165: can undo this

controllers/secret_controller.go
```diff
@@ -138,8 +147,9 @@ type SecretReconciler struct {
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
 func (r *SecretReconciler) Reconcile(ctx context.Context,
 	request ctrl.Request) (result reconcile.Result, reconcileErr error) {
-	log := r.log.WithValues(SecretController, request.NamespacedName)
+	_ = r.log.WithValues(SecretController, request.NamespacedName)
```
saifshaikh48, line 150: I think this the left side should become `r.log` since WithValues returns a new logr

https://sdk.operatorframework.io/docs/building-operators/golang/references/logging/#creating-a-structured-log-statement

pkg/windows/windows.go
```diff
@@ -545,7 +545,7 @@ func (vm *windows) Bootstrap(desiredVer, watchNamespace, wicdKubeconfigContents
 	if err := vm.transferFiles(); err != nil {
 		return fmt.Errorf("error transferring files to Windows VM: %w", err)
 	}
-
+
```
saifshaikh48, line 548: undo for linter

controllers/secret_controller.go
```diff
@@ -149,48 +159,101 @@ func (r *SecretReconciler) Reconcile(ctx context.Context,
...
 	if err != nil {
 		if k8sapierrors.IsNotFound(err) {
 			// Private key secret was not found, could have been deleted after reconcile request.
 			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
 			// Return and don't requeue
-			return reconcile.Result{}, nil
+			return nil
 		}
-		return reconcile.Result{}, fmt.Errorf("unable to get secret %s: %w", request.NamespacedName, err)
+		return fmt.Errorf("unable to get secret %s: %w", r.watchNamespace, err)
```
sebsoto, line 184: r.watchNamespace doesnt make sense here

pkg/nodeconfig/nodeconfig.go
```diff
@@ -169,11 +172,11 @@ func (nc *nodeConfig) Configure() error {
 	}
 
 	wmcoVersion := version.Get()
+
```
sebsoto, line 175: revert

pkg/nodeconfig/nodeconfig.go
```diff
@@ -169,11 +172,11 @@ func (nc *nodeConfig) Configure() error {
...
 	// Start all required services to bootstrap a node object using WICD
 	if err := nc.Windows.Bootstrap(wmcoVersion, nc.wmcoNamespace, wicdKC); err != nil {
 		return fmt.Errorf("bootstrapping the Windows instance failed: %w", err)
 	}
-
```
sebsoto, line 176: revert

pkg/nodeconfig/nodeconfig.go
```diff
@@ -615,6 +618,22 @@ func (nc *nodeConfig) UpdateTrustedCABundleFile(data string) error {
 	return nc.Windows.EnsureFileContent([]byte(data), fileName, dir)
 }
 
+func (nc *nodeConfig) createTLSCerts() error {
```
sebsoto, line 621: needs doc

sebsoto (approved):
Good work @mansikulkarni96
Approved but please address all open comments

PR #2163, opened 2024-05-07
WINC-1222: WMCO applies image mirroring settings to Windows instances

saifshaikh48:
- [WINC-1222](https://issues.redhat.com//browse/WINC-1222): WMCO applies image mirroring settings to Windows instances
- [WINC-1294](https://issues.redhat.com//browse/WINC-1294): Support token-based auth for mirror registries

This PR transfers containerd registry config files generated by WMCO to
each Windows instance based on the cluster's global image mirroring settings,
allowing containerd to pull Windows images from user-defined mirror locations.

The containerd daemon picks up the registry config without requiring restart.

Containerd also becomes auth-aware in order to support pulling images from secure registries.
We add a header field to each host listed in hosts.tomls with the repo's auth token.
The token for each mirror is taken from cluster's global pull secret.

### DIFF
--- a/controllers/registry_controller.go
+++ b/controllers/registry_controller.go
@@ -21,13 +21,20 @@ import (
 	"fmt"
 
 	config "github.com/openshift/api/config/v1"
+	core "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
 	"github.com/openshift/windows-machine-config-operator/pkg/cluster"
+	"github.com/openshift/windows-machine-config-operator/pkg/nodeconfig"
 	"github.com/openshift/windows-machine-config-operator/pkg/registries"
+	"github.com/openshift/windows-machine-config-operator/pkg/secrets"
+	"github.com/openshift/windows-machine-config-operator/pkg/signer"
+	"github.com/openshift/windows-machine-config-operator/pkg/windows"
 )
 
 //+kubebuilder:rbac:groups="config.openshift.io",resources=imagedigestmirrorsets,verbs=get;list;watch
@@ -68,19 +75,37 @@ func NewRegistryReconciler(mgr manager.Manager, clusterConfig cluster.Config,
 func (r *registryReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, err error) {
 	r.log = r.log.WithValues(RegistryController, req.NamespacedName)
 
-	// List all IDMS/ITMS resources
-	imageDigestMirrorSetList := &config.ImageDigestMirrorSetList{}
-	if err = r.client.List(ctx, imageDigestMirrorSetList); err != nil {
-		return ctrl.Result{}, fmt.Errorf("error getting IDMS list: %w", err)
-	}
-	imageTagMirrorSetList := &config.ImageTagMirrorSetList{}
-	if err = r.client.List(ctx, imageTagMirrorSetList); err != nil {
-		return ctrl.Result{}, fmt.Errorf("error getting ITMS list: %w", err)
+	configFiles, err := registries.GenerateConfigFiles(ctx, r.client)
+	if err != nil {
+		return ctrl.Result{}, err
 	}
 
-	_ = registries.NewRegistryConfig(imageDigestMirrorSetList.Items, imageTagMirrorSetList.Items)
-	// TODO: transfer generated config files to Windows nodes as part of WINC-1222
-
+	// Transfer the generated registry config folder to each Windows node, completely replacing any existing config
+	nodes := &core.NodeList{}
+	if err := r.client.List(ctx, nodes, client.MatchingLabels{core.LabelOSStable: "windows"}); err != nil {
+		return ctrl.Result{}, err
+	}
+	r.signer, err = signer.Create(types.NamespacedName{Namespace: r.watchNamespace, Name: secrets.PrivateKeySecret},
+		r.client)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("unable to create signer from private key secret: %w", err)
+	}
+	for _, node := range nodes.Items {
+		winInstance, err := r.instanceFromNode(&node)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("unable to create instance object from node: %w", err)
+		}
+		nc, err := nodeconfig.NewNodeConfig(r.client, r.k8sclientset, r.clusterServiceCIDR, r.watchNamespace,
+			winInstance, r.signer, nil, nil, r.platform)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to create new nodeconfig: %w", err)
+		}
+		// TODO: If this flakes for any one node, we have to loop over all nodes again and re-transfer the directory to
+		// all nodes. We should fix this as part of https://issues.redhat.com/browse/WINC-1306
+		if err := nc.Windows.ReplaceDir(configFiles, windows.ContainerdConfigDir); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
 	return ctrl.Result{}, nil
 }
 
--- a/controllers/registry_controller.go
+++ b/controllers/registry_controller.go
@@ -25,9 +25,12 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	"github.com/openshift/windows-machine-config-operator/pkg/cluster"
 	"github.com/openshift/windows-machine-config-operator/pkg/nodeconfig"
@@ -100,6 +103,8 @@ func (r *registryReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
 		if err != nil {
 			return ctrl.Result{}, fmt.Errorf("failed to create new nodeconfig: %w", err)
 		}
+
+		r.log.Info("updating containerd config", "directory", windows.ContainerdConfigDir, "node", node.Name)
 		// TODO: If this flakes for any one node, we have to loop over all nodes again and re-transfer the directory to
 		// all nodes. We should fix this as part of https://issues.redhat.com/browse/WINC-1306
 		if err := nc.Windows.ReplaceDir(configFiles, windows.ContainerdConfigDir); err != nil {
@@ -111,8 +116,22 @@ func (r *registryReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *registryReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	mirrorSetPredicate := predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			return true
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return true
+		},
+		GenericFunc: func(e event.GenericEvent) bool {
+			return true
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return true
+		},
+	}
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&config.ImageDigestMirrorSet{}).
-		Watches(&config.ImageTagMirrorSet{}, &handler.EnqueueRequestForObject{}).
+		For(&config.ImageDigestMirrorSet{}, builder.WithPredicates(mirrorSetPredicate)).
+		Watches(&config.ImageTagMirrorSet{}, &handler.EnqueueRequestForObject{}, builder.WithPredicates(mirrorSetPredicate)).
 		Complete(r)
 }
--- a/pkg/nodeconfig/nodeconfig.go
+++ b/pkg/nodeconfig/nodeconfig.go
@@ -35,6 +35,7 @@ import (
 	"github.com/openshift/windows-machine-config-operator/pkg/instance"
 	"github.com/openshift/windows-machine-config-operator/pkg/metadata"
 	"github.com/openshift/windows-machine-config-operator/pkg/nodeutil"
+	"github.com/openshift/windows-machine-config-operator/pkg/registries"
 	"github.com/openshift/windows-machine-config-operator/pkg/retry"
 	"github.com/openshift/windows-machine-config-operator/pkg/secrets"
 	"github.com/openshift/windows-machine-config-operator/pkg/windows"
@@ -150,6 +151,9 @@ func (nc *nodeConfig) Configure() error {
 	}
 
 	if err := nc.createBootstrapFiles(); err != nil {
+		return err
+	}
+	if err := nc.createRegistryConfigFiles(); err != nil {
 		return err
 	}
 	if cluster.IsProxyEnabled() {
@@ -368,6 +372,15 @@ func (nc *nodeConfig) write(pathToData map[string]string) error {
 		}
 	}
 	return nil
+}
+
+// createRegistryConfigFiles creates all files on the node required for containerd to mirror images
+func (nc *nodeConfig) createRegistryConfigFiles() error {
+	configFiles, err := registries.GenerateConfigFiles(context.TODO(), nc.client)
+	if err != nil {
+		return err
+	}
+	return nc.Windows.ReplaceDir(configFiles, windows.ContainerdConfigDir)
 }
 
 // createFilesFromIgnition returns the contents and write locations on the instance for any file it can create from
--- a/pkg/registries/registries.go
+++ b/pkg/registries/registries.go
@@ -140,19 +140,29 @@ func mergeMirrors(existingMirrors, newMirrors []mirror) []mirror {
 }
 
 // generateConfig is a serialization method that generates a valid TOML representation from a mirrorSet object.
-// Results in content usable as a containerd image registry configuration file.
+// Results in content usable as a containerd image registry configuration file. Returns empty string if no mirrors exist
 func (ms *mirrorSet) generateConfig() string {
-	result := ""
-
-	if ms.mirrorSourcePolicy == config.AllowContactingSource {
-		result += fmt.Sprintf("server = \"https://%s\"", ms.source)
-		result += "\r\n\r\n"
+	if len(ms.mirrors) == 0 {
+		return ""
 	}
 
+	result := ""
+
+	fallbackServer := ms.source
+	if ms.mirrorSourcePolicy == config.NeverContactSource {
+		// set the fallback server to the first mirror to ensure the source is never contacted, even if all mirrors fail
+		fallbackServer = ms.mirrors[0].host
+	}
+	result += fmt.Sprintf("server = \"https://%s\"", fallbackServer)
+	result += "\r\n\r\n"
+
+	// Each mirror should result in an entry followed by a set of settings for interacting with the mirror host
 	for _, m := range ms.mirrors {
 		result += fmt.Sprintf("[host.\"https://%s\"]", m.host)
 		result += "\r\n"
 
+		// Specify the operations the registry host may perform. IDMS mirrors can only be pulled by directly by digest,
+		// whereas ITMS mirrors have the additional resolve capability, which allows converting a tag name into a digest
 		var hostCapabilities string
 		if m.resolveTags {
 			hostCapabilities = "  capabilities = [\"pull\", \"resolve\"]"
--- a/pkg/registries/registries.go
+++ b/pkg/registries/registries.go
@@ -3,15 +3,54 @@ package registries
 import (
 	"fmt"
 	"sort"
+	"strings"
 
 	config "github.com/openshift/api/config/v1"
 )
 
+// imagePathSeparator separates the repo name, namespaces, and image name in an OCI-compliant image name
+const imagePathSeparator = "/"
+
 // mirror represents a mirrored image repo entry in a registry configuration file
 type mirror struct {
+	// host is the mirror image location. Can include the registry hostname/IP address, port, and namespace path
 	host string
 	// resolveTags indicates to the container runtime if this mirror is allowed to resolve an image tag into a digest
 	resolveTags bool
+}
+
+// newMirror constructs a new mirror object with proper host name structure to be used in containerd registry config
+func newMirror(sourceImageLocation, mirrorImageLocation string, resolveTags bool) mirror {
+	mirrorHost := ""
+	// containerd appends any shared namespaces between source and mirror locations to the mirror's host entry in the
+	// registry config file to construct the full mirror image location
+	if sourceImageLocation != mirrorImageLocation {
+		// truncate the mirror to drop any shared namespaces since containerd automatically appends them on image pull
+		mirrorHost = extractMirrorURL(sourceImageLocation, mirrorImageLocation)
+	} else {
+		// special case if source and mirror are the same. Do not drop the host repo name to avoid an empty host entry
+		mirrorHost = extractHostname(mirrorImageLocation)
+	}
+	return mirror{host: mirrorHost, resolveTags: resolveTags}
+}
+
+// extractMirrorURL drops the common suffix from the second repo, returning only the unique leading URL and namespaces
+func extractMirrorURL(source, mirror string) string {
+	sourceParts := strings.Split(source, imagePathSeparator)
+	mirrorParts := strings.Split(mirror, imagePathSeparator)
+	uniqueMirrorParts := mirrorParts
+
+	// Process until the end of either repo string
+	for i := 0; i < len(sourceParts) && i < len(mirrorParts); i++ {
+		// Check if suffix piece is equal, starting from the backs of the lists
+		if sourceParts[len(sourceParts)-1-i] != mirrorParts[len(mirrorParts)-1-i] {
+			// break when something different is found to retain all pieces after the last common element
+			break
+		}
+		// Remove common suffix piece
+		uniqueMirrorParts = uniqueMirrorParts[:len(uniqueMirrorParts)-1]
+	}
+	return strings.Join(uniqueMirrorParts, imagePathSeparator)
 }
 
 // mirrorSet holds the mirror registry information for a single source image repo
@@ -22,6 +61,23 @@ type mirrorSet struct {
 	mirrors []mirror
 	// mirrorSourcePolicy defines the fallback policy if fails to pull image from the mirrors
 	mirrorSourcePolicy config.MirrorSourcePolicy
+}
+
+// newMirrorSet constructs an object with proper source and mirror name structures to be used in containerd registry config
+func newMirrorSet(srcImage string, mirrorLocations []config.ImageMirror, resolveTags bool,
+	mirrorSourcePolicy config.MirrorSourcePolicy) mirrorSet {
+	truncatedMirrors := []mirror{}
+	for _, m := range mirrorLocations {
+		truncatedMirrors = append(truncatedMirrors, newMirror(srcImage, string(m), resolveTags))
+	}
+	return mirrorSet{source: extractHostname(srcImage), mirrors: truncatedMirrors, mirrorSourcePolicy: mirrorSourcePolicy}
+}
+
+// extractHostname extracts just the initial host repo from a full image location
+// e.g. mcr.microsoft.com would be extracted from mcr.microsoft.com/oss/kubernetes/pause:3.9
+func extractHostname(fullImage string) string {
+	parts := strings.Split(fullImage, imagePathSeparator)
+	return parts[0]
 }
 
 // registryConfig represents a system-wide image registry configuration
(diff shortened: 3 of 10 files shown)


saifshaikh48:
/test vsphere-disconnected-e2e-operator

for early feedback

build/Dockerfile
```diff
@@ -169,6 +170,10 @@ COPY --from=build /build/windows-machine-config-operator/csi-proxy/bin/csi-proxy
 RUN mkdir /payload/generated
 RUN chmod 0777 /payload/generated
 
+# Create directory for generated registry config files with open permissions
+RUN mkdir /payload/generated/cri_registry_config
+RUN chmod 0777 /payload/generated/cri_registry_config
```
saifshaikh48, line 175: Will need to update midstream Dockerfile

saifshaikh48:
/test aws-e2e-operator
/test vsphere-e2e-operator

azure, gcp, nutanix CI all passed. aws passed locally so seeing what's up in CI
vsphere-disconnected hitting issues since golden images are currently missing from devqe env. other vsphere jobs should not hit this though

pkg/registries/registries.go
```diff
@@ -140,14 +140,20 @@ func mergeMirrors(existingMirrors, newMirrors []mirror) []mirror {
...
-	if ms.mirrorSourcePolicy == config.AllowContactingSource {
-		result += fmt.Sprintf("server = \"https://%s\"", ms.source)
-		result += "\r\n\r\n"
+	fallbackServer := ms.source
+	if ms.mirrorSourcePolicy == config.NeverContactSource {
+		// set the fallback server to the first mirror to ensure the source is never contacted, even if all mirrors fail
+		fallbackServer = ms.mirrors[0].host
 	}
+	result += fmt.Sprintf("server = \"https://%s\"", fallbackServer)
+	result += "\r\n\r\n"
```
sebsoto, line 167: This is getting a bit confusing, please add some documentation to this clarifying the expected input/output.

pkg/registries/registries.go
```diff
@@ -24,6 +29,22 @@ type mirrorSet struct {
...
+	truncatedMirrors := []mirror{}
+	for _, m := range mirrorLocations {
+		truncatedMirrors = append(truncatedMirrors, newMirror(string(m), resolveTags))
+	}
+	return mirrorSet{source: extractHost(srcImage), mirrors: truncatedMirrors, mirrorSourcePolicy: mirrorSourcePolicy}
+}
+
+// extractHost extracts just the initial host repo from a full image location
+// e.g. mcr.microsoft.com would be extracted from mcr.microsoft.com/oss/kubernetes/pause:3.9
+func extractHost(fullImage string) string {
```
sebsoto, line 45: Host -> Hostname

build/Dockerfile
```diff
@@ -112,6 +112,7 @@ RUN make build-daemon
 #│   ├── csi-proxy.exe
 #├── ecr-credential-provider.exe
 #├── generated/
+#│   └── cri_registry_config/
```
sebsoto, line 115: Why does this need to be created?

test/e2e/mirror_test.go
```diff
@@ -0,0 +1,160 @@
...
+func testImageMirroring(t *testing.T) {
+	tc, err := NewTestContext()
+	require.NoError(t, err)
+
+	require.NoError(t, tc.loadExistingNodes(), "error getting the current Windows nodes in the cluster")
+
+	if !t.Run("Mirror settings applied to nodes", tc.testMirrorSettingsApplied) {
+		// No point in running the rest of the tests if settings aren't applied
+		return
+	}
```
sebsoto, line 39: This would be tested by the workload running sucessfully.

test/e2e/mirror_test.go
```diff
@@ -0,0 +1,160 @@
...
+	t.Run("Workload using a mirrored container image", tc.testWorkloadWithMirroredImage)
```
sebsoto, line 40: Please do not make this a separate test.
Rather setup the mirrors as part of the test setup, and we should see the tests on a disconnected job pass as normal.

You can make the mirror setup configurable via a CMD line option,

test/e2e/mirror_test.go
```diff
@@ -0,0 +1,160 @@
...
+	t.Run("Mirror settings cleared from nodes", tc.testMirrorSettingsCleared)
```
sebsoto, line 41: Please do this as a step in the deconfigure suite

controllers/registry_controller.go
```diff
@@ -68,19 +75,36 @@ func NewRegistryReconciler(mgr manager.Manager, clusterConfig cluster.Config,
...
+	nodes := &core.NodeList{}
+	if err := r.client.List(ctx, nodes, client.MatchingLabels{core.LabelOSStable: "windows"}); err != nil {
+		return ctrl.Result{}, err
+	}
+	for _, node := range nodes.Items {
+		r.signer, err = signer.Create(types.NamespacedName{Namespace: r.watchNamespace, Name: secrets.PrivateKeySecret},
+			r.client)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("unable to create signer from private key secret: %w", err)
+		}
```
sebsoto, line 93: This should only be done once, move it out of the loop

pkg/registries/registries.go
```diff
@@ -180,3 +182,27 @@ func (ms *mirrorSet) generateConfig() string {
...
+	imageDigestMirrorSetList := &config.ImageDigestMirrorSetList{}
+	if err := c.List(ctx, imageDigestMirrorSetList); err != nil {
+		return nil, fmt.Errorf("error getting IDMS list: %w", err)
+	}
+	imageTagMirrorSetList := &config.ImageTagMirrorSetList{}
+	if err := c.List(ctx, imageTagMirrorSetList); err != nil {
+		return nil, fmt.Errorf("error getting ITMS list: %w", err)
+	}
+
+	registryConf := newRegistryConfig(imageDigestMirrorSetList.Items, imageTagMirrorSetList.Items)
```
sebsoto, line 198: Please reevaluate this object and how it is being used.

As it is now, this function would be better as `getMirrorsets`

pkg/windows/windows.go
```diff
@@ -361,11 +369,16 @@ func (vm *windows) EnsureFile(file *payload.FileInfo, remoteDir string) error {
 			vm.log.Error(err, "error closing local file", "file", file.Path)
 		}
 	}()
-	vm.log.V(1).Info("copy", "local file", file.Path, "remote dir", remoteDir)
-	if err := vm.interact.transfer(f, filepath.Base(file.Path), remoteDir); err != nil {
+	vm.log.V(1).Info("copying", "local file", file.Path, "remote dir", remoteDir)
```
sebsoto, line 372: please revert

pkg/windows/windows.go
```diff
@@ -334,11 +337,16 @@ func (vm *windows) EnsureFileContent(contents []byte, filename string, remoteDir
 		// The file already exists with the expected content, do nothing
 		return nil
 	}
-	vm.log.V(1).Info("copy", "file content", filename, "remote dir", remoteDir)
-	if err := vm.interact.transfer(bytes.NewReader(contents), filename, remoteDir); err != nil {
+	vm.log.V(1).Info("copying", "file content", filename, "remote dir", remoteDir)
```
sebsoto, line 340: please revert

pkg/windows/windows.go
```diff
@@ -334,11 +337,16 @@ func (vm *windows) EnsureFileContent(contents []byte, filename string, remoteDir
...
+
+	c, err := vm.interact.createSFTPClient()
+	if err != nil {
+		return err
```
sebsoto, line 344: please wrap error for easier debugging

pkg/nodeconfig/nodeconfig.go
```diff
@@ -152,6 +153,9 @@ func (nc *nodeConfig) Configure() error {
 	if err := nc.createBootstrapFiles(); err != nil {
 		return err
 	}
+	if err := nc.createRegistryConfigFiles(); err != nil {
```
mansikulkarni96, line 156: we are creating the Containerd directory inside nc.Windows.Boostrap(), why don't we call this function after it is created? If needed, we can also split the bootstrap function into two. Let me know if there is another reason for doing this here.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -370,6 +374,15 @@ func (nc *nodeConfig) write(pathToData map[string]string) error {
 	return nil
 }
 
+// createBootstrapFiles creates all files on the node required for containerd to mirror images
```
mansikulkarni96, line 377: doc needs update

pkg/windows/connectivity.go
```diff
@@ -157,7 +165,27 @@ func (c *sshConnectivity) transfer(reader io.Reader, filename, remoteDir string)
...
+func (c *sshConnectivity) transferFiles(sftpClient *sftp.Client, files map[string][]byte, remoteDir string) error {
+	for workingPath, content := range files {
+		// Load file content into io.Reader
+		reader := bytes.NewReader(content)
+
+		// Construct the full destination path the file will be copied to
+		dstPath := remoteDir + "\\" + workingPath
+		// Get the directory that will be written into, without trailing slash, and the base file name
+		splitIndex := strings.LastIndexByte(dstPath, '\\')
+		writeDir, filename := dstPath[:splitIndex], dstPath[splitIndex+1:]
```
sebsoto, line 183: Please separate this into two lines for readability

pkg/windows/connectivity.go
```diff
@@ -157,7 +165,27 @@ func (c *sshConnectivity) transfer(reader io.Reader, filename, remoteDir string)
...
-		c.log.Error(err, "error closing remote file", "file", remoteFile)
+		//return fmt.Errorf("error closing remote file %s: %w", remoteFile, err)
+		c.log.Error(err, "error closing remote file", "file")
+	}
+	return nil
+}
+
+func (c *sshConnectivity) transferFiles(sftpClient *sftp.Client, files map[string][]byte, remoteDir string) error {
+	for workingPath, content := range files {
+		// Load file content into io.Reader
```
sebsoto, line 176: unecessary comment

sebsoto (changes requested):
Nice work. Please look at my comments.

pkg/windows/windows.go
```diff
@@ -391,6 +404,26 @@ func (vm *windows) FileExists(path, checksum string) (bool, error) {
 	return false, nil
 }
 
+func (vm *windows) ReplaceDir(files map[string][]byte, remoteDir string) error {
```
mansikulkarni96, line 407: needs doc

mansikulkarni96 (changes requested):

build/Dockerfile
saifshaikh48: ah it doesn't, vestige of the previous solution. I will remove it

pkg/nodeconfig/nodeconfig.go
saifshaikh48: My thoughts here were that createBootstrapFiles() sets up files needed for kubelet and createRegistryConfigFiles() basically does the same for containerd (and containerd is a pre-req service to kubelet) so the funcs should be called together

> we are creating the Containerd directory inside nc.Windows.Boostrap()

This is a good point though, I will explore your suggestion

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
```diff
@@ -23,7 +23,7 @@ metadata:
     operatorframework.io/suggested-namespace: openshift-windows-machine-config-operator
     operators.openshift.io/valid-subscription: '["Red Hat OpenShift support for Windows
       Containers"]'
-    operators.operatorframework.io/builder: operator-sdk-v1.32.0+git
+    operators.operatorframework.io/builder: operator-sdk-v1.32.0
```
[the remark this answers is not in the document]
saifshaikh48: Not required but is there any reason for the `+git`?

test/e2e/mirror_test.go
saifshaikh48: Not quite sure what you mean here

- Should we only test mirroring through the disconnected job?
- Do you forsee the mirror_test suite to fail in the disconnected job?

pkg/nodeconfig/nodeconfig.go
saifshaikh48: I thought about this more and decided to keep it where it is (though I can still be swayed)
Additional reasoning was 
- this way the registry config is created before containerd service actually starts so it doesn't have to do any additional config loading during runtime
- createRegistryConfigFiles() creates the required directories if needed so no issue

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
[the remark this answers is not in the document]
wgahnagl: Those are generated when running make bundle, and need to be reverted every time. Make bundle removes the +git.

pkg/registries/registries.go
saifshaikh48: added some comments. LMK if the unit tests are sufficient for expected in/output

test/e2e/mirror_test.go
saifshaikh48: Hmm, if we moved this we would no longer be testing the registry-controller's reactivity to ITMS changes

test/e2e/mirror_test.go
saifshaikh48: Should I remove this test altogether? IMO this case adds some value to validate there is the expected number of config files. I can add another entry to the test ITMS object to make it more clear if you prefer

sebsoto: > IMO this case adds some value to validate there is the expected number of config files

The expected config files will be tested through the workloads successfully running.
If you are saying you are trying to test that there are not more files than expected, that would be better tested within unit tests.

test/e2e/mirror_test.go
sebsoto: You would still be testing that if you remove the ITMS created for the tests to run properly and then wait for the registry directory to be cleared

test/e2e/mirror_test.go
sebsoto: > Should we only test mirroring through the disconnected job?

I don't see it necessary to test it elsewhere if the disconnected job is required.

> Do you forsee the mirror_test suite to fail in the disconnected job?

I'm not sure what you mean by this.

saifshaikh48:
/test vsphere-disconnected-e2e-operator

updated mirror_test suite to work with disconnected

saifshaikh48:
/test vsphere-disconnected-e2e-operator

realized mirror_test suite that’s being introduced is too naive to work with disconnected job as it’s written, skipping the suite just to see the other  suites run

saifshaikh48:
/test vsphere-disconnected-e2e-operator

I looked at the [ipi-conf-mirror-commands step we use](https://github.com/openshift/release/pull/50547/files#diff-4cf9971406e74fc92602aebc6533d32b1d94b9c4895c084cbc2d812933a10df3R10) and the user provided [CA bundle](https://github.com/openshift/release/blob/master/ci-operator/step-registry/ipi/conf/mirror/ipi-conf-mirror-commands.sh#L18-L43) also contains the mirror registry pull secret. So I thought we should point the mirrors to this ca-bundle (which we transfer onto Windows nodes) to support TLS/https endpoints

saifshaikh48:
/test vsphere-disconnected-e2e-operator

I looked at the [ipi-conf-mirror-commands step we use](https://github.com/openshift/release/pull/50547/files#diff-4cf9971406e74fc92602aebc6533d32b1d94b9c4895c084cbc2d812933a10df3R10) and the user provided [CA bundle](https://github.com/openshift/release/blob/master/ci-operator/step-registry/ipi/conf/mirror/ipi-conf-mirror-commands.sh#L18-L43) also contains the mirror registry pull secret. So I thought we should point the mirrors to this ca-bundle (which we transfer onto Windows nodes) to support TLS/https endpoints

saifshaikh48:
/test vsphere-disconnected-e2e-operator

I looked at the [ipi-conf-mirror-commands step we use](https://github.com/openshift/release/pull/50547/files#diff-4cf9971406e74fc92602aebc6533d32b1d94b9c4895c084cbc2d812933a10df3R10) and the user provided [CA bundle](https://github.com/openshift/release/blob/master/ci-operator/step-registry/ipi/conf/mirror/ipi-conf-mirror-commands.sh#L18-L43) also contains the mirror registry pull secret. So I thought we should point the mirrors to this ca-bundle (which we transfer onto Windows nodes) to support TLS/https endpoints

saifshaikh48:
/test vsphere-disconnected-e2e-operator

install issue during one of the qe steps

saifshaikh48:
/test vsphere-disconnected-e2e-operator

to see if QE step that was broken on branching day is still affecting cluster install

pkg/registries/registries.go
```diff
@@ -3,17 +3,48 @@ package registries
...
+	uniqueMirrorParts := mirrorParts
+
+	// Process until we hit the end of either repo string
+	for i := 0; i < len(sourceParts) && i < len(mirrorParts); i++ {
+		// Check if suffix piece is equal, starting from the backs of the lists
+		if sourceParts[len(sourceParts)-1-i] == mirrorParts[len(mirrorParts)-1-i] {
+			// Remove common suffix piece
+			uniqueMirrorParts = uniqueMirrorParts[:len(uniqueMirrorParts)-1]
+		}
+	}
```
sebsoto, line 49: Does this work as expected in the case where there are two strings
`foo1/fah/different/foo2`
`foo1/fah/something/foo2`
Tried this and its giving me `foo1`, when it really should give `foo1/fah` ?

I believe you should break when you hit something that is different.
I suggest doing it this way instead:
```
	// Process until we hit the end of either repo string
	var index int
	for index = 0; index < len(sourceParts) && index < len(mirrorParts); index++ {
		// Check if suffix piece is equal, starting from the backs of the lists
		if sourceParts[len(sourceParts)-1-index] != mirrorParts[len(mirrorParts)-1-index] {
			break
		}
	}

	fmt.Println(strings.Join(mirrorParts[:len(mirrorParts)-1-index], "/"))
```

But does this required that sourceParts and mirrorParts are of equal length, or that one specific one is longer than the other?

sebsoto (changes requested):
Partial review, will review rest of PR shortly

pkg/registries/registries.go
```diff
@@ -3,17 +3,48 @@ package registries
...
 	"fmt"
 	"sort"
+	"strings"
 
 	config "github.com/openshift/api/config/v1"
 )
 
 // mirror represents a mirrored image repo entry in a registry configuration file
 type mirror struct {
+	// host is the mirror image target location. Can include the registry hostname/IP address, port, and namespace path
```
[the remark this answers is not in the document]
saifshaikh48: target refers to the fact that containerd will target this location rather than a source location to pull the image. this is info beyond the scope of the struct though so I'll remove this wording

pkg/registries/registries.go
```diff
@@ -3,17 +3,48 @@ package registries
...
 	host string
 	// resolveTags indicates to the container runtime if this mirror is allowed to resolve an image tag into a digest
 	resolveTags bool
 }
 
+func newMirror(sourceImageLocation, mirrorImageLocation string, resolveTags bool) mirror {
+	mirrorHost := ""
+	if sourceImageLocation == mirrorImageLocation {
+		// special case if source and mirror are the same
```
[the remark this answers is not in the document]
saifshaikh48: Just for corner-case resiliency. If a user for some reason put the same source and mirror, while stupid, we don't want things to break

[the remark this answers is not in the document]
saifshaikh48: added some more comments around this

pkg/registries/registries.go
saifshaikh48: > when it really should give foo1/fah

it should give `foo1/fah/something` as that is the mirror location minus any common suffixes. Good catch though, you're right it doesn't work as intended in that case!

saifshaikh48: I did a combination of the two approaches to avoid the equal length concern and added some unit tests to prove it works now

sebsoto (changes requested):

controllers/registry_controller.go
```diff
@@ -68,19 +75,36 @@ func NewRegistryReconciler(mgr manager.Manager, clusterConfig cluster.Config,
...
+		nc, err := nodeconfig.NewNodeConfig(r.client, r.k8sclientset, r.clusterServiceCIDR, r.watchNamespace,
+			winInstance, r.signer, nil, nil, r.platform)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to create new nodeconfig: %w", err)
+		}
+
+		if err := nc.Windows.ReplaceDir(configFiles, windows.ContainerdConfigDir); err != nil {
+			return ctrl.Result{}, err
+		}
```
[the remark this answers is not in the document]
saifshaikh48: Is this something we can address as follow-up work? I am using an established pattern -- there are multiple places in the code base where we follow this pattern of looping over nodes for file transfer (configmap_controller, controllerconfig_controller etc), and all face the same issue of re-looping over all nodes if one flakes. There seems a greater effort is needed to tackle a file transfer overhaul.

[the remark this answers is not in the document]
sebsoto: Thats a fair point.
I am okay with that as long as no other reviewers think differently.

pkg/windows/windows.go
```diff
@@ -391,6 +404,26 @@ func (vm *windows) FileExists(path, checksum string) (bool, error) {
...
+	vm.log.V(1).Info("publishing", "remote destination dir", remoteDir, "number of files", len(files))
```
sebsoto, line 408: nit: s/publishing/overwriting

test/e2e/mirror_test.go
```diff
@@ -0,0 +1,152 @@
...
+	for _, node := range gc.allNodes() {
+		t.Run(node.Name, func(t *testing.T) {
+			addr, err := controllers.GetAddress(node.Status.Addresses)
+			require.NoError(t, err, "unable to get node address")
+
+			// retry to give time for registry controller to transfer generated config to all nodes
+			err = wait.PollUntilContextTimeout(context.TODO(), 15*time.Second, 5*time.Minute, true,
+				func(ctx context.Context) (done bool, err error) {
+					count, err := tc.countItemsInDir(windows.ContainerdConfigDir, addr)
+					require.NoErrorf(t, err, "error counting items in dir %s on node %s", windows.ContainerdConfigDir, addr)
```
sebsoto, line 79: As you are already going node by node, I think it would be valuable to add a check that the `test.registry.io/hosts.toml` file has the expected contents on the VM.

Could just be a file read and assert `strings.Contains(out, expectedContents)`

sebsoto (changes requested):
Good work @saifshaikh48.
Just a couple comments, otherwise LGTM

test/e2e/mirror_test.go
saifshaikh48: This to me is something covered by the `TestGenerateConfig` unit tests. 

The real e2e test validating something like this would be making containerd to consume the host file i.e. [deploying a mirrored workload](https://github.com/openshift/windows-machine-config-operator/compare/8e74444c13342a529cd7da9f97ce3165ba484399..ca7d85ebb17653a7d8f751aa58662de3b59f09bc#diff-099e3332f0e1e5fc59c8cf99775065fdc4b594e7b949b39e3873de36269e0722L40), but we decided this was redundant

pkg/registries/registries.go
jrvaldes: I recommend moving this logic into a separate helper function so that it can be unit-tested.

controllers/registry_controller.go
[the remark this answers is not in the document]
jrvaldes: @saifshaikh48, please create a follow-up card in Jira and drop a ref here in the code.

pkg/registries/registries.go
```diff
@@ -3,17 +3,53 @@ package registries
...
+		mirrorHost = dropCommonSuffix(sourceImageLocation, mirrorImageLocation)
+	} else {
+		// special case if source and mirror are the same. Do not drop the host repo name to avoid an empty host entry
+		mirrorHost = extractHostname(mirrorImageLocation)
+	}
+	return mirror{host: mirrorHost, resolveTags: resolveTags}
+}
+
+// dropCommonSuffix drops the common suffix from the second repo, returning only the unique leading URL and namespaces
+func dropCommonSuffix(source, mirror string) string {
```
jrvaldes, line 37: Consider naming this as `extractMirrorURL`

dropping the common suffix is just the "how" it achieved the goal: returning only the unique leading URL and namespaces of the mirror

pkg/registries/registries.go
```diff
@@ -3,17 +3,53 @@ package registries
...
+	for i := 0; i < len(sourceParts) && i < len(mirrorParts); i++ {
+		// Check if suffix piece is equal, starting from the backs of the lists
+		if sourceParts[len(sourceParts)-1-i] != mirrorParts[len(mirrorParts)-1-i] {
+			// break when something different is found to retain all pieces after the last common element
+			break
+		}
+		// Remove common suffix piece
+		uniqueMirrorParts = uniqueMirrorParts[:len(uniqueMirrorParts)-1]
+	}
+	return strings.Join(uniqueMirrorParts, "/")
```
jrvaldes, line 52: consider introducing a constant for `/`  and re-use it everywhere.

`MirrorPathSeparator = "/"`

pkg/registries/registries.go
```diff
@@ -24,6 +60,23 @@ type mirrorSet struct {
 	mirrorSourcePolicy config.MirrorSourcePolicy
 }
 
+// newMirror constructs an object with proper source and mirror name structures to be used in containerd registry config
+func newMirrorSet(srcImage string, mirrorLocations []config.ImageMirror, resolveTags bool,
+	mirrorSourcePolicy config.MirrorSourcePolicy) mirrorSet {
```
jrvaldes, line 67: consider passing down the whole entry (idms.Spec.ImageDigestMirrors) and let the function use what it needs from it, it will reduce the number of arguments and make it scale as needed.

pkg/registries/registries.go
```diff
@@ -24,6 +60,23 @@ type mirrorSet struct {
...
+	for _, m := range mirrorLocations {
+		truncatedMirrors = append(truncatedMirrors, newMirror(srcImage, string(m), resolveTags))
+	}
+	return mirrorSet{source: extractHostname(srcImage), mirrors: truncatedMirrors, mirrorSourcePolicy: mirrorSourcePolicy}
+}
+
+// extractHostname extracts just the initial host repo from a full image location
+// e.g. mcr.microsoft.com would be extracted from mcr.microsoft.com/oss/kubernetes/pause:3.9
+func extractHostname(fullImage string) string {
+	parts := strings.Split(fullImage, "/")
```
jrvaldes, line 78: should this account for the protocol in the full image URL? for example: `https://mcr.microsoft.com/oss/kubernetes/pause:3.9`

controllers/registry_controller.go
```diff
@@ -21,13 +21,20 @@ import (
 	"fmt"
```
jrvaldes, line 21: > Edits the way file transfers work for easier reuse of FTP clients.

this may warrant it's own commit.

jrvaldes (commented):
@saifshaikh48  good work on putting this together. Left minor suggestions. PTAL

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
```diff
@@ -11,7 +11,7 @@ metadata:
     features.operators.openshift.io/cnf: "false"
     features.operators.openshift.io/cni: "true"
     features.operators.openshift.io/csi: "false"
-    features.operators.openshift.io/disconnected: "false"
+    features.operators.openshift.io/disconnected: "true"
```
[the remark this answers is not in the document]
saifshaikh48: AFAIK CVP only checks for the presence of the annotation. Don't know of any further validation.

But now that you mention it, this commit should be dropped and introduced in one of the PRs that we've split off from this work as we don't fully support disconnected until those merge

pkg/registries/registries.go
```diff
@@ -3,17 +3,53 @@ package registries
...
 
 	config "github.com/openshift/api/config/v1"
 )
 
 // mirror represents a mirrored image repo entry in a registry configuration file
 type mirror struct {
+	// host is the mirror image location. Can include the registry hostname/IP address, port, and namespace path
 	host string
 	// resolveTags indicates to the container runtime if this mirror is allowed to resolve an image tag into a digest
 	resolveTags bool
```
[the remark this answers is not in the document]
saifshaikh48: I lean towards keeping the var name as is

pkg/registries/registries.go
saifshaikh48: Nope, having the web protocol in front of the repo is not accepted by the IDMS/ITMS resource so we'd never have to process such a URL

pkg/registries/registries.go
saifshaikh48: This wouldn't work as the function is used for ImageTagMirrors as well (and it's not worth it to introduce a generic type here)

controllers/registry_controller.go
saifshaikh48: the changes are tightly coupled, not sure I can split this

pkg/registries/registries.go
```diff
@@ -140,19 +140,28 @@ func mergeMirrors(existingMirrors, newMirrors []mirror) []mirror {
 }
 
 // generateConfig is a serialization method that generates a valid TOML representation from a mirrorSet object.
-// Results in content usable as a containerd image registry configuration file.
+// Results in content usable as a containerd image registry configuration file. Returns empty string if no mirrors exist
 func (ms *mirrorSet) generateConfig() string {
 	result := ""
+	if len(ms.mirrors) == 0 {
```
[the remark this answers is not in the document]
saifshaikh48: added

controllers/registry_controller.go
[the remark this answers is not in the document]
saifshaikh48: added as a TODO above

pkg/registries/registries.go
```diff
@@ -24,47 +65,52 @@ type mirrorSet struct {
 	mirrorSourcePolicy config.MirrorSourcePolicy
 }
 
-// registryConfig represents a system-wide image registry configuration
-type registryConfig struct {
-	sourceConfigs []mirrorSet
+// newMirror constructs an object with proper source and mirror name structures to be used in containerd registry config
```
mansikulkarni96, line 68: newMirrorSet*

pkg/registries/registries.go
```diff
@@ -24,47 +65,52 @@ type mirrorSet struct {
...
+// extractHostname extracts just the initial host repo from a full image location
+// e.g. mcr.microsoft.com would be extracted from mcr.microsoft.com/oss/kubernetes/pause:3.9
+func extractHostname(fullImage string) string {
+	parts := strings.Split(fullImage, imagePathSeparator)
+	return parts[0]
 }
 
-// NewRegistryConfig creates a new RegistryConfig object by extracting and merging the contents of the given mirror sets
-func NewRegistryConfig(idmsItems []config.ImageDigestMirrorSet, idtsItems []config.ImageTagMirrorSet) *registryConfig {
+// getMirrorSets extracts and merges the contents of the given mirror sets.
```
mansikulkarni96, line 85: This function is getting and merging the mirror sets consider either renaming it to denote that or doing the merge where this is being called.

saifshaikh48: sure, updated the name

mansikulkarni96:
/lgtm
will let @sebsoto approve for the changes requested.

saifshaikh48:
/retest-required

I went through each of the changes and there were no significant changes compared to the last time CI passed fully on all platforms, aside from the kube 1.30 rebase changes. Went through must gather's and cluster state looked as expected for image mirroring. Retesting to see if things are just flaky as things passed on platform none and failed at different spots across other jobs

jrvaldes (approved):
/lgtm

pkg/windows/windows.go
```diff
@@ -391,6 +404,26 @@ func (vm *windows) FileExists(path, checksum string) (bool, error) {
...
+	}
+
+	sftpClient, err := vm.interact.createSFTPClient()
+	if err != nil {
+		return fmt.Errorf("Failed to create SFTP client: %w", err)
+	}
+	if err = vm.interact.transferFiles(sftpClient, files, remoteDir); err != nil {
+		return err
+	}
+	return sftpClient.Close()
```
mansikulkarni96, line 424: shouldn't the client be closed even if we return error in between? This also applies to  a few other places in this commit.

mansikulkarni96 (changes requested):

jrvaldes:
/lgtm

@saifshaikh48   please fetch and rebase the latest from master before triggering the rest of the tests.

jrvaldes:
> > /lgtm
> > 
> > @saifshaikh48 please fetch and rebase the latest from master before triggering the rest of the tests.
> 
> @jrvaldes Why is rebasing against master necessary here?  Tests are always run against the latest master with the current PR applied?



No need, just a nice to have.

Lately, I'm noticing the `remaining-required` re-testing twice if there is a diff from the head of master.

jrvaldes:
/test aws-e2e-operator

ipi-install failed

PR #2167, opened 2024-05-10
OCPBUGS-22237: Use kubelet-ca.crt from ControllerConfig when certs rotate

sebsoto:
This commit changes how WMCO determines the expected state of the
kubelet-ca.crt file on Windows nodes.

The previous functionality had WMCO using the kubelet-ca.crt file
present in the worker MachineConfig as the initial state. When WMCO
reconciles the ConfigMap containing the
kube-apiserver-to-kubelet-client-ca, WMCO would then replace the
kubelet-ca.crt file with a combination of the clusters initial
kube-apiserver-to-kubelet-client-ca and the latest version.

With this commit, when the CA is changed in the cluster, WMCO will
instead look at the machine-config-controller ControllerConfig for the
new version of the kubelet-ca.crt. This way WMCO is consistent in
how it is sourcing the expected state with Linux Nodes, as the
Machine Config Daemon does the same. This stops WMCO from repeating work
that is already being done, and also ensures that we are including any
other certs that should be present within kubelet-ca.crt.


controllers/controllerconfig_controller.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/openshift/windows-machine-config-operator/pkg/cluster"
+	"github.com/openshift/windows-machine-config-operator/pkg/secrets"
+	"github.com/openshift/windows-machine-config-operator/pkg/signer"
+)
+
+//+kubebuilder:rbac:groups="machineconfiguration.openshift.io",resources=controllerconfigs,verbs=list;watch
+
+const (
+	// ControllerConfigController is the name of this controller in logs and other outputs.
+	ControllerConfigController = "controllerconfig"
```
saifshaikh48, line 44: nit: not sure if there is a way around this but I can see "controllerconfig controller" leading to some confusion. Is there any better naming scheme? All my ideas are falling flat

controllers/controllers.go
```diff
@@ -140,39 +139,6 @@ func (r *instanceReconciler) updateKubeletCA(node core.Node, contents []byte) er
...
-
-	// fetch all Windows nodes (Machine and BYOH instances)
-	winNodes := &core.NodeList{}
-	if err = r.client.List(ctx, winNodes, client.MatchingLabels{core.LabelOSStable: "windows"}); err != nil {
-		return fmt.Errorf("error listing Windows nodes: %w", err)
-	}
-	r.log.V(1).Info("processing", "node count", len(winNodes.Items))
-	// loop Windows nodes and trigger kubelet CA update
-	for _, winNode := range winNodes.Items {
-		if err := r.updateKubeletCA(winNode, kubeAPIServerServingCABytes); err != nil {
```
mansikulkarni96, line 169: shouldn't we also delete updateKubeletCA function definition?

mansikulkarni96: I see it being used in the new controller

mansikulkarni96 (approved):

saifshaikh48 (commented):
/lgtm

sebsoto:
/cherry-pick release-4.15,release-4.14,release-4.13,release-4.12

PR #2184, opened 2024-05-23
WINC-1224: OCP 4.17 branching day changes

sebsoto:


jrvaldes (approved):
/lgtm

build/Dockerfile
saifshaikh48, line 1: Do we also have to make this change in [ci-operator.yaml](https://github.com/openshift/windows-machine-config-operator/blob/master/.ci-operator.yaml#L5)?

.gitmodules
```diff
@@ -15,27 +15,28 @@
 [submodule "kubelet"]
 	path = kubelet
 	url = https://github.com/openshift/kubernetes
-	branch = release-4.16
+	branch = release-4.17
 [submodule "kube-proxy"]
 	path = kube-proxy
 	url = https://github.com/openshift/kubernetes
 	branch = sdn-4.16-kubernetes-1.29.0
```
saifshaikh48, line 22: Could you remind me when this is bumped/when we engage SDN to create the branch?

config/manifests/bases/windows-machine-config-operator.clusterserviceversion.yaml
saifshaikh48, line 1: There are a few places where we reference 4.16 in the CSV description section that should be bumped as well. Curious if the pre-release script didn't work for this?

.gitmodules
sebsoto: It needs to be created after the kube rebase

build/Dockerfile
sebsoto: You're correct, fixed

config/manifests/bases/windows-machine-config-operator.clusterserviceversion.yaml
sebsoto: The script did not work

.gitmodules
sebsoto: The proposal was for the WINC team to maintain them :)

PR #2186, opened 2024-05-30
OCPBUGS-33319: use FQDN for hostname in vSphere 

jrvaldes:
This PR introduces a better processing of the hostname so that it returns the FQDN of the instance, including the domain if the windows instance is part of a Active Directory domain controller. The latter, fixes an inconsistency with the configuration workflow of the operator that was not taking into account the DNS suffix in vSphere resulting in the cloud provider implementation (vsphere-cloud-provider) to fail while looking up the VM in the datacenter by name. 

```
[node_controller.go:236] error syncing 'win-worker-1': failed to get provider ID for node win-worker-1 at cloudprovider: failed to get instance ID from cloud provider: No VM found, requeuing
```

Finally, it passed the FQDN of the VM to the kubelet configuration to override the hostname resulting in a consistent naming convention on the node object.

### DIFF
--- a/pkg/csr/csr.go
+++ b/pkg/csr/csr.go
@@ -325,11 +325,7 @@ func findHostName(instanceInfo *instance.Info, instanceSigner ssh.Signer) (strin
 		return "", fmt.Errorf("error instantiating Windows instance: %w", err)
 	}
 	// get the instance host name  by running hostname command on remote VM
-	hostName, err := win.Run("hostname", true)
-	if err != nil {
-		return "", fmt.Errorf("error getting the host name, with stdout %s: %w", hostName, err)
-	}
-	return hostName, nil
+	return win.GetHostname()
 }
 
 // matchesDNS returns true if the node name passed matches with the instance address of any of the instances present
--- a/pkg/windows/windows.go
+++ b/pkg/windows/windows.go
@@ -139,6 +139,15 @@ const (
 	wicdKubeconfigPath = K8sDir + "\\wicd-kubeconfig"
 	// TrustedCABundlePath is the location of the trusted CA bundle file
 	TrustedCABundlePath = remoteDir + "\\ca-bundle.crt"
+	// GetHostnameFQDNCommand is the PowerShell command to get the FQDN hostname of the Windows instance
+	GetHostnameFQDNCommand = "$output = Invoke-Expression 'ipconfig /all'; " +
+		"$hostNameLine = ($output -split '`n') | Where-Object { $_ -match 'Host Name' }; " +
+		"$dnsSuffixLine = ($output -split '`n') | Where-Object { $_ -match 'Primary Dns Suffix' }; " +
+		"$hostName = ($hostNameLine -split ':')[1].Trim(); " +
+		"$dnsSuffix = ($dnsSuffixLine -split ':')[1].Trim(); " +
+		"if (-not $dnsSuffix) { return $hostName }; " +
+		"$fqdn = $hostName + '.' + $dnsSuffix; " +
+		"return $fqdn"
 )
 
 var (
@@ -227,6 +236,8 @@ func GetK8sDir() string {
 type Windows interface {
 	// GetIPv4Address returns the IPv4 address of the associated instance.
 	GetIPv4Address() string
+	// GetHostname returns the FQDN of the associated instance including the domain name, if any
+	GetHostname() (string, error)
 	// EnsureFile ensures the given file exists within the specified directory on the Windows VM. The file will be copied
 	// to the Windows VM if it is not present or if it has the incorrect contents. The remote directory is created if it
 	// does not exist.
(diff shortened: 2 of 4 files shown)


pkg/csr/csr.go
```diff
@@ -325,11 +325,7 @@ func findHostName(instanceInfo *instance.Info, instanceSigner ssh.Signer) (strin
 		return "", fmt.Errorf("error instantiating Windows instance: %w", err)
 	}
 	// get the instance host name  by running hostname command on remote VM
-	hostName, err := win.Run("hostname", true)
-	if err != nil {
-		return "", fmt.Errorf("error getting the host name, with stdout %s: %w", hostName, err)
-	}
-	return hostName, nil
+	return win.GetHostname()
```
sebsoto, line 328: What does this return normally, when not part of AD?
Is it the same as before, just the hostname?

sebsoto: And have you been able to see what the csr approval behavior is like when using AD?

sebsoto (commented):
Thanks @jrvaldes, this is an interesting find.

pkg/csr/csr.go
```diff
@@ -325,11 +325,7 @@ func findHostName(instanceInfo *instance.Info, instanceSigner ssh.Signer) (strin
 		return "", fmt.Errorf("error instantiating Windows instance: %w", err)
```
sebsoto, line 325: Please add why this needs to be done to the commit message.
I believe you said it was so vsphere cloud controller can match the node to a VM.

pkg/windows/windows.go
```diff
@@ -139,6 +139,15 @@ const (
 	wicdKubeconfigPath = K8sDir + "\\wicd-kubeconfig"
 	// TrustedCABundlePath is the location of the trusted CA bundle file
 	TrustedCABundlePath = remoteDir + "\\ca-bundle.crt"
+	// GetHostnameFQDNCommand is the PowerShell command to get the FQDN hostname of the Windows instance
+	GetHostnameFQDNCommand = "$output = Invoke-Expression 'ipconfig /all'; " +
+		"$hostNameLine = ($output -split '`n') | Where-Object { $_ -match 'Host Name' }; " +
+		"$dnsSuffixLine = ($output -split '`n') | Where-Object { $_ -match 'Primary Dns Suffix' }; " +
+		"$hostName = ($hostNameLine -split ':')[1].Trim(); " +
+		"$dnsSuffix = ($dnsSuffixLine -split ':')[1].Trim(); " +
+		"if (-not $dnsSuffix) { return $hostName }; " +
```
saifshaikh48, line 148: same question, if there is no dnsSuffix then will the output of this be equivalent to `vm.Run("hostname", true)`

pkg/windows/windows.go
```diff
@@ -613,11 +632,11 @@ func (vm *windows) ensureHostNameAndContainersFeature() error {
 
 // isHostNameChangeNeeded tells if we need to update the host name of the Windows VM
 func (vm *windows) isHostNameChangeNeeded() (bool, error) {
-	out, err := vm.Run("hostname", true)
+	hostName, err := vm.GetHostname()
 	if err != nil {
-		return false, fmt.Errorf("error getting the host name, with stdout %s: %w", out, err)
+		return false, err
```
wgahnagl, line 637: do we need to get rid of this fmt?

pkg/csr/csr.go
jrvaldes: > What does this return normally, when not part of AD? Is it the same as before, just the hostname?

correct, the proposed implementation keeps backward compatibility  the previous. that is, if no DNS suffix found just the hostname is returned.

jrvaldes: > And have you been able to see what the csr approval behavior is like when using AD?

Same, I had to change this here, given that the approver could not match a node by name using just the output of  `hostname,` it also helps to standardize the "get hostname" implementation to be consistent across the code base.

pkg/csr/csr.go
jrvaldes: The description and comments of the linked bug expand on the why. I added more context to the description of the PR to avoid pushing a new commit and re-running all those that passed already. In case there is a code change request, I will also update the commit message.

pkg/windows/windows.go
jrvaldes: correct, and the evidence in all e2e test job that passed, none are using a windows instance that is part of a domain controller, so dnsSuffix is empty and returns just the hostName

pkg/windows/windows.go
jrvaldes: it is not going away, it just got moved into the new helper function

```go
func (vm *windows) GetHostname() (string, error) {
	hostName, err := vm.Run(GetHostnameFQDNCommand, true)
	if err != nil {
		return "", fmt.Errorf("error getting the host name, with stdout %s: %w", hostName, err)
	}
	return strings.TrimSpace(hostName), nil
}
```

PR #2212, opened 2024-06-07
WINC-1291: Use all existing CAs for TLS auth when pulling images

mansikulkarni96:
This PR adds support for using TLS certs for mirror registries by watching MCO's controllerconfig which holds the ImageRegistryBundleUserData and ImageRegistryBundleData.
These certs are combined into a bundle file and transferred to the trusted bundle store location on Windows nodes. 
If proxy is enabled, these certs are combined with the proxy certs.

### DIFF
--- a/controllers/controllerconfig_controller.go
+++ b/controllers/controllerconfig_controller.go
@@ -33,6 +33,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	"github.com/openshift/windows-machine-config-operator/pkg/cluster"
+	"github.com/openshift/windows-machine-config-operator/pkg/nodeconfig"
 	"github.com/openshift/windows-machine-config-operator/pkg/secrets"
 	"github.com/openshift/windows-machine-config-operator/pkg/signer"
 )
@@ -101,16 +102,15 @@ func (r *ControllerConfigReconciler) Reconcile(ctx context.Context, req ctrl.Req
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *ControllerConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	mccName := "machine-config-controller"
 	mccPredicate := predicate.Funcs{
 		CreateFunc: func(e event.CreateEvent) bool {
-			return e.Object.GetName() == mccName
+			return e.Object.GetName() == nodeconfig.MccName
 		},
 		UpdateFunc: func(e event.UpdateEvent) bool {
-			return e.ObjectNew.GetName() == mccName
+			return e.ObjectNew.GetName() == nodeconfig.MccName
 		},
 		GenericFunc: func(e event.GenericEvent) bool {
-			return e.Object.GetName() == mccName
+			return e.Object.GetName() == nodeconfig.MccName
 		},
 		DeleteFunc: func(e event.DeleteEvent) bool {
 			return false
--- a/pkg/nodeconfig/nodeconfig.go
+++ b/pkg/nodeconfig/nodeconfig.go
@@ -12,6 +12,7 @@ import (
 	ignCfgTypes "github.com/coreos/ignition/v2/config/v3_4/types"
 	"github.com/go-logr/logr"
 	configv1 "github.com/openshift/api/config/v1"
+	mcfg "github.com/openshift/api/machineconfiguration/v1"
 	"github.com/vincent-petithory/dataurl"
 	"golang.org/x/crypto/ssh"
 	core "k8s.io/api/core/v1"
@@ -62,6 +63,8 @@ const (
 	mcoNamespace = "openshift-machine-config-operator"
 	// mcoBootstrapSecret is the resource name that holds the cert and token required to create the bootstrap kubeconfig
 	mcoBootstrapSecret = "node-bootstrapper-token"
+	// MccName is the name of the Machine Config Controller object
+	MccName = "machine-config-controller"
 )
 
 // nodeConfig holds the information to make the given VM a kubernetes node. As of now, it holds the information
@@ -156,10 +159,8 @@ func (nc *nodeConfig) Configure() error {
 	if err := nc.createRegistryConfigFiles(); err != nil {
 		return err
 	}
-	if cluster.IsProxyEnabled() {
-		if err := nc.ensureTrustedCABundle(); err != nil {
-			return err
-		}
+	if err := nc.SyncTrustedCABundle(); err != nil {
+		return err
 	}
 	wicdKC, err := nc.generateWICDKubeconfig()
 	if err != nil {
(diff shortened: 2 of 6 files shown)


controllers/controllerconfig_controller.go
```diff
@@ -85,14 +89,49 @@ func (r *ControllerConfigReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+	}
+	// merge registry certs with proxy certs, if enabled
+	if cluster.IsProxyEnabled() {
+		proxyCA := &core.ConfigMap{}
+		if err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: r.watchNamespace,
+			Name: certificates.ProxyCertsConfigMap}, proxyCA); err != nil {
+			return ctrl.Result{}, fmt.Errorf("unable to get ConfigMap %s: %w", certificates.ProxyCertsConfigMap, err)
+		}
+		caBundle += proxyCA.Data[certificates.CABundleKey]
+	}
```
sebsoto, line 109: Please break this out into a separate function

sebsoto: Having both this and `SyncTrustedCABundle` is confusing to me.
Will `SyncTrustedCABundle` not overwrite the certs from ImageRegistryBundleUserData + ImageRegistryBundleData? Shouldn't all of this be in that function?

pkg/registries/registries.go
```diff
@@ -229,8 +231,7 @@ func (ms *mirrorSet) generateConfig(secretsConfig credentialprovider.DockerConfi
 		result += hostCapabilities
 		result += "\r\n"
 
-		// TODO: Remove this when we support TLS auth using CA certs - https://issues.redhat.com/browse/WINC-1291
-		result += "  skip_verify = true"
+		result += fmt.Sprintf("  ca = \"%s\"", strings.ReplaceAll(windows.TrustedCABundlePath, "\\", "\\\\"))
```
sebsoto, line 234: Does this line need to be added for this to work?

sebsoto (changes requested):

pkg/registries/registries.go
mansikulkarni96: yes the bundle location needs to be specified

controllers/controllerconfig_controller.go
mansikulkarni96: updated PTAL if it's clear now.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -588,20 +587,38 @@ func (nc *nodeConfig) UpdateKubeletClientCA(contents []byte) error {
...
-	trustedCA := &core.ConfigMap{}
+// SyncTrustedCABundle builds the trusted CA ConfigMap from image registry certificates and the proxy trust bundle
+// and ensures the cert bundle on the instance has up-to-date data
+func (nc *nodeConfig) SyncTrustedCABundle() error {
+	caBundle := ""
+	var cc mcfg.ControllerConfig
 	if err := nc.client.Get(context.TODO(), types.NamespacedName{Namespace: nc.wmcoNamespace,
-		Name: certificates.ProxyCertsConfigMap}, trustedCA); err != nil {
-		return fmt.Errorf("unable to get ConfigMap %s: %w", certificates.ProxyCertsConfigMap, err)
+		Name: "machine-config-controller"}, &cc); err != nil {
```
sebsoto, line 596: Lets use a variable for `machine-config-controller` if it is being used within the controllerconfig controller as well

pkg/registries/registries_test.go
```diff
@@ -252,7 +252,7 @@ func TestGenerateConfig(t *testing.T) {
 				},
 				mirrorSourcePolicy: config.AllowContactingSource,
 			},
-			expectedOutput: "server = \"https://registry.access.redhat.com/ubi9/ubi-minimal\"\r\n\r\n[host.\"https://example.io/example/ubi-minimal\"]\r\n  capabilities = [\"pull\"]\r\n  skip_verify = true\r\n",
+			expectedOutput: "server = \"https://registry.access.redhat.com/ubi9/ubi-minimal\"\r\n\r\n[host.\"https://example.io/example/ubi-minimal\"]\r\n  capabilities = [\"pull\"]\r\n  ca = \"C:\\\\Temp\\\\ca-bundle.crt\"\r\n",
```
sebsoto, line 255: I know you are not introducing this, but putting this in `C:\Temp` is very odd.
As this is partly a refactor for the ca-bundle managment I think it is appropriate to move this to C:/k.
Please add a commit changing TrustedCABundlePath to use`k8sDir`, instead of `remoteDir`

sebsoto (changes requested):
Thanks @mansikulkarni96 LGTM besides from the comments I left

pkg/nodeconfig/nodeconfig.go
```diff
@@ -12,6 +12,7 @@ import (
 	ignCfgTypes "github.com/coreos/ignition/v2/config/v3_4/types"
 	"github.com/go-logr/logr"
 	configv1 "github.com/openshift/api/config/v1"
+	mcfg "github.com/openshift/api/machineconfiguration/v1"
```
wgahnagl, line 15: nit: maybe mcfg could be changed to something more descriptive

pkg/nodeconfig/nodeconfig.go
```diff
@@ -588,20 +587,38 @@ func (nc *nodeConfig) UpdateKubeletClientCA(contents []byte) error {
...
+	var cc mcfg.ControllerConfig
 	if err := nc.client.Get(context.TODO(), types.NamespacedName{Namespace: nc.wmcoNamespace,
-		Name: certificates.ProxyCertsConfigMap}, trustedCA); err != nil {
-		return fmt.Errorf("unable to get ConfigMap %s: %w", certificates.ProxyCertsConfigMap, err)
+		Name: "machine-config-controller"}, &cc); err != nil {
+		return err
+	}
+	for _, bundle := range cc.Spec.ImageRegistryBundleUserData {
+		// Accumulate the certs into a single bundle, adding a comment with the image repo name for observability
+		caBundle += fmt.Sprintf("# %s\n%s\n\n", strings.ReplaceAll(bundle.File, "..", ":"), bundle.Data)
```
wgahnagl, line 601: could this caBundle be moved to a function instead of duplicating the code?

pkg/nodeconfig/nodeconfig.go
jrvaldes: I think is fine, that is the existing pattern https://github.com/openshift/windows-machine-config-operator/blob/master/pkg/ignition/ignition.go#L11

pkg/nodeconfig/nodeconfig.go
jrvaldes: +1

and consider adding unit test coverage.

pkg/registries/registries.go
```diff
@@ -229,8 +231,7 @@ func (ms *mirrorSet) generateConfig(secretsConfig credentialprovider.DockerConfi
 		result += hostCapabilities
 		result += "\r\n"
 
-		// TODO: Remove this when we support TLS auth using CA certs - https://issues.redhat.com/browse/WINC-1291
-		result += "  skip_verify = true"
```
jrvaldes, line 233: What is the default value for `skip_verify`?  should it be explicitly set to not `true`

jrvaldes (commented):
@mansikulkarni96  thanks for working on this. Left minor suggestions, PTAL.

pkg/registries/registries.go
mansikulkarni96: Default value is false https://github.com/containerd/containerd/blob/main/docs/hosts.md

jrvaldes: ACK, thanks for checking this out.

jrvaldes (approved):
/lgtm

PR #2248, opened 2024-07-02
OCPBUGS-36395: [config] Increase controller manager resource limits

mansikulkarni96:
Increase memory resource requests and limits as the current limits are proving to be insufficient in an upgrade scenario, tested in QE clusters.


saifshaikh48:
/lgtm

if you choose update to only increase the limit, lmk and I’ll re-LGTM

jrvaldes:
LGTM

PR #2256, opened 2024-07-05
OCPBUGS-36643: Handle reverse lookup error correctly

sebsoto:
This fixes an issue where if reverse lookup fails due to an error, such as reverse lookup services being down, WMCO would not fall back to using the VM hostname to determine if a CSR should be approved.

### DIFF
--- a/pkg/csr/csr.go
+++ b/pkg/csr/csr.go
@@ -193,9 +193,8 @@ func (a *Approver) validateNodeName(nodeName string) (bool, error) {
 	// check if the node name matches the lookup of any of the instance addresses
 	hasEntry, err := matchesDNS(nodeName, windowsInstances)
 	if err != nil {
-		return false, fmt.Errorf("unable to map node name to the addresses of Windows instances: %w", err)
-	}
-	if hasEntry {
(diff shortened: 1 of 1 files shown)


pkg/csr/csr.go
```diff
@@ -193,9 +193,8 @@ func (a *Approver) validateNodeName(nodeName string) (bool, error) {
 	// check if the node name matches the lookup of any of the instance addresses
 	hasEntry, err := matchesDNS(nodeName, windowsInstances)
 	if err != nil {
-		return false, fmt.Errorf("unable to map node name to the addresses of Windows instances: %w", err)
-	}
-	if hasEntry {
+		a.log.Info("error occurred with reverse DNS lookup, falling back to hostname validation", "error", err)
```
jrvaldes, line 196:
```suggestion
		a.log.Info("reverse DNS lookup not available, falling back to hostname validation", "error", err)
```

I would spin it toward a info-like msg

jrvaldes (approved):
LGTM, left a minor suggestion.

saifshaikh48 (commented):
/lgtm

sebsoto:
/override ci/prow/nutanix-e2e-operator

Cloud agnostic issue.
Nutanix infrastructure is down in CI.

PR #2279, opened 2024-07-18
OCPBUGS-29253, WINC-1172: Always use created WICD SA token

sebsoto:
This is a behavior change which comes as a result of inconsistencies across OCP platforms and versions. Previously WMCO would check if the cluster created a SA token secret for the WICD SA, and if it exists would use the cluster created secret, else it would create its own secret.

We are finding that on upgrades from 4.15 to 4.16 the secret was missing on 4.15, but then created by an entity in the cluster on 4.16. This is throwing off WMCO's logic to get the secret as there will be multiple SA token secrets, one created by WMCO in 4.15, and an additional one when the 4.16 upgrade occurs.

This commit remedies this issue by having WMCO only use the secret it creates, ignoring any other service account token secrets in the WMCO namespace.

### DIFF
--- a/pkg/nodeconfig/nodeconfig.go
+++ b/pkg/nodeconfig/nodeconfig.go
@@ -16,6 +16,7 @@ import (
 	"github.com/vincent-petithory/dataurl"
 	"golang.org/x/crypto/ssh"
 	core "k8s.io/api/core/v1"
+	k8sapierrors "k8s.io/apimachinery/pkg/api/errors"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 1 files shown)


jrvaldes:
wicd-unit-vsphere job is failing consistently. Investigating.  

https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_windows-machine-config-operator/2279/pull-ci-openshift-windows-machine-config-operator-master-wicd-unit-vsphere/1814000801824116736

jrvaldes:
/retitle WINC-1172, OCPBUGS-29253: Always use created WICD SA token

saifshaikh48:
/test azure-e2e-upgrade

Operator install failed post upgrade, interesting it passed in a previous run

mansikulkarni96:
/test azure-e2e-upgrade 
passed on a recent run on master submodule update

mansikulkarni96:
/test azure-e2e-upgrade
cluster install failure

PR #2376, opened 2024-08-14
[release-4.15] OCPBUGS-38485: Always use created WICD SA token

sebsoto:
Manual cherry-pick of https://github.com/openshift/windows-machine-config-operator/pull/2285

This is a behavior change which comes as a result of inconsistencies across OCP platforms and versions. Previously WMCO would check if the cluster created a SA token secret for the WICD SA, and if it exists would use the cluster created secret, else it would create its own secret.

We are finding that on upgrades from 4.15 to 4.16 the secret was missing on 4.15, but then created by an entity in the cluster on 4.16. This is throwing off WMCO's logic to get the secret as there will be multiple SA token secrets, one created by WMCO in 4.15, and an additional one when the 4.16 upgrade occurs.

This commit remedies this issue by having WMCO only use the secret it creates, ignoring any other service account token secrets in the WMCO namespace.

### DIFF
--- a/pkg/nodeconfig/nodeconfig.go
+++ b/pkg/nodeconfig/nodeconfig.go
@@ -15,6 +15,7 @@ import (
 	"github.com/vincent-petithory/dataurl"
 	"golang.org/x/crypto/ssh"
 	core "k8s.io/api/core/v1"
+	k8sapierrors "k8s.io/apimachinery/pkg/api/errors"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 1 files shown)


sebsoto:
/jira cherrypick [OCPBUGS-37481](https://issues.redhat.com/browse/OCPBUGS-37481)

sebsoto:
/test azure-e2e-upgrade


OSDK is failing to upgrade the bundle, due to an error related to the index image. The actual error is being lost.
https://github.com/operator-framework/operator-sdk/blame/5d9a9203be4442b73912db54bd1380d31d58775a/internal/olm/operator/registry/index_image.go#L454

Hoping this is an infrastructure issue in some way. Retesting just in case.

mansikulkarni96 (approved):

sebsoto:
/override ci/prow/vsphere-proxy-e2e-operator
Test passed previously. This has no difference on proxy vs normal vsphere

PR #2379, opened 2024-08-14
WINC-710: replaces kube-proxy flags with config file

wgahnagl:
launching kube-proxy with flags is now deprecated. Changes to using a config file instead.
This adds the creation of the kube-proxy config file to the script run before setting up kube-proxy. This script also contains the post-configuration setup steps for  CNI. Ideally this script will be broken up or retired, and we will be able to generate the kube-proxy configuration in a more idiomatic way.

### DIFF
--- a/pkg/daemon/controller/controller.go
+++ b/pkg/daemon/controller/controller.go
@@ -429,7 +429,7 @@ func (sc *ServiceController) expectedServiceCommand(expected servicescm.Service)
 	var nodeVars, psVars map[string]string
 	var err error
 	if len(expected.NodeVariablesInCommand) > 0 {
-		nodeVars, err = sc.resolveNodeVariables(expected)
+		nodeVars, err = sc.resolveNodeVariables(expected.NodeVariablesInCommand)
 		if err != nil {
 			return "", err
 		}
@@ -453,14 +453,14 @@ func (sc *ServiceController) expectedServiceCommand(expected servicescm.Service)
 
 // resolveNodeVariables returns a map, with the keys being each variable, and the value being the string to replace the
 // variable with
-func (sc *ServiceController) resolveNodeVariables(svc servicescm.Service) (map[string]string, error) {
+func (sc *ServiceController) resolveNodeVariables(nodevars []servicescm.NodeCmdArg) (map[string]string, error) {
 	vars := make(map[string]string)
 	var node core.Node
 	err := sc.client.Get(sc.ctx, client.ObjectKey{Name: sc.nodeName}, &node)
 	if err != nil {
 		return nil, err
 	}
-	for _, nodeVar := range svc.NodeVariablesInCommand {
+	for _, nodeVar := range nodevars {
 		nodeParser := jsonpath.New("nodeParser")
 		if err := nodeParser.Parse(nodeVar.NodeObjectJsonPath); err != nil {
 			return nil, err
@@ -488,15 +488,36 @@ func (sc *ServiceController) resolveNodeVariables(svc servicescm.Service) (map[s
 func (sc *ServiceController) resolvePowershellVariables(svc servicescm.Service) (map[string]string, error) {
 	vars := make(map[string]string)
 	for _, script := range svc.PowershellPreScripts {
+		// replaces the values in the PowershellPreScripts path with the node args
+		script, err := sc.resolveVariablesInPath(script)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to resolve variables in path")
+		}
 		out, err := sc.psCmdRunner.Run(script.Path)
 		if err != nil {
-			return nil, fmt.Errorf("could not resolve PowerShell variable %s: %w", script.VariableName, err)
+			return nil, fmt.Errorf("could not run PowerShell script %s: %w", script.Path, err)
 		}
 		if script.VariableName != "" {
 			vars[script.VariableName] = strings.TrimSpace(out)
 		}
 	}
 	return vars, nil
+}
+
+// resolveVariablesInPath converts placeholder command line arguments passed through in the NodeArgs into their actual values
+func (sc *ServiceController) resolveVariablesInPath(script servicescm.PowershellPreScript) (servicescm.PowershellPreScript, error) {
+	if len(script.NodeArgs) != 0 {
+		psNodeVars, err := sc.resolveNodeVariables(script.NodeArgs)
+		if err != nil {
+			return script, err
+		}
+		pwshCommand := script.Path
+		for key, value := range psNodeVars {
+			pwshCommand = strings.ReplaceAll(pwshCommand, key, value)
+		}
+		script.Path = pwshCommand
+	}
+	return script, nil
 }
 
 // waitUntilNodeReady waits until the Node being configured is ready. Returns an error on timeout.
--- a/pkg/daemon/controller/controller_test.go
+++ b/pkg/daemon/controller/controller_test.go
@@ -24,6 +24,7 @@ import (
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/manager"
 	"github.com/openshift/windows-machine-config-operator/pkg/metadata"
 	"github.com/openshift/windows-machine-config-operator/pkg/servicescm"
+	"github.com/openshift/windows-machine-config-operator/pkg/windows"
 )
 
 var wmcoNamespace = "openshift-windows-machine-config-operator"
@@ -121,7 +122,7 @@ func TestResolveNodeVariables(t *testing.T) {
 				cmdRunner: &fakePSCmdRunner{},
 			})
 			require.NoError(t, err)
-			actual, err := c.resolveNodeVariables(test.service)
+			actual, err := c.resolveNodeVariables(test.service.NodeVariablesInCommand)
 			if test.expectErr {
 				require.Error(t, err)
 				return
@@ -134,19 +135,64 @@ func TestResolveNodeVariables(t *testing.T) {
 
 func TestResolvePowershellVariables(t *testing.T) {
 	testIO := []struct {
-		name      string
-		service   servicescm.Service
-		expected  map[string]string
-		expectErr bool
+		name            string
+		nodeName        string
+		nodeAnnotations map[string]string
+		nodeLabels      map[string]string
+		service         servicescm.Service
+		expected        map[string]string
+		expectErr       bool
 	}{
 		{
-			name:      "No Powershell variables to replace",
-			service:   servicescm.Service{},
-			expected:  map[string]string{},
+			name:            "No Powershell variables to replace",
+			nodeName:        "node",
+			nodeAnnotations: map[string]string{"desiredkey": "desiredvalue"},
+			nodeLabels:      map[string]string{"label": "labelvalue"},
+			service:         servicescm.Service{},
+			expected:        map[string]string{},
+			expectErr:       false,
+		},
+		{
+			name:            "Resolve powershell path variables",
+			nodeName:        "node",
+			nodeAnnotations: map[string]string{"desiredkey": "desiredvalue"},
+			nodeLabels:      map[string]string{"label": "labelvalue"},
+			service: servicescm.Service{
+				PowershellPreScripts: []servicescm.PowershellPreScript{{
+					VariableName: "CMD_REPLACE",
+					Path:         "c:\\k\\script.ps1" + " -hostnameOverride NODE_NAME -clusterCIDR NODE_SUBNET -kubeConfigPath KUBE_CONFIG_PATH -kubeProxyConfigPath KUBE_PROXY_CONFIG_PATH -verbosity VERBOSITY",
+					NodeArgs: []servicescm.NodeCmdArg{
+						{
+							Name:               "NODE_NAME",
+							NodeObjectJsonPath: "{.metadata.name}",
+						},
+						{
+							Name:               "KUBE_CONFIG_PATH",
+							NodeObjectJsonPath: windows.KubeconfigPath,
+						},
+						{
+							Name:               "KUBE_PROXY_CONFIG_PATH",
+							NodeObjectJsonPath: windows.KubeProxyConfigPath,
+						},
+						{
+							Name:               "VERBOSITY",
+							NodeObjectJsonPath: "0",
+						},
+						{
+							Name:               "NODE_SUBNET",
+							NodeObjectJsonPath: "nodesubnet",
+						},
+					},
+				},
+				}},
+			expected:  map[string]string{"CMD_REPLACE": "127.0.0.1"},
 			expectErr: false,
 		},
 		{
-			name: "Resolve variable with unknown path",
+			name:            "Resolve variable with unknown path",
+			nodeName:        "node",
+			nodeAnnotations: map[string]string{"desiredkey": "desiredvalue"},
+			nodeLabels:      map[string]string{"label": "labelvalue"},
 			service: servicescm.Service{
 				PowershellPreScripts: []servicescm.PowershellPreScript{{
 					VariableName: "CMD_REPLACE",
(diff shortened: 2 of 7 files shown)


wgahnagl:
https://github.com/wgahnagl/windows-machine-config-operator/blob/d3315ce91e66b5cb87c14fc51c6bfe176fb2556a/pkg/nodeconfig/payload/payload.go#L282
this could also use the method used here if the current strategy doesn't look good

pkg/services/services.go
```diff
@@ -136,33 +136,31 @@ func hybridOverlayConfiguration(vxlanPort string, debug bool) servicescm.Service
...
+			NodeArgs: []servicescm.NodeCmdArg{
+				{
+					Name:               "NODE_NAME",
+					NodeObjectJsonPath: "{.metadata.name}",
+				},
+				{
+					Name:               "NODE_SUBNET",
+					NodeObjectJsonPath: fmt.Sprintf("{.metadata.annotations.%s}", sanitizedSubnetAnnotation),
+				},
+			},
```
saifshaikh48, line 160: How do the node variables in the `Path` get replaced? Might need to have WICD do this before it runs the script

pkg/nodeconfig/payload/payload.go
```diff
@@ -161,9 +167,86 @@ if( $endpoint -eq $null) {
     $endpoint = New-HnsEndpoint -NetworkId $hns_network.ID -Name "VIPEndpoint"
     Attach-HNSHostEndpoint -EndpointID $endpoint.ID -CompartmentID 1
 }
+# Get HNS endpoint IP
+$ipv4Address = (Get-NetIPConfiguration -AllCompartments -All -Detailed | Where-Object { $_.NetAdapter.LinkLayerAddress -eq $endpoint.MacAddress }).IPV4Address.IPAddress.Trim()
```
saifshaikh48, line 171: does the script still return this value? if not, please update the comment at the top of the script with the new output

pkg/nodeconfig/payload/payload.go
```diff
@@ -81,6 +81,12 @@ const (
 	//       related. Correct that in: https://issues.redhat.com/browse/WINC-882
 	// networkConfTemplate is the template used to generate the network configuration script
 	networkConfTemplate = `# This script ensures the contents of the CNI config file is correct, and returns the HNS endpoint IP
+
+param(
+    [string]$NodeName,
+    [string]$NodeSubnet
```
jrvaldes, line 87: consider naming the input params according to the usage, i.e. 

```
    [string]$hostnameOverride,
    [string]$clusterCIDR
```


the    mapping should happened where the invocation takes place

pkg/services/services.go
wgahnagl: do you mean that the NODE_NAME and NODE_SUBNET are variables in the path? For the script, the NodeObjectJsonPath is basically a json gotten from an api call that contains things like the name and annotation.
this sets the args passed to the script to the value gotten from that API call

pkg/nodeconfig/payload/payload.go
wgahnagl: fixed!

pkg/services/services.go
```diff
@@ -136,33 +136,31 @@ func hybridOverlayConfiguration(vxlanPort string, debug bool) servicescm.Service
 
 // kubeProxyConfiguration returns the Service definition for kube-proxy
 func kubeProxyConfiguration(debug bool) servicescm.Service {
-	sanitizedSubnetAnnotation := strings.ReplaceAll(nodeconfig.HybridOverlaySubnet, ".", "\\.")
-	cmd := fmt.Sprintf("%s -log-file=%s %s --windows-service --proxy-mode=kernelspace --feature-gates=WinOverlay=true,WinDSR=true "+
-		"--hostname-override=NODE_NAME --kubeconfig=%s --cluster-cidr=NODE_SUBNET "+
-		"--network-name=%s --source-vip=ENDPOINT_IP --enable-dsr=true", windows.KubeLogRunnerPath, windows.KubeProxyLog,
-		windows.KubeProxyPath, windows.KubeconfigPath, windows.OVNKubeOverlayNetwork)
+	cmd := fmt.Sprintf("%s -log-file=%s %s --source-vip=ENDPOINT_IP --config %s", windows.KubeLogRunnerPath, windows.KubeProxyLog,
```
jrvaldes, line 139: is the `--source-vip=ENDPOINT_IP `  replacement needed? looks like it is happening in the powershell script already:

```
  sourceVip: $ipv4Addres
```

jrvaldes (changes requested):
@wgahnagl   Thanks for working on this; it's mostly LGTM. Please take a look at my comments.

pkg/services/services.go
```diff
@@ -136,33 +136,31 @@ func hybridOverlayConfiguration(vxlanPort string, debug bool) servicescm.Service
 
 // kubeProxyConfiguration returns the Service definition for kube-proxy
 func kubeProxyConfiguration(debug bool) servicescm.Service {
-	sanitizedSubnetAnnotation := strings.ReplaceAll(nodeconfig.HybridOverlaySubnet, ".", "\\.")
-	cmd := fmt.Sprintf("%s -log-file=%s %s --windows-service --proxy-mode=kernelspace --feature-gates=WinOverlay=true,WinDSR=true "+
```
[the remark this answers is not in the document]
wgahnagl: that was one of the flags for the kube-proxy. It's handled now by the config file :+1:

pkg/nodeconfig/payload/payload.go
```diff
@@ -161,9 +167,86 @@ if( $endpoint -eq $null) {
...
+$ipv4Address = (Get-NetIPConfiguration -AllCompartments -All -Detailed | Where-Object { $_.NetAdapter.LinkLayerAddress -eq $endpoint.MacAddress }).IPV4Address.IPAddress.Trim()
+
+#Kube Proxy configuration
+
+$kube_proxy_config=@'
+kind: KubeProxyConfiguration
+apiVersion: kube-proxy.config.k8s.io/v1alpha1
+featureGates:
+  WinDSR: true
+  WinOverlay: true
```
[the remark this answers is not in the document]
saifshaikh48: imo we could keep `WinOverlay: true` just to be explicit

pkg/services/services.go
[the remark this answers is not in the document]
jrvaldes: can you highlight how/where the config file indicates that kube-proxy should run as a service?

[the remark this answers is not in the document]
mansikulkarni96: +1 couldn't find this, WindowsRunAsService bool should be added in the config.

mansikulkarni96 (changes requested):

pkg/services/services.go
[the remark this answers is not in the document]
jrvaldes: WindowsRunAsService is available in kubelet 1.31+     , not in 1.30.  So the merge to `master` will be blocked by the branch day task and the 1.31 rebase. cc @mansikulkarni96  

hope no backport for this :)  

https://github.com/kubernetes/kube-proxy/blob/release-1.30/config/v1alpha1/types.go

pkg/services/services.go
saifshaikh48: Could you give an example of what WICD actually executes on the Windows node to run the script?

I guess don't see what mechanism is filling in the NODE_NAME and NODE_SUBNET values into
```
network-conf.ps1 -NodeName NODE_NAME -NodeSubnet NODE_SUBNET
```

mansikulkarni96 (commented):
Mostly LGTM, can we kick off a test here to see the logs and networking tests?

pkg/daemon/controller/controller.go
```diff
@@ -488,9 +488,21 @@ func (sc *ServiceController) resolveNodeVariables(svc servicescm.Service) (map[s
 func (sc *ServiceController) resolvePowershellVariables(svc servicescm.Service) (map[string]string, error) {
 	vars := make(map[string]string)
 	for _, script := range svc.PowershellPreScripts {
+		// replaces the values in the PowershellPreScripts path with the node args
+		if len(script.NodeArgs) != 0 {
```
mansikulkarni96, line 492: you can check if this is 0 and return early here. This will remove indentation.

wgahnagl: if we return early, then it won't get to the sc.psCmdRunner.Run, and the script won't be executed... This block should only run when NodeArgs are set

pkg/daemon/controller/controller.go
```diff
@@ -488,9 +488,21 @@ func (sc *ServiceController) resolveNodeVariables(svc servicescm.Service) (map[s
...
+			pwshCommand := script.Path
+			for key, value := range psNodeVars {
+				pwshCommand = strings.ReplaceAll(pwshCommand, key, value)
+			}
+			script.Path = pwshCommand
+		}
 		out, err := sc.psCmdRunner.Run(script.Path)
 		if err != nil {
-			return nil, fmt.Errorf("could not resolve PowerShell variable %s: %w", script.VariableName, err)
+			return nil, fmt.Errorf("could not run PowerShell script %s: %w", script.Path, err)
```
saifshaikh48, line 505: not sure how i feel about including the entire powershell script. might affect readability of the error msg. consider logging it instead

pkg/daemon/controller/controller.go
```diff
@@ -488,9 +488,21 @@ func (sc *ServiceController) resolveNodeVariables(svc servicescm.Service) (map[s
...
+			psNodeVars, err := sc.resolveNodeVariables(script.NodeArgs)
+			if err != nil {
+				return nil, err
+			}
+			pwshCommand := script.Path
+			for key, value := range psNodeVars {
+				pwshCommand = strings.ReplaceAll(pwshCommand, key, value)
+			}
+			script.Path = pwshCommand
+		}
```
saifshaikh48, line 502: consider breaking this out into a helper function to ease unit testing

pkg/daemon/controller/controller_test.go
```diff
@@ -134,19 +135,63 @@ func TestResolveNodeVariables(t *testing.T) {
...
+			expected:        map[string]string{},
+			expectErr:       false,
+		},
+		{
+			name:            "Resolve powershell path variables",
+			nodeName:        "node",
+			nodeAnnotations: map[string]string{"desiredkey": "desiredvalue"},
+			nodeLabels:      map[string]string{"label": "labelvalue"},
+			service: servicescm.Service{
+				PowershellPreScripts: []servicescm.PowershellPreScript{{
```
saifshaikh48, line 161: as is, this test is not doing anything.

In order to properly test that the replacements are occurring as expected, you need to add a `VariableName` and then add its entry to the test case's `expected` map

saifshaikh48 (commented):
nice work getting this to where it is. almost at the finish line

pkg/daemon/controller/controller.go
wgahnagl: Do you think we should break it out and test the individual function, instead of testing within ResolveNodeVariables?

pkg/services/services.go
```diff
@@ -136,33 +136,31 @@ func hybridOverlayConfiguration(vxlanPort string, debug bool) servicescm.Service
...
-		"--hostname-override=NODE_NAME --kubeconfig=%s --cluster-cidr=NODE_SUBNET "+
-		"--network-name=%s --source-vip=ENDPOINT_IP --enable-dsr=true", windows.KubeLogRunnerPath, windows.KubeProxyLog,
-		windows.KubeProxyPath, windows.KubeconfigPath, windows.OVNKubeOverlayNetwork)
+	cmd := fmt.Sprintf("%s -log-file=%s %s --source-vip=ENDPOINT_IP --config %s", windows.KubeLogRunnerPath, windows.KubeProxyLog,
+		windows.KubeProxyPath, windows.KubeProxyConfigPath)
 	// Set log level
 	cmd = fmt.Sprintf("%s %s", cmd, klogVerbosityArg(debug))
```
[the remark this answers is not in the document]
sebsoto: This is now doing verbosity via both the args and the config file.

pkg/daemon/controller/controller.go
sebsoto: Its not the entire powershell script. It is the path + args

sebsoto (changes requested):

pkg/daemon/controller/controller.go
saifshaikh48: Fair enough, though the script "Path" is a bit of a misnomer since it can be an actual [in-line powershell](https://github.com/openshift/windows-machine-config-operator/blob/master/pkg/services/services.go#L221-L222) commands

pkg/daemon/controller/controller.go
```diff
@@ -499,6 +504,21 @@ func (sc *ServiceController) resolvePowershellVariables(svc servicescm.Service)
 	return vars, nil
 }
 
+func (sc *ServiceController) resolveVaraiblesInPath(script servicescm.PowershellPreScript) (servicescm.PowershellPreScript, error) {
```
jrvaldes, line 507: Please add a comment documenting the function and consider adding a unit test in `controller_test.go` for the function as well, especially to cover the string replacement.

pkg/daemon/controller/controller.go
```diff
@@ -488,9 +488,14 @@ func (sc *ServiceController) resolveNodeVariables(svc servicescm.Service) (map[s
 func (sc *ServiceController) resolvePowershellVariables(svc servicescm.Service) (map[string]string, error) {
 	vars := make(map[string]string)
 	for _, script := range svc.PowershellPreScripts {
+		// replaces the values in the PowershellPreScripts path with the node args
+		script, err := sc.resolveVaraiblesInPath(script)
```
jrvaldes, line 492: s/resolveVaraiblesInPath/resolveVariablesInPath

pkg/nodeconfig/payload/payload.go
```diff
@@ -80,7 +80,36 @@ const (
...
+        [string]$ConfigPath,
+        [string]$NewConfigContent
+    )
+    
+    # Read existing config content
+    $existing_config = ""
+    if (Test-Path -Path $ConfigPath) {
+        $config_file_content = Get-Content -Path $ConfigPath -Raw
+        if ($config_file_content -ne $null) {
+` + "        $existing_config=$config_file_content.Replace(\"`r\",\"\")" + `
```
jrvaldes, line 104: >  $existing_config=$config_file_content.Replace(\"`r\",\"\")"

Is this replacement taking place in PowerShell? PowerShell uses 
```
`n`r
```
 for new lines on Windows.

Consider using the `[Environment]::NewLine`  resolver to easy to char escaping

jrvaldes (changes requested):
@wgahnagl, thanks for working throughout the reviews. I left a few recommendations. PTAL

As the `kube-proxy.conf` file grows in content, I feel this PR should add an e2e test to fetch the content of the file and check it has the correct/expected values.

@openshift/openshift-team-windows-containers  WDYT?

pkg/nodeconfig/payload/payload.go
jrvaldes: I follow, @wgahnagl. What is the goal here? To replace all the `new line` with the empty string `""`?

Does the `Replace` function replaces all occurrences?

sebsoto: Please don't change any of the replace logic with this ticket. False positives on this were the cause of a major networking bug. There is no easy way to test this logic given it is in powershell.

pkg/daemon/controller/controller.go
sebsoto: Would this potential unit test test anything that resolvePowershellVariables is not already testing?

sebsoto (approved):
Thanks for the PR @wgahnagl

I am of the personal opinion that an E2E test for the config file is not needed. E2E tests validating the functionality of k8s services and networking for windows workloads are more useful and will test that the config file is populated correctly.


If windows workloads have incorrect networking behavior that could be caused by a config file issue, we should be testing the end behavior to catch the config issues.

mansikulkarni96:
/test aws-e2e-operator
Kicking off to see the networking behavior

saifshaikh48:
/lgtm

assuming existing networking func has been validated by running the e2e tests locally

jrvaldes:
LGTM

/test remaining-required

PR #2388, opened 2024-08-20
WINC-1269: Expose WMCO /metrics endpoint via HTTPS

mansikulkarni96:
Use the new controller-runtime secureAccess flag and [filters.WithAuthenticationAndAuthorization](https://github.com/kubernetes-sigs/controller-runtime/pull/2407) when exposing metrics endpoint.
kube-rbac-proxy has been removed from controller-runtime scaffolding and its use is discouraged to drop the dependency on maintaining the image.


mansikulkarni96:
[WMCO logs](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_windows-machine-config-operator/2388/pull-ci-openshift-windows-machine-config-operator-master-aws-e2e-operator/1826696816788246528/artifacts/aws-e2e-operator/windows-e2e-operator-test/artifacts/wmco.log)
2024-08-22T21:39:04Z	INFO	controller-runtime.metrics	Serving metrics server	{"bindAddress": "0.0.0.0:8443", "secure": true}

wgahnagl:
I think you might've pushed some vendor changes. Should those be there?

mansikulkarni96:
> I think you might've pushed some vendor changes. Should those be there?

@wgahnagl they are generated with make vendor for the import `sigs.k8s.io/controller-runtime/pkg/metrics/filters`

mansikulkarni96:
> With this are we able to remove the kube-rbac-proxy deployment (`manager_auth_proxy_patch.yaml` config file)? I know it's not being used but not sure if it a generated file or a resource that we maintain

We are not using it, but we can consider removing it. It came with the Kubebuilder project scaffolding.

jrvaldes:
> > With this are we able to remove the kube-rbac-proxy deployment (`manager_auth_proxy_patch.yaml` config file)? I know it's not being used but not sure if it a generated file or a resource that we maintain
> 
> We are not using it, but we can consider removing it. It came with the Kubebuilder project scaffolding.

+1 to removing it as part of this PR

cmd/operator/main.go
```diff
@@ -156,7 +160,9 @@ func main() {
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme: scheme,
 		Metrics: metricsserver.Options{
-			BindAddress: fmt.Sprintf("%s:%d", metrics.Host, metrics.Port),
```
jrvaldes, line 159: Why `9182` port is no longer relevant?

https://github.com/openshift/windows-machine-config-operator/blob/master/pkg/metrics/metrics.go#L46

cmd/operator/main.go
```diff
@@ -61,8 +62,11 @@ func init() {
 
 func main() {
 	var debugLogging bool
+	var metricsAddr string
 
 	flag.BoolVar(&debugLogging, "debugLogging", false, "Log debug messages")
+	flag.StringVar(&metricsAddr, "metrics-bind-address", "0.0.0.0:8443",
+		"The address the metric endpoint binds to.")
```
jrvaldes, line 69: Please mention the proposed default (`8443`) in the argument description.

config/manager/manager.yaml
```diff
@@ -35,9 +35,14 @@ spec:
         - windows-machine-config-operator
         args:
         - "--debugLogging"
+        - "--metrics-bind-address=0.0.0.0:8443"
         image: controller:latest
         name: manager
         imagePullPolicy: IfNotPresent
+        ports:
+        - containerPort: 8443
```
jrvaldes, line 43: Given that the port is hard-coded in the manifest; What is the use case of the proposed flag `metrics-bind-address` ?

jrvaldes (commented):
@mansikulkarni96  thanks for working on this. Mostly LGTM, left minor suggestions. PTAL

config/manager/manager.yaml
mansikulkarni96: The flag is already available through controller-runtime's server package, the intention is to avoid hardcoding in code and pass it through the CSV.

cmd/operator/main.go
saifshaikh48: nit, if you still want to add this

config/default/manager_auth_proxy_patch.yaml
saifshaikh48, line 1: nit, if you want to add to the commit message why we can't/won't use it and how this work makes it obsolete anyway

saifshaikh48 (commented):
/lgtm

jrvaldes:
/lgtm

good work @mansikulkarni96  🎉

mansikulkarni96:
/approve
self-approve after 2 lgtm's

mansikulkarni96:
test passed de-provision timed out

PR #2408, opened 2024-09-04
OCPBUGS-16711, WINC-853: WICD best effort cleans up removes all managed services

saifshaikh48:
The best effort cleanup of WICD is expanded to look at service descriptions
to cleanup all services that WICD started, even if they are not in the curren
 iteration of the Windows services ConfigMap. This also prevents a user from
being able to sneak in services into the CM that WICD would never cleanup
since all services started through WICD/windows-services CM will have the
"OpenShift managed" description.

### DIFF
--- a/pkg/daemon/cleanup/cleanup.go
+++ b/pkg/daemon/cleanup/cleanup.go
@@ -21,6 +21,7 @@ package cleanup
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	core "k8s.io/api/core/v1"
 	"k8s.io/client-go/rest"
@@ -34,6 +35,7 @@ import (
 	"github.com/openshift/windows-machine-config-operator/pkg/daemon/powershell"
 	"github.com/openshift/windows-machine-config-operator/pkg/metadata"
 	"github.com/openshift/windows-machine-config-operator/pkg/servicescm"
+	"github.com/openshift/windows-machine-config-operator/pkg/windows"
 )
 
 // Deconfigure removes all managed services from the instance and the version annotation, if it has an associated node.
@@ -59,11 +61,11 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
 		klog.Exitf("could not create service manager: %s", err.Error())
 	}
 	defer svcMgr.Disconnect()
-	mergedCMData, err := getMergedCMData(ctx, directClient, configMapNamespace, node)
+	mergedCMData, removeAllTaggedServices, err := getMergedCMData(ctx, directClient, configMapNamespace, node)
 	if err != nil {
 		return err
 	}
-	if err = removeServices(svcMgr, mergedCMData.Services); err != nil {
+	if err = removeServices(svcMgr, mergedCMData.Services, removeAllTaggedServices); err != nil {
 		return err
 	}
 	envVarsRemoved, err := ensureEnvVarsAreRemoved(mergedCMData.WatchedEnvironmentVars)
@@ -93,17 +95,18 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
 
 // getMergedCMData attempts to get the latest and the version CM data specified by the node's version annotation
 // It returns the merged CM Data containing services and the watched environment variables
+// as well as a boolean indicating whether the latest services CM spec differs from that of the version annotation CM.
 func getMergedCMData(ctx context.Context, cli client.Client,
-	configMapNamespace string, node *core.Node) (*servicescm.Data, error) {
+	configMapNamespace string, node *core.Node) (*servicescm.Data, bool, error) {
 	// get data from the latest services ConfigMap
 	latestCM, err := servicescm.GetLatest(cli, ctx, configMapNamespace)
 	if err != nil {
-		return nil, fmt.Errorf("cannot get latest services ConfigMap from namespace %s: %w",
+		return nil, false, fmt.Errorf("cannot get latest services ConfigMap from namespace %s: %w",
 			configMapNamespace, err)
 	}
 	latestCMData, err := servicescm.Parse(latestCM.Data)
 	if err != nil {
-		return nil, err
+		return nil, false, err
 	}
 
 	// attempt to get the ConfigMap specified by the version annotation
@@ -131,18 +134,18 @@ func getMergedCMData(ctx context.Context, cli client.Client,
 	if err != nil {
 		klog.Infof("error getting services ConfigMap associated with version annotation, "+
 			"falling back to use latest services ConfigMap: %s", err)
-		return latestCMData, nil
+		return latestCMData, true, nil
 	}
 	// If the instance was configured using latestCM, return that
 	if versionCM.GetName() == latestCM.GetName() {
-		return latestCMData, nil
+		return latestCMData, false, nil
 	}
 	mergedServices := mergeServices(latestCMData.Services, versionCMData.Services)
 	mergedEnvVars := merge(latestCMData.WatchedEnvironmentVars, versionCMData.WatchedEnvironmentVars)
 	return &servicescm.Data{
 		Services:               mergedServices,
 		WatchedEnvironmentVars: mergedEnvVars,
-	}, nil
+	}, true, nil
 }
 
 // mergeServices combines the list of services, prioritizing the data given by s1
@@ -181,7 +184,8 @@ func merge(e1, e2 []string) []string {
 }
 
 // removeServices uses the given manager to remove all the given Windows services from this instance.
-func removeServices(svcMgr manager.Manager, services []servicescm.Service) error {
+// The removeAllTaggedServices flag is used to also remove OpenShift-managed services that may not be in the given slice
+func removeServices(svcMgr manager.Manager, services []servicescm.Service, removeAllTaggedServices bool) error {
 	// Build up log message and failures
 	servicesRemoved := []string{}
 	failedRemovals := []error{}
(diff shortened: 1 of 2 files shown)


pkg/daemon/cleanup/cleanup.go
```diff
@@ -194,6 +196,33 @@ func removeServices(svcMgr manager.Manager, services []servicescm.Service) error
 			servicesRemoved = append(servicesRemoved, service.Name)
 		}
 	}
+
+	if bestEffort {
```
jrvaldes, line 200: should this happend only if `len(failedRemovals) > 0` ?

pkg/daemon/cleanup/cleanup.go
```diff
@@ -194,6 +196,33 @@ func removeServices(svcMgr manager.Manager, services []servicescm.Service) error
...
+			winSvcObj, err := svcMgr.OpenService(name)
+			if err != nil {
+				return err
+			}
+			config, err := winSvcObj.Config()
+			if err != nil {
+				return err
+			}
+			if strings.Contains(config.Description, windows.ManagedTag) {
+				if err := svcMgr.DeleteService(name); err != nil {
```
jrvaldes, line 216: Will the priority order of the Windows services be a dependency issue given `existingSvcs` has no order?

jrvaldes (commented):
@saifshaikh48  thanks for working on this, mostly LGTM. PTAL at my comments.

pkg/daemon/cleanup/cleanup.go
saifshaikh48: Nope those failures would be from the services CM, while the best effort removals are a separate concept: any services NOT in the services CM anymore but still running

pkg/daemon/cleanup/cleanup.go
saifshaikh48: This is a good point and something I went back and forth about — I chose to not look into dependencies in best effort cleanup because of a particular scenario that could enable a DDOS:
- user sneaks in service X into configmap spec
- Service X depends on a crucial system service like system32
- WICD goes to delete service X
- Traverses dependency tree and first stops system32…

saifshaikh48:
/test aws-e2e-operator
/test vsphere-e2e-operator
/test gcp-e2e-operator

Not sure where I’m going wrong with the windows service manager api interaction

pkg/daemon/cleanup/cleanup.go
mansikulkarni96: I agree but there is a caveat here. We are also using the best effort bool to clean the services from the latest config map if the version config map is not found. Wouldn't we want to delete the services by dependency for that?

mansikulkarni96 (changes requested):

pkg/daemon/cleanup/cleanup.go
saifshaikh48: > services from the latest config map

These are passed in to removeServices and so will be deleted in proper dependency order. The bestEffort block will only affect lingering OpenShift managed services that haven't been passed into the func (i.e. services in neither latest CM nor version CM)

saifshaikh48:
Issue I'm trying to fix: WICD cleanup is failing with Process exited with status 1

cleanup command output shows:
```
PS C:\Users\Administrator> C:\\k\\windows-instance-config-daemon.exe cleanup --kubeconfig C:\\k\\wicd-kubeconfig --namespace openshift-windows-machine-config-operator
I0909 16:50:06.935061    1652 cleanup.go:46] starting deconfigure() with cmNS: openshift-windows-machine-config-operator      
I0909 16:50:07.008534    1652 cleanup.go:59] no associated node found
I0909 16:50:07.035531    1652 cleanup.go:137] error getting services ConfigMap associated with version annotation, falling back to use latest services ConfigMap: no node object present
I0909 16:50:07.040658    1652 cleanup.go:208] best effort...
F0909 16:50:07.044232    1652 cleanup.go:51] could not open service ClipSVC: Access is denied.
```
so some system services cannot even be opened (need to open to read its description, as description is used to determine if we should delete it or not). I may have to tweak the logic to not error out if we cannot open a service during bestEffort cleanup

saifshaikh48:
/test aws-e2e-operator
/test vsphere-e2e-operator
/test gcp-e2e-operator

Manually brought up a node with this change, looked fine

pkg/daemon/cleanup/cleanup.go
sebsoto: https://github.com/openshift/windows-machine-config-operator/pull/2408/files#r1747421558

I'm not sure if this logic is correct.
Why should we stop system32 if Service X depends on it?
We should have no problem stopping service X. The issue would be if system32 depended on service X. In that case we would have to stop system32 before stopping service x.

But in order for that to be the case, the person doing this would need permissions to change the system32 service in the first place.

pkg/daemon/cleanup/cleanup.go
```diff
@@ -92,18 +94,18 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
 }
 
 // getMergedCMData attempts to get the latest and the version CM data specified by the node's version annotation
-// It returns the merged CM Data containing services and the watched environment variables
+// It returns the merged CM Data containing services, the watched environment variables, and if best effort is desired
 func getMergedCMData(ctx context.Context, cli client.Client,
```
sebsoto, line 98: This function seems a bit overloaded with this addition.
A teller of this is that 3 things are being returned.

sebsoto: Its difficult to tell the criteria of whether a cleanup should be best effort or not.

pkg/daemon/cleanup/cleanup.go
```diff
@@ -131,18 +133,18 @@ func getMergedCMData(ctx context.Context, cli client.Client,
...
-		return latestCMData, nil
+		return latestCMData, false, nil
 	}
 	mergedServices := mergeServices(latestCMData.Services, versionCMData.Services)
 	mergedEnvVars := merge(latestCMData.WatchedEnvironmentVars, versionCMData.WatchedEnvironmentVars)
 	return &servicescm.Data{
 		Services:               mergedServices,
 		WatchedEnvironmentVars: mergedEnvVars,
-	}, nil
+	}, true, nil
```
sebsoto, line 147: Why is a best effort cleanup needed here?

pkg/daemon/cleanup/cleanup.go
```diff
@@ -92,18 +94,18 @@ func Deconfigure(cfg *rest.Config, ctx context.Context, configMapNamespace strin
...
+		return nil, false, fmt.Errorf("cannot get latest services ConfigMap from namespace %s: %w",
 			configMapNamespace, err)
 	}
 	latestCMData, err := servicescm.Parse(latestCM.Data)
 	if err != nil {
-		return nil, err
+		return nil, false, err
 	}
 
 	// attempt to get the ConfigMap specified by the version annotation
```
sebsoto, line 111: Depending on what is done to this function to address the overload, breaking this inline func out into its own function could be a good idea.

pkg/daemon/cleanup/cleanup.go
sebsoto: Can best effort cleanup always be done? Is a bool even required?

sebsoto (changes requested):

pkg/daemon/cleanup/cleanup.go
sebsoto: >  A teller of this is that 3 things are being returned.

Note: not saying this is always the case, but at the least it means we should inspect whats happening here closely to see if its truely required.

jrvaldes:
> Issue I'm trying to fix: WICD cleanup is failing with Process exited with status 1
> 
> cleanup command output shows:
> 
> ```
> PS C:\Users\Administrator> C:\\k\\windows-instance-config-daemon.exe cleanup --kubeconfig C:\\k\\wicd-kubeconfig --namespace openshift-windows-machine-config-operator
> I0909 16:50:06.935061    1652 cleanup.go:46] starting deconfigure() with cmNS: openshift-windows-machine-config-operator      
> I0909 16:50:07.008534    1652 cleanup.go:59] no associated node found
> I0909 16:50:07.035531    1652 cleanup.go:137] error getting services ConfigMap associated with version annotation, falling back to use latest services ConfigMap: no node object present
> I0909 16:50:07.040658    1652 cleanup.go:208] best effort...
> F0909 16:50:07.044232    1652 cleanup.go:51] could not open service ClipSVC: Access is denied.
> ```
> 
> so some system services cannot even be opened (need to open to read its description, as description is used to determine if we should delete it or not). I may have to tweak the logic to not error out if we cannot open a service during bestEffort cleanup

I looked online and couldn't find a better way to list/filter the services, I need to dig deeper.

Ideally, the logic should list (fetch) the services, filtering by the description, so that the heavy lifting happens on the server side (Windows), not in Go. @saifshaikh48, have you looked into this?

saifshaikh48:
> filtering by the description, so that the heavy lifting happens on the server side (Windows)

@jrvaldes that's a good call out, I had considered this but I decided against it as it is a Powershell-based solution which would require processing the powershell text output then running a bunch of powershell commands to stop the services

pkg/daemon/cleanup/cleanup.go
saifshaikh48: explored an alternative with [these drafted commits](https://github.com/openshift/windows-machine-config-operator/compare/master...saifshaikh48:windows-machine-config-operator:managed-svc-cleanup2?expand=1). 

resolving the convo as we decided to keep this approach. I've since improved the docs around the logic/decisions

pkg/daemon/cleanup/cleanup.go
saifshaikh48: @sebsoto was correct. As a bonus, we realized DeleteService() does a "safe delete" under the hood meaning we already implemented the dependency tree traversal logic behind the scenes :)

pkg/daemon/cleanup/cleanup.go
saifshaikh48: addressed with a clearer function doc

sebsoto: Going off the doc the return value should be flipped.
True should be returned only if the instance was configured using latestCM

sebsoto (approved):
Good work, approved but please address the comment I left before allowing this to merge

jrvaldes (approved):
/lgtm

Thanks @saifshaikh48  for fixing this.

sebsoto:
Passed previously already

PR #2422, opened 2024-09-17
WINC-1337: Use upstreamed dependency code

sebsoto:
Changes the code to list Windows services dependencies to use the function which was upstreamed into the go windows svc libraries.


saifshaikh48 (commented):
/lgtm

PR #2282, opened 2024-07-22
OCPBUGS-36305: Add reboot docs and clean HNS networks before rebooting in AWS

jrvaldes:
This PR introduces documentation to reboot a Windows node configured by WMCO properly. 

Direct access to the underlying Windows VM to trigger a restart is not recommended, the existing node annotation must be used, instead. 

In addition, this PR introduces a fix for AWS as a particular case where the EC2 Agent fails to configure the metadata endpoint when the HNS are fully configured with the virtual network interfaces.

PR #2433, opened 2024-09-26
OCPBUGS-35285: Remove HNS networks before rebooting in AWS

jrvaldes:
This PR introduces a fix for AWS as a particular case where the EC2 Agent fails to configure the metadata endpoint when the HNS are fully configured with the virtual network interfaces.

followup-to
- https://github.com/openshift/windows-machine-config-operator/pull/2282

### DIFF
--- a/pkg/nodeconfig/nodeconfig.go
+++ b/pkg/nodeconfig/nodeconfig.go
@@ -276,6 +276,14 @@ func (nc *nodeConfig) SafeReboot(ctx context.Context) error {
 		return fmt.Errorf("unable to drain node %s: %w", nc.node.Name, err)
 	}
 
+	// HNS networks conflicts with the persistent route to the metadata endpoint in AWS. Explicitly remove them
+	// to allow the same VM to be configured as a node again.
+	if nc.platformType == configv1.AWSPlatformType {
+		if err := nc.Windows.EnsureHNSNetworksAreRemoved(); err != nil {
+			return err
+		}
+	}
+
 	if err := nc.Windows.RebootAndReinitialize(); err != nil {
 		return err
 	}
--- a/pkg/nodeconfig/nodeconfig.go
+++ b/pkg/nodeconfig/nodeconfig.go
@@ -277,7 +277,7 @@ func (nc *nodeConfig) SafeReboot(ctx context.Context) error {
 	}
 
 	// HNS networks conflicts with the persistent route to the metadata endpoint in AWS. Explicitly remove them
-	// to allow the same VM to be configured as a node again.
+	// to allow the same instance to be configured as a node again.
 	if nc.platformType == configv1.AWSPlatformType {
 		if err := nc.Windows.EnsureHNSNetworksAreRemoved(); err != nil {
 			return err
--- a/test/e2e/reconfigure_test.go
+++ b/test/e2e/reconfigure_test.go
@@ -3,12 +3,16 @@ package e2e
 import (
 	"context"
 	"encoding/json"
+	"fmt"
+	"log"
 	"testing"
 
+	config "github.com/openshift/api/config/v1"
 	"github.com/stretchr/testify/require"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
+	"github.com/openshift/windows-machine-config-operator/pkg/metadata"
 	"github.com/openshift/windows-machine-config-operator/pkg/patch"
 	"github.com/openshift/windows-machine-config-operator/pkg/wiparser"
 )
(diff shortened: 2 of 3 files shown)


test/e2e/reconfigure_test.go
```diff
@@ -3,23 +3,51 @@ package e2e
...
 
+	"github.com/openshift/windows-machine-config-operator/pkg/metadata"
 	"github.com/openshift/windows-machine-config-operator/pkg/patch"
 	"github.com/openshift/windows-machine-config-operator/pkg/wiparser"
 )
 
 func reconfigurationTestSuite(t *testing.T) {
 	tc, err := NewTestContext()
 	require.NoError(t, err)
+	t.Run("Remove version annotation", tc.testRemoveVersionAnnotation)
```
saifshaikh48, line 22: Only thing I am worried about is how much time this may add to our test runs. Consider running this only on 1 or 2 platforms (AWS + something else) or just not on azure since azure is right up against the timeout

test/e2e/reconfigure_test.go
```diff
@@ -3,23 +3,51 @@ package e2e
...
 	t.Run("Re-add removed instance", tc.testReAddInstance)
 	t.Run("Change private key", testPrivateKeyChange)
 }
 
+// testRemoveVersionAnnotation tests the case where the version annotation is removed from the node.
+func (tc *testContext) testRemoveVersionAnnotation(t *testing.T) {
+	ctx := context.TODO()
+	require.NoError(t, tc.loadExistingNodes(), "error getting the current Windows nodes in the cluster")
+	// machines instances
+	for _, node := range gc.machineNodes {
```
saifshaikh48, line 32: consider doing this only to only 1 machine node and 1 BYOH node. Would give a good enough test signal IMO

test/e2e/reconfigure_test.go
saifshaikh48, line 1: I think we had a [similar test removed](https://github.com/openshift/windows-machine-config-operator/pull/2211/commits/1a9503657a5c63181ee6f968f75aee13dbf6f4e3) in the past, any idea why the behavior described in that commit message is not being seen anymore?

pkg/nodeconfig/nodeconfig.go
```diff
@@ -276,6 +276,14 @@ func (nc *nodeConfig) SafeReboot(ctx context.Context) error {
 		return fmt.Errorf("unable to drain node %s: %w", nc.node.Name, err)
 	}
 
+	// HNS networks conflicts with the persistent route to the metadata endpoint in AWS. Explicitly remove them
+	// to allow the same VM to be configured as a node again.
+	if nc.platformType == configv1.AWSPlatformType {
+		if err := nc.Windows.EnsureHNSNetworksAreRemoved(); err != nil {
```
saifshaikh48, line 282: Just to understand more thoroughly, is the consequence of this that nodes on AWS will be fully reconfigured any time WMCO reboots them?

test/e2e/reconfigure_test.go
```diff
@@ -73,3 +101,16 @@ func (tc *testContext) testReAddInstance(t *testing.T) {
 	require.NoError(t, err, "error waiting for the Windows node to be re-added")
 	tc.testNodesBecomeReadyAndSchedulable(t)
 }
+
```
mansikulkarni96, line 104: missing doc

pkg/nodeconfig/nodeconfig.go
mansikulkarni96: +1, we use safeReboot for quite  a few use cases, will this take of network re-configuration?

pkg/nodeconfig/nodeconfig.go
```diff
@@ -276,6 +276,14 @@ func (nc *nodeConfig) SafeReboot(ctx context.Context) error {
 		return fmt.Errorf("unable to drain node %s: %w", nc.node.Name, err)
 	}
 
+	// HNS networks conflicts with the persistent route to the metadata endpoint in AWS. Explicitly remove them
+	// to allow the same VM to be configured as a node again.
```
mansikulkarni96, line 280: same instance*

mansikulkarni96 (changes requested):

test/e2e/reconfigure_test.go
jrvaldes: AFAIK, there is only 1 machine node and 1 BYOH node in the e2e test.

test/e2e/reconfigure_test.go
jrvaldes: > WICD
immediately replaces as the instance is full configured.

such a statement is still valid

test/e2e/reconfigure_test.go
jrvaldes: makes sense, updated to run only in AWS

pkg/nodeconfig/nodeconfig.go
jrvaldes: addressed

test/e2e/reconfigure_test.go
jrvaldes: addressed

test/e2e/reconfigure_test.go
```diff
@@ -3,23 +3,55 @@ package e2e
...
 func reconfigurationTestSuite(t *testing.T) {
 	tc, err := NewTestContext()
 	require.NoError(t, err)
+	t.Run("Remove version annotation", tc.testRemoveVersionAnnotation)
 	t.Run("Re-add removed instance", tc.testReAddInstance)
 	t.Run("Change private key", testPrivateKeyChange)
 }
 
+// testRemoveVersionAnnotation tests the case where the version annotation is removed from the node.
+func (tc *testContext) testRemoveVersionAnnotation(t *testing.T) {
```
sebsoto, line 29: Where is the check that the node gets rebooted?

sebsoto: Is removing the version annotation guaranteed to cause a reboot?
If so the mechanics of that should be documented in the test

jrvaldes: > Where is the check that the node gets rebooted?

That is not what this test is "testing", the expectation is that WMCO/WICD restores the version annotation in the node where the "how" may be out of the scope for this test.

`err = tc.waitForConfiguredWindowsNodes()` with the second argument set to `true` means the version annotation must be present and valid.

saifshaikh48 (commented):
/lgtm

saifshaikh48:
there may be a few other bugs in the backlog that are fixed by this change btw

PR #2439, opened 2024-10-03
WINC-1339: Use kube 1.31

sebsoto:
Submodule changes generated with hack/update-submodules.sh


mansikulkarni96:
Should we also bump the kube-proxy version to use 1.31 tag with this?

saifshaikh48:
/lgtm

I don’t think a 1.31 tag has been published yet in o/k but there is one is k/k. Anyway IMO we should move our submodule to track upstream k/k

mansikulkarni96 (approved):

sebsoto:

The minimum version needs to be bumped in 4.17 and here before this can pass. Overriding to unblock CI and will 'backport' the change

PR #2471, opened 2024-10-11
Hardcode Azure SKU version for 2019 nodes

saifshaikh48:
This PR hardcodes Azure VMs to a specific version as the latest build released by MSFT is unstable. Any Windows Server 2019 VMs created from the October 2024 release `"version": "17763.6414.241004"` has issues preventing SSH connections.

Required to unblock CI. Should be reverted when patched Windows Server images are released.

### DIFF
--- a/hack/machineset.sh
+++ b/hack/machineset.sh
@@ -152,10 +152,12 @@ get_azure_ms() {
   local byoh=$6
 
   local sku="2022-datacenter-smalldisk"
+  local release="latest"
   if [ "$winver" == "2019" ]; then
 		# 2019 images without the containers feature pre-installed cannot be used due to
 		# https://issues.redhat.com/browse/OCPBUGS-13244
     sku="2019-datacenter-with-containers-smalldisk"
+    release="17763.6293.240905"   
   fi
 
   cat <<EOF
@@ -171,7 +173,7 @@ $(get_spec $infraID $az $provider $byoh)
             publisher: MicrosoftWindowsServer
             resourceID: ""
             sku: $sku
-            version: latest
+            version: $release
           kind: AzureMachineProviderSpec
           location: ${region}
           managedIdentity: ${infraID}-identity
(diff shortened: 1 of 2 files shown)


jrvaldes:
> This commit hardcodes Azure VMs to a specific version as the latest build released by MSFT is unstable. Any Windows Server 2019 VMs created from the October 2024 release `"version": "17763.6414.241004"` has issues preventing SSH connections.
> 
> Required to unblock CI. Should be reverted when patched Windows Server images are released.



How does the proposed fix unblock CI ?


I feel this must be addressed in the test code as well.

https://github.com/openshift/windows-machine-config-operator/blob/master/test/e2e/providers/azure/azure.go#L27

hack/machineset.sh
```diff
@@ -171,7 +171,7 @@ $(get_spec $infraID $az $provider $byoh)
             publisher: MicrosoftWindowsServer
             resourceID: ""
             sku: $sku
-            version: latest
+            version: "17763.6293.240905"
```
jrvaldes, line 174: Using the latest with Windows Server 2022 seems safe; the fix version should be used with Windows Server 2019 only.

saifshaikh48:
yep sorry the test code change slipped for some reason, that's what I get for trying to use the Github web UI for the first time

test/e2e/providers/azure/azure.go
```diff
@@ -52,6 +52,11 @@ func New(clientset *clusterinfo.OpenShift, infraStatus *config.InfrastructureSta
 
 // newAzureMachineProviderSpec returns an AzureMachineProviderSpec generated from the inputs, or an error
 func (p *Provider) newAzureMachineProviderSpec(location string, zone string, windowsServerVersion windows.ServerVersion) (*mapi.AzureMachineProviderSpec, error) {
+	imageVersion := defaultImageVersion
+	if windowsServerVersion == windows.Server2019 {
+		// TODO: remove when VM SSH issue is patched in Azure cloud
```
jrvaldes, line 57: Should we track this issue as a bug and add a link here?

jrvaldes (approved):
/lgtm

mansikulkarni96 (approved):

PR #2483, opened 2024-10-17
WINC-897: Regenerate servicesmanifest on rendered-worker update 

wgahnagl:
This PR reacts to rendered-worker changes by regenerating the ServicesManifest when the rendered-worker changes. This ensures that the ServicesManifest always stays up to date with the full state of the rendered-worker.

### DIFF
--- a/controllers/configmap_controller.go
+++ b/controllers/configmap_controller.go
@@ -25,6 +25,8 @@ import (
 	"strings"
 
 	config "github.com/openshift/api/config/v1"
+	oconfig "github.com/openshift/api/config/v1"
+	mcfgv1 "github.com/openshift/api/machineconfiguration/v1"
 	core "k8s.io/api/core/v1"
 	rbac "k8s.io/api/rbac/v1"
 	k8sapierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -83,6 +85,7 @@ type ConfigMapReconciler struct {
 	instanceReconciler
 	servicesManifest *servicescm.Data
 	proxyEnabled     bool
+	VXLANPort        string
 }
 
 // NewConfigMapReconciler returns a pointer to a ConfigMapReconciler
@@ -103,18 +106,9 @@ func NewConfigMapReconciler(mgr manager.Manager, clusterConfig cluster.Config, w
 	if err != nil {
 		return nil, err
 	}
-	ign, err := ignition.New(directClient)
-	if err != nil {
-		return nil, fmt.Errorf("error creating ignition object: %w", err)
-	}
-	argsFromIgnition, err := ign.GetKubeletArgs()
+	svcData, err := generateServicesManifest(directClient, clusterConfig.Network().VXLANPort(), clusterConfig.Platform())
 	if err != nil {
 		return nil, err
-	}
-	svcData, err := services.GenerateManifest(argsFromIgnition, clusterConfig.Network().VXLANPort(),
-		clusterConfig.Platform(), ctrl.Log.V(1).Enabled())
-	if err != nil {
-		return nil, fmt.Errorf("error generating expected Windows service state: %w", err)
 	}
 
 	return &ConfigMapReconciler{
@@ -130,6 +124,7 @@ func NewConfigMapReconciler(mgr manager.Manager, clusterConfig cluster.Config, w
 		},
 		servicesManifest: svcData,
 		proxyEnabled:     proxyEnabled,
+		VXLANPort:        clusterConfig.Network().VXLANPort(),
 	}, nil
 }
 
@@ -157,6 +152,12 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context,
 	if err != nil {
 		return ctrl.Result{}, fmt.Errorf("unable to create signer from private key secret: %w", err)
 	}
+
+	servicesManifest, err := generateServicesManifest(r.client, r.VXLANPort, r.platform)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	r.servicesManifest = servicesManifest
 
 	// Fetch the ConfigMap. The predicate will have filtered out any ConfigMaps that we should not reconcile
 	// so it is safe to assume that all ConfigMaps being reconciled are one of:
@@ -414,7 +415,26 @@ func (r *ConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			builder.WithPredicates(outdatedWindowsNodePredicate(true))).
 		Watches(&core.Node{}, handler.EnqueueRequestsFromMapFunc(r.mapToServicesConfigMap),
 			builder.WithPredicates(windowsNodeVersionChangePredicate())).
+		Watches(&mcfgv1.MachineConfig{}, handler.EnqueueRequestsFromMapFunc(r.mapToServicesConfigMap),
+			builder.WithPredicates(machineConfigCreatedPredicate())).
 		Complete(r)
+}
+
+func machineConfigCreatedPredicate() predicate.Funcs {
+	return predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			return strings.HasPrefix(e.Object.GetName(), ignition.RenderedWorkerPrefix)
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return false
+		},
+		GenericFunc: func(e event.GenericEvent) bool {
+			return false
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return false
+		},
+	}
 }
 
 // isValidConfigMap returns true if the ConfigMap object is the InstanceConfigMap or a WMCO-managed ConfigMap
@@ -670,4 +690,23 @@ func (r *ConfigMapReconciler) ensureWICDClusterRoleBinding(ctx context.Context) 
 		r.log.Info("Created resource", "ClusterRoleBinding", expectedCRB.Name)
 	}
 	return err
+}
+
+// generateServicesManifest generates and regenerates the services manifest.
+// this gets called when the configmap reconciler is first created, to create the services manifest,
+// and also when the rendered-worker configmap is changed, to regenerate it.
+func generateServicesManifest(client client.Client, port string, platform oconfig.PlatformType) (*servicescm.Data, error) {
+	ign, err := ignition.New(client)
+	if err != nil {
+		return nil, fmt.Errorf("error creating ignition object: %w", err)
+	}
+	argsFromIgnition, err := ign.GetKubeletArgs()
+	if err != nil {
+		return nil, fmt.Errorf("Error getting kubelet args from ignition: %w", err)
+	}
+	svcData, err := services.GenerateManifest(argsFromIgnition, port, platform, ctrl.Log.V(1).Enabled())
+	if err != nil {
+		return nil, fmt.Errorf("error generating expected Windows service state: %w", err)
+	}
+	return svcData, nil
 }


controllers/configmap_controller.go
```diff
@@ -414,6 +436,12 @@ func (r *ConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			builder.WithPredicates(outdatedWindowsNodePredicate(true))).
 		Watches(&core.Node{}, handler.EnqueueRequestsFromMapFunc(r.mapToServicesConfigMap),
 			builder.WithPredicates(windowsNodeVersionChangePredicate())).
+		Watches(&core.Node{}, handler.EnqueueRequestsFromMapFunc(r.mapToServicesConfigMap),
+			builder.WithPredicates(nodeLabelChangedPredicate())).
+		Watches(&core.Node{}, handler.EnqueueRequestsFromMapFunc(r.mapToServicesConfigMap),
+			builder.WithPredicates(nodeconfigChangedPredicate())).
```
saifshaikh48, line 442: you can consolidate this into 1 by creating a smarter predicate func

controllers/configmap_controller.go
```diff
@@ -186,7 +187,21 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context,
...
+    // get the machineConfig
+    machineConfig := &mcfgv1.MachineConfig{}
+    if err := r.client.Get(ctx, kubeTypes.NamespacedName{
+        Namespace: req.Namespace}, machineConfig); err != nil {
+        return ctrl.Result{}, err
+    }
+    node := &core.Node{}
+    if err := r.client.Get(ctx, kubeTypes.NamespacedName{
+      Namespace: "", 
+      Name: req.Name},
```
saifshaikh48, line 199: which node would you get by doing this? I thought the req.Name would be the name of a configmap given the usage of `handler.EnqueueRequestsFromMapFunc(r.mapToServicesConfigMap)` in your added Watches

controllers/configmap_controller.go
saifshaikh48: please explain why we this watch is needed as well. Do we use node annotations to create (or modify) kubelet args? If so, please point to where we do that because i can't find where in our code

jrvaldes:
/retitle [WIP] WINC-897: React to kubelet arg changes

controllers/configmap_controller.go
```diff
@@ -216,10 +230,50 @@ func (r *ConfigMapReconciler) reconcileServices(ctx context.Context, windowsServ
 			kubeTypes.NamespacedName{Namespace: r.watchNamespace, Name: windowsServices.Name}, "Error", err.Error())
 		return nil
 	}
+	r.setValues(node, data)
```
mansikulkarni96, line 233: what is expected to happen after you update the kubelet command?

controllers/configmap_controller.go
```diff
@@ -216,10 +230,50 @@ func (r *ConfigMapReconciler) reconcileServices(ctx context.Context, windowsServ
...
 	return nil
 }
 
+func (r *ConfigMapReconciler) setValues(node *core.Node, data *servicescm.Data) {
+	// machineconfig.Spec.Config
+	// unmarshal the json, and set the data to the machineconfig's value
+
+	// check the node.annotations and node.label
+
+	for i := range data.Services {
```
mansikulkarni96, line 244: This can be simplified, consider this:
```
for _, svc := range data.Services {
		if svc.Name == "kubelet" {
			if strings.Contains(svc.Command, "--node-labels"){
				
			}
		}
	}

```

mansikulkarni96 (changes requested):

controllers/configmap_controller.go
wgahnagl: I actually can't use a for x in range y for this loop because for some reason if you do that, it edits a copy instead of the original object! It's a weird go quirk that you have to use the index if you want to edit the thing the loop is referencing. However, strings.Contains would totally be way better lol

controllers/configmap_controller.go
```diff
@@ -22,9 +22,11 @@ import (
 	"fmt"
```
sebsoto, line 22: To make this more reviewable it would be helpful to describe what you are attempting to do in the commit message.

controllers/configmap_controller.go
```diff
@@ -216,10 +230,50 @@ func (r *ConfigMapReconciler) reconcileServices(ctx context.Context, windowsServ
 			kubeTypes.NamespacedName{Namespace: r.watchNamespace, Name: windowsServices.Name}, "Error", err.Error())
 		return nil
 	}
+	r.setValues(node, data)
 	// TODO: actually react to changes to the services ConfigMap
 	return nil
 }
 
+func (r *ConfigMapReconciler) setValues(node *core.Node, data *servicescm.Data) {
```
sebsoto, line 238: It should be somewhat clear as to what a function is doing from the name/doc.
`setValues` isn't descriptive enough. As this is a method of `ConfigMapReconciler` it implies that this is setting the values for the reconciler itself.

controllers/configmap_controller.go
```diff
@@ -414,6 +479,8 @@ func (r *ConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			builder.WithPredicates(outdatedWindowsNodePredicate(true))).
 		Watches(&core.Node{}, handler.EnqueueRequestsFromMapFunc(r.mapToServicesConfigMap),
 			builder.WithPredicates(windowsNodeVersionChangePredicate())).
+		Watches(&core.Node{}, handler.EnqueueRequestsFromMapFunc(r.mapToServicesConfigMap),
```
sebsoto, line 482: Where are you watching the machineconfig?
Is this supposed to be `MachineConfig`, not node?

controllers/configmap_controller.go
```diff
@@ -216,10 +230,50 @@ func (r *ConfigMapReconciler) reconcileServices(ctx context.Context, windowsServ
...
+func (r *ConfigMapReconciler) setValues(node *core.Node, data *servicescm.Data) {
+	// machineconfig.Spec.Config
+	// unmarshal the json, and set the data to the machineconfig's value
+
+	// check the node.annotations and node.label
+
+	for i := range data.Services {
+		value := &data.Services[i]
+		if value.Name == "kubelet" {
+			re := regexp.MustCompile(`(--node-labels=[^\s]+)`)
```
sebsoto, line 247: This is duplicating logic from https://github.com/openshift/windows-machine-config-operator/blob/a66229cec4c125dc81eaa1b44e99db5812a51c42/pkg/services/services.go#L248
while also missing everything except node-labels.

Do both need to exist? If so the code should be maintained through shared functions.
It is not clear to me how that linked code works in tandem with this code.

controllers/configmap_controller_test.go
```diff
@@ -13,6 +15,231 @@ import (
 	"github.com/openshift/windows-machine-config-operator/pkg/wiparser"
 )
 
+func TestReconcileWindowsServices(t *testing.T) {
```
sebsoto, line 18: Go unit  test functions should have the naming convention of Test($FUNCTION_NAME)
So in this case it would be TestSetValues

controllers/configmap_controller.go
```diff
@@ -158,6 +162,21 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context,
...
+	}
+  argsFromIgnition, err := ign.GetKubeletArgs()
+  if err != nil {
+		return ctrl.Result{}, err
+	}
+  svcData, err := services.GenerateManifest(argsFromIgnition, r.VXLANPort, r.platform, ctrl.Log.V(1).Enabled())
+  if err != nil {
+		return ctrl.Result{}, err
+	}
+  r.servicesManifest = svcData
```
sebsoto, line 178: This is duplicate code https://github.com/openshift/windows-machine-config-operator/blob/84de66e4398af19af6d60b6608c60a03c8373750/controllers/configmap_controller.go#L99-L111

it should be shared via a common helper function. The reason for this is if we need to update the logic then we don't want to have to do it in multiple places. This leads to code churn and potential errors.

sebsoto (changes requested):
Left some comments.
In general please read through your PRs before asking for reviews. Theres a lot of changes here that should not have been made.

Additionally the commit message should describe what you are trying to do, the functional changes to the controller. If this is WIP then you should call out what still needs to be done in either the commit message, or a comment on the PR. Its hard to review the logic changes otherwise.

controllers/configmap_controller.go
```diff
@@ -130,9 +124,26 @@ func NewConfigMapReconciler(mgr manager.Manager, clusterConfig cluster.Config, w
 		},
 		servicesManifest: svcData,
 		proxyEnabled:     proxyEnabled,
+		VXLANPort:        clusterConfig.Network().VXLANPort(),
 	}, nil
 }
 
+func generateServicesManifest(client client.Client, port string, platform oconfig.PlatformType) (*servicescm.Data, error) {
```
mansikulkarni96, line 131: Please add doc around this and expand on the commit msg + the PR message.

controllers/configmap_controller.go
```diff
@@ -130,9 +124,26 @@ func NewConfigMapReconciler(mgr manager.Manager, clusterConfig cluster.Config, w
 		},
 		servicesManifest: svcData,
 		proxyEnabled:     proxyEnabled,
+		VXLANPort:        clusterConfig.Network().VXLANPort(),
```
mansikulkarni96, line 127: It's not clear why we need this, is this for the e2e test?

mansikulkarni96 (changes requested):

controllers/configmap_controller.go
sebsoto, line 131: For style reasons, please move helpers to the end of the file, not between the method declarations

controllers/configmap_controller.go
```diff
@@ -130,9 +124,26 @@ func NewConfigMapReconciler(mgr manager.Manager, clusterConfig cluster.Config, w
...
 }
 
+func generateServicesManifest(client client.Client, port string, platform oconfig.PlatformType) (*servicescm.Data, error) {
+	ign, err := ignition.New(client)
+	if err != nil {
+		return nil, fmt.Errorf("error creating ignition object: %w", err)
+	}
+	argsFromIgnition, err := ign.GetKubeletArgs()
+	if err != nil {
+		return nil, err
```
sebsoto, line 138: Lets wrap this error for easier debugging

controllers/configmap_controller.go
```diff
@@ -414,9 +431,28 @@ func (r *ConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			builder.WithPredicates(outdatedWindowsNodePredicate(true))).
 		Watches(&core.Node{}, handler.EnqueueRequestsFromMapFunc(r.mapToServicesConfigMap),
 			builder.WithPredicates(windowsNodeVersionChangePredicate())).
+		Watches(&mcfgv1.MachineConfig{}, handler.EnqueueRequestsFromMapFunc(r.mapToServicesConfigMap),
+			builder.WithPredicates(machineConfigChangedPredicate())).
 		Complete(r)
 }
 
+func machineConfigChangedPredicate() predicate.Funcs {
```
sebsoto, line 439: machineConfigChangedPredicate doesnt accurately describe what is happening.

sebsoto (changes requested):
Thanks @wgahnagl, this looks better

controllers/configmap_controller.go
```diff
@@ -671,3 +691,19 @@ func (r *ConfigMapReconciler) ensureWICDClusterRoleBinding(ctx context.Context)
 	}
 	return err
 }
+
```
mansikulkarni96, line 694: Please add  a doc here

mansikulkarni96 (approved):

jrvaldes:
/test gcp-e2e-operator

ipi-install failed

controllers/configmap_controller.go
```diff
@@ -414,9 +415,28 @@ func (r *ConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 		Complete(r)
 }
 
+func machineConfigCreatedPredicate() predicate.Funcs {
+	return predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			return strings.HasPrefix(e.Object.GetName(), ignition.RenderedWorkerPrefix)
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return false
```
jrvaldes, line 429: Just curious; does the predicate should react to the `UpdateFunc`?

controllers/configmap_controller.go
```diff
@@ -414,9 +415,28 @@ func (r *ConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			builder.WithPredicates(outdatedWindowsNodePredicate(true))).
 		Watches(&core.Node{}, handler.EnqueueRequestsFromMapFunc(r.mapToServicesConfigMap),
 			builder.WithPredicates(windowsNodeVersionChangePredicate())).
+		Watches(&mcfgv1.MachineConfig{}, handler.EnqueueRequestsFromMapFunc(r.mapToServicesConfigMap),
+			builder.WithPredicates(machineConfigCreatedPredicate())).
 		Complete(r)
 }
 
+func machineConfigCreatedPredicate() predicate.Funcs {
```
jrvaldes, line 423: missing docs; would be helpful to comment what the predicates are  about.

controllers/configmap_controller.go
```diff
@@ -83,6 +85,7 @@ type ConfigMapReconciler struct {
 	instanceReconciler
 	servicesManifest *servicescm.Data
 	proxyEnabled     bool
+	VXLANPort        string
```
jrvaldes, line 88: Is there a requirement to export `VXLANPort`?

controllers/configmap_controller.go
```diff
@@ -158,6 +153,12 @@ func (r *ConfigMapReconciler) Reconcile(ctx context.Context,
 		return ctrl.Result{}, fmt.Errorf("unable to create signer from private key secret: %w", err)
 	}
 
+	servicesManifest, err := generateServicesManifest(r.client, r.VXLANPort, r.platform)
+	if err != nil {
+		return ctrl.Result{}, err
```
jrvaldes, line 158: I recommend wrapping the error for context.

```
errors.wrapf(err, "error generating the services manifest")
```

jrvaldes (approved):
@wgahnagl  thanks for putting this together. Mostly LGTM.  Left minor suggestions. PTAL

controllers/configmap_controller.go
wgahnagl: we need to export VXLANPort here in order to use the generateServicesManifest function on line 156
By that point it won't have access to the whole object like it does on 109

PR #2485, opened 2024-10-18
WINC-545: Improve handling of metrics configuration 

mansikulkarni96:
This PR adds a new metric controller that watches the WMCO namespace to create a service monitor when monitoring is enabled. This PR also removes the code where WMCO creates and manages endpoints object. An endpoint named kubelet in the kube-system namespace contains all the node IP addresses. WMCO can rely on this endpoint and create the serviceMonitor which will change the port from 10250 to 9182 which is our desired metric port for the windows-exporter service.


bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
```diff
@@ -211,6 +211,8 @@ spec:
           - namespaces
           verbs:
           - get
+          - list
```
sebsoto, line 214: Please squash this into the commit the kubebuilder directives are being added.
Else this will cause revert issues if one of the commits needs to be reverted

pkg/metrics/metrics.go
```diff
@@ -2,30 +2,25 @@ package metrics
...
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
-	"github.com/openshift/windows-machine-config-operator/pkg/nodeconfig"
-	"github.com/openshift/windows-machine-config-operator/pkg/patch"
+	monv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	monclient "github.com/prometheus-operator/prometheus-operator/pkg/client/versioned/typed/monitoring/v1"
 )
 
 //+kubebuilder:rbac:groups="",resources=services;services/finalizers,verbs=create;get;delete
-//+kubebuilder:rbac:groups="",resources=endpoints,verbs=create;get;delete;update;patch
```
sebsoto, line 28: I dont see this change reflected in the bundle. Is another file still making endpoint API calls?

pkg/metrics/metrics.go
```diff
@@ -2,30 +2,25 @@ package metrics
...
 	k8sclient "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
-	"github.com/openshift/windows-machine-config-operator/pkg/nodeconfig"
-	"github.com/openshift/windows-machine-config-operator/pkg/patch"
+	monv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	monclient "github.com/prometheus-operator/prometheus-operator/pkg/client/versioned/typed/monitoring/v1"
```
sebsoto: Sort imports. This will fail lint otherwise

pkg/metrics/metrics.go
```diff
@@ -325,12 +187,15 @@ func (c *Config) ensureServiceMonitor() error {
...
+			AttachMetadata: &monv1.AttachMetadata{
+				Node: &attachMetadataBool,
+			},
 			Endpoints: []monv1.Endpoint{
 				{
 					HonorLabels:     true,
 					Interval:        "30s",
 					Path:            "/metrics",
-					Port:            "metrics",
+					Port:            "https-metrics",
```
sebsoto: Does this have a functionality change?

pkg/metrics/metrics.go
```diff
@@ -343,18 +208,42 @@ func (c *Config) ensureServiceMonitor() error {
...
 							Action:      "replace",
 							Regex:       "(.*)",
-							Replacement: &replacement,
+							Replacement: &instanceLabel,
 							TargetLabel: "instance",
 							SourceLabels: []monv1.LabelName{
 								"__meta_kubernetes_endpoint_address_target_name",
 							},
 						},
+						{
```
sebsoto: There is a lot of 'magic' happening here, requiring maintainers to be familiar with these actions and what they may be doing. Please add a short comment for each explaining what they are doing and why they need to be done, if it is not too obvious.

sebsoto (changes requested):
Thanks @mansikulkarni96, its good to see the amount of code this removes :smile: 

Made an initial review, PTAL at my comments

pkg/metrics/metrics_test.go
saifshaikh48, line 1: this deletion would be better suited in the next commit

pkg/metrics/metrics.go
```diff
@@ -1,38 +1,13 @@
...
-	"k8s.io/client-go/tools/record"
-	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/manager"
-)
-
 //+kubebuilder:rbac:groups="",resources=services;services/finalizers,verbs=create;get;delete
 //+kubebuilder:rbac:groups="",resources=namespaces,verbs=get
 //+kubebuilder:rbac:groups="",resources=nodes,verbs=list
 //+kubebuilder:rbac:groups="monitoring.coreos.com",resources=servicemonitors,verbs=create;get;delete
 //+kubebuilder:rbac:groups="",resources=events,verbs=*
```
saifshaikh48, line 7: can we move the ones that are still relevant to metric controller?

pkg/metrics/metrics.go
```diff
@@ -1,38 +1,13 @@
...
-	// metricsEnabled specifies if metrics are enabled in the current cluster
-	metricsEnabled = true
-)
-
 const (
-	// metricsPortName specifies the portname used for Prometheus monitoring
+	// PortName specifies the portname used for Prometheus monitoring
 	PortName = "metrics"
 	// Host is the host address used by Windows metrics
 	Host = "0.0.0.0"
```
saifshaikh48, line 13: is this used anywhere? if not we can delete

pkg/metrics/metrics.go
```diff
@@ -311,11 +168,15 @@ func (c *Config) ensureServiceMonitor() error {
...
-		return fmt.Errorf("error retrieving %s serviceMonitor: %w", WindowsMetricsResource, err)
+		return fmt.Errorf(
+			"error retrieving %s serviceMonitor: %w", WindowsMetricsResource, err)
 	}
 
 	serverName := fmt.Sprintf("%s.%s.svc", WindowsMetricsResource, c.namespace)
-	replacement := "$1"
+	instanceLabel := "$1"
+	portLabel := "$1:9182"
+	jobLabel := "windows-exporter"
```
jrvaldes: I recommend to use the exist constant

https://github.com/openshift/windows-machine-config-operator/blob/05e38ebd7ad67898aa4d8873cb982f97b45dad96/pkg/metrics/metrics.go#L49

pkg/metrics/metrics.go
```diff
@@ -311,11 +168,15 @@ func (c *Config) ensureServiceMonitor() error {
...
 	if err != nil && !apierrors.IsNotFound(err) {
-		return fmt.Errorf("error retrieving %s serviceMonitor: %w", WindowsMetricsResource, err)
+		return fmt.Errorf(
+			"error retrieving %s serviceMonitor: %w", WindowsMetricsResource, err)
 	}
 
 	serverName := fmt.Sprintf("%s.%s.svc", WindowsMetricsResource, c.namespace)
-	replacement := "$1"
+	instanceLabel := "$1"
+	portLabel := "$1:9182"
```
jrvaldes: same here

https://github.com/openshift/windows-machine-config-operator/blob/05e38ebd7ad67898aa4d8873cb982f97b45dad96/pkg/metrics/metrics.go#L46

jrvaldes (commented):
@mansikulkarni96  thanks for refactoring this. Mostly LGTM, just minor suggestions.

jrvaldes:
@ mansikulkarni96 Would a new e2e test be created to cover the new job?

jrvaldes (approved):

pkg/metrics/metrics.go
mansikulkarni96: no but this needs to be same as the kubelet service port

pkg/metrics/metrics.go
mansikulkarni96: This change is difficult to make as that is an int32 constant and needs to be plugged in with the replaced metrics URL. It caused parsing issues when I tested.

controllers/metric_controller.go
```diff
@@ -0,0 +1,262 @@
...
+// Node object and aims to move the current state of the cluster closer to the desired state.
+func (r *metricReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, err error) {
+	r.log = r.log.WithValues(MetricController, req.NamespacedName)
+	// Prevent WMCO upgrades while Node objects are being processed
+	if err := condition.MarkAsBusy(r.client, r.watchNamespace, r.recorder, MetricController); err != nil {
+		return ctrl.Result{}, err
+	}
+	defer func() {
+		err = markAsFreeOnSuccess(r.client, r.watchNamespace, r.recorder, MetricController, result.Requeue, err)
+	}()
```
sebsoto, line 75: This isn't needed, this controller is not configuring nodes.

The purpose of blocking upgrades is to prevent nodes from falling behind from the rest of the cluster on OCP upgrades.

controllers/metric_controller.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+	return ctrl.Result{}, nil
+}
+
+// validate will verify if cluster monitoring is enabled in the operator namespace. If the label is set to false or not
+// present, it will log and send warning events to the user. If the label holds a non-boolean value, returns an error.
+func (r *metricReconciler) validate(ctx context.Context) (bool, error) {
+	// validate if metrics label is added to namespace
+	labelValue := false
+	var err error
```
sebsoto, line 96: Dont think this variable declaration is needed

controllers/metric_controller.go
```diff
@@ -0,0 +1,262 @@
...
+		},
+		GenericFunc: func(e event.GenericEvent) bool {
+			return isMonitoringEnabled(e.Object, r.watchNamespace)
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return false
+		},
+	}
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&v1.Namespace{}, builder.WithPredicates(metricsPredicate)).
```
sebsoto, line 241: I believe you need to watch the servicemonitor resource as well in order to ensure that any changes made to the monitor are correct.

sebsoto (changes requested):

controllers/metric_controller.go
mansikulkarni96: Agreed, I missed this.

controllers/metric_controller.go
```diff
@@ -0,0 +1,256 @@
...
+		Watches(&monv1.ServiceMonitor{}, &handler.EnqueueRequestForObject{}).
```
sebsoto, line 235: This is going to pass the servicemonitor object to the Reconcile function. This is also going to bypass any predicates. Lets be consistent on the object being reconciled. You could use `EnqueueRequestsFromMapFunc` to map the servicemonitor to the namespace it is in.

controllers/metric_controller.go
```diff
@@ -0,0 +1,267 @@
...
+		return fmt.Errorf("error creating service monitor: %w", err)
+	}
+	return nil
+}
+
+// mapToWatchNamespace fulfills the MapFn type, while always returning a request to the operator watch namespace
+func (r *metricReconciler) mapToWatchNamespace(_ context.Context, _ client.Object) []reconcile.Request {
+	return []reconcile.Request{{
+		NamespacedName: kubeTypes.NamespacedName{Namespace: r.watchNamespace, Name: metrics.WindowsMetricsResource},
+	}}
```
sebsoto, line 225: Shouldn't this only enqueue a request in the  case that client.Object.Namespace == watchNamespace?
Else any servicemonitor in the cluster changing will cause a reconcile to trigger.
That is assuming we are watching all namespaces and not just the wmco watch namespace

mansikulkarni96: yes that makes sense, I missed this.

sebsoto (approved):
Thanks @mansikulkarni96 changes LGTM

PR #2533, opened 2024-11-13
WINC-1352: Build CNI plugins with CGO_ENABLED

mansikulkarni96:
CNI submodule is compiled by explicitly setting CGO_ENABLED=0 which is not FIPS compliant. This commit remidaiates this by using a Windows-compatible C compiler “CC=gcc-mingw-w64” in the go build command.

### DIFF
--- a/build/Dockerfile.base
+++ b/build/Dockerfile.base
@@ -70,8 +70,7 @@ RUN env -u VERSION GOOS=windows make ecr-credential-provider
 # Build CNI plugins
 WORKDIR /build/windows-machine-config-operator/containernetworking-plugins/
 COPY containernetworking-plugins/ .
-ENV CGO_ENABLED=0
-RUN ./build_windows.sh
+RUN CGO_ENABLED=1 CC=x86_64-w64-mingw32-gcc ./build_windows.sh
 
 # Build csi-proxy
 WORKDIR /build/windows-machine-config-operator/csi-proxy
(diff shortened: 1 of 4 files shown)


jrvaldes (approved):
/lgtm

Thanks for updating this @mansikulkarni96

build/Dockerfile.ci
```diff
@@ -78,8 +78,7 @@ RUN env -u VERSION GOOS=windows make ecr-credential-provider
 # Build CNI plugins
 WORKDIR /build/windows-machine-config-operator/containernetworking-plugins/
 COPY containernetworking-plugins/ .
-ENV CGO_ENABLED=0
-RUN ./build_windows.sh
+RUN CGO_ENABLED=1 CC=x86_64-w64-mingw32-gcc ./build_windows.sh
```
jrvaldes, line 81: do we need this change in the midstream dockerfiles as well?

mansikulkarni96: yes we will

mansikulkarni96: unless we know that 4.18 will be released with konflux

build/Dockerfile.base
```diff
@@ -70,8 +70,7 @@ RUN env -u VERSION GOOS=windows make ecr-credential-provider
 # Build CNI plugins
 WORKDIR /build/windows-machine-config-operator/containernetworking-plugins/
 COPY containernetworking-plugins/ .
-ENV CGO_ENABLED=0
```
jrvaldes, line 73: Do we need the update here 

https://github.com/openshift/windows-machine-config-operator/blob/master/build/Dockerfile#L74

sebsoto (approved):

build/Dockerfile.ci
sebsoto: > unless we know that 4.18 will be released with konflux
That is the plan but please do it anyway so QE is able to test before the transition occurs

mansikulkarni96:
Failed to deprovision, test passed
/override ci/prow/vsphere-e2e-operator

PR #2566, opened 2024-12-02
OCPBUGS-23121: Restart operator pod on malformed RBAC

jrvaldes:
This PR triggers a restart after deleting malformed ClusterRoleBinding, so that the new instance of the operator pod reconciles with the newly created resources.

### DIFF
--- a/controllers/configmap_controller.go
+++ b/controllers/configmap_controller.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
+	"os"
 	"reflect"
 	"strings"
 
@@ -651,6 +652,9 @@ func (r *ConfigMapReconciler) ensureWICDClusterRoleBinding(ctx context.Context) 
 		}
 		r.log.Info("Deleted malformed resource", "ClusterRoleBinding", existingCRB.Name,
 			"RoleRef", existingCRB.RoleRef.Name, "Subjects", existingCRB.Subjects)
+
+		r.log.Info("Process will restart to reconcile resources")
+		defer os.Exit(1)
 	}
 	// create proper resource if it does not exist
 	_, err = r.k8sclientset.RbacV1().ClusterRoleBindings().Create(ctx, expectedCRB, meta.CreateOptions{})


jrvaldes:
/test aws-e2e-operator

to gather another data point

controllers/configmap_controller.go
```diff
@@ -71,8 +72,8 @@ const (
 	UsernameAnnotation = "windowsmachineconfig.openshift.io/username"
 	// ConfigMapController is the name of this controller in logs and other outputs.
 	ConfigMapController = "configmap"
-	// wicdRBACResourceName is the name of the resources associated with WICD's RBAC permissions
-	wicdRBACResourceName = "windows-instance-config-daemon"
+	// WicdRBACResourceName is the name of the resources associated with WICD's RBAC permissions
+	WicdRBACResourceName = "windows-instance-config-daemon"
```
sebsoto, line 76: Wicd -> WICD

hack/run-ci-e2e-test.sh
```diff
@@ -110,6 +110,16 @@ if  ! oc get deploy/windows-machine-config-operator -n $WMCO_DEPLOY_NAMESPACE >
...
+
+  # OCPBUGS-23121: For AWS, test the operator installed in a different namespace after a successful deployment in the
+  # provided namespace
+  platform="$(oc get infrastructure cluster -ojsonpath={.spec.platformSpec.type})"
+  if [[ "$platform" = "AWS" ]]; then
+    cleanup_WMCO $OSDK
+    export WMCO_DEPLOY_NAMESPACE="${WMCO_DEPLOY_NAMESPACE}-customized"
+    run_WMCO $OSDK
+    get_WMCO_logs
+  fi
```
sebsoto, line 122: Do we really need to add this test?
Our test suite is so complicated already and this is a true edge case

jrvaldes: I had mixed feelings about the value of adding e2e coverage for this corner case. So, I proposed the test for one platform only (AWS).  

Happy to remove it from the PR if that is the consensus. @openshift/openshift-team-windows-containers please chime in.

mansikulkarni96: I am fine with removing it.

jrvaldes: Removed the test commits.

PR #2605, opened 2025-01-09
OCPBUGS-45027: No Events Generated after changing cluster-monitoring Label

mansikulkarni96:
This commit fixes the controller to generate events when the cluster-monitoring label is not present or removed from the operator namespace.

### DIFF
--- a/controllers/metric_controller.go
+++ b/controllers/metric_controller.go
@@ -36,11 +36,14 @@ import (
 const (
 	// MetricController is the name of this controller in logs and other outputs.
 	MetricController = "metrics"
+	// monitoringLabel is the label added to the watch namespace to indicate cluster monitoring is enabled.
+	monitoringLabel = "openshift.io/cluster-monitoring"
 )
 
 type metricReconciler struct {
 	*monclient.MonitoringV1Client
 	instanceReconciler
+	monitoringEnabled bool
 }
 
 func NewMetricReconciler(mgr manager.Manager, clusterConfig cluster.Config, cfg *rest.Config,
@@ -70,7 +73,6 @@ func NewMetricReconciler(mgr manager.Manager, clusterConfig cluster.Config, cfg 
 // Node object and aims to move the current state of the cluster closer to the desired state.
 func (r *metricReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, err error) {
 	r.log = r.log.WithValues(MetricController, req.NamespacedName)
-
 	// validate if cluster monitoring is enabled in the operator namespace
 	enabled, err := r.validate(ctx)
 	if err != nil {
@@ -89,25 +91,31 @@ func (r *metricReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
 // validate will verify if cluster monitoring is enabled in the operator namespace. If the label is set to false or not
 // present, it will log and send warning events to the user. If the label holds a non-boolean value, returns an error.
 func (r *metricReconciler) validate(ctx context.Context) (bool, error) {
+	labelBool := false
 	// validate if metrics label is added to the operator namespace
-	labelValue := false
 	wmcoNamespace, err := r.k8sclientset.CoreV1().Namespaces().Get(ctx, r.watchNamespace, metav1.GetOptions{})
 	if err != nil {
 		return false, fmt.Errorf("error getting operator namespace: %w", err)
 	}
-	// if the label exists, update value from default of false
-	if value, ok := wmcoNamespace.Labels["openshift.io/cluster-monitoring"]; ok {
-		labelValue, err = strconv.ParseBool(value)
+	if value, ok := wmcoNamespace.Labels[monitoringLabel]; ok {
+		labelBool, err = strconv.ParseBool(value)
 		if err != nil {
 			return false, fmt.Errorf("monitoring label must have a boolean value: %w", err)
 		}
 	}
-	if !labelValue {
-		r.recorder.Eventf(wmcoNamespace, v1.EventTypeWarning, "labelValidationFailed",
-			"Cluster monitoring openshift.io/cluster-monitoring=true label is not enabled in %s namespace",
-			r.watchNamespace)
+	if r.monitoringEnabled != labelBool && labelBool {
+		// If the label exists, log the updated value
+		r.recorder.Eventf(wmcoNamespace, v1.EventTypeNormal, "monitoringEnabled",
+			"Cluster monitoring %s label is enabled in %s namespace",
+			monitoringLabel, r.watchNamespace)
+	} else if r.monitoringEnabled != labelBool && !labelBool {
+		// If the label is removed, log the removal
+		r.recorder.Eventf(wmcoNamespace, v1.EventTypeWarning, "monitoringDisabled",
+			"Cluster monitoring %s label is not enabled in %s namespace",
+			monitoringLabel, r.watchNamespace)
 	}
-	return labelValue, nil
+	r.monitoringEnabled = labelBool
+	return r.monitoringEnabled, nil
 }
 
 // ensureServiceMonitor creates a serviceMonitor object in the operator namespace if it does not exist.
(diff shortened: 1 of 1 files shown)


controllers/metric_controller.go
```diff
@@ -232,13 +232,23 @@ func (r *metricReconciler) mapToWatchNamespace(_ context.Context, obj client.Obj
...
-			return isMonitoringEnabled(e.ObjectNew, r.watchNamespace)
+			oldNamespace, okOld := e.ObjectOld.(*v1.Namespace)
+			newNamespace, okNew := e.ObjectNew.(*v1.Namespace)
+			if !okOld || !okNew {
+				return false
+			}
+			if oldNamespace.GetName() != r.watchNamespace && newNamespace.GetName() != r.watchNamespace {
+				return false
+			}
+			oldValue := oldNamespace.Labels["openshift.io/cluster-monitoring"]
```
jrvaldes, line 246: consider adding a const for `openshift.io/cluster-monitoring`, and reuse it

jrvaldes (approved):
@mansikulkarni96  thanks for fixing this. 

Do you anticipate adding e2e test coverage for this scenario?

mansikulkarni96:
> @mansikulkarni96 thanks for fixing this.
> 
> Do you anticipate adding e2e test coverage for this scenario?
I don't lean towards it, as adding an e2e for an event log message may not be useful . But I am open to it and need inputs from the team.

controllers/metric_controller.go
```diff
@@ -250,21 +260,11 @@ func (r *metricReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+// isOperatorNamespace returns true if the given object is the operator namespace
+func isOperatorNamespace(obj runtime.Object, watchNamespace string) bool {
 	namespace, ok := obj.(*v1.Namespace)
 	if !ok {
 		return false
 	}
-	if namespace.GetName() != watchNamespace {
-		return false
-	}
-	if value, ok := namespace.Labels["openshift.io/cluster-monitoring"]; ok {
```
sebsoto, line 262: I'm not following why the label check is no longer being made

mansikulkarni96: @sebsoto label check is being made in the reconciler, so we can generate events if the label has not been added to the namespace.

sebsoto: I guess I am not understanding the issue that is being solved.
It seems like the underlying bug is complaining about `event` generation indicating that monitoring is being disabled/enabled.

sebsoto: So in possibly in addition to this work, you would want to keep track of the current state, and if it changes, generate an event indicating it has changed. Right now an event is generated indicating it is not enabled.
https://github.com/openshift/windows-machine-config-operator/blob/641adab042a1c97b18832d3fb02da176e60b91f4/controllers/metric_controller.go#L93-L110


The followup question is, is this something that we want to do. I see some value in it.

mansikulkarni96: Correct that's the bug, this event was generated before the controller changes went in. Agreed, I see value in generating an event when there is a change.

controllers/metric_controller.go
```diff
@@ -96,16 +97,22 @@ func (r *metricReconciler) validate(ctx context.Context) (bool, error) {
...
+	if labelValue {
+		// If the label exists, log the updated value
+		r.recorder.Eventf(wmcoNamespace, v1.EventTypeNormal, "monitoringEnabled",
+			"Cluster monitoring %s label is enabled in %s namespace",
+			monitoringLabel, r.watchNamespace)
+	} else {
+		// If the label is removed, log the removal
+		r.recorder.Eventf(wmcoNamespace, v1.EventTypeWarning, "monitoringDisabled",
+			"Cluster monitoring %s label is not enabled in %s namespace",
+			monitoringLabel, r.watchNamespace)
```
sebsoto, line 115: This is fine, but wouldnt keeping track of the previous state in the metricReconciler object make more sense? Generating an event if the prev state != new state.
This would be more defensive, that way this function doesn't completely rely on the predicate to not generate unnecessary events

jrvaldes:
LGTM, will defer to @sebsoto  for the final review

controllers/metric_controller.go
```diff
@@ -36,11 +37,14 @@ import (
...
 	// MetricController is the name of this controller in logs and other outputs.
 	MetricController = "metrics"
+	// monitoringLabel is the label added to the watch namespace to indicate cluster monitoring is enabled.
+	monitoringLabel = "openshift.io/cluster-monitoring"
 )
 
 type metricReconciler struct {
 	*monclient.MonitoringV1Client
 	instanceReconciler
+	prevState map[string]string
```
sebsoto, line 47: Can't this just be `monitoringEnabled bool`?
The namespace really shouldn't be changing, that requires removing and reinstalling the operator into a different namespace.
Having this be a map is adding a bit of unneeded complexity.

controllers/metric_controller.go
```diff
@@ -70,9 +74,30 @@ func NewMetricReconciler(mgr manager.Manager, clusterConfig cluster.Config, cfg
...
+	// Get the current monitoring label state
+	currentLabelValue := namespace.Labels[monitoringLabel]
+	prevLabelValue, exists := r.prevState[namespace.Name]
+
+	// Skip processing if there's no change
+	if exists && currentLabelValue == prevLabelValue {
+		r.log.Info("No change", "namespace", namespace.Name)
+		return ctrl.Result{}, nil
+	}
+	r.prevState[namespace.Name] = currentLabelValue
```
sebsoto, line 98: This is duplicating logic in r.validate, which checks if the namespace has the label or not

controllers/metric_controller.go
```diff
@@ -88,24 +113,27 @@ func (r *metricReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
...
-			r.watchNamespace)
+	if labelValue {
+		// If the label exists, log the updated value
+		r.recorder.Eventf(wmcoNamespace, v1.EventTypeNormal, "monitoringEnabled",
+			"Cluster monitoring %s label is enabled in %s namespace",
+			monitoringLabel, r.watchNamespace)
+	} else {
+		// If the label is removed, log the removal
+		r.recorder.Eventf(wmcoNamespace, v1.EventTypeWarning, "monitoringDisabled",
+			"Cluster monitoring %s label is not enabled in %s namespace",
```
sebsoto, line 135: labelBool := false
	if value, ok := wmcoNamespace.Labels[monitoringLabel];  ok {
		labelBool, err = strconv.ParseBool(value)
		if err != nil {
			return false, fmt.Errorf("monitoring label must have a boolean value: %w", err)
		}
	}
	if r.monitoringEnabled != labelBool && labelBool {
		// If the label exists, log the updated value
		r.recorder.Eventf(wmcoNamespace, v1.EventTypeNormal, "monitoringEnabled",
			"Cluster monitoring %s label is enabled in %s namespace",
			monitoringLabel, r.watchNamespace)
	} else if r.monitoringEnabled != labelBool && !labelBool {
		// If the label is removed, log the removal
		r.recorder.Eventf(wmcoNamespace, v1.EventTypeWarning, "monitoringDisabled",
			"Cluster monitoring %s label is not enabled in %s namespace",		
	}
	r.monitoringEnabled = labelBool
	return r.monitoringEnabled, nil

sebsoto: Example as to how you could achieve this without the extra code in NewMetricReconciler

sebsoto (changes requested):

sebsoto (approved):

mansikulkarni96:
/override ci/prow/azure-e2e-operator  ci/prow/nutanix-e2e-operator

mansikulkarni96:
/override ci/prow/vsphere-disconnected-e2e-operator ci/prow//test azure-e2e-operator

mansikulkarni96:
/override ci/prow/vsphere-disconnected-e2e-operator ci/prow/azure-e2e-operator

mansikulkarni96:
/override "Red Hat Konflux / windows-machine-config-operator-staging-master-ec / windows-machine-config-operator-master"

mansikulkarni96:
/override "Red Hat Konflux / windows-machine-config-operator-staging-master-ec / windows-machine-config-operator-master"

PR #2643, opened 2025-01-23
Run with GOEXPERIMENT var strictfipsruntime

mansikulkarni96:
This PR sets the value of GOEXPERIMENT variable in Containerfile to strictfipsruntime.
This change is required to build fips compliant images and is in line with rest of the container files.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -10,6 +10,8 @@ ENV GO_COMPLIANCE_DEBUG=0
 # downloading the latest version
 ENV GOTOOLCHAIN=local
 
+ENV GOEXPERIMENT=strictfipsruntime
+
 WORKDIR /build/windows-machine-config-operator/
 COPY .git .git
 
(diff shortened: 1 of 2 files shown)


jrvaldes (approved):
/lgtm

Containerfile
```diff
@@ -10,6 +10,8 @@ ENV GO_COMPLIANCE_DEBUG=0
 # downloading the latest version
 ENV GOTOOLCHAIN=local
 
+ENV GOEXPERIMENT=strictfipsruntime
```
jrvaldes, line 13: @mansikulkarni96  Isn't the  `Containerfile`  used only by konflux? if so, consider overriding the e2e test jobs.

mansikulkarni96: yes, the konflux test isn't passing. Looking into it.

jrvaldes:
> /retest

@sebsoto   does this command works for `Red Hat Konflux` jobs too?

PR #2689, opened 2025-02-05
WINC-1362: Update fips task ref

sebsoto:

PR #2710, opened 2025-02-13
Include tag in fips check tekton bundle image ref

simonbaird:
Without the tag in the image ref, the Conforma lookup fails and the task is considered untrusted.

See https://issues.redhat.com/browse/EC-712 for more details about the root case.

This should fix "Code tampering violation" problem described in https://issues.redhat.com/browse/KFLUXSPRT-1798


simonbaird:
The change is included in #2689. Abandoning this one.

PR #2733, opened 2025-02-25
OCPBUGS-1721: adds retries to hns endpoint creation

wgahnagl:
solves ENDPOINT_IP: error running command with output New-HnsEndpoint

### DIFF
--- a/pkg/nodeconfig/payload/payload.go
+++ b/pkg/nodeconfig/payload/payload.go
@@ -110,6 +110,34 @@ param(
     }
   }
 
+# This retries getting the HNS-network until it succeeds
+function Retry-GetHnsNetwork {
+     $retryCount = 20
+     $retryDelaySeconds = 1
+
+     $attempt = 1
+ 
+     while ($attempt -le $retryCount) {
+         try {
+             $hns_network = Get-HnsNetwork | Where-Object { $_.Name -eq 'HNS_NETWORK' }
+             
+             # Check if hns_network is null
+             if ($null -eq $hns_network) {
+                 Write-Host "Attempt $attempt returned null. Retrying in $retryDelaySeconds seconds..."
+             } else {
+                 Write-Host "Found HNS_NETWORK on attempt $attempt"
+                 return $hns_network 
+             }
+         } catch {
+             Write-Host "Attempt $attempt failed: $_"
+         }
+         Start-Sleep -Seconds $retryDelaySeconds
+         $attempt++
+     }
+     Write-Host "Max retry attempts reached."
+     return $null
+}
+
 $ErrorActionPreference = "Stop"
 Import-Module -DisableNameChecking HNS_MODULE_PATH
 
@@ -165,8 +193,13 @@ $cni_template=@'
 }
 '@
 
+$hns_network = Retry-GetHnsNetwork
+# If the HNS network is never found, quit.
+if ($null -eq $hns_network) {
+  throw "cannot find HNS network with name HNS_NETWORK"
+}
+
 # Generate CNI Config
-$hns_network=Get-HnsNetwork  | where { $_.Name -eq 'HNS_NETWORK'}
 $subnet=$hns_network.Subnets.AddressPrefix
 $cni_template=$cni_template.Replace("ovn_host_subnet",$subnet)
 $provider_address=$hns_network.ManagementIP
--- a/pkg/nodeconfig/payload/payload_test.go
+++ b/pkg/nodeconfig/payload/payload_test.go
@@ -38,6 +38,34 @@ param(
     }
   }
 
+# This retries getting the HNS-network until it succeeds
+function Retry-GetHnsNetwork {
+     $retryCount = 20
+     $retryDelaySeconds = 1
+
+     $attempt = 1
+ 
+     while ($attempt -le $retryCount) {
+         try {
+             $hns_network = Get-HnsNetwork | Where-Object { $_.Name -eq 'OVNKubernetesHNSNetwork' }
+             
+             # Check if hns_network is null
+             if ($null -eq $hns_network) {
+                 Write-Host "Attempt $attempt returned null. Retrying in $retryDelaySeconds seconds..."
+             } else {
+                 Write-Host "Found OVNKubernetesHNSNetwork on attempt $attempt"
+                 return $hns_network 
+             }
+         } catch {
+             Write-Host "Attempt $attempt failed: $_"
+         }
+         Start-Sleep -Seconds $retryDelaySeconds
+         $attempt++
+     }
+     Write-Host "Max retry attempts reached."
+     return $null
+}
+
 $ErrorActionPreference = "Stop"
 Import-Module -DisableNameChecking c:\k\hns.psm1
 
@@ -93,8 +121,13 @@ $cni_template=@'
 }
 '@
 
+$hns_network = Retry-GetHnsNetwork
+# If the HNS network is never found, quit.
+if ($null -eq $hns_network) {
+  throw "cannot find HNS network with name OVNKubernetesHNSNetwork"
+}
+
 # Generate CNI Config
-$hns_network=Get-HnsNetwork  | where { $_.Name -eq 'OVNKubernetesHNSNetwork'}
 $subnet=$hns_network.Subnets.AddressPrefix
 $cni_template=$cni_template.Replace("ovn_host_subnet",$subnet)
 $provider_address=$hns_network.ManagementIP


pkg/nodeconfig/payload/payload.go
```diff
@@ -175,7 +175,34 @@ $cni_template=$cni_template.Replace("provider_address",$provider_address)
...
+$retryDelay = 10
+$attempt = 0
+
+while ($attempt -lt $retryCount) {
+  try {
+    $hns_network=Get-HnsNetwork  | where { $_.Name -eq 'HNS_NETWORK'}
+    $endpoint = Invoke-HNSRequest GET endpoints | where { $_.Name -eq 'VIPEndpoint'}
+
+    # Check if endpoint is null
+    if ($null -eq $endpoint) {
```
sebsoto, line 189: Im confused as to why this is being done, especially given the code below,
```
if( $endpoint -eq $null) {
    $endpoint = New-HnsEndpoint -NetworkId $hns_network.ID -Name "VIPEndpoint"
    Attach-HNSHostEndpoint -EndpointID $endpoint.ID -CompartmentID 1
}
```
And the comment above `# Create HNS endpoint if it doesn't exist`

sebsoto: Shouldn't you be checking if hns_network is null?

sebsoto: Please make this a function after you have corrected the behavior.

sebsoto: Specifically the function would be to encapsulate the retry logic, returning the object we care about so the script is more easily readable.

pkg/nodeconfig/payload/payload.go
```diff
@@ -110,6 +110,39 @@ param(
...
+             } else {
+                 Write-Host "Found HNS_NETWORK on attempt $($attempt + 1)"
+                 return $hns_network 
+             }
+         } catch {
+             Write-Host "Attempt $($attempt + 1) failed: $_"
+             if ($attempt -eq ($retryCount - 1)) {
+                 Write-Host "Max retry attempts reached. Continuing."
+                 exit 1
+             }
```
sebsoto, line 138: Please move this outside the while
The while will complete at attempt == retry and it should execute this, without needing another comparison.

sebsoto: exit 1 doesn't seem correct here.
You should return from within a function and let the caller handle the error, rather than exiting the process from within the function.

pkg/nodeconfig/payload/payload.go
```diff
@@ -165,8 +198,14 @@ $cni_template=@'
 }
 '@
 
+$hns_network = Retry-GetHnsNetwork
+# If the HNS network is never found, quit.
+if ($null -eq $hns_network) {
+  Write-Host "HNS_NETWORK not found after retries."
+  exit 1
```
sebsoto, line 205: Seems like [throw](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_throw?view=powershell-7.5) should be used rather than Write-Host + exit

sebsoto (changes requested):
Thanks for making this a function. Changes look good, just a few comments.

pkg/nodeconfig/payload/payload.go
```diff
@@ -110,6 +110,36 @@ param(
...
+                 Write-Host "Found HNS_NETWORK on attempt $($attempt + 1)"
+                 return $hns_network 
+             }
+         } catch {
+             Write-Host "Attempt $($attempt + 1) failed: $_"
+         }
+         Start-Sleep -Seconds $retryDelay
+         $attempt++
+     }
+     Write-Host "Max retry attempts reached. Continuing."
```
sebsoto, line 139: I feel dropping `Continuing` would be apt here
In case of a timeout, in the logs you would see `Continuing` and then the fatal error, which is not really continuing :laughing:

sebsoto (approved):
Approved, but please address the comment I left

pkg/nodeconfig/payload/payload.go
wgahnagl: FAIR, LOL

pkg/nodeconfig/payload/payload.go
```diff
@@ -110,6 +110,36 @@ param(
     }
   }
 
+# This retries getting the HNS-network until it succeeds
+function Retry-GetHnsNetwork {
+     param (
+         [int]$retryCount = 20,
+         [int]$retryDelay = 10
```
jrvaldes, line 117: s/retryDelay/retryDelaySeconds

so that it is clear the time unit

pkg/nodeconfig/payload/payload.go
```diff
@@ -110,6 +110,36 @@ param(
     }
   }
 
+# This retries getting the HNS-network until it succeeds
+function Retry-GetHnsNetwork {
+     param (
```
jrvaldes, line 115: >        [int]$retryCount = 20,
>         [int]$retryDelay = 10

Do you anticipate a case where the input params are needed? Otherwise, consider defining just local variables instead of params.

pkg/nodeconfig/payload/payload.go
```diff
@@ -110,6 +110,36 @@ param(
...
+ 
+     $attempt = 0
+ 
+     while ($attempt -lt $retryCount) {
+         try {
+             $hns_network = Get-HnsNetwork | Where-Object { $_.Name -eq 'HNS_NETWORK' }
+             
+             # Check if hns_network is null
+             if ($null -eq $hns_network) {
+                 Write-Host "Attempt $($attempt + 1) returned null. Retrying..."
```
jrvaldes, line 128:
```suggestion
     $attempt = 1
 
     while ($attempt -le $retryCount) {
         try {
             $hns_network = Get-HnsNetwork | Where-Object { $_.Name -eq 'HNS_NETWORK' }
             
             # Check if hns_network is null
             if ($null -eq $hns_network) {
                 Write-Host "Attempt $attempt returned null. Retrying..."
```

Consider starting the counter at `1` to avoid the arithmetic operation on each iteration.

jrvaldes, line 128: >                  Write-Host "Attempt $attempt: error getting HNS Network. Retrying in $retryDelay seconds..."

Consider rewording the message to capture the failed operation and providing feedback on when the next attempt will takes place

pkg/nodeconfig/payload/payload.go
```diff
@@ -110,6 +110,36 @@ param(
...
+             } else {
+                 Write-Host "Found HNS_NETWORK on attempt $($attempt + 1)"
+                 return $hns_network
```
jrvaldes, line 131: in your tests, how many attempts does it take?

pkg/nodeconfig/payload/payload.go
```diff
@@ -110,6 +110,36 @@ param(
...
+             if ($null -eq $hns_network) {
+                 Write-Host "Attempt $($attempt + 1) returned null. Retrying..."
+             } else {
+                 Write-Host "Found HNS_NETWORK on attempt $($attempt + 1)"
+                 return $hns_network 
+             }
+         } catch {
+             Write-Host "Attempt $($attempt + 1) failed: $_"
+         }
+         Start-Sleep -Seconds $retryDelay
```
jrvaldes, line 136: Please add to the commit message why `20 seconds` is the recommended value? Does lowering the time result in faster performance?

pkg/nodeconfig/payload/payload.go
```diff
@@ -165,8 +195,13 @@ $cni_template=@'
 }
 '@
 
+$hns_network = Retry-GetHnsNetwork
+# If the HNS network is never found, quit.
+if ($null -eq $hns_network) {
+  throw "HNS_NETWORK not found after retries."
```
jrvaldes, line 201:
```suggestion
  throw "cannot find HNS_NETWORK with name $hns_network"
```

The error should be a concluding statement and provide supporting information for debugging.

jrvaldes (changes requested):
Mostly LGTM, please take a look at my comments

pkg/nodeconfig/payload/payload.go
jrvaldes, line 128: @wgahnagl  Does this message appear in the WICD log file?

pkg/nodeconfig/payload/payload.go
wgahnagl: if I use $hns_network there, it would always be null. Maybe return the error from the function?

pkg/nodeconfig/payload/payload.go
wgahnagl: I was thinking it might be good to have if we want to change them later, for testing or something? IDK.

pkg/nodeconfig/payload/payload.go
wgahnagl: It appears in the log, but only if it errors. I think it's a good idea to keep because you'll only really be thinking about this file when you're debugging.

wgahnagl: the failure information is captured on the catch. Nothing is really failing here, so there's nothing to capture.

pkg/nodeconfig/payload/payload.go
jrvaldes: https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it

pkg/nodeconfig/payload/payload.go
jrvaldes: by "failed operation " I meant the `null` response.

pkg/nodeconfig/payload/payload.go
jrvaldes: >  error from the function

+1

pkg/nodeconfig/payload/payload.go
jrvaldes: agree

pkg/nodeconfig/payload/payload.go
```diff
@@ -110,6 +110,36 @@ param(
...
+ 
+     $attempt = 1
+ 
+     while ($attempt -le $retryCount) {
+         try {
+             $hns_network = Get-HnsNetwork | Where-Object { $_.Name -eq 'HNS_NETWORK' }
+             
+             # Check if hns_network is null
+             if ($null -eq $hns_network) {
+                 Write-Host "Attempt $($attempt + 1) returned null. Retrying in $retryDelaySeconds seconds..."
```
jrvaldes, line 128: Not following here, why keeping the arithmetic operation on each iteration?

pkg/nodeconfig/payload/payload.go
jrvaldes: WRT to the message content, we can use:

```
cannot find HNS network with name OVNKubernetesHNSNetwork
```

pkg/nodeconfig/payload/payload.go
```diff
@@ -110,6 +110,34 @@ param(
...
+ 
+     while ($attempt -le $retryCount) {
+         try {
+             $hns_network = Get-HnsNetwork | Where-Object { $_.Name -eq 'HNS_NETWORK' }
+             
+             # Check if hns_network is null
+             if ($null -eq $hns_network) {
+                 Write-Host "Attempt $attempt returned null. Retrying in $retryDelaySeconds seconds..."
+             } else {
+                 Write-Host "Found HNS_NETWORK on attempt $($attempt + 1)"
```
jrvaldes, line 128: @wgahnagl   I think you can now remove the `$($attempt + 1)` , otherwise the attempt counter will start at `2` = `(1+1)`

pkg/nodeconfig/payload/payload.go
jrvaldes: See https://github.com/openshift/windows-machine-config-operator/pull/2733/commits/da87782443049b253879f43331ef53787eee139d#r1983512639

jrvaldes (approved):
Good work on this @wgahnagl  LGTM

jrvaldes:
[vsphere-e2e-operator](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_windows-machine-config-operator/2733/pull-ci-openshift-windows-machine-config-operator-master-vsphere-e2e-operator/1897768004821716992) — Job succeeded 

WICD logs does not shows the error

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_windows-machine-config-operator/2733/pull-ci-openshift-windows-machine-config-operator-master-vsphere-e2e-operator/1897768004821716992/artifacts/vsphere-e2e-operator/windows-e2e-operator-test/artifacts/nodes/e2e-wm-rdlrs/wicd/windows-instance-config-daemon.exe.INFO

pkg/nodeconfig/payload/payload.go
sebsoto: This is incorrect. Please revert. `OVNKubernetesHNSNetwork` should not be hardcoded.

The original value `throw "HNS_NETWORK not found after retries."`
Evaluates to `OVNKubernetesHNSNetwork not found after retries."`

This message is fine.

sebsoto:
/approve cancel
Please address

https://github.com/openshift/windows-machine-config-operator/pull/2733#discussion_r1986334973

sebsoto (approved):

mtnbikenc:
@wgahnagl I see you keep retesting.  What are the reasons for the failures?  Is there something more to investigate?  Specifically, the [full PR history](https://prow.ci.openshift.org/pr-history?org=openshift&repo=windows-machine-config-operator&pr=2733) shows that pull-ci-openshift-windows-machine-config-operator-master-vsphere-disconnected-e2e-operator has never passed for this PR.

wgahnagl:
> @wgahnagl I see you keep retesting. What are the reasons for the failures? Is there something more to investigate? Specifically, the [full PR history](https://prow.ci.openshift.org/pr-history?org=openshift&repo=windows-machine-config-operator&pr=2733) shows that pull-ci-openshift-windows-machine-config-operator-master-vsphere-disconnected-e2e-operator has never passed for this PR.

 error: unable to copy layer sha256:9d44b32aeb89ba236839c3c799cf86b2bb3a40fb5e8a82d155f9ae8ee7aba185 to ci-op-r1zqxz9f-69f44-bastion.vmc-ci.devcluster.openshift.com:5000/pipeline: Patch "https://ci-op-r1zqxz9f-69f44-bastion.vmc-ci.devcluster.openshift.com:5000/v2/pipeline/blobs/uploads/c0f07005-24f4-4690-881d-e4c6c0733ecd?_state=4Nna8NoiaORecgbtasvofRdDcZTQyjpp-EFXmuoSxE17Ik5hbWUiOiJwaXBlbGluZSIsIlVVSUQiOiJjMGYwNzAwNS0yNGY0LTQ2OTAtODgxZC1lNGM2YzA3MzNlY2QiLCJPZmZzZXQiOjAsIlN0YXJ0ZWRBdCI6IjIwMjUtMDMtMTdUMjE6MzM6MTQuNDcxMzA4ODgxWiJ9": unexpected EOF
error: unable to copy layer sha256:65442b90c1456946b65a513148c3e504c3e1ae3889d427d5b4526ba9cec3ea0e to ci-op-r1zqxz9f-69f44-bastion.vmc-ci.devcluster.openshift.com:5000/pipeline: Patch "https://ci-op-r1zqxz9f-69f44-bastion.vmc-ci.devcluster.openshift.com:5000/v2/pipeline/blobs/uploads/86646c2f-7107-4428-89b8-22325ae71fd2?_state=4CnLUFHvEEvXhKONx6ZX1pAExsAC7c76isqq9N11Nbt7Ik5hbWUiOiJwaXBlbGluZSIsIlVVSUQiOiI4NjY0NmMyZi03MTA3LTQ0MjgtODliOC0yMjMyNWFlNzFmZDIiLCJPZmZzZXQiOjAsIlN0YXJ0ZWRBdCI6IjIwMjUtMDMtMTdUMjE6MzM6MTcuMjQyOTAxNTU1WiJ9": unexpected EOF
error: unable to copy layer sha256:ff546100e8c6594552faf4f514daac52d5ac97d211e811ad2cd9f6c3a39a46ec to ci-op-r1zqxz9f-69f44-bastion.vmc-ci.devcluster.openshift.com:5000/pipeline: Patch "https://ci-op-r1zqxz9f-69f44-bastion.vmc-ci.devcluster.openshift.com:5000/v2/pipeline/blobs/uploads/97a1ffa6-837e-4702-8a46-871ac7c87428?_state=aF1pNiK7UMdQrRSLkbx2NllVsysMxipl13xreVkkjj57Ik5hbWUiOiJwaXBlbGluZSIsIlVVSUQiOiI5N2ExZmZhNi04MzdlLTQ3MDItOGE0Ni04NzFhYzdjODc0MjgiLCJPZmZzZXQiOjAsIlN0YXJ0ZWRBdCI6IjIwMjUtMDMtMTdUMjE6MzM6MjMuNDcwNDg0NjM3WiJ9": unexpected EOF
error: unable to copy layer sha256:2bc423a1170190fe7338036610495da9f090e5608bb228b05068053d0f481aff to ci-op-r1zqxz9f-69f44-bastion.vmc-ci.devcluster.openshift.com:5000/pipeline: Patch "https://ci-op-r1zqxz9f-69f44-bastion.vmc-ci.devcluster.openshift.com:5000/v2/pipeline/blobs/uploads/3f4c11bc-55d6-4645-a38b-61f3554f28a7?_state=F0ii2NsJLogvfPvkZ-fau6wbWBDedauV095j6rKJZ3F7Ik5hbWUiOiJwaXBlbGluZSIsIlVVSUQiOiIzZjRjMTFiYy01NWQ2LTQ2NDUtYTM4Yi02MWYzNTU0ZjI4YTciLCJPZmZzZXQiOjAsIlN0YXJ0ZWRBdCI6IjIwMjUtMDMtMTdUMjE6MzM6MjcuMzgyNzkxMDY3WiJ9": unexpected EOF
error: unable to copy layer sha256:684f7cdc119a9948f01f852b14b9ca087e8ee89e732da10cddc0ff7b373f6cac to ci-op-r1zqxz9f-69f44-bastion.vmc-ci.devcluster.openshift.com:5000/pipeline: Patch "https://ci-op-r1zqxz9f-69f44-bastion.vmc-ci.devcluster.openshift.com:5000/v2/pipeline/blobs/uploads/a6ce28a9-449c-4b28-a82b-7b6c9d2afb14?_state=DAKjN0-nCtSVcM5cbWrFhXvR3k9qn9Ja__e2xraM4lN7Ik5hbWUiOiJwaXBlbGluZSIsIlVVSUQiOiJhNmNlMjhhOS00NDljLTRiMjgtYTgyYi03YjZjOWQyYWZiMTQiLCJPZmZzZXQiOjAsIlN0YXJ0ZWRBdCI6IjIwMjUtMDMtMTdUMjE6MzM6MzMuMDM3MzYxNzAxWiJ9": unexpected EOF
info: Mirroring completed in 48.32s (231.2MB/s)
error: one or more errors occurred while uploading images 

it's failing on the mirroring step

mtnbikenc:
@wgahnagl What needs to be done to re-run the Konflux job?

sebsoto:
/override "ci/prow/aws-e2e-operator"

Known issue with OCP CI, test passed before.

sebsoto:
/override ci/prow/vsphere-disconnected-e2e-operator ci/prow/gcp-e2e-operator ci/prow/aws-e2e-operator

PR #2743, opened 2025-02-28
WIP: Collect packet tracing on network_test failures

jrvaldes:
Moved the stopPacketTrace calls into a defer function to guarantee collection when tests fail or panic, improving test reliability.

### DIFF
--- a/test/e2e/network_test.go
+++ b/test/e2e/network_test.go
@@ -44,14 +44,15 @@ func testNetwork(t *testing.T) {
 	for _, node := range gc.allNodes() {
 		require.NoError(t, tc.startPacketTrace(&node))
 	}
+	defer func() {
+		for _, node := range gc.allNodes() {
+			require.NoError(t, tc.stopPacketTrace(&node))
+		}
+	}()
 
 	t.Run("East West Networking", tc.testEastWestNetworking)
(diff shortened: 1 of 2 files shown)


test/e2e/network_test.go
```diff
@@ -44,14 +44,15 @@ func testNetwork(t *testing.T) {
 	for _, node := range gc.allNodes() {
 		require.NoError(t, tc.startPacketTrace(&node))
 	}
+	defer func() {
+		for _, node := range gc.allNodes() {
+			require.NoError(t, tc.stopPacketTrace(&node))
+		}
+	}()
```
sebsoto, line 51: This doesn't actually do what is described in the PR. It is already executing on failure. Test failures in the `t.Run()`'s do not return from the testNetwork function.

See [example](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_windows-machine-config-operator/2730/pull-ci-openshift-windows-machine-config-operator-master-nutanix-e2e-operator/1895563510717353984) where the east/west networking failed. The rest of  the function executed.

I changing this to a defer is fine with me, but it is only helpful to make the function less fragile, so ordering doesn't matter if another test is added that should be traced.

Does a require in `defer` work properly? The packet trace should be a requirement for the networking test to succeed

PR #2796, opened 2025-04-04
[WINC-1380] increases resource limits to avoid OOMkilling

wgahnagl:
increasing our resource request to 10gi to avoid OOMkill. Should probably be lowered. 
based on the docs here 
https://konflux.pages.redhat.com/docs/users/building/overriding-compute-resources.html

### DIFF
--- a/.tekton/windows-machine-config-operator-bundle-master-pull-request.yaml
+++ b/.tekton/windows-machine-config-operator-bundle-master-pull-request.yaml
@@ -35,6 +35,15 @@ spec:
     value: quay.io/redhat-user-workloads/windows-machine-conf-tenant/windows-machine-config-operator/windows-machine-config-operator-bundle-master:on-pr-{{revision}}
   - name: revision
     value: '{{revision}}'
+  taskRunSpecs:
+    - pipelineTaskName: build-source-image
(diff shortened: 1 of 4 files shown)


.tekton/windows-machine-config-operator-bundle-master-pull-request.yaml
```diff
@@ -35,6 +35,15 @@ spec:
     value: quay.io/redhat-user-workloads/windows-machine-conf-tenant/windows-machine-config-operator/windows-machine-config-operator-bundle-master:on-pr-{{revision}}
   - name: revision
     value: '{{revision}}'
+  taskRunSpecs:
+    - pipelineTaskName: build-container
```
sebsoto, line 39: Looks like this is bumping build-container.
Aren't we OOMing on `build-source-image`?

sebsoto: Do we need to increase both?

wgahnagl: I think you're right, that the source image should be the one that's getting bumped. Let's see how the test runs :+1:

PR #2797, opened 2025-04-04
WINC-1380 [release-4.18] - increases resource limits to avoid OOMkilling

wgahnagl:
increasing our resource request to 10gi to avoid OOMkill. Should probably be lowered.
based on the docs here
https://konflux.pages.redhat.com/docs/users/building/overriding-compute-resources.html


jrvaldes (approved):
/lgtm

PR #2811, opened 2025-04-17
WINC-1383: Increase memory for konflux tasks that OOM

sebsoto:
Memory needs to be bumped for all tasks that use SOURCE_ARTIFACT


sebsoto:
/override ci/prow/ci-bundle-wmco-bundle ci/prow/images
Not valid for .tekton changes

PR #2817, opened 2025-04-17
WINC-1383: Increase build-container task memory

sebsoto:
To prevent OOM


jrvaldes:
LGTM

jrvaldes:
> /override ci/prow/ci-bundle-wmco-bundle ci/prow/images

@sebsoto  can we exclude those in the skip-if-only-changed ?

sebsoto:
@jrvaldes 
I tried to, but wasn't able. Not sure if it is possible or not, would be a good change.

PR #2833, opened 2025-04-23
WINC-1383: Bump use-trusted-artifact step memory

sebsoto:
This step is the one in particular that is going OOM. Only it should be given the memory increase.


mansikulkarni96 (approved):

jrvaldes (approved):
LGTM

PR #2967, opened 2025-05-29
Replace dos2unix from Dockerfile.ci with jq

sebsoto:
dos2unix is not required for building, jq is used by the test process and should be explicitly required.


build/Dockerfile.ci
```diff
@@ -17,8 +17,7 @@ ENV GOTOOLCHAIN=local
 
 ENV GOEXPERIMENT=strictfipsruntime
 
-# dos2unix is needed to build CNI plugins
```
jrvaldes, line 20: is dos2unix no longer needed to build CNI plugins ?

sebsoto: At the least it hasnt been required in the konflux builds.

https://github.com/openshift/windows-machine-config-operator/blob/master/Containerfile

jrvaldes: I follow

build/Dockerfile.ci
```diff
@@ -17,8 +17,7 @@ ENV GOTOOLCHAIN=local
...
-RUN yum install -y dos2unix
+RUN yum install -y jq
```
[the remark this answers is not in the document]
sebsoto: Thanks for pointing that out.
It looks like cli is injected into the base image used. https://steps.ci.openshift.org/ci-operator-reference#line795

I think this is required still, but I moved the jq install to the final stage of the containerfile. Otherwise JQ will only be present for the building stage.

PR #2991, opened 2025-06-03
[release-4.18] submodule update 06 03

wgahnagl:
git push origin release-4.15-submodule-update-06-03


jrvaldes:
@wgahnagl  looks like there are missing deps, run build and unit test locally

PR #3030, opened 2025-06-10
bump  `sftp` to v1.13.9 and `x/crypto` to v0.39.0

jrvaldes:


mansikulkarni96:
@jrvaldes Changes look good, but both commits have overlapping vendor changes. I see some sftp changes in first commit and some x/crypto in the second one.

PR #3133, opened 2025-07-22
WIP WINC-1147

mansikulkarni96:


bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
```diff
@@ -385,6 +386,14 @@ spec:
           - patch
           - update
           - watch
+        - apiGroups:
+          - ""
+          resources:
+          - serviceaccounts
+          verbs:
+          - create
+          - delete
```
jrvaldes, line 395: I have mixed feelings about granting create and delete for serviceaccounts, an attacker could create or remove critical accounts, how to prevent being exploited?

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
```diff
@@ -393,6 +402,14 @@ spec:
           - create
           - delete
           - get
+        - apiGroups:
+          - rbac.authorization.k8s.io
+          resources:
+          - clusterroles
```
jrvaldes, line 408: same  create and delete for clusterroles

pkg/nodeconfig/nodeconfig.go
```diff
@@ -184,6 +186,17 @@ func (nc *nodeConfig) Configure(ctx context.Context) error {
 			}
 		}
 
+		// Ensure node-specific RBAC exists
+		if err := nc.ensureNodeSpecificRBAC(ctx); err != nil {
```
jrvaldes, line 190: During the creation and deletion of RBAC resources (per-node) there is a change to race conditions if not handled atomically. A node may be compromised before RBAC is updated and retain broader permissions. Is it worth adding a test for this case?

controllers/controllers.go
```diff
@@ -167,6 +168,12 @@ func (r *instanceReconciler) deconfigureInstance(ctx context.Context, node *core
 	if err = nc.Deconfigure(ctx); err != nil {
 		return err
 	}
+
+	// Clean up node-specific RBAC after WICD cleanup is complete, before node deletion
+	if err = wmcorbac.CleanupNodeSpecificRBAC(ctx, r.client, r.k8sclientset, r.watchNamespace, instance.Node.GetName()); err != nil {
```
jrvaldes, line 173: If a node is force-removed or crashes, RBAC resources may remain. The cleanup function attempts to fix this, but isn't fail-proof. Consider a mechanism to clean up orphaned RBAC resources.

pkg/rbac/node_rbac.go
```diff
@@ -0,0 +1,272 @@
...
+	if err := ensureNodeSpecificClusterRoleBinding(ctx, k8sClient, wmcoNamespace, nodeName); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// CleanupNodeSpecificRBAC removes node-specific RBAC resources for a deleted node
+func CleanupNodeSpecificRBAC(ctx context.Context, ctrlRuntimeClient client.Client, k8sClient kubernetes.Interface, wmcoNamespace, nodeName string) error {
+	// Clean up ServiceAccounts by label (handles both node-name and IP-based ServiceAccounts)
```
jrvaldes, line 42: If users can override node names, how  to check they do not allow label injection or resource name spoofing?

bundle/manifests/windows-machine-config-operator.clusterserviceversion.yaml
mansikulkarni96: Any suggestions on how to handle this better?
I could create a different role that is namespace scoped instead?
config/rbac/wmco_serviceaccount_role.yaml
kind: Role 
namespace: system
resources: ["serviceaccounts"]
verbs: ["create", "delete", "get", "list"]

mansikulkarni96: or add application level security where we make sure the request is coming from valid account
if !strings.HasPrefix(roleName, "windows-instance-config-daemon-") {
	return fmt.Errorf("security violation: attempted to delete ClusterRole with invalid name pattern: %s", roleName)
	}

jrvaldes:
>Webhook implementation in follow-up PR to address cross node access 

@mansikulkarni96, why the need for a follow-up PR? Is there any direct dependency or risk?

jrvaldes:
> > Webhook implementation in follow-up PR to address cross node access
> 
> @mansikulkarni96, why the need for a follow-up PR? Is there any direct dependency or risk?

disregard, I see https://github.com/openshift/windows-machine-config-operator/pull/3133#pullrequestreview-3137062600

bundle/manifests/system-wicd-nodes_rbac.authorization.k8s.io_v1_clusterrole.yaml
```diff
@@ -0,0 +1,42 @@
...
+- apiGroups:
+  - ""
+  resources:
+  - nodes
+  verbs:
+  - list
+- apiGroups:
+  - ""
+  resources:
+  - nodes
```
jrvaldes, line 27: looks like `nodes` resource was defined in L#21 and extended in L#27 that grants additional verbs (get, patch, update, watch) on both nodes and nodes/status.

sebsoto (changes requested):
Overall good work @mansikulkarni96,
I've focusing this review on the WICD portions, lets straighten those out a bit and once that's done I will look closer at the CSR controller in WMCO.

pkg/daemon/controller/controller.go
```diff
@@ -134,28 +135,68 @@ func (sc *ServiceController) Bootstrap(desiredVersion string) error {
 }
 
 // RunController is the entry point of WICD's controller functionality
-func RunController(ctx context.Context, watchNamespace, kubeconfig, caBundle string) error {
-	cfg, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
+func RunController(ctx context.Context, watchNamespace, kubeconfig, caBundle string, certDir, certDuration string) error {
+	var cfg *rest.Config
+	var err error
+
+	duration, err := time.ParseDuration(certDuration)
```
sebsoto, line 142: If there is no reason for this function to have a string representation of time, please do this when parsing the cmd line flag and pass in a time.Duration to this function instead.

pkg/daemon/certs/wicd_cert_manager.go
```diff
@@ -0,0 +1,145 @@
...
+		bootstrapKubeconfigPath: bootstrapKubeconfigPath,
+		apiServer:               apiServer,
+		certDuration:            certDuration,
+		currentCertPath:         filepath.Join(certDir, rbac.WICDCertNamePrefix+"-current.pem"),
+		currentKeyPath:          filepath.Join(certDir, rbac.WICDCertNamePrefix+"-current.pem"),
+	}
+}
+
+// StartCertificateManagement starts certificate management for WICD.
+// The certificate manager automatically handles renewal at 80% of certificate lifetime using built-in goroutines.
```
sebsoto, line 69: Please dont mention it does it using goroutines. That is a normal part of the language. It is like saying the code manages this using if statements :P

pkg/daemon/certs/wicd_cert_manager.go
```diff
@@ -0,0 +1,145 @@
...
+// NOTE: Kubernetes does not support certificate revocation for CSR-issued certificates.
+// Once issued, certificates remain valid until expiration. For security, certificates should have
+// short lifetimes and be rotated frequently.
+func (wcm *WICDCertificateManager) StartCertificateManagement(ctx context.Context) error {
```
sebsoto, line 73: ctx is passed in but not used.

pkg/daemon/certs/wicd_cert_manager.go
```diff
@@ -0,0 +1,145 @@
...
+		return nil
+	}
+
+	// Load bootstrap kubeconfig for initial authentication
+	bootstrapConfig, err := clientcmd.BuildConfigFromFlags("", wcm.bootstrapKubeconfigPath)
+	if err != nil {
+		return fmt.Errorf("failed to load bootstrap kubeconfig from %s: %w", wcm.bootstrapKubeconfigPath, err)
+	}
+
+	defaultConfig := &rest.Config{
```
sebsoto, line 84: Seems like this should be constructed as part of the New() method. Much the values in the wcm struct are used solely to create this config. Moving this to New() would simplify the struct.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -214,10 +219,9 @@ func (nc *NodeConfig) Configure(ctx context.Context) error {
 			return fmt.Errorf("error waiting for proper %s annotation for node %s: %w", metadata.VersionAnnotation,
 				nc.node.GetName(), err)
 		}
-
 		// Now that the node has been fully configured, update the node object in NodeConfig once more
 		if err := nc.setNode(ctx, false); err != nil {
-			return fmt.Errorf("error getting node object: %w", err)
+			return fmt.Errorf("error setting node object: %w", err)
```
sebsoto, line 224: nit: Good change, really should be in a separate commit, not worth the effort to change it for this PR, but try to do so in the future.

pkg/windows/windows.go
```diff
@@ -237,7 +242,14 @@ func GetK8sDir() string {
 	return K8sDir
 }
 
-// Windows contains all the methods needed to configure a Windows VM to become a worker node
+// WICDCertConfig contains certificate authentication configuration for WICD
+type WICDCertConfig struct {
```
sebsoto, line 246: I don't think this needs to be made public. Especially based on how its being used in Nodeconfig.

pkg/windows/windows.go
```diff
@@ -267,7 +279,7 @@ type Windows interface {
 	// Bootstrap prepares the Windows instance and runs the WICD bootstrap command
 	Bootstrap(context.Context, string, string, string) error
 	// ConfigureWICD ensures that the Windows Instance Config Daemon is running on the node
-	ConfigureWICD(string, string) error
+	ConfigureWICD(string, string, *WICDCertConfig) error
```
sebsoto, line 282: Is this interface change necessary? What arguments would change between calls to ConfigureWICD?

controllers/wicd_csr_controller.go
```diff
@@ -0,0 +1,167 @@
...
+		},
+		validator: validator,
+	}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager
+func (r *wicdCSRReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	// Build controller for CertificateSigningRequests
+	return builder.ControllerManagedBy(mgr).
+		For(&certificatesv1.CertificateSigningRequest{}).
```
sebsoto, line 82: Any possibility of predicates here, so that we don't call reconcile on all CSR events in the cluster? I think at the least you need to filter based on event type.

controllers/wicd_csr_controller.go
```diff
@@ -0,0 +1,167 @@
...
+		Status:         corev1.ConditionTrue,
+		Reason:         "WICDAutoApproved",
+		Message:        "This CSR was approved by the WICD certificate controller",
+		LastUpdateTime: metav1.Now(),
+	})
+
+	if _, err := r.k8sclientset.CertificatesV1().CertificateSigningRequests().UpdateApproval(ctx, csr.Name, csr, metav1.UpdateOptions{}); err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to approve CSR: %w", err)
+	}
+	log.Info("CSR approved", "CSR", csr.Name)
```
sebsoto, line 145: Please move the log message out of this function, that way the duplicate 
```
	log := r.log.WithValues("CertificateSigningRequest", csr.Name)
```
can be removed

test/e2e/wicd_rbac_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+// wicdRBACTestSuite tests WICD RBAC manifests and certificate functionality
+func wicdRBACTestSuite(t *testing.T) {
+	testCtx, err := NewTestContext()
+	if err != nil {
+		t.Fatalf("Failed to create test context: %v", err)
+	}
+
+	// Get all Windows nodes to determine if RBAC is relevant
+	windowsNodes, err := testCtx.client.K8s.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{
```
sebsoto, line 22: This is going to pick up Nodes not created by the test suite. Please use one of the existing helpers

test/e2e/wicd_rbac_test.go
```diff
@@ -0,0 +1,139 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/openshift/windows-machine-config-operator/pkg/rbac"
+)
+
+// wicdRBACTestSuite tests WICD RBAC manifests and certificate functionality
+func wicdRBACTestSuite(t *testing.T) {
+	testCtx, err := NewTestContext()
+	if err != nil {
+		t.Fatalf("Failed to create test context: %v", err)
```
sebsoto, line 18: Please use require here to match rest of tests

test/e2e/wicd_rbac_test.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	// Test WICD CSR approval functionality
+	t.Run("WICD CSR Approval", func(t *testing.T) {
+		testWICDCSRApproval(t, testCtx, windowsNodes.Items)
+	})
+}
+
+// testWICDCSRApproval tests that WICD CSRs created during node configuration are properly approved
+func testWICDCSRApproval(t *testing.T, testCtx *testContext, windowsNodes []corev1.Node) {
```
sebsoto, line 41: Does this test validate anything that the rest of the tests do not already do?
In order for a node to be functional it must have had its CSRs approved right?

pkg/nodeconfig/nodeconfig.go
```diff
@@ -403,13 +407,51 @@ func (nc *NodeConfig) generateBootstrapKubeconfig(ctx context.Context) (string,
 
 // generateWICDKubeconfig returns the contents of a kubeconfig created from the WICD ServiceAccount
 func (nc *NodeConfig) generateWICDKubeconfig(ctx context.Context) (string, error) {
+	// First, try to use certificate-based authentication if available
```
sebsoto, line 410: Comment doesn't make much sense unless there is an alternative to cert authentication.

sebsoto (changes requested):

test/e2e/wicd_rbac_test.go
mansikulkarni96: not really , test 1 may be redundant but this test is also testing
// Test 2: Verify it was approved by WICD controller with correct reason
// Test 3: Verify CSR username matches expected WICD identity format

sebsoto: ack, I am fine with that as long as it doesn't add noticeably additional time to the tests.

sebsoto (approved):
LGTM :tada:
In the future please try and move any refactors into a separate commit. It know it can be tedious for the author but it helps a lot for reviewers.

test/e2e/wicd_rbac_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+	"github.com/stretchr/testify/require"
+	certificatesv1 "k8s.io/api/certificates/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/openshift/windows-machine-config-operator/pkg/rbac"
+)
+
+// wicdRBACTestSuite tests WICD RBAC manifests and certificate functionality
+func wicdRBACTestSuite(t *testing.T) {
```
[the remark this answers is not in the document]
mansikulkarni96: The counter-example test you're suggesting is not possible right now because it would require webhook implementation to enforce node-specific access, cross-node write will be permitted in current state.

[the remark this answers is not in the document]
jrvaldes: I follow

mansikulkarni96:
/test azure-e2e-operator
test failure reason and point of failure is not consistent

PR #3135, opened 2025-07-31
Retry SSH server initialization

sebsoto:
Attempts to fix an issue where Windows Server 2019 machines were not having a running SSH server due to network instability.


jrvaldes (approved):
/lgtm

PR #3162, opened 2025-08-13
[WIP] Compress payload

sebsoto:


pkg/nodeconfig/payload/payload.go
```diff
@@ -338,3 +391,27 @@ func generateNetworkConfigScript(clusterCIDR, hnsNetworkName, hnsPSModulePath,
 	}
 	return networkConfScript, nil
 }
+
+// Creates a .tar.gz archive from file data
+func createTarGzFile(data []byte, fileName string, outWriter io.Writer) error {
+	// Chain writers: File -> Gzip -> Tar
+	gzipWriter := gzip.NewWriter(outWriter)
```
jrvaldes, line 398: consider using func `NewWriterLevel(w io.Writer, level int) (*Writer, error)`   with `BestCompression`

https://pkg.go.dev/compress/gzip#BestCompression

jrvaldes: actually we can do  "Level 1 (Fastest)" for testing and CI,
and "Level 9 (Best Compression)" for final release.

build/Dockerfile.ci
```diff
@@ -107,7 +105,6 @@ RUN make build-daemon
 # /payload/
 #├── azure-cloud-node-manager.exe
 #├── cni/
-#│   ├── flannel.exe
```
jrvaldes, line 110: there is another instance in 

https://github.com/openshift/windows-machine-config-operator/blob/ea36c7b1a691bd5273aace390fb58e4f2e0e3c00/build/Dockerfile#L101

Containerfile
```diff
@@ -119,7 +118,68 @@ RUN make build-daemon
...
+pushd /build/windows-machine-config-operator/pkg/internal/ && tar -cf - gcp-get-hostname.ps1 | gzip -9 > /payload/powershell/gcp-get-hostname.ps1.tar.gz && popd && \
+sha256sum /build/windows-machine-config-operator/pkg/internal/gcp-get-hostname.ps1 >> sha256sum && \
+pushd /build/windows-machine-config-operator/pkg/internal/ && tar -cf - windows-defender-exclusion.ps1 | gzip -9 > /payload/powershell/windows-defender-exclusion.ps1.tar.gz && popd && \
+sha256sum /build/windows-machine-config-operator/pkg/internal/windows-defender-exclusion.ps1 >> sha256sum && \
+pushd /build/windows-machine-config-operator/pkg/internal/ && tar -cf - hns.psm1 | gzip -9 > /payload/powershell/hns.psm1.tar.gz && popd && \
+sha256sum /build/windows-machine-config-operator/pkg/internal/hns.psm1 >> sha256sum && \
+chmod -R 644 /payload && chmod -R +X /payload &&\
+# Create directory for generated files with open permissions, this allows WMCO to write to this directory
+mkdir /payload/generated && \
+chmod 0777 /payload/generated
```
jrvaldes, line 182: for my own education:

why is Read + Write + Execute needed for group and others?

pkg/nodeconfig/payload/payload.go
```diff
@@ -296,21 +300,63 @@ Compare-And-Replace-Config -ConfigPath $kubeProxyConfigPath -NewConfigContent $k
...
+		return err
+	}
+	return nil
+}
+
+// parseShaFile returns a mapping of file names to SHA256 sums
+// input should be of the format generated by the sha256sum binary: `SHA256SUM  filepath`
+func parseShaFile(fileData []byte) (map[string]string, error) {
+	m := make(map[string]string)
+	for _, line := range strings.Split(string(fileData), "\n") {
```
jrvaldes, line 325: I recommend to  trim spaces and check if `string(fileData)` is empty first, is so return error

pkg/nodeconfig/payload/payload.go
```diff
@@ -296,21 +300,63 @@ Compare-And-Replace-Config -ConfigPath $kubeProxyConfigPath -NewConfigContent $k
...
+		if len(line) == 0 {
```
jrvaldes, line 326: trim spaces in `line` before calculating the length

pkg/nodeconfig/payload/payload.go
```diff
@@ -296,21 +300,63 @@ Compare-And-Replace-Config -ConfigPath $kubeProxyConfigPath -NewConfigContent $k
...
+			continue
+		}
+		columnSplit := strings.SplitN(line, "  ", 2)
```
jrvaldes, line 329: consider a constant for `"  "`, e.g. sha256sumColumnSeparator

pkg/nodeconfig/payload/payload.go
```diff
@@ -296,21 +300,63 @@ Compare-And-Replace-Config -ConfigPath $kubeProxyConfigPath -NewConfigContent $k
...
+		if len(columnSplit) != 2 {
+			return nil, fmt.Errorf("unexpected sha256sum line format %s", line)
+		}
+		fileName := filepath.Base(columnSplit[1])
+		if _, present := m[fileName]; present {
+			return nil, fmt.Errorf("duplicate sha256sum entry for %s", fileName)
```
jrvaldes, line 335: Consider logging the checksum as well, they may be different.

pkg/nodeconfig/payload/payload.go
```diff
@@ -296,21 +300,63 @@ Compare-And-Replace-Config -ConfigPath $kubeProxyConfigPath -NewConfigContent $k
...
+		}
+		m[fileName] = columnSplit[0]
```
jrvaldes, line 337: I recommend to check  `columnSplit[0]`  is not empty

pkg/nodeconfig/payload/payload_test.go
```diff
@@ -227,3 +227,18 @@ Compare-And-Replace-Config -ConfigPath $kubeProxyConfigPath -NewConfigContent $k
 	require.NoError(t, err)
 	assert.Equal(t, string(expectedOut), actual)
 }
+
+func TestParseShaFile(t *testing.T) {
```
jrvaldes, line 231: consider including other negative test cases, missing filename, blank lines, etc.

sebsoto:
@jrvaldes 


> @sebsoto, why not get all the directories and files into a single compressed file? So that there is only one SSH connection to transfer the payload and one checksum validation?

@jrvaldes
Doing one big transfer has two major downsides:

1) If any file in the payload changes, the entire payload needs to be transferred again. This would mainly affect upgrades.
2) If there is a network interruption during the transfer process, the whole transfer begins again. By breaking up the payload into smaller chunks, the files that have been completely transferred already can be skipped, removing the duplicate work.

jrvaldes:
> @jrvaldes That feels a bit like premature optimization. I dont think the gains we would get by grouping the files would outweigh the additional complexity to map multiple files with different decompressed SHA256's to an archive.
> 
> As we share the same SSH connection for all transfers, the downside of having many small files is that each transfer will cause an exec call to `tar` to unzip. This will only happen if the file needs to be updated. That will probably result in 10s of milliseconds added, I think its worth it to keep things simple.

true, thanks for the explanation.

cmd/operator/main.go
```diff
@@ -148,6 +148,10 @@ func main() {
 		os.Exit(1)
 	}
 
+	if err := payload.PopulateSHAMap(); err != nil {
+		setupLog.Error(err, "error using payload checksums")
```
wgahnagl: most annoying nitpick in the world: Maybe capitalize this "Error"

pkg/nodeconfig/payload/payload.go
```diff
@@ -1,79 +1,84 @@
...
 
-// Payload files
 const (
+	// sha25sumColumnSeparator is the separator used in the utility sha256sum. This is between the file name and the sha256 sum value.
+	sha256sumColumnSeparator = "  "
 	// payloadDirectory is the directory in the operator image where are all the binaries live
 	payloadDirectory = "/payload/"
 	// WICDPath is the path to the Windows Instance Config Daemon exe
-	WICDPath = payloadDirectory + "windows-instance-config-daemon.exe"
+	WICDPath = payloadDirectory + "windows-instance-config-daemon.exe.tar.gz"
```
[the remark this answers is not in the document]
sebsoto: I actually looked into this before making the change, its preferred to maintain the .exe, as it makes it clear that this is an exe, which has been tarred, and then gzipped.

PR #3168, opened 2025-08-13
Reduce node retry interval in e2e tests

sebsoto:
This reduces the retry interval from 60 -> 15 seconds. This change reduces the potential wasted time of a Node finishing configuration immediately after a retry.


jrvaldes (approved):
/lgtm

jrvaldes:
@sebsoto  consider backporting this to other release branches.

PR #3225, opened 2025-08-25
OCPBUGS-60866: Fix HNS network removal error log level

sebsoto:
A failure to remove an HNS network is a recoverable error, and should be logged at debug level. This commit fixes an issue where this being logged at info level.

log.V(1).Info needs to be used to log to debug level. log.V(1).Error does not work. logger.Error has the documentation: "The log message will always be emitted, regardless of verbosity level"

### DIFF
--- a/pkg/windows/windows.go
+++ b/pkg/windows/windows.go
@@ -970,7 +970,7 @@ func (vm *windows) ensureHNSNetworksAreRemoved() error {
 		err = wait.PollImmediate(retry.Interval, retry.Timeout, func() (bool, error) {
 			// reinitialize and retry on failure to avoid connection reset SSH errors
 			if err := vm.removeHNSNetwork(network); err != nil {
-				vm.log.V(1).Error(err, "error removing %s HNS network", "network", network)
+				vm.log.V(1).Info("error removing HNS network", "network", network, "err", err.Error())
 				if err := vm.reinitialize(); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/windows/windows.go
```diff
@@ -970,7 +970,7 @@ func (vm *windows) ensureHNSNetworksAreRemoved() error {
 		err = wait.PollImmediate(retry.Interval, retry.Timeout, func() (bool, error) {
 			// reinitialize and retry on failure to avoid connection reset SSH errors
 			if err := vm.removeHNSNetwork(network); err != nil {
-				vm.log.V(1).Error(err, "error removing %s HNS network", "network", network)
+				vm.log.V(1).Info("error removing HNS network", "network", network, "err", err.Error())
```
jrvaldes, line 973: > A failure to remove an HNS network is a recoverable error,

True, and should also be captured in the error message.

jrvaldes: Lowering the verbosity of the message to debug will force the users to restart the operator in debug mode, to see the details of the error and then investigate.

sebsoto: This is an expected outcome, so I don't see there to be a reason for investigation.
The reason we retry is because removing the HNS network will cause the network to go down.

The intention was for this to only be logged in debug logging, thus the existing `V(1)`.
Printing errors is alarming for users.

jrvaldes (approved):
makes sense

PR #3253, opened 2025-09-02
OCPBUGS-61122: Fix secret controller logger mutation

sebsoto:
On each secret reconciliation the secret controller is incorrectly being changed to add data for the current reconciliation. This data is persisting and snowballing on each reconciliation loop.


controllers/secret_controller.go
```diff
@@ -147,8 +147,6 @@ type SecretReconciler struct {
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
 func (r *SecretReconciler) Reconcile(ctx context.Context,
 	request ctrl.Request) (result reconcile.Result, reconcileErr error) {
-	r.log = r.log.WithValues(SecretController, request.NamespacedName)
```
jrvaldes, line 150: is this issue also present in other controllers?

https://github.com/openshift/windows-machine-config-operator/blob/b21500c103eaa5692ccebc58ee0f6792120f9c85/controllers/node_controller.go#L74



https://github.com/openshift/windows-machine-config-operator/blob/b21500c103eaa5692ccebc58ee0f6792120f9c85/controllers/registry_controller.go#L81

controllers/node_controller.go
```diff
@@ -93,6 +92,7 @@ func (r *nodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 		return ctrl.Result{}, err
 	}
 
+	r.log.V(1).Info("reconciling", req.NamespacedName.String())
```
sebsoto, line 95: Changed this to a debug log to be in line with other controllers.

jrvaldes (commented):
/lgtm

controllers/secret_controller.go
jrvaldes: ack

jrvaldes:
@sebsoto   I don't see how the proposed changes are affecting the failed e2e tests, but they all failed twice in different spots, though

mansikulkarni96:
@sebsoto missed this earlier but the metric controller needs update as well
https://github.com/openshift/windows-machine-config-operator/blob/12d0058f5ff746b5e225613b851749d7e3ce539e/controllers/metric_controller.go#L75

PR #3411, opened 2025-10-07
[build] Remove build/Dockerfile

sebsoto:
Removes the Dockerfile previously used for CPaaS builds. This has been superseded by `Containerfile`, used for Konflux builds.


jrvaldes:
@sebsoto   do we need to backport this to cleanup lower branches too?

PR #3516, opened 2025-11-03
Fix renovate artifact update

mansikulkarni96:
This PR removes the allowedCommands as
they have to be enabled by konflux team on
their end and also removes postupdateoptions
as the config exists by default
https://github.com/konflux-ci/mintmaker/blob/main/config/renovate/renovate.json#L603C6-L603C11
Additionally there is an issue with downgrading
go packages to a particular version so the jwt
package needs to be excluded from upgrades.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -121,7 +121,6 @@
     "enabled": true,
     "additionalBranchPrefix": "{{baseBranch}}/",
     "branchPrefix": "konflux/mintmaker/",
-    "postUpdateOptions": ["gomodTidy", "gomodUpdateImportPaths"],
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -294,6 +294,7 @@
     "^go mod vendor$",
     "^go mod tidy$",
     "^go mod verify$"
+    "^go mod vendor$"
```
jrvaldes: looks like is there already; Line 294

jrvaldes: @mansikulkarni96  PTAL ^

PR #3913, opened 2026-03-30
[hack] Add create-release-tag.py script

mtnbikenc:
## Summary

- Adds \`hack/create-release-tag.py\` to create consistent annotated release tags for WMCO
- Resolves the commit SHA from the bundle image in the Red Hat Container Catalog,
  preferring the \`org.opencontainers.image.revision\` OCI label (full SHA) and falling
  back to the short-SHA image tag
- Sets the tag date to the bundle image's \`push_date\` so the tag timestamp reflects
  the actual release date rather than when the tag was created, consistent with the
  commit source
- Validates \`--commit\` input against a hex SHA pattern at parse time to prevent git
  option injection
- Prompts for confirmation before creating the tag, showing the resolved commit, date,
  and their sources
- Documents the team rationale (*why*) behind each process decision in the script
  header: why the bundle image commit is tagged, why the OCI label is preferred over
  the short SHA tag, why annotated tags, and why the push command is shown rather than
  executed
- Includes a PREREQUISITES section covering runtime requirements, package dependencies,
  environment expectations, and permissions

## Usage

    python3 hack/create-release-tag.py <version> [--commit SHA] [--date YYYY-MM-DD]

\`--commit\` is required for releases not present in the bundle catalog (e.g. backport-only
releases). \`--date\` allows overriding the publish date when needed.

## Test plan

- [x] Run \`python3 hack/create-release-tag.py --help\` and verify usage output
- [x] Run against an existing version (e.g. \`10.21.1\`) and verify it exits early with "tag already exists"
- [x] Run against a new version and verify the catalog-resolved commit and date match expectations
- [x] Run with \`--commit\` and \`--date\` overrides and verify the confirmation prompt reflects them
- [x] Verify the created tag is annotated (\`git cat-file -t vX.Y.Z\` returns \`tag\`) with correct message and date

### DIFF
--- a/hack/create-release-tag.py
+++ b/hack/create-release-tag.py
@@ -1,0 +1,445 @@
+#!/usr/bin/env python3
+"""
+create-release-tag.py — Create an annotated release tag for WMCO.
+
+The commit SHA is resolved from the bundle image in the Red Hat Container
+Catalog, preferring the org.opencontainers.image.revision label (full SHA)
+and falling back to the short-SHA image tag.  The tag date is set to the
+operator image's push_date so the timestamp reflects the actual release date.
+
+Usage:
+    python3 hack/create-release-tag.py <version>
+    python3 hack/create-release-tag.py <version> --commit <SHA>
+    python3 hack/create-release-tag.py <version> --date YYYY-MM-DD
+    python3 hack/create-release-tag.py <version> --commit <SHA> \
+        --date YYYY-MM-DD
+
+    --commit is required when the version has no entry in the bundle catalog
+    (e.g. backport releases that were never shipped as a container image).
+
+Note:
+    Pushing tags to the upstream openshift/windows-machine-config-operator
+    repository requires write access granted by the repository administrators.
+    Contact the WMCO team to have your GitHub account added before pushing.
+
+━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
+DATA SOURCES
+━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
+
+  Red Hat Container Catalog  (catalog.redhat.com)
+    Bundle image:    openshift4-wincw/windows-machine-config-operator-bundle
+    Provides: build commit SHA.  Newer images carry the full 40-character SHA
+    in the org.opencontainers.image.revision OCI label.  Older images that
+    predate the label are handled by falling back to the short (7-character)
+    hex SHA that Konflux publishes as an image tag alongside the version tag.
+
+    Operator image:  openshift4-wincw/windows-machine-config-rhel9-operator
+    Provides: push_date — the UTC timestamp when the image was published to
+    the catalog.  This is used as the tag date so the annotated tag timestamp
+    reflects the actual release date rather than the date the tag was created.
+
+  Local git repository
+    Provides: existing tag detection (prevents accidental overwrites), full
+    SHA expansion from short SHAs via git rev-parse, commit existence
+    verification to catch catalog/repo mismatches before the tag is written,
+    and upstream remote detection for the post-creation push instruction.
+
+━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
+LOGIC
+━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
+
+1. COMMIT RESOLUTION
+   Source:   Bundle image catalog
+   Data:     org.opencontainers.image.revision label (preferred) or short hex
+             image tag (fallback for images that predate the OCI label).
+   Logic:
+     • Scans the bundle catalog for the image whose tags include vX.Y.Z.
+     • Prefers the full SHA from the OCI label; falls back to the short SHA
+       tag.
+     • Any SHA (full or short) is expanded to a full 40-character commit via
+       git rev-parse and verified to exist locally.  This catches catalog/repo
+       mismatches (e.g. if the local clone is stale) before writing anything.
+     • --commit skips the catalog lookup entirely and uses the provided value.
+       Required for backport releases that were never shipped as a container
+       image and therefore have no bundle catalog entry.
+
+2. TAG DATE RESOLUTION
+   Source:   Operator image catalog
+   Data:     push_date field from the repository entry for the matching
+             version.
+   Logic:
+     • The push_date is the UTC timestamp when the operator image was
+       published.
+     • Only the date portion (YYYY-MM-DD) is used; the time is fixed to noon
+       UTC (12:00:00+0000) to produce an unambiguous, timezone-neutral
+       timestamp.
+     • The fixed noon time is set via GIT_COMMITTER_DATE when git tag is
+       invoked, so it appears as the tag's creation date in git log and GitHub.
+     • --date skips the catalog lookup and uses the provided date instead.
+
+3. TAG CREATION
+   Logic:
+     • Tag name:    vX.Y.Z
+     • Tag message: "Windows Machine Config Operator vX.Y.Z"
+     • Tag type:    annotated (git tag -a), never lightweight.
+     • The user is shown all resolved details — tag, message, full commit SHA
+       with its source, and date with its source — and must confirm before the
+       tag is written.  This surfaces any unexpected catalog values before they
+       become permanent.
+
+4. PUSH INSTRUCTION
+   Logic:
+     • After creation, scans git remote -v for a remote whose URL contains
+       "openshift/windows-machine-config-operator" and uses its configured
+       name (commonly "upstream") in the suggested push command.
+     • Falls back to the canonical SSH URL if no matching remote is found, so
+       the command is always complete and immediately usable.
+"""
+
+# pylint: disable=invalid-name  # hyphenated script name is intentional
+
+import argparse
+from datetime import date
+import os
+import re
+import shutil
+import subprocess
+import sys
+
+import requests
+
+OPERATOR_CATALOG_API = (
+    "https://catalog.redhat.com/api/containers/v1/repositories/"
+    "registry/registry.access.redhat.com/repository/"
+    "openshift4-wincw/windows-machine-config-rhel9-operator/images"
+)
+BUNDLE_CATALOG_API = (
+    "https://catalog.redhat.com/api/containers/v1/repositories/"
+    "registry/registry.access.redhat.com/repository/"
+    "openshift4-wincw/windows-machine-config-operator-bundle/images"
+)
+
+_VERSION_TAG_RE = re.compile(r"^v(\d+\.\d+\.\d+)$")
+_HEX_RE = re.compile(r"^[0-9a-f]{7,40}$")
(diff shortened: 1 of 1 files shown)


hack/create-release-tag.py
```diff
@@ -0,0 +1,412 @@
...
+# ---------------------------------------------------------------------------
+
+def git(*args, env=None) -> str:
+    """Run a git command and return stdout, raising on failure."""
+    result = subprocess.run(
+        ["git", *args], capture_output=True, text=True, env=env, check=False
+    )
+    if result.returncode != 0:
+        raise RuntimeError(result.stderr.strip())
+    return result.stdout.strip()
```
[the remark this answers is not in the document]
mtnbikenc: Addressed in commit ae23adf40a61e9abc4c2505dd8f8f032d4769d08: wrapped `subprocess.run` in a `try/except OSError` and added a `shutil.which("git")` check so callers always receive `RuntimeError` rather than an unhandled `FileNotFoundError`.

hack/create-release-tag.py
```diff
@@ -0,0 +1,412 @@
...
+
+
+def tag_exists(tag: str) -> bool:
+    """Return True if the git tag already exists in this repository."""
+    try:
+        git("rev-parse", tag)
+        return True
+    except RuntimeError:
+        return False
```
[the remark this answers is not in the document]
mtnbikenc: Addressed in commit ae23adf40a61e9abc4c2505dd8f8f032d4769d08: changed `git("rev-parse", tag)` to `git("rev-parse", "-q", "--verify", f"refs/tags/{tag}")` so only actual tags (not branches with the same name) return `True`.

hack/create-release-tag.py
```diff
@@ -0,0 +1,412 @@
...
+    parser.add_argument("--date", metavar="YYYY-MM-DD",
+                        help="Override published date")
+    args = parser.parse_args()
+
+    # Validate inputs
+    if not re.fullmatch(r"\d+\.\d+\.\d+", args.version):
+        parser.error(f"version must be X.Y.Z, got: {args.version!r}")
+
+    if args.date and not re.fullmatch(r"\d{4}-\d{2}-\d{2}", args.date):
+        parser.error(f"--date must be YYYY-MM-DD, got: {args.date!r}")
```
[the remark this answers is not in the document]
mtnbikenc: Addressed in commit ae23adf40a61e9abc4c2505dd8f8f032d4769d08: replaced the `re.fullmatch` check with `date.fromisoformat()` inside a `try/except ValueError` so invalid calendar dates like `2026-02-31` are rejected at argument-parse time.

hack/create-release-tag.py
```diff
@@ -0,0 +1,412 @@
...
+                file=sys.stderr,
+            )
+            sys.exit(1)
+
+    if args.date:
+        published_date = args.date
+        date_source = "provided manually"
+    else:
+        published_date = fetch_operator_push_date(version)
+        date_source = "operator image push date"
```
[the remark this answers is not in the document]
mtnbikenc: Addressed in commit ae23adf40a61e9abc4c2505dd8f8f032d4769d08: wrapped both `fetch_bundle_info()` and `fetch_operator_push_date()` calls in `try/except Exception` blocks that print a clean `ERROR:` message to stderr and call `sys.exit(1)`, matching the existing git-failure error style.

hack/create-release-tag.py
```diff
@@ -0,0 +1,445 @@
...
+        git("rev-parse", "-q", "--verify", f"refs/tags/{tag}")
+        return True
+    except RuntimeError:
+        return False
+
+
+def resolve_commit(ref: str) -> str:
+    """Return the full commit SHA for ref, or raise RuntimeError."""
+    try:
+        return git("rev-parse", f"{ref}^{{commit}}")
```
jrvaldes, line 275: consider validating the  `--commit` input; the user-supplied SHA is passed to git rev-parse without format checking, enabling argument injection.

hack/create-release-tag.py
```diff
@@ -0,0 +1,445 @@
...
+        # Each line: "<name>\t<url> (fetch|push)"
+        parts = line.split()
+        if (len(parts) >= 2
+                and "openshift/windows-machine-config-operator" in parts[1]):
+            return parts[0]
+    return ""
+
+
+# pylint: disable=too-many-locals,too-many-branches,too-many-statements
+def main():
```
jrvaldes, line 383: this func is big and complex, consider generalizing orthogonal logic using helper functions, for example

`resolve_commit_sha()` and `resolve_date()`

hack/create-release-tag.py
```diff
@@ -0,0 +1,445 @@
...
+    "registry/registry.access.redhat.com/repository/"
+    "openshift4-wincw/windows-machine-config-operator-bundle/images"
+)
+
+_VERSION_TAG_RE = re.compile(r"^v(\d+\.\d+\.\d+)$")
+_HEX_RE = re.compile(r"^[0-9a-f]{7,40}$")
+
+
+# ---------------------------------------------------------------------------
+# Catalog helpers (shared pattern with verify-release.py)
```
jrvaldes: still have mixed feelings about introducing helper tools in Python; anyhow, consider extracting a shared module under hack/lib/  to avoid duplicating this implementation

jrvaldes (commented):
@mtnbikenc  thanks for working on this, LGTM. just minor suggestions, PTAL

mtnbikenc:
Thanks for the feedback and suggestions, all good points.  I'll look into these and come back with some updates.

hack/create-release-tag.py
mtnbikenc: Addressed in commit ebb6fdb23: added a `_HEX_RE` (`^[0-9a-f]{7,40}$`) check against the `--commit` value at argument-parse time via `parser.error()`. Any value that isn't a lowercase hex string of 7–40 characters is rejected before reaching `git rev-parse`, preventing git option injection (e.g. `--upload-pack=evil-cmd` or relative refs like `HEAD~1`). Note: since `subprocess.run()` uses a list rather than `shell=True`, shell metacharacter injection was never possible — the real risk was git option injection, which `_HEX_RE` now covers.

hack/create-release-tag.py
mtnbikenc: Addressed in commit ebb6fdb23: extracted `resolve_commit_sha(tag, version, override)` and `resolve_date(tag, version, override)` from `main()`. Each helper encapsulates the catalog lookup, exception handling, and missing-value error messaging for its respective concern. The `too-many-branches` pylint suppression on `main()` was also removed as it is no longer needed.

hack/create-release-tag.py
mtnbikenc: Agreed on the value. Deferring the `hack/lib/` extraction until team direction on future hack script organization is decided.

mtnbikenc:
I didn't squash commits to allow easier review of the changes.  If the team decides to move forward with using this script, and other python scripts, I can add python linting tests to this PR since it is the first.

hack/create-release-tag.py
```diff
@@ -0,0 +1,473 @@
...
+            file=sys.stderr,
+        )
+        sys.exit(1)
+
+    return published_date, "operator image push date"
+
+
+# pylint: disable=too-many-locals,too-many-statements
+def main():
+    """Parse args, resolve tag details, confirm with user, create the tag."""
```
mansikulkarni96, line 384: I like that the script asks for confirmation to avoid accidental tagging.

mansikulkarni96 (approved):
I have used this script and it works for me. I am okay with having this in python as this is easy to maintain and right for the job. I agree with adding a linter like pylint to help further updates to this.

mtnbikenc:
As part of this PR I reviewed the script against what I believe should be a baseline standard for hack scripts. Noting it here for team visibility and feedback.

**Hack script checklist**

- [ ] **Prerequisites / requirements** — Python/Go/tool version, third-party packages, any required binaries (e.g. `git` in PATH)
- [ ] **Usage instructions** — all supported invocation forms with argument descriptions
- [ ] **Permissions and access tokens** — what credentials or access rights are needed, and whether external APIs require authentication
- [ ] **Data sources** — every external system the script reads from, what it provides, and the API or mechanism used
- [ ] **Logic documentation** — for each major step: *Source*, *Data*, *Why* (the team decision or rationale), and *Logic* (how it works)
- [ ] **Lint clean** — passes the linter appropriate for the language (`pylint`/`pyflakes` for Python, `golangci-lint` for Go)

The *Why* is the most important piece — these scripts codify team process decisions, and without the reasoning future maintainers have no basis for knowing whether a detail is intentional or accidental.

Open to guidance from the team if there are other items that should be on this list.

hack/create-release-tag.py
mtnbikenc: Having this confirmation is good however, it seems it could be better have it create the tag then print the tag for the user to confirm and then push.  Confirming the details of what would be created is not as valuable as confirming what what actually created.  If the tag it created needed to be recreated, it could be deleted since it only exists on the local clone.  This keeps the user in control of publishing the tag.  Thoughts?

hack/create-release-tag.py
```diff
@@ -0,0 +1,498 @@
...
+     • Tag name:    vX.Y.Z
+     • Tag message: "Windows Machine Config Operator vX.Y.Z"
+     • Tag type:    annotated (git tag -a), never lightweight.
+     • The user is shown all resolved details — tag, message, full commit SHA
+       with its source, and date with its source — and must confirm before the
+       tag is written.  This surfaces any unexpected catalog values before they
+       become permanent.
+
+4. PUSH INSTRUCTION
+   Why show command rather than push: This script runs interactively in a
```
wgahnagl, line 117: these "why do xyz" comments feel a little strange to keep in the comments of the script. I think a lot of this could be cut down, since it's kind of explaining a lot of information that a user would already get the gist of. 
"Any action that modifies public,
     shared content is shown as a command for the developer to review and
     execute deliberately, rather than performed automatically." is kind of implied.

mtnbikenc: > The Why is the most important piece — these scripts codify team process decisions, and without the reasoning future maintainers have no basis for knowing whether a detail is intentional or accidental.

Are you saying this statement is not important?  Just trying to get a feel for what you mean by 'already get the gist of.'

mtnbikenc:
Closing this for now.  I'm happy to open it back up or have someone else take this on if the team decides this would be beneficial.

mtnbikenc:
I'm happy to take this out of draft and get it merged if that's what the team wants.  :wave:

mtnbikenc:
Rebased.

jrvaldes:
/override ci/prow/aws-e2e-operator ci/prow/azure-e2e-operator ci/prow/azure-e2e-upgrade ci/prow/gcp-e2e-operator ci/prow/nutanix-e2e-operator ci/prow/platform-none-vsphere-e2e-operator ci/prow/vsphere-disconnected-e2e-operator ci/prow/vsphere-e2e-operator ci/prow/vsphere-proxy-e2e-operator ci/prow/images ci/prow/lint ci/prow/security ci/prow/unit ci/prow/wicd-unit-vsphere ci/prow/ci-bundle-wmco-bundle

jrvaldes:
Thanks, @mtnbikenc, for helping with this. Appreciate all your amazing contributions!

jrvaldes:
/override ci/prow/aws-e2e-operator ci/prow/azure-e2e-operator ci/prow/azure-e2e-upgrade ci/prow/gcp-e2e-operator ci/prow/nutanix-e2e-operator ci/prow/platform-none-vsphere-e2e-operator ci/prow/vsphere-disconnected-e2e-operator ci/prow/vsphere-e2e-operator ci/prow/vsphere-proxy-e2e-operator ci/prow/images ci/prow/lint ci/prow/security ci/prow/unit ci/prow/wicd-unit-vsphere ci/prow/ci-bundle-wmco-bundle

PR #4278, opened 2026-07-07
OCPBUGS-65625: Clean up C:\var\lib\cni directory

ranjithrajaram:
The C:\var\lib\cni directory is created dynamically by CNI plugins (like host-local IPAM) but was left behind during node deconfiguration. Unlike C:\var\lib\kubelet (which is also left behind but contains active pod volume mounts that could result in data loss if deleted), the CNI state directory only contains transient network lease information and can be safely removed.

Fixes https://issues.redhat.com/browse/OCPBUGS-65625

For C:\var\lib\kubelet , I will  be filing a documentation bug so that administrators can review and clean them

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows node teardown to also clean up leftover CNI state/IPAM data by removing the CNI state directory with best-effort handling.
  * Expanded end-to-end cleanup verification on Windows to ensure the CNI state directory is absent after deletion.

### DIFF
--- a/pkg/windows/windows.go
+++ b/pkg/windows/windows.go
@@ -49,6 +49,8 @@ const (
 	wicdLogDir = logDir + "\\wicd"
 	// cniDir is the directory for storing CNI binaries
 	cniDir = K8sDir + "\\cni"
+	// CniStateDir is the directory for storing CNI state/IPAM information
+	CniStateDir = "C:\\var\\lib\\cni"
 	// CniConfDir is the directory for storing CNI configuration
 	CniConfDir = cniDir + "\\config"
 	// ContainerdDir is the directory for storing Containerd binary
@@ -560,6 +562,11 @@ func (vm *windows) Bootstrap(ctx context.Context, desiredVer, watchNamespace, wi
 		return fmt.Errorf("unable to cleanup the Windows instance: %w", err)
 	}
 
+	// Remove stale CNI state from a previous configuration on a best-effort basis
+	if out, err := vm.Run(rmDirCmd(CniStateDir), true); err != nil {
+		vm.log.V(1).Error(err, "unable to remove directory", "dir", CniStateDir, "out", out)
+	}
+
 	if err := vm.ensureHostNameAndContainersFeature(ctx); err != nil {
 		return err
 	}
@@ -726,6 +733,10 @@ func (vm *windows) removeDirectories() error {
 	// At this point, we don't want to retry reconciliation again as the WICD files could have been deleted.
 	if out, err := vm.Run(rmDirCmd(K8sDir), true); err != nil {
 		vm.log.V(1).Error(err, "unable to remove directory", "dir", K8sDir, "out", out)
+	}
+	// Clean up CNI state directory on a best-effort basis
+	if out, err := vm.Run(rmDirCmd(CniStateDir), true); err != nil {
+		vm.log.V(1).Error(err, "unable to remove directory", "dir", CniStateDir, "out", out)
 	}
 	return nil
 }
--- a/test/e2e/delete_test.go
+++ b/test/e2e/delete_test.go
@@ -128,7 +128,10 @@ func (tc *testContext) getProxyCABundle() (string, error) {
 func (tc *testContext) checkDirsDoNotExist(address string) (bool, error) {
 	command := ""
 
-	for _, dir := range windows.RequiredDirectories {
+	dirs := make([]string, 0, len(windows.RequiredDirectories)+1)
+	dirs = append(dirs, windows.RequiredDirectories...)
+	dirs = append(dirs, windows.CniStateDir)
+	for _, dir := range dirs {
 		command += fmt.Sprintf("if ((Test-Path %s) -eq $true) { Write-Output %s exists}", dir, dir)
 	}
 	command += "exit 0"


ranjithrajaram:
@coderabbitai full review

ranjithrajaram:
@coderabbitai full review

pkg/windows/windows.go
```diff
@@ -51,6 +49,8 @@ const (
 	wicdLogDir = logDir + "\\wicd"
 	// cniDir is the directory for storing CNI binaries
 	cniDir = K8sDir + "\\cni"
+	// CniStateDir is the directory for storing CNI state/IPAM information
+	CniStateDir = "C:\\var\\lib\\cni"
```
jrvaldes, line 53: can you list the files found in C:\var\lib\cni?

pkg/windows/windows.go
```diff
@@ -740,6 +729,10 @@ func (vm *windows) removeDirectories() error {
 	if out, err := vm.Run(rmDirCmd(K8sDir), true); err != nil {
 		vm.log.V(1).Error(err, "unable to remove directory", "dir", K8sDir, "out", out)
 	}
+	// Clean up CNI state directory on a best-effort basis
+	if out, err := vm.Run(rmDirCmd(CniStateDir), true); err != nil {
```
jrvaldes, line 738: the directory `C:\\var\\lib\\cni` will be always managed by WMCO/CNI. Instead of removing during cleanup, I'd rather do a preventive cleanup pre-configuration, so that a clean state is guarantee.

jrvaldes (commented):
@ranjithrajaram  thanks for working on this. PTAL at the comments.

pkg/windows/windows.go
ranjithrajaram: PS C:\Users\Administrator> Get-ChildItem c:\var\lib\cni


    Directory: C:\var\lib\cni


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d-----        11/19/2025   3:47 AM                networks
d-----         5/25/2026   4:33 AM                results


PS C:\Users\Administrator> Get-ChildItem c:\var\lib\cni -Recurse


    Directory: C:\var\lib\cni


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d-----        11/19/2025   3:47 AM                networks
d-----         5/25/2026   4:33 AM                results


    Directory: C:\var\lib\cni\networks


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d-----          7/7/2026   9:20 AM                OVNKubernetesHybridOverlayNetwork


    Directory: C:\var\lib\cni\networks\OVNKubernetesHybridOverlayNetwork


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a----          7/6/2026   4:03 AM             10 last_reserved_ip.0
-a----        11/19/2025   3:47 AM              0 lock

pkg/windows/windows.go
ranjithrajaram: @jrvaldes 	Thanks for the suggestion
	
	I did review this option during bootstrap, but i thought during Post-deconfiguration leaving an artifact behind was the one to fix.
	
	We should do both. it protects against stale CNI state on re-join (e.g., BYOH re-add, failed prior configuration). If the node is not re-added back to the OCP cluster, leaving such an artifact behind on node is undesirable . Let me know if we can do both or just do the cleanup during bootstrap

jrvaldes: >We should do both

agreed

cc @mansikulkarni96

mansikulkarni96: I agree lets do pre-cleanup,  every configuration should start with clean state.

mansikulkarni96 (changes requested):

ranjithrajaram:
@coderabbitai  full review

mansikulkarni96 (approved):
Thanks for working on this @ranjithrajaram

jrvaldes:
@mansikulkarni96  IMO, a e2e test with two (2) machine instances  is required to validate this case.

ranjithrajaram:
> @mansikulkarni96 IMO, a e2e test with two (2) machine instances is required to validate this case.

@jrvaldes  e2e test case with two machines was for PR 4155 ?

jrvaldes:
> > @mansikulkarni96 IMO, a e2e test with two (2) machine instances is required to validate this case.
> 
> @jrvaldes e2e test case with two machines was for PR 4155 ?

correct, had my wires crossed 🫓

jrvaldes:
The reason[ I asked ](https://github.com/openshift/windows-machine-config-operator/pull/4278#discussion_r3537375088)for the content of the C:\var\lib\cni , I haven't see anything in this dir on any test cluster.

jrvaldes:
in the current e2e tests we are doing reconfiguration of both machine and BYOH intances, even with upgrades. @ranjithrajaram  why are we not getting this issue on the current test? what is the diff in your test environment that populate the C:\var\lib\cni

jrvaldes:
@coderabbitai full review

jrvaldes:
@ranjithrajaram  please adjust the Jira fields, see https://github.com/openshift/windows-machine-config-operator/pull/4278#issuecomment-4934931024

rrasouli:
@CodeRabbit 
The test change in `checkDirsDoNotExist` added an assertion for `CniStateDir` (`C:\var\lib\cni`), but the actual removal in removeDirectories() is best-effort -- it logs the error and continues without failing. This mismatch caused the `vsphere-proxy-e2e-operator` job to fail on BYOH node cleanup (delete_test.go:87 "directories not removed").

The test was asserting that a directory was removed, while the production code intentionally tolerates removal failure. This inconsistency should have been caught during review.

  Fixed in ranjithrajaram@fcf16d8 by reverting the test change back to only checking RequiredDirectories.

PR #4284, opened 2026-07-07
OCPBUGS-90077: Fix infinite reboot on SSH exit command status

mansikulkarni96:
## Summary by CodeRabbit

* **New Features**
  * Nodes now track reboot attempts and stop retrying after a limit is reached.

* **Bug Fixes**
  * Improved reboot handling so an expected SSH disconnect during restart is no longer treated as a failure.
  * When reboot attempts are exhausted, the reboot request is cleared and a warning is recorded.
  * Reboot-related cleanup now removes both the reboot request and any stored attempt count.

### DIFF
--- a/pkg/windows/windows.go
+++ b/pkg/windows/windows.go
@@ -540,7 +540,9 @@ func (vm *windows) Run(cmd string, psCmd bool) (string, error) {
 // RebootAndReinitialize restarts the Windows instance and re-initializes the SSH connection for further configuration
 func (vm *windows) RebootAndReinitialize(ctx context.Context) error {
 	vm.log.Info("rebooting instance")
-	if _, err := vm.Run("Restart-Computer -Force", true); err != nil {
+	// SSH connection may terminate during reboot before returning exit status - this is expected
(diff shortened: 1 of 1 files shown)


mansikulkarni96:
@coderabbit full review

pkg/metadata/metadata.go
```diff
@@ -23,6 +23,10 @@ const (
 	DesiredVersionAnnotation = "windowsmachineconfig.openshift.io/desired-version"
 	// RebootAnnotation indicates the node's underlying instance needs to be restarted
 	RebootAnnotation = "windowsmachineconfig.openshift.io/reboot-required"
```
jrvaldes, line 25: Kubernetes annotations are just map[string]string, so a value like a reboot-required having the counter is perfectly valid. for example reboot-required: "2"

Nothing anywhere inspects the value, so overloading it with the attempt count is a non-breaking change.

 I'd lean toward consolidating to the single annotation, since it also reduces the total annotation surface area on the Node object

jrvaldes (approved):
/lgtm

PR #4359, opened 2026-07-21
[WINC-1979] Implement log rotation for windows-exporter

wgahnagl:
Adds log rotation for windows-exporter 

This allows windows-exporter logs to be managed by kube-log-runner, and rotated automatically.

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,11 +260,11 @@ The log rotation functionality is disabled by default, causing log files to grow
 Managed Windows services with log rotation capabilities:
 - kubelet
 - kube-proxy
+- windows_exporter
 
 Not yet supported:
 - containerd
 - csi-proxy
-- windows_exporter
 - hybrid-overlay-node
 - azure-cloud-node-manager
 
--- a/pkg/services/services.go
+++ b/pkg/services/services.go
@@ -32,9 +32,15 @@ const (
 // will be enabled for services that support it.
 func GenerateManifest(kubeletArgsFromIgnition map[string]string, apiServerEndpoint string, vxlanPort string,
 	platform config.PlatformType, debug bool) (*servicescm.Data, error) {
-	windowsExporterServiceCommand := fmt.Sprintf("%s --collectors.enabled "+
-		"cpu,logical_disk,net,os,service,system,container,memory,cpu_info --web.config.file %s",
-		windows.WindowsExporterPath, windows.TLSConfPath)
+	windowsExporterLogLevel := "info"
+	if debug {
+		windowsExporterLogLevel = "debug"
+	}
+	windowsExporterServiceCommand := logconfig.GetKubeLogRunnerCmd(windows.KubeLogRunnerPath,
+		windows.WindowsExporterPath, windows.WindowsExporterLogPath)
+	windowsExporterServiceCommand = fmt.Sprintf("%s --collectors.enabled "+
+		"cpu,logical_disk,net,os,service,system,container,memory,cpu_info --web.config.file %s --log.level %s --log.file stdout",
+		windowsExporterServiceCommand, windows.TLSConfPath, windowsExporterLogLevel)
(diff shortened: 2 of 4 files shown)


mansikulkarni96:
@code-rabbit full review

pkg/services/services.go
```diff
@@ -32,9 +32,15 @@ const (
...
-		"cpu,logical_disk,net,os,service,system,container,memory,cpu_info --web.config.file %s",
-		windows.WindowsExporterPath, windows.TLSConfPath)
+	windowsExporterLogLevel := "info"
+	if debug {
+		windowsExporterLogLevel = "debug"
+	}
+	windowsExporterServiceCommand := logconfig.GetKubeLogRunnerCmd(windows.KubeLogRunnerPath,
+		windows.WindowsExporterPath, windows.WindowsExporterLogPath)
+	windowsExporterServiceCommand = fmt.Sprintf("%s --collectors.enabled "+
+		"cpu,logical_disk,net,os,service,system,container,memory,cpu_info --web.config.file %s --log.level %s --log.file stdout",
```
mansikulkarni96, line 42: do we need to send the logs to stdout if they are going to a log file? 
whats the benefit of doing this and whats the pattern for other services?

pkg/services/services.go
```diff
@@ -32,9 +32,15 @@ const (
...
 func GenerateManifest(kubeletArgsFromIgnition map[string]string, apiServerEndpoint string, vxlanPort string,
 	platform config.PlatformType, debug bool) (*servicescm.Data, error) {
-	windowsExporterServiceCommand := fmt.Sprintf("%s --collectors.enabled "+
-		"cpu,logical_disk,net,os,service,system,container,memory,cpu_info --web.config.file %s",
-		windows.WindowsExporterPath, windows.TLSConfPath)
+	windowsExporterLogLevel := "info"
+	if debug {
+		windowsExporterLogLevel = "debug"
+	}
+	windowsExporterServiceCommand := logconfig.GetKubeLogRunnerCmd(windows.KubeLogRunnerPath,
```
mansikulkarni96, line 39: 1. what is the default log max file size for kubelogrunner at which the logs will start rotating?
2. There is a env variable added by Jose to pass user defined log file size, will that be taken into account for windows exporter?

README.md
```diff
@@ -260,11 +260,11 @@ The log rotation functionality is disabled by default, causing log files to grow
 Managed Windows services with log rotation capabilities:
 - kubelet
 - kube-proxy
+- windows_exporter
```
mansikulkarni96, line 263: Is this PR implementing log rotation as well?

mansikulkarni96 (changes requested):

mansikulkarni96:
@code-rabbit review

mansikulkarni96:
@code-rabbit full review

pkg/services/services.go
wgahnagl: Adding this stdout flag doesn't technically do anything, it just lets you know that the loging is getting sent to stdout for kube-log-runner to capture

pkg/services/services.go
wgahnagl: https://github.com/openshift/windows-machine-config-operator/blob/740351c65a5d08c5450500908a1ac9327b3c0e5e/pkg/logconfig/logconfig.go#L34 
this is where that is all set 
it uses the env variable added by Jose because it uses the GetKubeLogRunnerCmd function which populates all of that by default. This also implements log rotation by using that command. 
The default max log file is set in that function as well

README.md
wgahnagl: yes! KubeLogRunner handles that. It rotates when the max size is reached.

pkg/services/services.go
mansikulkarni96: If its not doing anything I would skip it and follow the pattern for rest of the services

mansikulkarni96 (changes requested):

PR #4360, opened 2026-07-22
WINC-1966: Add setup-only mode and fix SKIP_NODE_DELETION in CI script

rrasouli:
Line 30 unconditionally wiped the SKIP_NODE_DELETION env var, ignoring any value set by the caller (export SKIP_NODE_DELETION=true from CI config was destroyed before getopts ran). Fix with parameter expansion to preserve the caller's value.

Add setup-only as a new -t option: deploys WMCO and creates Windows nodes, then exits before running any e2e tests. This allows the aws-e2e-ote CI job to provision the cluster without running the full e2e suite, reducing the setup step from ~1.5h to ~20m. Invoked as hack/run-ci-e2e-test.sh -t setup-only -s.

**Related PR:** https://github.com/openshift/release/pull/82317 (uses -t setup-only -s in aws-e2e-ote job). This PR must merge first.

### DIFF
--- a/hack/run-ci-e2e-test.sh
+++ b/hack/run-ci-e2e-test.sh
@@ -27,7 +27,7 @@ source $WMCO_ROOT/hack/common.sh
 
 MACHINE_NODE_COUNT_OPTION=""
 BYOH_NODE_COUNT_OPTION=""
-SKIP_NODE_DELETION=""
+SKIP_NODE_DELETION=${SKIP_NODE_DELETION:-""}
 
 export CGO_ENABLED=0
 
@@ -56,10 +56,10 @@ while getopts ":m:c:b:st:w:" opt; do
     s ) # process option for skipping deleting Windows VMs created by test suite
       SKIP_NODE_DELETION="true"
       ;;
-    t ) # test to run. Defaults to basic. Other options are basic, upgrade-setup, and upgrade-test.
+    t ) # test to run. Defaults to basic. Other options are basic, upgrade-setup, upgrade-test, and setup-only.
       TEST=$OPTARG
-      if [[ "$TEST" != "basic" && "$TEST" != "upgrade-setup" && "$TEST" != "upgrade-test" ]]; then
-        echo "Invalid -t option $TEST. Valid options are basic, upgrade-setup, and upgrade-test"
+      if [[ "$TEST" != "basic" && "$TEST" != "upgrade-setup" && "$TEST" != "upgrade-test" && "$TEST" != "setup-only" ]]; then
+        echo "Invalid -t option $TEST. Valid options are basic, upgrade-setup, upgrade-test, and setup-only"
         exit 1
       fi
       ;;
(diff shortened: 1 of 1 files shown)


hack/run-ci-e2e-test.sh
```diff
@@ -56,6 +57,10 @@ while getopts ":m:c:b:st:w:" opt; do
     s ) # process option for skipping deleting Windows VMs created by test suite
       SKIP_NODE_DELETION="true"
       ;;
+    S ) # setup only: deploy WMCO and create Windows nodes, skip all e2e tests
+      SETUP_ONLY="true"
+      SKIP_NODE_DELETION="true"
+      ;;
```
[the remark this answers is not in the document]
rrasouli: The upgrade-test-setup Makefile target intentionally uses -s (skip node deletion), not -S (setup-only). It needs to run the upgrade-setup e2e tests (-t upgrade-setup) while preserving nodes for the subsequent upgrade-test step. Using -S would skip all tests entirely, breaking the upgrade workflow.

hack/run-ci-e2e-test.sh
```diff
@@ -56,6 +57,10 @@ while getopts ":m:c:b:st:w:" opt; do
     s ) # process option for skipping deleting Windows VMs created by test suite
       SKIP_NODE_DELETION="true"
       ;;
+    S ) # setup only: deploy WMCO and create Windows nodes, skip all e2e tests
+      SETUP_ONLY="true"
+      SKIP_NODE_DELETION="true"
```
mansikulkarni96: This is already set to true using s flag, why is this trying to combine functionality of another flag?

hack/run-ci-e2e-test.sh
```diff
@@ -27,7 +27,8 @@ source $WMCO_ROOT/hack/common.sh
 
 MACHINE_NODE_COUNT_OPTION=""
 BYOH_NODE_COUNT_OPTION=""
-SKIP_NODE_DELETION=""
+SKIP_NODE_DELETION=${SKIP_NODE_DELETION:-""}
```
mansikulkarni96, line 30: @rrasouli I dont see any point doing this.
line 89 sets SKIP_NODE_DELETION=${SKIP_NODE_DELETION:-"false"}. 
The :- substitution only applies when the variable is unset or empty.

hack/run-ci-e2e-test.sh
```diff
@@ -45,7 +46,7 @@ TEST="basic"
 # WINDOWS_SERVER_VERSION will be set in the CI config. If it is not set, default to 2022.
 WIN_VER=${WINDOWS_SERVER_VERSION:-"2022"}
 
-while getopts ":m:c:b:st:w:" opt; do
+while getopts ":m:c:b:st:w:S" opt; do
```
mansikulkarni96: having both -s (skip node deletion) and -S (setup only mode) is confusing , they're visually similar but very unrelated.
I would recommend using  -t setup-only since -t already handles test mode selection like basic, upgrade-setup

mansikulkarni96 (changes requested):

hack/run-ci-e2e-test.sh
mansikulkarni96: @rrasouli This change is a no-op, doesnt do anything.

mansikulkarni96:
/approve
@rrasouli is ote test supposed to fail?

rrasouli:
> /approve @rrasouli is ote test supposed to fail?

Yes, this fix is still having the problem 
`private key file not found: ../internal/config/keys/openshift-qe.pem
    {
        s: "private key file not found: ../internal/config/keys/openshift-qe.pem",
    }`
Which we fixed in commit https://github.com/openshift/windows-machine-config-operator/pull/4319/changes/47b8d9348760742cde543c928b856cd86427d5ba 

Can you add lgtm as well?

PR #4429, opened 2026-08-06
WINC-1968: Add OTE Batch 3 Smokerun tests

rrasouli:
Depends-On: #4375

## Summary

Migrates 5 Smokerun tests from OTP to OTE as part of the ongoing OTE migration ([WINC-1536](https://redhat.atlassian.net/browse/WINC-1536)).
This is Batch 3, following Batch 2 (#4319) which added the initial 10 tests and core helpers.

## Tests Added

| OCP ID | Priority | Description |
|--------|----------|-------------|
| OCP-42204 | Medium | Create Windows pod with a Projected Volume -- deploys a pod with projected volume mounting two secrets, verifies file contents via exec |
| OCP-25593 | Critical | Prevent scheduling non-Windows workloads on Windows nodes -- verifies os=Windows:NoSchedule taint, deploys untolerated pod, confirms scheduling rejection |
| OCP-73752 | Medium | Monitor Network In/Out graphs for WMCO pods -- queries Prometheus for pod:network_receive_bytes_total:sum and pod:network_transmit_bytes_total:sum |
| OCP-84267 | Critical | Hybrid-overlay cert rotation -- verifies binary/CA on host, checks 6 log patterns for cert rotation, creates 10 fake certs and confirms WICD prunes to 2 (OCPBUGS-86246) |
| OCP-88278 | Medium | controllerConfig certificates -- retrieves kubelet CA from controllerConfig, verifies kubelet-ca.crt on each Windows node matches, optionally checks cloud provider CA and additional trust bundle |

## CI Fixes

- **runDebugNodePS namespace fix**: Added explicit `-n openshift-windows-machine-config-operator` to `oc debug node` calls. In CI, the kubeconfig context defaults to an ephemeral `ci-op-*` namespace that doesn't exist, causing `unable to get namespace` errors. This affected OCP-88278 and OCP-84267.
- **Prometheus metric polling (OCP-73752, OCP-70922)**: Recording rules (`pod:network_receive_bytes_total:sum`, `instance:node_cpu_utilisation:rate1m`) require EndpointSlice propagation, scrape completion, recording rule evaluation, and 2+ data points for rate functions before returning results. Replaced single-shot queries with `wait.Poll` retry loops (10s interval, 5min timeout) matching the established WMCO e2e pattern (`test/e2e/metrics_test.go:279`). Polls inline `gjson.Parse` instead of calling `extractMetricValue` (which contains `o.Expect` that would panic mid-retry), and track `lastReason` for diagnostic timeout messages.
[...]

PR #4439, opened 2026-08-10
WINC-1968: [ote] Fix OCP-73752 network metric assertion

rrasouli:
## Summary

Fixes OCP-73752 OTE test that fails 100% of the time in CI with "Timed out waiting for metric pod:network_receive_bytes_total:sum" after 5 minutes of polling.

## Root Cause

The recording rule `pod:network_receive_bytes_total:sum` does not produce results for the WMCO operator pod because it is a Linux control-plane pod, not a Windows workload pod. The original OTP test only verifies the Prometheus query executes successfully (status=success) without asserting the metric value is non-empty. PR #4429 made the test stricter than the original by adding a non-empty assertion inside a poll loop, which caused the test to time out every run.

## Fix

Revert to the original OTP behavior: query Prometheus, assert no error, log the value. Do not require the value to be non-empty.

## Testing

- Compiles and passes gofmt
- Matches the original OTP test behavior at openshift-tests-private/test/extended/winc/winc.go

## Summary by CodeRabbit

* **Tests**
  * Streamlined network metrics validation by executing each Prometheus query once and immediately evaluating the returned metric.
  * Removed repeated polling and retry handling for query errors, unsuccessful responses, and empty results.

### DIFF
--- a/ote/test/e2e/winc.go
+++ b/ote/test/e2e/winc.go
@@ -580,30 +580,9 @@ spec:
 		for _, podName := range podNames {
 			for _, metric := range networkMetrics {
 				g.By(fmt.Sprintf("Verifying %s for pod %s", metric, podName))
-				var metricValue, lastReason string
-				pollErr := wait.Poll(10*time.Second, 5*time.Minute, func() (bool, error) {
-					queryResult, err := mon.SimpleQuery(fmt.Sprintf("%s{pod=\"%s\"}", metric, podName))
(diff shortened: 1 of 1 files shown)


jrvaldes:
@coderabbitai full review

jrvaldes:
/override "Red Hat Konflux / windows-machine-config-operator-release-5-0-ec / windows-machine-config-operator-release-5-0"

PR #4456, opened 2026-08-12
WINC-2047: [ote] Set test lifecycle to blocking

rrasouli:
## Summary

All OTE tests had their lifecycle hardcoded to `LifecycleInforming` in `ote/cmd/wmco-tests-ext/main.go`. The openshift-tests framework treats informing tests as non-blocking, so the `aws-e2e-ote` Prow job reported "all tests passed" even when individual tests failed. PRs were merged based on this false signal.

This changes the lifecycle to `LifecycleBlocking` so that test failures properly fail the CI job.

## Changes

- `ote/cmd/wmco-tests-ext/main.go`: `LifecycleInforming` -> `LifecycleBlocking`

## Summary by CodeRabbit

* **Tests**
  * Component test specifications now block lifecycle progression when they fail, rather than reporting failures as informational only.

### DIFF
--- a/ote/cmd/wmco-tests-ext/main.go
+++ b/ote/cmd/wmco-tests-ext/main.go
@@ -73,7 +73,7 @@ func main() {
 		if match := re.FindStringSubmatch(spec.Name); match != nil {
 			spec.Include(et.PlatformEquals(match[1]))
 		}
-		spec.Lifecycle = et.LifecycleInforming
+		spec.Lifecycle = et.LifecycleBlocking
 	})
 
 	ext.AddSpecs(componentSpecs)


jrvaldes:
@rrasouli  on this PR, please include the corresponding fixes so that aws-e2e-ote  job passes.

rrasouli:
The 4 OTE test failures (56354, 50924, 76765, 79100) are pre-existing, they were failing silently before this change because all tests had LifecycleInforming, which meant the job always reported success regardless of test results.

This PR makes those failures visible by switching to LifecycleBlocking. The fixes for those tests are in PR #4446 (HostProcess container support for service management operations).

Merging this first is important so that future test failures are caught instead of silently passing. PR #4446 should follow to resolve the now-visible failures.

PR #4490, opened 2026-08-18
WINC-1972: Add Serial tags to reduce OTE CI load

rrasouli:
[WINC-1972](https://redhat.atlassian.net/browse/WINC-1972): Fix OTE Serial test timeouts, cluster-state pollution, and node-count assumptions

## Summary

The OTE Serial suite was passing at roughly 40 percent. The failures were not flakes. Three distinct problems were compounding, and a single failing test cascaded into skips and failures for everything that ran after it:

1. Long-running tests were being killed at 15 minutes by the runner, because `g.SpecTimeout()` alone does not raise the runner's cap.
2. OCP-33794 was deleting the cluster's own Windows MachineSet, permanently destroying Windows nodes mid-suite.
3. Cleanup paths across the suite hardcoded the Windows node count as `2` or `3`, so on any cluster with a different shape they waited for a state that could never be reached.

This PR fixes all three, adds the diagnostics needed to tell them apart in CI logs, and unblocks one test that was skipped on a misdiagnosis.

Files changed: `ote/test/e2e/winc.go`, `ote/test/e2e/utils.go`, `ote/test/e2e/proxy.go`.

## 1. Test scheduling: Serial tags and dual timeout enforcement

OTE enforces two independent timeouts. `g.SpecTimeout()` bounds the Ginkgo spec, but the test runner applies its own 15-minute cap that is only raised by a `[Timeout:Xm]` tag in the test NAME. Four tests in `winc.go` and all five in `proxy.go` had `SpecTimeout` but no name tag, so they were terminated at 15 minutes with "Interrupted by User" regardless of the value passed to `SpecTimeout`.

`winc.go` - added `[Timeout:Xm]` to nine tests; five of those also gained a `SpecTimeout` they never had:

| Test | Name tag | SpecTimeout |
|---|---|---|
| OCP-84267 hybrid-overlay-node cert rotation | `[Timeout:45m]` | added, 40m |
| OCP-50403 Windows services ConfigMap | `[Timeout:35m]` | added, 30m |
| OCP-56354 stop dependent services in WICD | `[Timeout:45m]` | added, 40m |
| OCP-50924 kubelet CA rotation | `[Timeout:50m]` | added, 45m |
| OCP-87809 node drain with DaemonSet | `[Timeout:35m]` | added, 30m |
| OCP-76765 WICD-Remove-Services | `[Timeout:35m]` | existing, 30m |
| OCP-33794 watch cloud private key secret | `[Timeout:35m]` | existing, 30m |
| OCP-39451 access Windows workload through clusterIP | `[Timeout:35m]` | existing, 30m |
| OCP-39030 re-queue on Windows machines edge cases | `[Timeout:35m]` | existing, 30m |

`proxy.go` - added `[Timeout:Xm]` to the five tests that already carried a `SpecTimeout`: OCP-90290 (50m), OCP-90289 (65m), OCP-66670 (35m), OCP-68320 (50m), OCP-71173 (35m). No `SpecTimeout` values changed.

Eight `winc.go` tests that mutate cluster-wide state also gained `[Serial]` (4 tagged at base, 12 now), so they no longer run concurrently with tests that read the same state.

## 2. OCP-33794 was deleting the cluster's Windows MachineSet

Root cause, `winc.go`:

```go
sourceMSName := getWindowsMachineSetName(oc, defaultWindowsMS, iaasPlatform, zone)
cloneMSName := strings.ReplaceAll(sourceMSName, "winworker", "winc-worker")
```

On the AWS CI cluster `getWindowsMachineSetName` returns `ci-op-<id>-e2e-wm`, matched by the `strings.HasSuffix(ms, "-wm")` branch. That name contains no `winworker`, so `ReplaceAll` is a no-op and `cloneMSName == sourceMSName`.

Every subsequent operation then targeted the live MachineSet. `cloneWindowsMachineSet` used `oc apply`, which patched the existing object rather than creating a clone:

```
Warning: resource machinesets/ci-op-...-e2e-wm is missing the kubectl.kubernetes.io/last-applied-configuration annotation
machineset.machine.openshift.io/ci-op-...-e2e-wm configured
machineset.machine.openshift.io/ci-op-...-e2e-wm scaled
```

The deferred cleanup then deleted it:

```
machineset.machine.openshift.io "ci-op-...-e2e-wm" deleted from openshift-machine-api namespace
Windows nodes ready: 2/3
Windows nodes ready: 1/3    <- for the next 14 minutes
```

Nothing could recreate those nodes. The test hit its 30-minute `SpecTimeout` and leaked a goroutine, and the cluster stayed at one Windows node for the rest of the suite. OCP-77777 skipped nine minutes later with "Need at least 2 Windows nodes to test cross-node HTTP rejection" purely as a consequence.

Fix, in `winc.go`:

```go
// Must always differ from sourceMSName: the deferred cleanup below deletes it.
cloneMSName := sourceMSName + "-clone"
```

Fix, in `cloneWindowsMachineSet` (`utils.go`) - the invariant is now enforced at the layer that owns it, so this class of bug fails loudly instead of silently destroying the cluster:

- Assert `cloneName != sourceName` before doing anything.
- Use `oc create` instead of `oc apply`, so the helper can never mutate an existing MachineSet.
- Strip server-managed fields (`status`, and `metadata.resourceVersion`, `uid`, `creationTimestamp`, `generation`, `selfLink`, `managedFields`, last-applied annotation) so the object is accepted as a genuine create. The previous code only worked because `apply` took the update path; a real create with `resourceVersion` set would have been rejected.
- Set `spec.replicas: 0` in the manifest instead of creating at full size and scaling down immediately afterwards, removing both the extra call and the provision/deprovision churn in between.

## 3. Cleanup paths no longer assume a fixed node count

Cleanup code across the suite restored the cluster to a hardcoded 2 or 3 Windows nodes. On any cluster that did not happen to match, `waitWindowsNodesReady` polled until its timeout for a state that could never be reached, failing the test and leaving the cluster short for everything after it.

Every such site now captures the real value before mutating anything and restores to that:

| Test | Was | Now |
|---|---|---|
| OCP-89616 log rotation | `waitWindowsNodesReady(oc, 2, ...)` | `expectedNodes` from `len(winInternalIPs)` |
| OCP-84267 cert rotation | no restore | `defer waitWindowsNodesReady(oc, expectedNodes, 15m)` |
| OCP-88278 controllerConfig certs | `2` | `len(winInternalIPs)` |
| OCP-56354 WICD dependent services | `2` | `expectedWindowsNodes` from `getWindowsHostNames` |
| OCP-76765 WICD-Remove-Services (3 sites) | `2` | `len(winHostNames)` |
| OCP-50924 kubelet CA rotation | scale back to `2` | `initialReplicas` + `originalWindowsNodeCount` |
| OCP-33794 private key secret | `2` | `expectedOriginalNodes` |
| OCP-39451 clusterIP access | `2` restore, `3` assertion | `initialReplicas` / `originalWindowsNodeCount+1` |
| OCP-39030 re-queue edge cases | `2` restore, `3` scale and assertion | `initialReplicas` / `originalWindowsNodeCount+1` |
| OCP-87809 node drain | `2` | `expectedCount` from `getWindowsHostNames` |

In OCP-87809 the deferred cleanups were also moved to after the node-count capture, so they register with the correct value.

New helper in `utils.go`:

```go
// getMachineSetReplicas returns the desired replica count of the given MachineSet. Tests that scale
// a MachineSet must capture this before scaling so cleanup restores the cluster's original size
// instead of assuming a fixed count.
func getMachineSetReplicas(oc *exutil.CLI, machineSetName string) int
```

The remaining literal counts are deliberate. OCP-50924 scales to exactly one node and asserts `waitWindowsNodesReady(oc, 1, ...)` because single-node CA rotation is the behaviour under test, not a cleanup assumption.

## 4. OCP-87809 unskipped

OCP-87809 was skipped in this PR with `Blocked by OCPBUGS-114564: WMCO reconfiguration hangs indefinitely during defer cleanup after annotation change`. The run that prompted that skip does not support it. All nine steps passed:

```
isNone: Windows machines (...os-id=Windows)="ci-op-83qv3mx2-dec2f-p94qh-e2e-lx5ng"   <- ONE node
DaemonSet ready: 1/1 pods
STEP: Step 3: Manually drain node ... Step 5: Uncordon
DaemonSet recovery after manual drain: 1/1 pods ready
STEP: Step 7: Trigger WMCO reconciliation by setting invalidVersion
STEP: Step 8: Wait for WMCO to reconcile the node
DaemonSet recovery after WMCO reconciliation: 1/1 pods ready       <- Step 9 passed
```

The only failure was in cleanup: `timed out waiting for 2 Windows nodes to be Ready after 15m0s`, from the old hardcoded `2` on a cluster that only ever had one Windows node. The node-ready progression over that 15 minutes shows WMCO was healthy throughout: Ready for about 2 minutes, NotReady for about 3 minutes while the node rebooted for reconfiguration, then Ready again and stable for the remaining 8.5 minutes. Reconfiguration completed; there was no hang.

That cleanup now uses the captured count, so the skip is removed.

## 5. OCP-50924 remains skipped

OCP-50924 keeps its `OCPBUGS-114564` skip. Unlike OCP-87809, the bug carries concrete logs for this test and they describe a real WMCO defect that this PR does not address:

```
1. MachineSet scales 1 to 2   [SUCCESS]
2. New Machine is created     [SUCCESS]
3. VM provisions and boots    [SUCCESS]
4. Node joins cluster         [SUCCESS]
5. WMCO starts reconfiguration on new node -- HANGS INDEFINITELY
6. Node stuck Ready=False:
     "container runtime network not ready: NetworkReady=false
      reason:NetworkPluginNotReady message:Network plugin returns error:
      cni plugin not initialized"
```

That cluster genuinely had two Windows nodes, so restoring to two was the correct target; the newly provisioned node hung in WMCO's HNS network recreation at +3:40 and never recovered. The dynamic-replica change above is still correct for this test, but it does not unblock it. The bug also states explicitly that raising timeouts is not a valid workaround, so the added post-scale node wait is bounded at 10 minutes rather than matching the 20-minute scale wait, keeping total cleanup inside the 45-minute spec budget.

## 6. Bounded, cancellable diagnostics for NotReady nodes

When `waitWindowsNodesReady` has been waiting more than 5 minutes, it captures node status, conditions, WMCO and WICD pod status, event summaries, machine status, and Windows service status. Without bounds this ran synchronously inside the `wait.Poll` callback, so the poll timeout was not enforced until it returned, and it could consume the caller's entire remaining budget.

`debugWindowsNodesNotReady` now takes a `context.Context`, and the caller bounds it:

```go
ctx, cancel := context.WithTimeout(context.Background(), diagnosticTimeout) // 2 minutes
debugWindowsNodesNotReady(ctx, oc)
cancel()
```

The per-node loop checks `ctx.Done()` before each capture and stops when the deadline expires. A goroutine with `time.After` was deliberately not used: the callback would return while the helper kept issuing `oc` commands, overlapping later polling and cleanup.

The capture logs summaries only (condition types, event types and reasons, pod phases) rather than raw logs, annotations, or event messages, and uses `runHostProcessPS` for service status instead of `runDebugNodePS`.

## 7. Windows MachineSet discovery on CI clusters

`getWindowsMachineSetName` matched only `winworker`, the `windows` keyword, or a `-wm` suffix, and called `e2e.Failf` otherwise. Two fallbacks were added for CI naming:

1. A MachineSet whose name contains `-e2e` but not `worker`, confirmed as Windows via its `machine.openshift.io/os-id` template label.
2. Resolution through a Windows node's `machine.openshift.io/machine` annotation to the Machine, then to its `cluster-api-machineset` label.

## Test results

Latest AWS OTE run, before the MachineSet-clone fix in this PR: 28 of 30 passing. The two non-passes were OCP-33794 (timed out, root cause section 2) and OCP-77777 (skipped as a direct consequence).

vSphere proxy OTE: 5 pass, 1 fail. The failure is OCP-65980, a HostProcess pod that did not complete within its wait on vSphere. It is environmental and unrelated to these changes.

## Known remaining work

`time.Sleep` and `wait.Poll` inside test callbacks in OCP-50403, OCP-50924, OCP-87809, and OCP-84267 are not yet context-aware and should move to `wait.PollWithContext` and `select` on `ctx.Done()`. The diagnostics half of that work is done in section 6; the test-body half is not.

### DIFF
--- a/ote/test/e2e/proxy.go
+++ b/ote/test/e2e/proxy.go
@@ -46,7 +46,7 @@ var _ = g.Describe("[OTP][sig-windows][apigroup:config.openshift.io] Windows_Con
 		o.Expect(err).NotTo(o.HaveOccurred(), "trusted-ca configmap not found")
 	})
 
-	g.It("Smokerun-Author:rrasouli-Longduration-Critical-90290-[node-proxy]-Remove trusted CA from cluster proxy and verify propagation [Serial][Disruptive][Slow]",
+	g.It("Smokerun-Author:rrasouli-Longduration-Critical-90290-[node-proxy]-Remove trusted CA from cluster proxy and verify propagation [Timeout:50m][Serial][Disruptive][Slow]",
 		g.SpecTimeout(45*time.Minute),
 		func(ctx g.SpecContext) {
 			defer restoreProxyEnvironment(oc, initialProxySpec)
@@ -72,7 +72,7 @@ var _ = g.Describe("[OTP][sig-windows][apigroup:config.openshift.io] Windows_Con
 			}
 		})
 
-	g.It("Smokerun-Author:rrasouli-Longduration-Critical-90289-[node-proxy]-Remove proxy variables and verify WMCO propagation [Serial][Disruptive][Slow]",
+	g.It("Smokerun-Author:rrasouli-Longduration-Critical-90289-[node-proxy]-Remove proxy variables and verify WMCO propagation [Timeout:65m][Serial][Disruptive][Slow]",
 		g.SpecTimeout(60*time.Minute),
 		func(ctx g.SpecContext) {
 			winNodes := getWindowsNodeNames(oc)
@@ -131,7 +131,7 @@ var _ = g.Describe("[OTP][sig-windows][apigroup:config.openshift.io] Windows_Con
 			}
 		})
 
-	g.It("Smokerun-Author:rrasouli-Longduration-Critical-66670-[node-proxy]-Cluster-wide proxy trusted-ca configmap tests [Serial][Disruptive][Slow]",
+	g.It("Smokerun-Author:rrasouli-Longduration-Critical-66670-[node-proxy]-Cluster-wide proxy trusted-ca configmap tests [Timeout:35m][Serial][Disruptive][Slow]",
 		g.SpecTimeout(30*time.Minute),
 		func(ctx g.SpecContext) {
 			defer restoreProxyEnvironment(oc, initialProxySpec)
@@ -183,7 +183,7 @@ var _ = g.Describe("[OTP][sig-windows][apigroup:config.openshift.io] Windows_Con
 			waitForCM(oc, trustedCACM, trustedCACM, wmcoNamespace)
 		})
 
-	g.It("Smokerun-Author:rrasouli-Critical-68320-[node-proxy]-Import custom CA certificates into Windows node system store [Serial][Disruptive]",
+	g.It("Smokerun-Author:rrasouli-Critical-68320-[node-proxy]-Import custom CA certificates into Windows node system store [Timeout:50m][Serial][Disruptive]",
 		g.SpecTimeout(45*time.Minute),
 		func(ctx g.SpecContext) {
 			const (
@@ -240,7 +240,7 @@ var _ = g.Describe("[OTP][sig-windows][apigroup:config.openshift.io] Windows_Con
 			checkUserCertificatesOnNodes(oc, userSelfSignedCommonName, 0)
 		})
 
-	g.It("Author:rrasouli-Smokerun-Longduration-Critical-71173-[node-proxy]-Test connectivity from Windows nodes behind proxy [Serial][Disruptive][Slow]",
+	g.It("Author:rrasouli-Smokerun-Longduration-Critical-71173-[node-proxy]-Test connectivity from Windows nodes behind proxy [Timeout:35m][Serial][Disruptive][Slow]",
 		g.SpecTimeout(30*time.Minute),
 		func(ctx g.SpecContext) {
 			o.Expect(getClusterProxy(oc, "status.noProxy")).ToNot(o.BeEmpty(), "status.noProxy is not set on the cluster")
--- a/ote/test/e2e/utils.go
+++ b/ote/test/e2e/utils.go
@@ -1400,6 +1400,20 @@ func getWindowsMachineSetName(oc *exutil.CLI, name, platform, zone string) strin
 	return machinesetName
 }
 
+// getMachineSetReplicas returns the desired replica count of the given MachineSet. Tests that scale
+// a MachineSet must capture this before scaling so cleanup restores the cluster's original size
+// instead of assuming a fixed count.
+func getMachineSetReplicas(oc *exutil.CLI, machineSetName string) int {
+	replicas, err := oc.AsAdmin().WithoutNamespace().Run("get").Args(
+		"machinesets.machine.openshift.io", machineSetName, "-n", mcoNamespace,
+		"-o=jsonpath={.spec.replicas}").Output()
+	o.Expect(err).NotTo(o.HaveOccurred(), "Failed to get replicas of MachineSet %s", machineSetName)
+
+	count, err := strconv.Atoi(strings.TrimSpace(replicas))
+	o.Expect(err).NotTo(o.HaveOccurred(), "Unexpected replica count %q for MachineSet %s", replicas, machineSetName)
+	return count
+}
+
 // scaleWindowsMachineSet scales the Windows MachineSet to the specified replica count.
 func scaleWindowsMachineSet(oc *exutil.CLI, machineSetName string, deadTime, replicas int, skipWait bool) {
 	err := oc.AsAdmin().WithoutNamespace().Run("scale").Args(
--- a/ote/test/e2e/utils.go
+++ b/ote/test/e2e/utils.go
@@ -352,8 +352,147 @@ func getNodeNameFromIP(oc *exutil.CLI, nodeIP string) string {
 	return nodeName
 }
 
+// debugWindowsNodesNotReady captures comprehensive debug information when Windows nodes fail to become Ready
+func debugWindowsNodesNotReady(oc *exutil.CLI) {
+	e2e.Logf("===== DEBUG: Windows Nodes Not Ready - Capturing Diagnostic Information =====")
+
+	// 1. Get detailed status for all Windows nodes
+	e2e.Logf("=== Windows Node Status ===")
+	nodeOutput, err := oc.AsAdmin().WithoutNamespace().Run("get").Args(
+		"nodes", "-l", windowsNodeLabel,
+		"-o=custom-columns=NAME:.metadata.name,STATUS:.status.conditions[?(@.type==\"Ready\")].status,REASON:.status.conditions[?(@.type==\"Ready\")].reason,MESSAGE:.status.conditions[?(@.type==\"Ready\")].message").Output()
+	if err != nil {
+		e2e.Logf("ERROR: Failed to get node status: %v", err)
+	} else {
+		e2e.Logf("Node status:\n%s", nodeOutput)
+	}
+
+	// 2. Get all node conditions for each Windows node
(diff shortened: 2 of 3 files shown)


jrvaldes:
@coderabbitai full review

rrasouli:
@coderabbitai regarding the **Microshift Test Compatibility** warning on OCP-87809:

This test is intentionally OCP-only—Windows node support doesn't exist on MicroShift. The test is already correctly scoped:

1. **Location**: Inside `[OTP][sig-windows]` block with Windows-only tests
2. **Dependencies**: Requires multiple Windows nodes (`getWindowsHostNames(oc)`) and WMCO reconfiguration
3. **Natural incompatibility**: MicroShift is single-node; Windows multi-node reconciliation is not applicable

Since OCP excludes MicroShift at the suite level (via job variant selection), adding an explicit skip is redundant. The test will naturally not run on MicroShift because there are no Windows nodes in that environment.

Also note: the test is currently blocked by OCPBUGS-114564, so it skips before reaching any node interaction code.

rrasouli:
@coderabbitai review

rrasouli:
@coderabbitai review

rrasouli:
Regarding CodeRabbit's Microshift Test Compatibility finding on OCP-87809:

This test is intentionally OCP-only—Windows node support doesn't exist on MicroShift. The test is already correctly scoped:

1. **Location**: Inside `[OTP][sig-windows]` block with Windows-only tests
2. **Dependencies**: Requires multiple Windows nodes and WMCO reconfiguration
3. **Natural incompatibility**: MicroShift is single-node; Windows multi-node reconciliation is not applicable

Job variants handle platform selection at the suite level (OCP vs MicroShift jobs are separate), so explicit MicroShift guards are redundant. The test will naturally not run on MicroShift because there are no Windows nodes in that environment.

Additionally, OCP-87809 is currently blocked by OCPBUGS-114564, so it's skipped before reaching any node interaction code.

rrasouli:
**Re: Diagnostic timeout (Lines R498-R503)**

✅ ADDRESSED in commit 16e44d39d:

- Added `diagnosticTimeout := 2 * time.Minute` to bound diagnostic capture
- Wrapped `debugWindowsNodesNotReady` call with `context.WithTimeout(2 * time.Minute)`
- Added `ctx.Done()` check in the diagnostic loop to stop node captures early when deadline expires
- Updated function signature: `debugWindowsNodesNotReady(ctx context.Context, oc *exutil.CLI)`

The diagnostic helper now respects a hard 2-minute deadline instead of potentially consuming the caller's entire timeout budget via unbounded `oc debug` operations.

rrasouli:
**Re: Context propagation to blocking operations (Lines R1237-R1238)**

Partial fix applied in commit 16e44d39d:

✅ **Diagnostics**: Diagnostic context now cancels synchronously with 2-minute deadline

⏳ **Remaining work**: Replace `time.Sleep` and `wait.Poll` in test callbacks with context-aware alternatives

The three callback tests (OCP-50403, OCP-50924, OCP-87809, OCP-84267) still use:
- `time.Sleep(3 * time.Minute)` which ignores SpecTimeout cancellation
- `wait.Poll` without context awareness

These should be replaced with:
- `select { case <-time.After(...): case <-ctx.Done(): return }` 
- `wait.PollWithContext(ctx, ...)`

This is a heavy lift spanning multiple tests and will be addressed in a follow-up.

jrvaldes (commented):
/lgtm

PR #4594, opened 2026-09-10
OCPBUGS-97792: Add SSHAddressAnnotation support to GetAddress

emagtala:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved node address selection by prioritizing configured SSH addresses.
  - Internal IP addresses are now preferred over internal DNS names.
  - IPv6 addresses are now supported when selecting node connection addresses.

### DIFF
--- a/controllers/controllers.go
+++ b/controllers/controllers.go
@@ -3,7 +3,6 @@ package controllers
 import (
 	"context"
 	"fmt"
-	"net"
 	"sync"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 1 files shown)


controllers/controllers.go
```diff
@@ -32,6 +31,11 @@ const (
 	MaxParallelUpgrades = 1
 )
 
+const (
```
jrvaldes, line 34: consider using the existing const block

jrvaldes (changes requested):
@emagtala   thanks for opening the PR, see https://github.com/openshift/windows-machine-config-operator/blob/master/CONTRIBUTION.md and format the commit message  mentioning what/why this PR resolves OCPBUGS-97792

